Issue #503, opened 2026-08-08
A columnar scan is priced with no decode cost, so the planner declines a parallel projection that is 2.9x faster


OffgridwithJD:
A row-returning columnar scan is priced without any decode cost, so the planner declines a parallel plan that is **2.9x faster** — and the plan it picks loses to heap on the same query, despite the columnar table being 4.4x smaller.

## Measured

20,000,000 rows, identical data in both, PG 18.4, `max_parallel_workers_per_gather = 8`, median of three. Query returns 798,020 rows:

```sql
SELECT id, k, v FROM c WHERE k BETWEEN 100 AND 40000
```

| plan | median |
| --- | ---: |
| columnar serial — **what the planner picks** | 351.9 ms |
| columnar parallel — **what it refuses** | **121.3 ms** |
| heap parallel — what the planner picks for heap | 314.2 ms |

Columnar is 363 MB, heap is 1611 MB. **The columnar plan the planner chooses is slower than the heap plan it chooses**, on 4.4x less data, purely because heap gets parallelised and columnar does not.

This is a pricing decision, not a missing capability. Setting `parallel_tuple_cost = 0` produces `Gather -> Parallel Custom Scan` immediately, so the partial path is offered and costed; it just never wins.

## Why it never wins

```
serial:               Custom Scan (PgColumnarScan) on c  (cost=0.00..71444.47 rows=847435)
parallel, tuple=0:    Gather  (cost=1000.00..36722.24 rows=847435)
                        ->  Parallel Custom Scan (cost=0.00..35722.24 rows=423718)
```

The parallel path is genuinely cheaper on scan cost, 36,722 against 71,444. At the default `parallel_tuple_cost = 0.1` the Gather then adds `0.1 * 847435 = 84,743` for row shipping, which alone exceeds the entire serial estimate. Serial wins by arithmetic.

## The cost model has no decode term

`pgcolumnar_cost_custom_path` (`src/columnar_customscan.c:1247`) builds the estimate as the heap-equivalent full scan cost, minus the I/O for unread columns, scaled by zone-map survival. Its only per-row terms are core's `cpu_tuple_cost` and the qual cost:

```c
run = seq_page_cost * (double) rel->pages;
run += (cpu_tuple_cost + qcost.per_tuple) * ntuples;      /* :778-779 */
```

[...]

Issue #565, opened 2026-08-10
docs/benchmarks.md says the q18/q31 accelerator regressions are "run to run variation" with an "identical" plan; the plans differ and the effect is 3.9x-5.6x, stable to 1%


OffgridwithJD:
## The claim

`docs/benchmarks.md`, in "The accelerations are off by default, and turning them
on is not a clear win":

> Three queries improve and three get worse. This is a reason to investigate
> rather than a conclusion, and one caution belongs with it. For `q18` and `q31`
> the plan is **identical** with these settings on and off. Their rows are
> therefore run to run variation, and not an effect of the settings.

Both halves of that are false. I am filing it because the page's *conclusion* —
do not turn these on by default — is correct and well supported, and an incorrect
reason attached to a correct conclusion is the kind of thing that gets repaired
into the wrong fix later.

## The plans are not identical

`q18`, on the ClickBench table (11,110,833 rows), `main` at `744b7ca`,
non-assert pg18n. `EXPLAIN (COSTS OFF, VERBOSE)`, GUCs off then on, nothing else
changed:

**Accelerators OFF — 19 rows, parallel:**

```
 Limit
   ->  Finalize GroupAggregate
         Group Key: hits_col.userid, hits_col.searchphrase
         ->  Gather Merge
               Workers Planned: 7
               ->  Sort
                     ->  Partial HashAggregate
                           ->  Parallel Custom Scan (PgColumnarScan) on public.hits_col
                                 Columnar Projected Columns: 2
```

**Accelerators ON — 8 rows, serial:**

```
 Limit
   ->  Custom Scan (PgColumnarScan)
         Columnar Vectorized Group Keys: 2
         Columnar Vectorized Aggregates: 1
```

A seven-worker parallel plan is replaced by a serial one. That is not an
identical plan by any reading.

## The effect is not run-to-run variation

Interleaved off/on, three passes each, so a warming trend cannot favour either
arm:

| query | GUC off | GUC on | ratio |
| --- | ---: | ---: | ---: |
| q18 | 1540.4 / 1534.4 / 1534.8 ms | 8726.5 / 8502.4 / 8516.9 ms | **5.6x** |
| q31 | 1521.6 / 1507.4 / 1507.7 ms | 5975.1 / 5877.0 / 5924.0 ms | **3.9x** |
| q32 | 1601.0 / 1602.7 / 1597.0 ms | 6328.3 / 6376.1 / 6246.9 ms | **3.9x** |

[...]

Issue #171, opened 2026-07-27
Point lookup regressed 23.75 ms to 1251.88 ms once ANALYZE has run, and ANALYZE itself is pathologically slow on some shapes


jdatcmd:
Two observations from the full benchmark run at `7a9c9f7`. They may share a
cause. I am filing them together with what is confirmed and what is not, rather
than guessing.

## 1. Point lookup is 53x slower than the previous run

`docs/benchmarks.md` at `1be027b` recorded 23.75 ms. The same query, same
harness, same machine, at `7a9c9f7`:

| query | heap | columnar |
| --- | --- | --- |
| point lookup by indexed id | 0.01 ms | **1251.88 ms** |

**It is not the lazy slot (#169).** I A/B'd `SELECT *` across that merge on a
6,000,000-row table with no `ANALYZE`:

| shape | before #169 | after #169 |
| --- | --- | --- |
| `SELECT *` | 16.42 ms | 13.73 ms |
| one column | 17.46 ms | 21.43 ms |
| two columns | 24.66 ms | 30.20 ms |
| `id` only | 0.04 ms | 0.03 ms |

So the fetch path is fine. The difference between my probe and the benchmark is
that **the benchmark runs `ANALYZE cz` before measuring** (`bench/run_bench.sh`
line 121) and my probe did not. Statistics exist now that did not before #159, so
the planner is choosing differently for `SELECT * WHERE id = $MID`, and choosing
worse.

I have not captured the two plans side by side: see below for why.

## 2. ANALYZE runs for tens of minutes on a wide table, in isolation

Trying to capture those plans, `ANALYZE` on a **250,000-row**, 8-column table
(one bigint, three int, one timestamptz, three ~20-character text) ran for
**23 minutes** before I killed it. On a 6,000,000-row table of the same shape it
ran **40 minutes** before I killed it. Both on an otherwise idle box,
`wait_event_type` null throughout, so it was running rather than blocked.

**This did not reproduce inside the benchmark**, which completed all three
harnesses including `ANALYZE cz` on 6,000,000 rows in 26 minutes total. So
something differs between the two setups and I have not isolated it. Candidates I
have not ruled out: the exact column set, whether the index exists before
`ANALYZE`, or an interaction with a preceding `EXPLAIN (ANALYZE)` in my script.

`test/analyze_stats.sh` uses 500,000 rows and four narrow columns and passes in
seconds, which is consistent with cost being driven by column count and width
[...]

PR #580, opened 2026-08-11
fix: price per-column decode CPU so the planner sees the parallel win (#503)

jdatcmd:
Closes the plan-side of #503. The measured half was @ChronicallyJD's; this is the
fix built on that measurement.

## The defect

A columnar scan inherited the heap seqscan cost, whose CPU term
(`cpu_tuple_cost`) is per row, paid once whatever the projection width. So
decoding nine columns was priced the same as decoding one, and the planner could
not see the decode CPU that a parallel plan divides across workers. It declined
the partial path (which `add_path` builds, `columnar_customscan.c:1649`) even
where the parallel scan is faster.

## Isolated cleanly, not conflated with cardinality

@ChronicallyJD measured the issue's 4M shape and controlled the confound I
flagged: **after `ANALYZE` the row estimate is accurate (156,596 vs 159,604
actual) and the plan is STILL serial** — so it is the cost model, not the
estimate. Interleaved A/B on an idle bench: **serial 64.4 ms vs forced-parallel
25.1 ms = 2.56x**, `Workers Launched: 4`.

Distinct from #565: that is the aggregate `parallel_safe = false` foreclosing the
plan structurally; here the plan exists and is only mispriced.

## The fix

`pgcolumnar_projected_ncols` counts decoded columns, and the cost site prices one
`cpu_operator_cost` per decoded value, `ncols` per row, scaled by survival — the
same per-value shape the index-fetch penalty already uses.

## #171 safety, proven not asserted

Raising the full-scan cost is the safe direction for #171 (a pricier full scan
keeps the index scan more often, never less). The new suite pins a point-lookup
arm directly, and `analyze_stats` (the #171 regression pin), `zonemap_cost` and
`planner_choice_quality` all stay green — so a too-large term cannot push the
planner off the index either.

## Effect confirmed in-container

On the 4M shape the planner now chooses `Gather -> Parallel Custom Scan` where it
previously chose serial. The magnitude — does the flipped plan match the measured
2.56x on real ClickBench data — is a bench follow-up @ChronicallyJD can take; the
plan flip, the #171 safety, and the cost scaling are proven here.

## TDD

`test/scan_decode_cost.sh`: a wide projection must cost >= 1.5x a narrow one.
RED on unfixed main (`got [flat]`, the two are ~1.4% apart), GREEN after. Point-
lookup arm pins #171. Broad plan-dependent sweep green:
native_agg/index/index_projection, differential, parallel/parallel_degree,
pushdown_report, column_projection, native_projection.

## Gate

PG18 non-assert for the plan/timing reasoning; the suite is version-independent
plan logic. **I have not run the full PG15-19 matrix.** Suite registered sorted,
runner reports 146, harness_selftest 107. Not merging this.

### DIFF
--- a/src/columnar_customscan.c
+++ b/src/columnar_customscan.c
@@ -954,6 +954,40 @@ pgcolumnar_full_scan_cost(RelOptInfo *rel, Path *seqpath)
  *		Both the target list and the restriction clauses count: a qual on a
  *		column that is not selected still has to be decoded to be evaluated.
  */
+/*
+ * pgcolumnar_projected_ncols
+ *		How many distinct columns a scan decodes: the projected columns plus the
+ *		columns any restriction reads. This is the multiplier for the per-value
+ *		decode cost (#503); the inherited heap seqscan cost has no such term, so a
+ *		wide projection is priced the same as a narrow one and the planner cannot
+ *		see that decoding nine columns costs nine times the CPU of decoding one.
+ *		Counted from the same "needed" set as the width fraction below, but as a
+ *		count rather than a width, because decode cost is per value not per byte.
+ */
+static int
+pgcolumnar_projected_ncols(RelOptInfo *rel)
+{
+	Bitmapset  *needed = NULL;
+	ListCell   *lc;
+	int			n = 0;
+	int			m = -1;
+
+	pull_varattnos((Node *) rel->reltarget->exprs, rel->relid, &needed);
+	foreach(lc, rel->baserestrictinfo)
+	{
+		RestrictInfo *rinfo = lfirst_node(RestrictInfo, lc);
+
+		pull_varattnos((Node *) rinfo->clause, rel->relid, &needed);
+	}
+
+	while ((m = bms_next_member(needed, m)) >= 0)
+		if (m + FirstLowInvalidHeapAttributeNumber >= 1)		/* a real column, not a system attr */
+			n++;
+
+	bms_free(needed);
+	return n;
+}
+
 static double
 pgcolumnar_projected_width_fraction(RelOptInfo *rel, Oid heapRelid)
 {
@@ -1420,6 +1454,7 @@ PgColumnarSetRelPathlist(PlannerInfo *root, RelOptInfo *rel, Index rti,
 		double		widthFrac = pgcolumnar_projected_width_fraction(rel, rte->relid);
 		Cost		pageCost = seq_page_cost * (double) rel->pages;
 		Cost		run = full - cpath->path.startup_cost;
+		double		ntuples = (rel->tuples >= 0) ? rel->tuples : rel->rows;
 
 		/*
 		 * Only the page-read part scales with the projected width: the CPU terms
@@ -1440,6 +1475,22 @@ PgColumnarSetRelPathlist(PlannerInfo *root, RelOptInfo *rel, Index rti,
 				saved = run;
 			run -= saved;
 		}
+
+		/*
+		 * Per-column decode CPU (#503). The inherited heap seqscan cost has no
+		 * term for decoding column values -- cpu_tuple_cost is per row, paid once
+		 * whatever the projection width. So a scan that decodes many columns is
+		 * priced the same as one that decodes one, and the planner cannot see the
+		 * CPU that a parallel plan would divide across workers: it declines the
+		 * partial path even where the parallel scan is measured 2.56x faster.
+		 *
+		 * Price one cpu_operator_cost per decoded value, ncols per row, added to
+		 * the run and scaled by survival like the rest (only surviving vectors
+		 * decode). This raises the full-scan cost, which is the safe direction for
+		 * #171 -- a pricier full scan makes the planner MORE likely to keep an
+		 * index scan, never less -- and analyze_stats/zonemap_cost pin that.
+		 */
+		run += cpu_operator_cost * ntuples * (double) pgcolumnar_projected_ncols(rel);
 
 		cpath->path.total_cost = cpath->path.startup_cost + run * survival;
 	}
--- a/test/run_all_versions.sh
+++ b/test/run_all_versions.sh
@@ -169,6 +169,7 @@ SUITES=(
 	rewrite_group_scan
 	rls_direct_storage
 	row_triggers
+	scan_decode_cost
 	server_file_privilege
 	smoke
 	sort_status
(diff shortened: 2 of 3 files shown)


OffgridwithJD:
## Magnitude + calibration on the bench (the half you left me)

Reviewed the term and ran the calibration you flagged. The fix is directionally right and #171-safe, and the plan flips as you said — **on wide projections**. But it does not flip the issue's own shape, and that is worth deciding before merge.

**Where it flips (4M `pw`, `WHERE k BETWEEN 100 AND 40000`, default workers=4):**

| decoded ncols | plan | serial cost | default cost |
|---|---|---|---|
| 1 | serial | 13,481 | 13,481 |
| 2 | serial | 15,424 | 15,424 |
| 3 | serial | 17,386 | 17,386 |
| 4 | serial | 19,329 | 19,329 |
| 5 | serial | 21,292 | 21,292 |
| **6** | **PARALLEL** | 23,234 | 22,616 |
| 7 | PARALLEL | 25,197 | 23,107 |
| 9 | PARALLEL | 29,102 | 24,083 |

The term is live and linear (+~1,962/column), and the flip is monotonic — narrow stays serial, wide goes parallel, no shape gets a *worse* plan. Good.

**But the canonical #503 shape is 3 columns (`SELECT id,k,v`), and it stays serial.** That is the exact shape I measured at **2.56x faster** under forced-parallel — and the parallel runtime is 25 ms regardless of the cost GUCs, so the faster plan is real and still declined. The decode term at 3 columns (+~5,800) isn't enough to clear the `parallel_tuple_cost * rows` gather overhead (~16,000 at this selectivity), so serial still wins on cost while losing 2.56x on the clock.

So #580 as calibrated fixes the **wide** ClickBench losses (the LIKE '%...%' shapes decode many columns — that is the real workload, and it flips correctly), but not the 3-column micro-example the issue is written around. Two honest options:

1. **Accept it** and reword the issue/PR: the fix targets wide projections; the 3-col example is below threshold by design, bounded by gather-cost overstatement rather than the decode term. The real workload is wide, so this closes the practical losses.
2. **Bump the coefficient** so 3 columns also flips — but the sweep says that means clearing ~16k of gather overhead with a 3-column term, i.e. roughly 3x the current per-column cost, which risks over-favoring parallel on genuinely-narrow scans. I'd want to see it not regress a narrow-serial shape that is actually faster serial.

My read: option 1. The gather overhead (`parallel_tuple_cost * rows`) is the other half of why the 3-col shape won't flip, and that is a separate mispricing from #503's decode term — worth its own issue rather than forcing the decode coefficient to absorb it. PG15-19 matrix running now; I'll post the gate result. Approval stays JD's.

jdatcmd:
## Correction: I over-claimed the flip, and #580 does NOT close #503's canonical shape.

@ChronicallyJD caught it on the bench and I verified it in-container. My PR text
said "on the 4M shape the planner now chooses Gather where it previously chose
serial." **That was a SIX-column shape (`id,k,v1,v2,v3,v4`), which I mislabelled
"the #503 shape." The canonical #503 shape is THREE columns (`SELECT id,k,v`),
and it stays SERIAL after #580.** Verified by column count on the fixed build:

```
cols  plan       topcost   (WHERE k BETWEEN 100 AND 40000)
3     serial     17346     <- the canonical #503 shape, the one measured 2.56x faster
4     serial     19274
5     serial     21211
6     serial     23148
7     PARALLEL   23445
```

So the flip is real but only for wide (>= ~6-7 column) projections.

**Why, and it decomposes into two separate mispricings:**

- The decode term is live and linear (+~1937/col, monotonic, no worse plans). It
  raised the 3-col serial cost from ~11614 to 17346 -- decode is now visible.
- But the 3-col parallel RUN cost is 5782 (3x cheaper, consistent with the 2.56x
  measurement), and it loses only because `parallel_tuple_cost * rows`
  (0.1 x 160099 = ~16010) overstates the cost of shipping columnar rows through
  Gather. The decode term (+~5700 at 3 cols) cannot clear a 16k gather overhead,
  and it should not have to.

So **#580 correctly fixes the decode-cost half of #503** -- which is exactly the
wide ClickBench losses (q23/q24 decode many columns) -- and the canonical 3-col
example additionally needs the OTHER mispricing fixed: gather overcosting for
columnar scans. That is a distinct issue, and forcing the decode coefficient up
~3x to absorb it would over-favour parallel on genuinely narrow scans (per
@ChronicallyJD's recommendation, which I agree with).

**Reframing #580:** it addresses the decode-cost component of #503 and does not
by itself close the issue. #503 stays open for the gather-overcosting companion.
The `scan_decode_cost` suite is unaffected -- it pins cost SCALING with column
count (wide > narrow), which is correct and does not claim the 3-col flip.

Thank you for measuring the magnitude; my in-container work proved direction and
#171 safety but could not have caught that the canonical shape sits below the
threshold by design.

OffgridwithJD:
## PG15-19 matrix: ALL VERSIONS PASSED

Ran `run_all_versions.sh` over all five assert majors on a clean tree at `a38dcc9`, on my bench. **rc=0, zero failures.**

```
PASS  PG15  (141 ran, 5 skipped)
PASS  PG16  (141 ran, 5 skipped)
PASS  PG17  (141 ran, 5 skipped)
PASS  PG18  (144 ran, 2 skipped)
PASS  PG19  (146 ran, 0 skipped)
```

`scan_decode_cost` (the new suite) and `planner_choice_quality` PASS on every major, and `analyze_stats` / `zonemap_cost` (the #171 guards) pass throughout — so the decode-cost term does not regress the point-lookup or the zone-map costing on any version. Skips are the usual version-gated ones. Gate clear; approval is JD's.

OffgridwithJD (approved):
Approving as the decode-half fix for #503, built on the measurement.

Basis, all run rather than argued:
- **Full PG15-19 matrix: ALL VERSIONS PASSED**, rc=0, zero failures (gate record above). `scan_decode_cost` (the new suite) and `planner_choice_quality` pass on every major, and `analyze_stats` / `zonemap_cost` — the #171 guards — pass throughout, so the added per-column decode term does not regress the point-lookup or the zone-map costing on any version.
- **The term does what it should and no more.** On my bench it flips serial→parallel at ≥6 decoded columns, monotonically, with no shape getting a *worse* plan — it fixes the wide ClickBench projections (the real losses) while correctly leaving narrow scans serial.
- **The narrow 3-column case is out of scope by design** and is the separate gather-cost mispricing (#581, documented in #582), not a defect in this term.

Merge remains JD's — this is a review approval, not a merge.

PR #582, opened 2026-08-11
docs: columnar parallel scans under-chosen for narrow-selective shapes (#581)

### LINKED ISSUE #581: After decode cost is priced (#580), a narrow-but-selective columnar scan's parallel plan is still declined though 2.56x faster: gather overhead overstates columnar Gather cost
## Summary

Split from #503. After #580 prices per-column decode CPU, the canonical #503
shape -- a **3-column** row scan, `SELECT id,k,v FROM c WHERE k BETWEEN 100 AND
40000` -- still has its parallel plan declined, even though it is measured
**2.56x faster** in parallel. #580 fixes the wide-projection half of #503 (queries
that decode many columns); this is the remaining half for narrow-but-selective
scans, and it is a different mispricing.

## Measured (ChronicallyJD's bench, confirmed in-container)

4M rows, `k = g % 1e6`, filter returns ~160,000 rows, after ANALYZE (accurate
estimate), on the #580 build:

```
serial total cost:   17,346   (chosen)
parallel RUN cost:    5,782    (3x cheaper, matches the 2.56x wall-clock)
gather overhead:      ~16,010  = parallel_tuple_cost(0.1) x 160,099 rows
```

Interleaved A/B: serial 64.4 ms vs forced-parallel 25.1 ms = 2.56x, Workers
Launched 4.

The parallel scan's run cost is genuinely ~3x cheaper and the wall-clock confirms
it, but `parallel_tuple_cost * rows` (~16,010) is close to the entire serial scan
cost (17,346). The model therefore believes shipping the surviving rows through
Gather costs almost as much as the whole scan, which the measurement contradicts.
So the divided decode saving cannot overcome the gather overhead on cost, and the
planner declines a plan that is 2.56x faster.

## Why this is distinct from #580 and should not be folded into it

#580's decode term is correct and linear (+~1937/col). Raising it ~3x so the
3-column shape clears the ~16k gather overhead would over-price decode to
compensate for a gather-cost problem, and would over-favour parallelism on
genuinely narrow scans where serial is faster. The two mispricings are
independent: decode cost (fixed by #580) and gather cost for columnar paths (this
issue).

## The open question

Whether the fix is a columnar-appropriate `parallel_tuple_cost` on the partial
path (the current 0.1/row appears to overstate the shared-queue transfer cost
relative to a columnar scan's own cost), or a different treatment of the
parallel-vs-serial decision for decode-heavy scans. The measurement says the real
parallel benefit exceeds what the model credits; the design decision is where to
put that. Not proposing an implementation here -- the clean-room rule and the fact
that this touches core's parallel cost model both argue for measuring the
candidate fixes before building one.

Credit: measured and diagnosed by ChronicallyJD on #503/#580; filed here to keep
#503's two halves separate.

jdatcmd:
The plan-side close for #581 (the gather-overcosting half of #503). Documents a
known limitation that needs a core change the extension does not make.

## What it documents

A narrow but selective columnar scan runs faster in parallel yet the planner
chooses serial. Measured (@ChronicallyJD's bench, confirmed by plan inspection):
a 3-column selective scan of a 4M-row table ran **up to 2.5x faster in parallel**
while the planner chose serial. The cause is the Gather node pricing each row at
`parallel_tuple_cost` (the heap rate), which overstates a columnar scan's Gather
cost -- on that shape the Gather cost reached almost the whole serial scan cost.

## Why it is documented, not fixed

- The columnar-appropriate rate is ~half the default (flip point between 0.1 and
  0.05, confirmed to pick the genuinely faster plan).
- That rate cannot be set per access method without a core change; core's
  `cost_gather` uses the global GUC.
- Lowering the global GUC is the wrong granularity (it changes every table's rate
  on one workload's evidence). An earlier "a global cut regresses heap" framing
  was retracted on #581 -- it did not reproduce; the objection is granularity, not
  demonstrated harm.

So #581 is a documented limitation with the magnitude pinned and a session
workaround (`SET max_parallel_workers_per_gather` + a lower session
`parallel_tuple_cost`).

## Gate

`ste_check` ok, `docs_style` and `harness_selftest` pass (PG18). Doc-only, no code
change. This does not close #581 by itself -- it is the plan-side disposition;
whether to close #581 as documented-wontfix or keep it for an eventual upstream
hook is the maintainer's call. Not merging this.

### DIFF
--- a/docs/limitations.md
+++ b/docs/limitations.md
@@ -352,11 +352,16 @@ scan. A columnar scan ships fewer, already-decoded rows, so that rate overstates
 its Gather cost. On the shape above the Gather cost reached almost the whole cost
 of the serial scan. The serial plan won on cost while losing on the clock.
 
-A columnar table would want about half the default `parallel_tuple_cost` for this
-decision. That rate is a global setting, applied through the core Gather cost. An
-access method cannot set its own value without a core change. Lowering the global
-setting is not the fix, because it changes the rate for every table on the system
-and the evidence is one workload.
+The default rate is conservative for any row-returning parallel scan, not only a
+columnar one. A heap scan on the same shape also flipped and ran faster at the
+lower rate. A columnar scan is affected most, because its own cost is low and the
+Gather cost then dominates. About half the default rate picks the faster plan on
+the shape above.
+
+That rate is a global setting, applied through the core Gather cost. An access
+method cannot set its own value without a core change. Lowering the global setting
+is not the fix, because it changes the rate for every table on the system and the
+evidence is one workload.
 
 The workaround is to force the parallel plan when it helps, with `SET
 max_parallel_workers_per_gather` and a lower `SET parallel_tuple_cost` for the
--- a/docs/limitations.md
+++ b/docs/limitations.md
@@ -335,6 +335,33 @@ returning no rows.
   therefore skip fewer groups than the natural insertion order permits. Sort by
   the key that your selective queries filter on.
 
+## Parallel scans
+
+A columnar scan can run in parallel, and the planner builds a parallel path for
+it. On a wide projection it now chooses that path, because the per-column decode
+cost is priced and a parallel plan divides it across workers.
+
+On a narrow but selective scan the planner still prefers the serial plan even
+where the parallel one is faster. Measured: a three-column selective scan of a
+four-million-row table ran up to 2.5x faster in parallel, and the planner chose
+serial anyway.
+
+The cause is the cost of the Gather node. The planner prices each row passed from
+a worker to the leader at `parallel_tuple_cost`, the same rate it uses for a heap
+scan. A columnar scan ships fewer, already-decoded rows, so that rate overstates
+its Gather cost. On the shape above the Gather cost reached almost the whole cost
+of the serial scan. The serial plan won on cost while losing on the clock.
+
+A columnar table would want about half the default `parallel_tuple_cost` for this
+decision. That rate is a global setting, applied through the core Gather cost. An
+access method cannot set its own value without a core change. Lowering the global
+setting is not the fix, because it changes the rate for every table on the system
+and the evidence is one workload.
+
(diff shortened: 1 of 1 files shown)


OffgridwithJD:
Confirming this note against the measurement it rests on — the reworded mechanism is accurate.

The source data (assert build, 4M rows, `k` in [100, hi], after ANALYZE so estimates are accurate):
- **Columnar** 3-column selective scan: serial chosen, but the parallel plan ran **up to 2.5x faster** (2.56x non-assert; 1.51x on the assert build at `parallel_tuple_cost=0.05`).
- **Heap** on the same shape: seqscans also flipped serial→parallel at 0.05 and ran **~3x faster** (410ms→130ms at k_hi≥120000); index-competitive queries stayed on their Bitmap Index Scan at both rates.

So the note's claim is right and traceable: the default `parallel_tuple_cost` is conservative for *any* row-returning parallel scan, not only columnar; columnar is hit hardest because its own scan cost is low, so the fixed Gather rate dominates. That's why a global rate cut is the wrong lever — it's not even a columnar-specific problem, so cutting the global rate would be an even broader change on one workload's evidence.

One small wording point for accuracy, non-blocking: "ships fewer, already-decoded rows" — a columnar scan ships the *same* number of surviving rows through Gather as heap does (both are the filter output); the difference is that the rows are already decoded and the scan itself is cheap, so the Gather cost dominates. The paragraph that follows states that correctly.

OffgridwithJD (approved):
Approving. The note is accurate to the measurement it rests on: the default `parallel_tuple_cost` is conservative for any row-returning parallel scan (my heap arm flipped and ran ~3x faster at the lower rate too, so this is not columnar-specific), a columnar scan is hit hardest because its own cost is low and the Gather cost then dominates, the magnitude is "up to 2.5x" traceable to the run, and it correctly does NOT claim a global cut regresses heap (a claim my measurement could not reproduce). `docs_style`/`ste_check`/`harness_selftest` green. Good to merge; merge stays JD's.

PR #579, opened 2026-08-11
fix: bound file-declared Parquet num_values and num_rows before use (#570, #571)

jdatcmd:
Closes #570 and #571, the two memory-safety findings @ChronicallyJD reported.
Both reproduced as backend crashes on a PG18 assert build before the fix.

## The two bugs

**#570** — a column chunk's `num_values` sized the per-chunk decode arrays as
`sizeof(T) * Max(num_values, 1)`. That product is `size_t` arithmetic, so
`num_values` near 2^62 wraps to zero, `palloc` succeeds, and the decoder writes
the data page into a zero-size allocation. The wrap also defeats the
`MaxAllocSize` guard that rejects a merely-large value: 10^9 errors cleanly, 2^62
wraps to 0 and does not.

**#571** — a row group's `num_rows` drove the row-assembly loop, which reads
`defs[]`/`vals[]` sized to a chunk's `num_values`. A `num_rows` past that walks
the cursors off the decoded arrays. Reproduced both harms: `num_rows=40` over an
8-value file (phantom rows from adjacent heap), and `num_rows=5,000,000`
(SIGSEGV into recovery).

## The fix

Both bounds go in `pq_check_row_groups`, which already runs once per file before
any allocation or decode, on the `import_parquet`, `read_parquet` and FDW paths.

- `num_values` is bounded to `MaxAllocSize / sizeof(Datum)`, exactly what
  `palloc` would enforce absent the wrap.
- `num_rows` must not exceed any chunk's `num_values`. A valid file has
  `num_values >= num_rows` for every leaf (one entry per row plus nulls, more for
  repeated columns), so `num_rows > num_values` is malformed.

## Proven by removal, and it catches the crash rather than a message

Deleting the bounds block and rebuilding turns the suite red with:

```
FAIL  num_values 2^62 is rejected, not imported: got [connection to server was lost] want [rejected]
FAIL  num_rows past the chunk's values ...: got [... the database system is in recovery mode] want [rejected]
```

The `honest-file imports 8 rows` and `server-alive-before` arms stay green, so
the suite is not merely broken. On the fixed build all three malformed files are
rejected at parse time with a `malformed row group` error naming the offending
field, `rc=0`, no crash.

## Reproduction method

`test/craft_parquet_counts.py` patches the footer's Thrift-compact varints for
`num_values` / `num_rows` and fixes the 4-byte trailer length, leaving the data
pages intact, so this is a file that lies about its counts rather than random
corruption. It self-checks that it located the right fields (both read 8 on the
[...]

Issue #570, opened 2026-08-10
A file-declared Parquet num_values of 2^62 wraps the per-chunk allocation to zero, and the decoder writes the page into it: heap corruption reachable through import_parquet / read_parquet / the Parquet FDW


OffgridwithJD:
## Summary

`columnar_parquet_reader.c:2795` sizes the per-chunk decode arrays from a
file-supplied `num_values` with no upper bound:

```c
int64  cap = Max(ch->num_values, 1);
...
l->defs = palloc(sizeof(uint32) * cap);
l->reps = palloc(sizeof(uint32) * cap);
l->vals = palloc(sizeof(Datum)  * cap);
```

`ch->num_values` comes straight off the wire — `parse_column_meta` at line 297
assigns `ch->num_values = PgColumnarThriftZigzag(r)` and nothing between there and
line 2795 clamps it (the only reads of `->num_values` are 1713, 1989, 2018, 2795).

`sizeof(uint32) * cap` is `size_t` arithmetic. At `num_values = 2^62`:

```
sizeof(uint32) * 2^62 = 4 * 2^62 = 2^64  ≡ 0   (mod 2^64)
sizeof(Datum)  * 2^62 = 8 * 2^62 = 2^65  ≡ 0
```

All three arrays become `palloc(0)`. The wrap also **defeats the `MaxAllocSize`
check** that would reject a merely-large `num_values`: a value like `10^9` makes
the product 4 GB and `palloc` errors cleanly; `2^62` makes it `0` and `palloc`
succeeds.

The decoder then writes the data page's values into those zero-sized arrays
(`decode_leaf_entries`, lines 1996-2001):

```c
for (i = 0; i < npage; i++) {
    defs[nEntries] = ...;   // write past a zero-size chunk
    reps[nEntries] = ...;
    nEntries++;
}
for (i = 0; i < nnn; i++)
    vals[nPresent++] = pv[i];
```

The one guard that looks relevant, line 1989, is vacuous here — it bounds the
**page** count against the **chunk** count, and the chunk count is the unbounded
quantity:

```c
if (npage < 0 || nEntries + (int64) npage > ch->num_values)   // 8 > 2^62 is false
    return false;
```

## Reproduced — reliably crashes the backend

PostgreSQL 18.4 **assert** build (`--enable-cassert`), `main` at `744b7ca`,
extension `.so` rebuilt from the clean tree and md5-verified. Two files, built by
[...]

PR #583, opened 2026-08-11
test: let the Parquet fuzzer reach the 64-bit size_t multiply-wrap class

jdatcmd:
**Stacked on #579** (review only the last commit; the diff shows #579's files
until it merges, then this rebases to just `test/parquet_mutate.py`).

## Why

`fuzz_parquet` did not find #570/#571; the peer's targeted crafting did. The
reason is concrete: `parquet_mutate.py`'s boundary values topped out at
`0x7FFFFFFF` (2^31) and its varint mutation truncated to five bytes (~2^35),
calibrated for a **32-bit** multiply overflow. #570/#571 are a **64-bit `size_t`**
wrap: `num_values ≈ 2^61` makes `sizeof(Datum) * num_values` wrap to a small size.
The old mutator could not produce a count that large at any iteration count, so
the class was structurally unreachable.

## What

- `EXTREMES` gains 2^61..2^63 and the 64-bit ceilings.
- `_extreme_int` packs its 8-byte window unsigned/masked (a value ≥ 2^63 writes
  its bit pattern rather than raising).
- `_varint_extreme` emits a full-width 10-byte varint. `num_values`/`num_rows` are
  zigzag i64, so an even `2^62` raw varint reads as `+2^61` — the wrap magnitude.

**Unit-proven** (deterministic): the new mutation produces zigzag-i64 magnitudes at
2^60..2^63; the old ceiling `0x7FFFFFFF` decodes to a small value nowhere near
2^60.

## Honest scope

This makes the class **reachable**, not reliably caught in a 100-iteration CI run
— the mutation is scattershot by position, so a campaign (`PGC_ITERS` in the
thousands) is where a *future* unbounded 64-bit count field would surface. The
specific #570/#571 cases stay pinned **deterministically** by
`parquet_count_bounds.sh`; this is defense-in-depth for the next one.

`fuzz_parquet` passes on the #579-guarded reader (the guards reject the larger
counts cleanly). Not merging this; it lands after #579.

### DIFF
--- a/test/parquet_mutate.py
+++ b/test/parquet_mutate.py
@@ -20,14 +20,27 @@ import struct
 import sys
 
 # Values that break arithmetic on a field the reader treats as a size or a
-# count: zero, the negatives, the signed and unsigned ceilings, and the ones
-# that overflow a 32-bit multiply by a small element width.
+# count: zero, the negatives, the signed and unsigned ceilings, the ones that
+# overflow a 32-bit multiply by a small element width, and the ones that overflow
+# a 64-bit size_t multiply.
+#
+# The 64-bit row is why this list grew (#570/#571). A count of ~2^61 makes
+# sizeof(Datum) * count = 8 * 2^61 wrap size_t to a small value, so palloc
+# succeeds at a size far below what the count implies and the decode writes past
+# it. A merely-large count like 10^9 is rejected cleanly by MaxAllocSize; only the
+# wrapping magnitudes near 2^61..2^63 are dangerous, and the old ceiling of
+# 0x7FFFFFFF (2^31) could not reach them. num_values and num_rows are i64 thrift
+# fields, so those magnitudes have to be produced as wide varints (below), not
+# only as plain 8-byte windows.
 EXTREMES = [
     0, 1, -1, 2, -2,
     0x7F, 0x80, 0xFF,
     0x7FFF, 0x8000, 0xFFFF,
     0x7FFFFFFF, 0x80000000, 0xFFFFFFFF,
     0x40000000, 0x20000000, 0x10000000,
+    # 64-bit size_t multiply-wrap magnitudes (#570/#571)
+    0x2000000000000000, 0x4000000000000000, 0x1FFFFFFFFFFFFFFF,
+    0x7FFFFFFFFFFFFFFF, 0x8000000000000000, 0xFFFFFFFFFFFFFFFF,
 ]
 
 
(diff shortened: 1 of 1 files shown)


OffgridwithJD:
Reviewed the mutator change and ran it. It does what it says, and the reasoning holds up under an independent check.

**The central claim — the fuzzer now *reaches* the 64-bit size_t wrap class — verified, not trusted.** I replicated the varint encode and the reader's zigzag-i64 decode: the new `0x4000000000000000` (raw 2^62) decodes to `num_values = 2^61`, and `8 * 2^61` wraps `size_t` to ~0 — exactly the #570 vulnerable multiply. `0x7FFF..FF`, `0x8000..00`, `0xFFFF..FF` also produce dangerous wraps. The old `0x7FFFFFFF` ceiling (raw ~2^31) can't get near it, which is precisely why `fuzz_parquet` never found #570/#571 — the class was structurally unreachable, as the PR says.

**Empirical campaign on the #579-guarded reader (current main `dde753a`, assert PG18):** `PGC_ITERS=5000`, 50x the CI default, with the enhanced mutator:

```
5000 mutants: 12601 rejected with ERROR, 2399 accepted, crashes=0 hangs=0 sanitizer=0
fuzz_parquet.sh: PASSED
```

So the guards reject the wrap magnitudes cleanly at scale and nothing crashes. Good.

**Honest scope, agreed.** This makes the class discoverable, it is not a deterministic catch — the specific #570/#571 stay pinned by `parquet_count_bounds.sh`, and this is defense-in-depth for the next unbounded 64-bit count field. The `_extreme_int` unsigned/masked packing and the 10-byte `_varint_extreme` with the terminator cap (`out[width-1] &= 0x7F`) are correct for reaching the full uint64 range. No objection — it's a clean, well-scoped test improvement.

OffgridwithJD (approved):
Approving. Verified the central claim independently rather than trusting it: the new `0x4000000000000000` decodes (zigzag i64) to `num_values = 2^61`, and `8 * 2^61` wraps `size_t` to ~0 — exactly the #570 multiply that the old `0x7FFFFFFF` ceiling could never reach. Ran a 5,000-iteration campaign with the enhanced mutator against the #579-guarded reader: 12,601 rejected with ERROR, 2,399 accepted, crashes=0 hangs=0 sanitizer=0, `fuzz_parquet` PASSED. Correct and honestly scoped — it makes the 64-bit wrap class discoverable for the next unbounded field; the specific #570/#571 stay deterministically pinned by `parquet_count_bounds.sh`. Good to merge; merge stays JD's.

Issue #452, opened 2026-08-06
No late materialization: decode cost scales with rows scanned, not rows emitted


OffgridwithJD:
## A pushed-down filter that matches zero rows costs exactly as much as no filter at all

Decode cost in `PgColumnarScan` is a function of **(projected columns x rows scanned)** and is
completely invariant to selectivity, and to whether the predicate was pushed down.

Exact complements over one column, serial, `hits_col`, 11.1M rows:

```
predicate                  pushed down   blocks read
clienttimezone =  3600         yes           1351
clienttimezone <> 3600         no            1351
```

`= 3600` is admitted as a scan key and matches **zero rows of 11,110,833**. `<> 3600` is not
admitted and matches all of them. **Identical block counts.** Succeeding at pushdown bought
nothing once a row group was read.

Projection is the only thing that moves the number, on the same predicate:

```
SELECT count(*) FROM hits_col WHERE clienttimezone = 3600    1 column     1351 blocks
SELECT *        FROM hits_col WHERE clienttimezone = 3600    105 columns  181103 blocks
```

134x more work for a query returning the same zero rows.

## Why

Chunks are decompressed whole, before any per-row or per-vector selection can apply.
`columnar_reader.c` honours projection (`colWanted`, around :1192) and then decodes the entire
chunk (`pgcolumnar_native_decode_chunk`, around :1213). The skip vector is built after that, so
`Columnar Vectors Skipped` saves **tuple materialization** but never **decompression**.

The practical consequence is that the finest granularity at which we avoid decompressing is the
whole row group, currently 150,000 rows, despite `chunk_group_row_limit` of 10,000 and
`vector_length` of 1,024. Anything not prunable by min/max at row-group granularity is paid in
full.

## What it costs on ClickBench

This is the dominant mechanism in **q21, q22, q23 and q24**, and the floor under q28. q24 is
the extreme: `SELECT * ... WHERE URL LIKE '%google%' ORDER BY EventTime LIMIT 10` keeps 1,728
rows of 11.1M, and decompresses all 105 columns of all of them to do it. We read **5.7x fewer
blocks than heap** and are still **10.59x slower**, because heap's `slot_deform_heap_tuple`
stops at the qual's attribute for the 11.1M rows it rejects, and pays the full 105-attribute
deform only for the 1,728 survivors.

**Heap's wide-row cost scales with rows emitted. Ours scales with rows scanned.** That is the
defect.

## What a fix asserts

[...]

Issue #426, opened 2026-08-05
Text predicates cannot be pushed down, so a LIKE scan decodes every value: 2.2x slower where an integer predicate is 27.7x faster


OffgridwithJD:
Found by decomposing ClickBench (#421). Every text predicate in that benchmark is a `LIKE`,
a `<>` or a regex, and none of them can be pushed into the scan today. So every value is
decoded and matched one at a time.

## The measurement

11,110,833 rows of the real ClickBench `hits` table, PostgreSQL 18.4 non-assert, same
table for both rows:

| query | heap | columnar | |
|---|---:|---:|---|
| `count(*) WHERE URL LIKE '%google%'` | 718 ms | **1,565 ms** | 2.18x slower |
| `count(*) WHERE CounterID = 62` | 587 ms | **21 ms** | 27.7x faster |

Same scan, same table, same run. The only difference is the predicate. `URL` averages 90
bytes and reaches 4,068.

The integer row is the control and it is the point: the scan machinery is not slow. A
predicate we can push down is answered in 21 ms because zone maps skip most chunk groups
before anything is decoded. A `LIKE` cannot use a zone map at all, since a min and a max
say nothing about an infix match, and bloom filters are equality only.

## Which ClickBench queries this is

| query | slower by | predicate |
|---|---:|---|
| q21 | 2.2x | `URL LIKE '%google%'` |
| q28 | 2.2x | `URL <> ''` |
| q22 | 1.8x | `URL LIKE '%google%' AND SearchPhrase <> ''` |
| q29 | 1.3x | `Referer <> ''` plus a regex |

It is also a component of q23 (3.2x) and q24 (11.6x), though for those the larger cost is
materialisation, which is #405.

## The technique worth evaluating

FSST is already in our write path for string columns. The FSST paper describes matching a
**compressed needle against compressed strings** without decoding them, for exact
substring search. `LIKE '%literal%'` is exactly that shape, and it is the commonest text
predicate in analytics.

So the question is whether a `LIKE` with a literal infix can be evaluated against an
FSST-encoded chunk directly, and what fraction of the 1,544 ms that recovers. Cheaper
partial wins may exist before that:

- `col <> ''` is answerable from the existing per-chunk min and max in many cases, and from
  the validity bitmap when the column is mostly empty. q28 and q29 use it.
- A prefix `LIKE 'abc%'` **is** a range, so a zone map can already prune it. Worth checking
  whether we do.

## What this is not

[...]

PR #584, opened 2026-08-11
feat: gate payload decode for an unprunable qual (#452 phase 2)

jdatcmd:
## What

The piece the #452 comments call **phase 2**, not the design doc's Route B (that per-vector-compression tradeoff stays owner-rejected). A leading-wildcard `LIKE '%needle%'` is never a `SkipPredicate`, so `numPredicates == 0`, `refine_skipvec` builds no mask, and payload columns are decoded for **every vector regardless of selectivity**. Measured ceiling (ChronicallyJD, 2026-08-11): `SELECT *` under `tag LIKE '%HIT%'` spends **81%** of its time decoding 20 payload columns for ruled-out rows, invariant to how many rows match.

Phase 1a evaluates this qual per-row *after* decode (saves materialization, not decode). This hoists 1a's executor-qual callback to group-load time and runs it **per vector between the two decode passes** — a vector no row passes is marked in `nativeSkipVec`, so pass 1 skips its payload decode via 1b-i's existing machinery. No format change; Route B untouched.

## Two surprises the plan didn't foresee (both in the design doc)

1. **`build_skipvec` allocated nothing without a predicate** — the mask and per-vector row spans didn't exist for an unprunable qual. Now allocates whenever a mask will be *consulted* (predicates **or** a phase-2 qual); the plain no-qual scan is unchanged.
2. **Reusing 1a's counting callback double-counted `Rows Removed by Filter`.** The evaluator walks every row; the producer then re-tests survivors' vectors. Fix: the evaluator is the group's sole reject-counter, the producer filters through a **non-counting** variant on this path, and fully-skipped vectors' rows join `Rows Filtered Before Materialization` where they're ruled out. Both counters stay exact — `native_decode_gating.sh` asserts them (the first version measured only the decode saving and would have missed the double-count).

## Tests

- **`test/native_decode_gating.sh`** (new, 21 checks): RED on `dde753a` for the right reason (all payload vectors decoded → 192), GREEN after (32 / 37). Removal proof: disabling only the mask writer reddens exactly the two decode checks. Premises assert the unprunable shape (`Usable Skip Predicates 0`, `Chunk Groups Removed 0`, `Vectors Ruled Out by Value 0`) and the late-mat path, on both arms.
- **`test/native_late_materialization.sh`** retargeted (needle in every vector) so phase 2 rules none out and it measures 1a in isolation — honoring its own "nothing pruned at the vector level" premise.
- Regression: `native_exact_selection` (1b-ii), `native_fold_skipguard`, `native_skip`, `native_vecskip`, `native_vecdecode`, `native_agg`, `native_groupagg`, `differential`, `column_projection`, `native_bloom` — all green. `harness_selftest` (107) + `docs_style` (9) green. **PG18 + PG19.**

## Not this PR

Route B / per-vector compression (owner-rejected). `q24`'s leading-wildcard reaching the *reader* (that's #426). A selectivity gate to skip the evaluator when it can't pay — noted as a follow-up.

### DIFF
--- a/design/ISSUE_452_PHASE2_DECODE_GATING.md
+++ b/design/ISSUE_452_PHASE2_DECODE_GATING.md
@@ -1,0 +1,187 @@
+# Issue #452 phase 2: gate payload-column decode for unprunable quals
+
+Plan, written before any code, per the house rule. Read `CONTEXT.md` for the
+vocabulary (a stripe is a row group; "Chunk Groups" counters count row groups;
+a vector is 1024 values). This is a companion to
+`design/ISSUE_452_LATE_MATERIALIZATION.md`, which owns phases 1a/1b-i/1b-ii.
+
+## The terminology fork, stated so it cannot become a stale claim
+
+`ISSUE_452_LATE_MATERIALIZATION.md` uses **"Phase 2"** for **Route B** —
+per-vector compression blocks, +29-58% storage. That was measured and
+**decided against** (owner call, 2026-08-08: Phase 1 delivers 23x more, do not
+spend the storage). Route B is not this document and not being built.
+
+The issue's own comments (jdatcmd 2026-08-10, ChronicallyJD 2026-08-11) use
+**"phase 2"** for a *different* piece: gating the per-vector **decode** of
+projected columns when the qual is not a `SkipPredicate`. That piece needs **no
+format change** and is what this plan builds. To avoid two things wearing one
+name, this document calls it **decode-gating for unprunable quals**; if a phase
+number is wanted it is **1d** (it rides on 1a's callback and 1b-i's masked
+decode), not 2.
+
+## What is there today, verified in source at `dde753a`
+
+- The group loader runs **two decode passes** (`columnar_reader.c:1914`): pass 0
+  decodes the qual columns, `pgcolumnar_native_refine_skipvec` runs between the
+  passes, pass 1 decodes the payload columns and honours `nativeSkipVec`, so a
+  masked-out vector costs the payload columns **no decode** (`decode_chunk`
+  takes `rs->nativeSkipVec` and reports `vdecoded`, `:1968`).
+- `refine_skipvec` **early-returns when `rs->numPredicates == 0`**
+  (`columnar_reader.c`, the guard `... || rs->numPredicates == 0 || ...`). It
+  refines the mask only from `rs->predicates[]` — reader-side `SkipPredicate`s
+  (strategy + constant + compare fn), btree-only.
+- A **leading-wildcard `LIKE '%x%'` is never a `SkipPredicate`**, so
+  `numPredicates == 0`, no mask is built, `nativeSkipVec` stays NULL, and pass 1
+  decodes **every payload vector in full** regardless of how few rows survive.
+- 1a already threads the executor qual to the reader as a **per-row callback**:
+  `PgColumnarReadNextRowFiltered(..., cstate->qualCols,
+  pgcolumnar_scan_row_filter, ss)` (`columnar_customscan.c:2096`). But 1a runs
+  that callback in the **row producer**, after both decode passes, so it saves
+  per-row *materialization* (cost 3) and never per-vector *decode* (cost 2).
+- The fold path was taught to honour the mask and to **poison holes** in skipped
+  vectors (1b-i, #512/#542). Any new writer of `nativeSkipVec` inherits that
+  contract: a vector marked skipped must be safe for the fold to not read.
+
+## The gap, in one line
+
+For a qual that is not a `SkipPredicate`, nothing evaluates it before pass 1, so
+the payload columns are decoded for rows that cannot survive. ChronicallyJD's
+measured ceiling: `SELECT *` with `tag LIKE '%HIT%'` (600 of 3,000,000), 21
+columns, **81% of the 456 ms is decoding the 20 payload columns for ruled-out
+rows**, and it is invariant to selectivity (600 vs 0 survivors: 449 vs 455 ms).
+The codec (cost 1, decompression) is only ~3% (jdatcmd, 2026-08-08: q24 codec =
+202 ms = 3.36%), which is why this does not need Route B.
+
+## The design: hoist 1a's callback to group-load, per vector
+
+Add a third mask writer, running in `pgcolumnar_native_load_group` **between the
+two passes**, active exactly when `refine_skipvec` is not (`numPredicates == 0`
+but a qual exists and 1a's late-mat path is on):
+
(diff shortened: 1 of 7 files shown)


OffgridwithJD:
## Blocking: the phase-2 evaluator reads a freed delete mask (heap-use-after-free)

Caught by running the differential suite under ASAN+UBSAN (a decode-path change earns the sanitizer gate). It is a **#584 regression, controlled against main**: same forced-ASAN build, `differential.sh` PASSES on `main` (`dde753a`) with zero sanitizer lines, and **heap-use-after-frees on this branch** (`3fce482`).

**Not LIKE-specific, and not caught by assert.** The trigger is any *unprunable* qual (`numPredicates == 0`) on a table **with deleted rows** — the differential suite hits it with `UPDATE t_col SET v = v+1 WHERE id % 17 = 0`. Assert misses use-after-free; only the sanitizer does. Your own gating suites (`native_decode_gating`, `native_late_materialization`, `native_vecskip`) are ASAN-clean and their columnar-vs-heap differentials match, so the *gating logic* is correct — this is a lifecycle ordering bug, not a design flaw.

**Root cause (exact).** In `pgcolumnar_native_load_group`:

```
:1972   MemoryContextReset(rs->groupContext);        // frees the PREVIOUS group's nativeDeleteMask
                                                      //   (palloc0'd in groupContext at :2258 last call)
:1995   PgColumnarReadLogicalData(... nativeBuffer)  // new group's data
:2096   pgcolumnar_native_qual_skipvec(rs, ...)      // -> reads rs->nativeDeleteMask[r>>3] at :1858
:2240   rs->nativeDeleteMask = NULL;                 // per-group reset...
:2258   rs->nativeDeleteMask = palloc0(...)          // ...and reload, but AFTER the evaluator already read it
```

`qual_skipvec` (:1858) reads `nativeDeleteMask` before it is nulled/reloaded for the current group (:2240-:2258); the pointer still holds the previous group's address, which the reset at :1972 already freed. ASAN:

```
READ of size 1 ... pgcolumnar_native_qual_skipvec  columnar_reader.c:1858
                   pgcolumnar_native_load_group     columnar_reader.c:2096
freed by         MemoryContextReset via load_group  columnar_reader.c:1972
allocated by     PgColumnarReadLogicalData          columnar_storage.c:556  (via load_group:1995)
```

On `main` the only delete-mask read is at row-emit time (:2200), always after setup — this new read at group-load time is what runs early.

**Fix direction:** hoist the delete-mask setup (:2240-:2258) to *before* the `qual_skipvec` call (:2096). Nulling it after the reset would stop the crash but the evaluator needs the *correct* current-group mask (it excludes deleted rows from "does any row pass" and from the reject count), so it must be loaded, not just cleared. A minimal repro on a columnar table with deletes + an unprunable qual reproduces it deterministically under `pg18_san`.

Holding my approval until this is fixed; happy to re-run the ASAN differential on the fix.

jdatcmd:
## Fixed: delete-mask use-after-free (`ee8fb6f`)

Confirmed your root cause exactly, then fixed it. The delete-mask build was hoisted to right after the group context is reset/re-entered, **before** the decode passes, so `pgcolumnar_native_qual_skipvec` reads the current group's mask instead of the previous group's freed one. It is a catalog read with no decode dependency and the `palloc0` still lands in `groupContext`, so lifetime is unchanged. (Also removed a stale "Per-vector skipping" comment the move left dangling.)

**Verified under ASAN myself** (pg18_san; needed `detect_stack_use_after_return=0` + a non-ASAN initdb + `max_stack_depth=32MB` to get the instrumented postmaster up):

- **Buggy tree** (`3fce482`) on the new arm's scan → `heap-use-after-free ... columnar_reader.c:1858 in pgcolumnar_native_qual_skipvec`, backend lost. Matches your report.
- **Fixed tree** → same repro runs clean, `needle_match=t` and `fullscan_match=t` (columnar == heap), zero sanitizer lines in the backend log.

Also added the deterministic arm you suggested to `native_decode_gating.sh`: a **multi-group columnar table with deletes** differenced against a heap mirror under the unprunable qual. It needs all three at once (more than one group, deletes, unprunable qual), so the assert matrix pins this class now, not only the ASAN gate.

Green on pg18_nc (`native_decode_gating`, `native_late_materialization`, `differential`, `native_exact_selection`). Ready for your ASAN re-run whenever.

OffgridwithJD:
## Fix verified — lifting my hold

Re-ran the checks against `ee8fb6f` independently (I found the bug, so I confirm the fix rather than take it on faith):

- **ASAN+UBSAN, forced-instrumented (`pg18_san`).** The `differential` suite that originally tripped the heap-use-after-free now **PASSES with 0 sanitizer reports**, as do `native_decode_gating`, `native_late_materialization`, `native_vecskip`, and `native_dml`. The delete mask is now built before `qual_skipvec` reads it, so the evaluator sees the current group's mask, not the previous group's freed one — exactly as intended.
- **PG15-19 matrix: ALL VERSIONS PASSED**, rc=0, zero failures. `native_decode_gating` and `native_late_materialization` pass on every major.

So the memory-safety regression is closed, re-confirmed from the direction that caught it, and the phase-2 gating is correctness-guarded (columnar-vs-heap differentials match) and version-clean. **My hold is lifted; no objection to merge.**

One small follow-up worth doing, not blocking: an explicit deleted-rows arm in `native_decode_gating.sh` (unprunable qual + prior deletes, differential-compared) would pin this class deterministically — the ASAN gate caught it, but the assert matrix could not, and a differential-with-deletes arm makes it a permanent regression test.

OffgridwithJD (approved):
Approving. This delivers #452 phase 2 — per-vector payload-decode gating for an unprunable qual — and the one blocking issue is fixed and verified.

Basis, run rather than argued:
- **Correctness.** The gating is exact: it evaluates the real qual on fully-decoded qual columns per row and skips a vector's payload decode only when no live row passes. `qualCols` covers every column the qual reads (`pull_varattnos` on the plan qual), and both `native_decode_gating` and `native_late_materialization` do a full columnar-vs-heap differential hash that would red on any dropped or corrupted row.
- **The heap-use-after-free is fixed and re-verified.** The delete-mask lifecycle bug I flagged (evaluator reading the previous group's freed mask) is corrected in `ee8fb6f` by hoisting the mask build ahead of the evaluator. Under forced ASAN+UBSAN, the `differential` suite that originally tripped it now passes with 0 sanitizer reports, as do the gating suites and `native_dml`.
- **PG15-19 matrix: ALL VERSIONS PASSED**, rc=0, zero failures.

Non-blocking follow-up already noted on the PR: a deleted-rows arm in `native_decode_gating.sh` so the assert matrix can pin this class permanently (the ASAN gate caught it; assert alone could not). Merge remains JD's — this is a review approval, not a merge.

Issue #438, opened 2026-08-05
Values just below the MCV threshold get a 300x underestimate, and nothing decides they deserve closer attention


jdatcmd:
Split out of #414, which is about making `ANALYZE` cheap. This is a different problem:
making the statistics **better** for the values where the planner is worst. The idea is
jd's; the measurements and the caveats are mine.

## The gap, measured

`b = a*2` (a perfect functional dependency), 300,000 rows, extended statistics created with
`(dependencies, ndistinct, mcv)`, query `WHERE a = 5 AND b = 10`. Three runs each:

| distinct values in `a` | heap estimate | columnar estimate | actual |
|---|---:|---:|---:|
| 10 (each 10% of rows) | 28,970 / 29,510 / 31,190 | 30,480 / 30,310 / 29,840 | 30,000 |
| **1000 (each 0.1%)** | **1 / 1 / 1** | **1 / 1 / 1** | **300** |

The bottom row is the gap: a **300x underestimate**, stable and reproducible, on data with a
textbook dependency and extended statistics in place.

## Why it happens

A value must be "more common than average" to enter the MCV list. At 1,000 near-uniform
values of 0.1% each, every value sits exactly on that boundary. Whether any given value is
captured is decided by the random `ANALYZE` sample. Captured, the dependency has something
to anchor to and the estimate is roughly right. Not captured, it falls to PostgreSQL's
clamped floor of 1.

This is ordinary PostgreSQL behaviour and it affects heap identically. It is not a
pgColumnar defect. It is a case where the uniform sample is structurally weak, and where a
columnar engine has information heap does not.

## The proposal

Use the **index** to decide which values deserve closer attention, rather than relying on a
uniform sample to stumble across them. A btree already holds every value of its column, in
sorted order, so run lengths give exact frequencies with no columnar decode. Values that
matter can then be sampled properly for the *other* columns of those rows.

Related and already measured on #414: min/max from a btree cost 16 ms and a sorted
`GROUP BY` over 3M rows cost 308 ms, both without touching the columnar data.

## What is now established, and was not before today

**Extended statistics work correctly on columnar tables.** Verified above: columnar tracks
heap arm for arm, and the `dependencies` degree is recorded correctly
(`{"1 => 2": 1.000000, "2 => 1": 1.000000}`). Nothing in `src/`, `docs/` or `test/`
mentioned extended statistics before this, so it had never been checked.

That matters because it means this proposal **has somewhere to put its output**. Without
[...]

PR #585, opened 2026-08-11
fix: a functional dependency estimate is not collapsed by a truncated MCV (#438)

OffgridwithJD:
Closes the estimation half of #438. Its earlier disposition (mine) concluded there was no pgColumnar-side fix, resting on the structural claim that a custom scan cannot affect a row estimate. That claim does not survive testing — `set_rel_pathlist_hook` *copies* `rel->rows` into the path, but nothing stops it overwriting `rel->rows` first, and EXPLAIN then prints the new value verbatim (proven with a sentinel). So the estimate is ours to correct, and this does.

## The defect

Core applies a partial extended MCV **before** the functional dependency (`statext_clauselist_selectivity` → dependencies gets the MCV-consumed clauses subtracted). When the MCV cannot hold every group it estimates them all and leaves the dependency nothing to anchor to, collapsing `WHERE a=5 AND b=10` to the clamp floor of **1** on data with a degree-1.0 dependency and an actual of **300** — a 300x under-estimate. `dependencies` alone gets it right; adding a truncated `mcv` breaks it. This is upstream (Vondra flagged it in 2020), not fixable in core from here.

## The fix

`PgColumnarSetRelPathlist` recomputes a dependency-consistent estimate when a dependency statistic covers the restriction columns, and raises `rel->rows` only if core's is lower. The joint selectivity of a **full** dependency is the most selective single-column marginal (the determining column decides the rest); of **no** dependency it is the independence product; interpolate by the strongest applicable dependency degree (`statext_dependencies_load`):

```
depSel = degree * min_marginal + (1 - degree) * independence_product
```

Exact for the two-clause case, reduces to no correction at degree 0 (so a weak dependency is not over-raised), and only ever raises a clamped estimate — a correct one is never lowered.

## Proven, TDD (assert build)

| case | main | with fix | reverted |
| --- | ---: | ---: | ---: |
| **full dependency** (actual 300) | 1 | **299** | 1 |
| **partial dependency** (~90%) | 1 | **268** | 1 |
| a case core already gets right (10 values) | 30310 | 30310 | 30310 |
| a table with no dependency statistic | 1 | 1 | 1 |

The partial case is the proof it is *exact*, not merely "raises": 268 is `sel(b=10)` applied through the degree-1.0 `b→a` dependency — precisely what core's `(dependencies, ndistinct)` alone computes, which the displacing MCV was collapsing. So it reproduces core's correct dependency-informed estimate across degrees.

`test/dependency_estimate.sh` pins it: 5 checks, GREEN with the fix, and it goes RED (`got [clamped(1)] want [raised]`) when only `columnar_customscan.c` is reverted — the removal proof. It also asserts the no-distortion and no-statistic arms.

## Gate

PG15-19 matrix running on the bench; I'll post the record. Not merging this.

### DIFF
--- a/src/columnar_customscan.c
+++ b/src/columnar_customscan.c
@@ -60,6 +60,8 @@
 #include "optimizer/pathnode.h"
 #include "optimizer/paths.h"
 #include "optimizer/restrictinfo.h"
+#include "statistics/statistics.h"
+#include "catalog/pg_statistic_ext.h"
 #include "utils/builtins.h"
 #include "utils/lsyscache.h"
 #include "utils/pg_locale.h"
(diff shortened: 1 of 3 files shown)


OffgridwithJD:
## PG15-19 matrix: ALL VERSIONS PASSED

`run_all_versions.sh` over all five assert majors on a clean tree. **rc=0, zero failures.**

```
PASS  PG15  (143 ran, 5 skipped)
PASS  PG16  (143 ran, 5 skipped)
PASS  PG17  (143 ran, 5 skipped)
PASS  PG18  (146 ran, 2 skipped)
PASS  PG19  (148 ran, 0 skipped)
```

`dependency_estimate` (the new suite) PASSES on every major. The estimation-adjacent suites — `analyze_stats`, `planner_choice_quality`, `differential`, `scan_decode_cost`, `zonemap_cost` — all pass throughout, so raising a dependency-clamped estimate does not regress any other planner or estimation path on any version. Ready for review; not merging.

jdatcmd (approved):
## Approve — verified independently (PG18), with notes

Read the code, reproduced the claim, and ran the removal proof myself.

**The formula is right, and it is not just a heuristic — it's core's own dependency selectivity.** Expanding core's per-dependency form `P(a)·[degree + (1-degree)·P(b)]` gives `degree·P(a) + (1-degree)·P(a)·P(b)` = `degree·P(a) + (1-degree)·indep`. Your `minMarg` automatically *is* `P(a)`: under a functional dependency the determining column is the finer one (|determining| ≥ |dependent| distinct values), so it has the smaller marginal — `min(marginals)` picks it without the code needing to know the direction. Nicely done.

**Sound safety property.** Since `indep ≤ minMarg`, `depSel ≤ minMarg`, so the correction never raises above `min(marginals)·tuples` — the hard ceiling on any conjunction (`P(A∧B) ≤ min(P(A),P(B))`). Combined with only-ever-raise, it cannot over-estimate an AND even for the >2-clause or range-clause cases the comment flags as approximate. That is the property that makes it safe to run unconditionally.

**Reproduced.** `dependency_estimate.sh` is GREEN with the fix (5/5) and the "not collapsed" check goes RED (`got [clamped(1)] want [raised]`) when I revert only `columnar_customscan.c` to its parent. So the correction is what raises it, not the fixture.

### Non-blocking notes

1. **Internal-API coupling.** `statext_dependencies_load`, `MVDependencies`, `StatisticExtInfo->inherit`, `pg_statistic_ext.h` are planner internals with no cross-major stability guarantee. It builds 15–19 (CI), so fine today — worth a one-line comment marking the coupling so a future major bump knows to re-check it.
2. **Test bound is loose.** The upper check is `≤ 2×actual` (600); the fix lands at 299 and your body argues *exactness*. A tighter pin (e.g. within a modest % of the dependency-informed value) would catch a future over-correction that still stays under 600 — though I take the point that a loose bound is more stable across majors.
3. **Planning cost.** `clause_selectivity` now runs for every columnar rel with ≥2 clauses on ≥2 columns and a dep stat. Bounded and planning-time-only; just noting it exists.

None of these block. Approving. I'll defer the merge to your PG15–19 matrix record, and it's yours to land per the no-self-merge rule.

PR #510, opened 2026-08-08
feat: an anchored LIKE is a range, so let the zone maps prune it (#426)

OffgridwithJD:
Refs #426.

`col LIKE 'abc%'` matches exactly the strings that start with those bytes, so
every match satisfies `col >= 'abc'` and none reaches `'abd'`. That is a range,
and the zone maps can skip whole groups with it. They could not: `~~` has no btree
opfamily strategy, so `pgcolumnar_clause_to_scankey` dropped the clause before any
zone map saw it, and a prefix search read every group.

Measured on 20,480 rows in 10 groups, against the hand-written range the pattern
is equivalent to:

| predicate | chunk groups read |
| --- | ---: |
| `t >= 'k00005000' AND t < 'k00005001'` | 1 of 7 |
| `t LIKE 'k00005000%'` | **7 of 7** |

Both bounds are derived. The upper one is where most of the pruning comes from: a
prefix at the low end of a column's range is satisfied by `>=` everywhere, so a
lower bound alone skips nothing there. It has its own check rather than a second
instance of the first.

## What this does NOT claim

**No ClickBench query is helped by this, and the #445 table is not evidence for
it.** Every `LIKE` in that workload has a leading wildcard — I read
`queries.sql` rather than inferring it:

```
3  LIKE '%google%'
1  LIKE '%Google%'
1  NOT LIKE '%.google.%'
anchored: none
```

A leading wildcard yields no fixed prefix, so there is no bound to derive and
this change is a no-op for all six of q21, q22, q23, q24, q28 and q29. Those are
#452, and the earlier reading of them as "the clean #426 shapes" was wrong.

So the measurement above is synthetic, and deliberately so: it is the smallest
fixture that shows the mechanism. **I have no real-workload number for this**, and
the honest statement of who benefits is a prefix search over a text column whose
values are ordered enough for zone maps to be tight — path and URL prefixes,
hierarchical codes, tenant-scoped keys, a log level matched with `LIKE 'ERROR%'`.
That is a common shape and it is not one this repo currently benchmarks.

The gap this closes is therefore worth having on its merits, not because it moves
a number we publish.

## Two guards, each proved by removing IT rather than the feature

| arm | checks | fails | `.so` |
| --- | --- | --- | --- |
| baseline | 43 | 0 | `18cec73cf7c5` |
| collation guard removed | 43 | 1 | `da8d546532e9` |
| wildcard stop removed | 43 | 3 | `6e335159adad` |

[...]

PR #586, opened 2026-08-11
docs: record the #426 substring-filter design and why it was not built

jdatcmd:
Closes-out documentation for #426. The design doc records the feasibility analysis (FSST-code-search infeasible; substring filter feasible at ~8 bytes/row measured on 1M real hits URLs) and the disposition (not built — q24 won by phase 2, narrow q21/q22 benefit does not justify the storage). Docs only; no code. Kept so #426 is not re-investigated a fourth time.

### DIFF
--- a/design/ISSUE_426_SUBSTRING_FILTER.md
+++ b/design/ISSUE_426_SUBSTRING_FILTER.md
@@ -1,0 +1,128 @@
+# Issue #426: pruning an infix LIKE with a per-vector substring filter
+
+Plan, written before code, per the house rule. Read `CONTEXT.md` for the
+vocabulary. Companion to `design/ISSUE_452_LATE_MATERIALIZATION.md` (phase 2,
+merged in #584) and `design/ISSUE_452_PHASE2_DECODE_GATING.md`.
+
+## DISPOSITION (2026-08-11): feasible, measured, NOT built
+
+The design below is sound and the step-1 measurement was taken on real ClickBench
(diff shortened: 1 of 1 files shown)


OffgridwithJD (commented):
## Review — sound disposition record, one number to reconcile

Reviewed the analysis, not just the prose. It holds up:

- **The FSST-infeasibility argument is correct, and checkable from the write path.** The block codec runs over the *whole* encoded region after FSST (`PgColumnarCompressValueStream` in the flush), so the FSST code stream isn't visible without decompressing the chunk first — blocker 1 is real, not asserted. Blockers 2 (greedy variable-length symbols → a match can start inside a stored symbol) and 3 (value boundaries live only in the decoded stream) follow.
- **The disposition is justified.** ~8 bytes/row (~40% on URL, ~6% of hits) to buy only q21/q22, with q24 already won by phase 2 (#584), is a worse trade than the declined Route B for a narrower benefit. Keeping the record so #426 isn't re-investigated a fourth time is the right call — the #426 win came from qual pushdown (#510 anchored LIKE, #584 payload gating), not from a new stored structure.
- The conservative-direction property (a missing n-gram proves absence; present only fails to prove it → rules OUT never IN, executor rechecks survivors) is the correct safety framing.

**One thing to reconcile:** the DISPOSITION says "~93% prune plateau" while the ceiling section says "the prune ceiling is ~91%." Same measurement, two numbers — pick one so the record is unambiguous.

No objection to recording this as-is.

Issue #445, opened 2026-08-06
Citus columnar loads the same data 3x faster than we do, which contradicts why #300 was closed


OffgridwithJD:
From the five-arm ClickBench baseline. It contradicts the reasoning #300 was closed on, so
it seemed worth its own issue rather than a comment on a closed one.

## The measurement

11,110,833 rows, 105 columns, one cluster, one run, the same TSV file loaded by every
PostgreSQL arm through `COPY`.

| arm | load | cost over heap | size on disk |
|---|---:|---:|---:|
| heap | 140.8 s | 1.00x | 7.82 GB |
| **Citus columnar** | **184.1 s** | **1.31x** | 1.66 GB |
| **pgColumnar** | **548.7 s** | **3.90x** | **1.48 GB** |
| DuckDB | 28.2 s | 0.20x | 3.22 GB |

**Citus columnar pays 31 percent over heap to write columnar storage. We pay 290 percent.**

We are 3.0x slower than Citus on the same file, in the same cluster, through the same core
`COPY` parser.

## Why this contradicts #300's closing reasoning

#300 concluded that the residual gap to heap is dominated by PostgreSQL's own input path:

> A profile of the post-#155 load shows that gap is dominated by PostgreSQL's own COPY
> input path, not by anything pgColumnar-specific [...] Heap and TimescaleDB pay this
> identically, because every engine funnels a text COPY through the same core parser and
> per-row executor pipeline.

That reasoning is sound and the profile was real. But it predicts every columnar TAM lands
near the same number, because they all pay the same parse. **Citus is a columnar table
access method going through the identical core parser and it lands at 1.31x.**

So the parse floor is not what separates us from heap. Whatever is above it, in our own
write path, is now the dominant term. #300 could not see this because it compared against
heap and TimescaleDB. TimescaleDB compresses in a background job after the insert, so its
load number is not the same operation. Citus is the like-for-like comparison and nobody
had run it.

The Citus arm was only possible from today: before #429 a server with both extensions
preloaded refused to start (#428).

## The honest caveat

We compress better: 1.48 GB against 1.66 GB, 12 percent smaller. Some of the extra time
buys that. But 12 percent smaller does not explain 3 times slower, and the trade as it
stands is a poor one to offer a user.

I have not profiled where our time goes, and I am not going to guess. #300's own method,
a profile of the load, is the right next step and it now has a specific question to answer:
what do we do per row that Citus does not.

## Why it matters beyond the number

[...]

PR #587, opened 2026-08-11
perf: detoast each varlena once per row on the write path (#445)

jdatcmd:
## What

`PgColumnarWriteRow` flattened every varlena value **up to four times per row** — once in `PgColumnarEncodeValue`, once for the bloom hash, and once per min/max comparison. For a **toasted (pglz-compressed)** value each of those is a full decompression; #445's write-path profile saw it as `detoast_attr`.

This detoasts once at the top of the row's column handling and reuses the flat `Datum` for the encoder, bloom hash, and min/max compares, freeing the copy after the row. `pg_detoast_datum` returns the same pointer for an already-flat value, so the uncompressed common case copies and frees nothing and non-varlena types skip it — **output is byte-identical either way**.

## Measured

300,000-row load of a ~3.8 kB `pglz`-compressed text column, pg18 non-assert, median of 3:

| | load |
|---|---:|
| before | 3193 ms |
| after | **2836 ms** |

**11.2% faster.** The saving scales with how compressed the values are — small for short strings (ClickBench URLs), larger for wide text. This is one item in #445's broad write-path tail, not its headline (that is compression, which parallelises).

## Tests

`test/native_toasted_write.sh` (new): loads a heap table's **toasted** values into columnar and differences against a heap mirror — values byte-identical, min/max range prune correct, equality (bloom + recheck) correct, and the short non-toasted column unaffected. It guards the flatten-once path specifically, which the other write suites don't reach (they use short, non-toasted values). Registered, `harness_selftest` green (107).

Correctness unaffected: `differential`, `native_writer`, `native_zonemap`, `native_bloom` all pass. The removal proof for the optimisation is the load-time number above (deleting it stays green — the output is identical — so the perf win is a timing, not a check).

### DIFF
--- a/src/columnar_write_state.c
+++ b/src/columnar_write_state.c
@@ -727,8 +727,27 @@ PgColumnarWriteRow(PgColumnarWriteState *writeState, Relation rel,
 		}
 		else
 		{
+			Datum		v = values[c];
+			struct varlena *flat = NULL;
+
+			/*
+			 * Detoast a varlena ONCE per row and reuse the flattened Datum for
+			 * everything below. It is otherwise flattened up to four times -- by
+			 * the encoder, the bloom hash, and each of the two min/max
+			 * comparisons -- and for a toasted (compressed) value each of those
+			 * is a full decompression, which #445's profile saw as detoast_attr
+			 * on the write path. pg_detoast_datum returns the same pointer for an
+			 * already-flat value, so the uncompressed common case copies and
+			 * frees nothing; non-varlena types skip it entirely.
+			 */
+			if (att->attlen == -1)
+			{
+				flat = pg_detoast_datum((struct varlena *) DatumGetPointer(v));
+				v = PointerGetDatum(flat);
+			}
+
 			appendStringInfoChar(&col->existsStream, 1);
-			PgColumnarEncodeValue(&col->valueStream, att, values[c]);
+			PgColumnarEncodeValue(&col->valueStream, att, v);
 			col->valueCount++;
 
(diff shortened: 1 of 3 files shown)


OffgridwithJD (approved):
## Approve — verified on four dimensions, prove-not-trust

Reviewed independently of the author. The detoast-once change is correct, memory-safe, faster, and adequately guarded.

- **Correctness vs heap oracle.** `differential` + `native_toasted_write` pass. I added an **external-TOAST probe** the base fixture doesn't reach — 4160-byte values with `STORAGE EXTERNAL`, premise-verified `is_external=t` — and columnar matches heap on values, min/max pruning, and bloom equality. That closes the one nuance in the change: `datumCopy` on the zone-map min/max now copies the flattened value rather than a possibly-external pointer, and it prunes identically to heap on the external branch.
- **Memory-safety.** Forced ASAN+UBSAN (instrumented `.so` confirmed): `native_toasted_write`, `differential`, `native_writer`, `native_dml`, `write_fsst_compressed`, `encode_invariants`, `corruption`, `native_reclaim` — **0 sanitizer reports**. The new per-row `pfree(flat)` is clean.
- **The guard is non-vacuous.** Removal proof: corrupting the flatten (marker verified live in the `.so`, not stale) flips `native_toasted_write` to FAILED on the value/bloom/non-toasted checks.
- **Perf reproduces.** Interleaved A/B, 300k-row toasted-text load, pg18n non-assert, 6 reps/arm: ~8.3% faster, every rep separated (your 11.2% on a more concentrated wide-text column; mine is diluted by a 3-column table). Direction unambiguous.

Merge remains yours — this is a review approval, not a merge.

PR #590, opened 2026-08-12
feat: round-trip flush_one_column through a dsm segment, serial (#445 slice 2)

OffgridwithJD:
## #445 slice 2: round-trip `flush_one_column` through a dsm segment (serial, no workers)

Second slice of the in-COPY parallelism design (#588), on top of slice 1 (#589). **Still serial in the backend; output is byte-identical.** No workers yet — the whole job of this slice is to prove the per-column input and result can cross a dsm segment byte-identically, so slice 3 can move the call to a worker with the serialization already settled.

> Depends on #589 (`flush_one_column`). Branch is stacked on it; will rebase to main once #589 merges.

### What changed

The per-column flush loop now, for each column:
1. **serializes** that column's `flush_one_column` input — its per-chunk-group buffers (`existsStream`, `valueStream`, `hashBuf`, `valueCount`, `sum`, `hasMinMax`, and the min/max `Datum`s via `datumSerialize`) — into a `dsm_create`d segment behind a `uint32` length prefix;
2. **deserializes** it back into a reconstructed per-column `chunkGroups` (every buffer copied *out* of the dsm into palloc'd memory), and runs `flush_one_column` on the round-tripped input;
3. **serializes** the result (`chunk` bytes, descriptor, block codec, zone-map rows incl. the numeric `sum` via `datumSerialize`, bloom filter) into a second segment, **deserializes** it, and assembles into the stripe exactly as slice 1 did;
4. `dsm_detach`es both segments.

`flush_one_column`'s slice-1 signature is **untouched**: the read side reconstructs a minimal `chunkGroups` whose `columns` array is sized `(columnIndex+1)` with only `[columnIndex]` populated (the only entry the function reads). Four static helpers (`serialize_column_input`/`deserialize_column_input`/`serialize_column_result`/`deserialize_column_result`) carry a fixed little wire format; `datumSerialize`/`datumRestore` handle the min/max and numeric-sum Datums.

### Why it's safe

- **Copy-out before detach.** Every buffer and Datum is copied from the dsm into palloc'd memory (`initStringInfo`+append for streams; `palloc`+`memcpy` for descriptor/min/max/filter; `datumRestore` for Datums), so nothing references the segment after `dsm_detach`. Lifetimes match slice 1 (flush context).
- **Self-checking format.** Each deserializer asserts the cursor consumed exactly the payload length (`Assert(cursor == base + 4 + payloadLen)`).

### Verification (prove-not-trust)

- **Byte-identical vs heap oracle:** `differential.sh` PASSES.
- **Zone-map / min-max / bloom serialization specifically:** `native_zonemap`, `native_bloom`, `write_minmax_fastpath` PASS — these exercise the parts of the wire format most likely to drift.
- **Write path across types/codecs:** `native_writer`, `native_dml`, `write_fsst_compressed`, `encode_invariants`, `encode_effort`, `corruption`, `native_reclaim`, `native_roundtrip` PASS.
- **Two toolchains:** forced **ASAN+UBSAN** (pg18_san) — 0 sanitizer reports on every suite (the dsm round-trip + copy-outs are the memory surface); and **assert** (pg18a) — no TRAP/Assert/crash. `-Wshadow=compatible-local` / `-Werror` clean.

### Next

Slice 3: dispatch `flush_one_column` across `min(natts, N)` background workers reading the slice-2 dsm, backend collects in column order, **degrading to the serial path when worker slots are unavailable** (never a wrong row count). Slice 4: GUC + measure vs the ~17% ceiling, then the default flip.

### DIFF
--- a/src/columnar_write_state.c
+++ b/src/columnar_write_state.c
@@ -906,6 +906,438 @@ PgColumnarBufferedRowByNumber(Relation rel, uint64 rowNumber,
 	return false;
 }
 
+typedef struct FlushColumnResult
+{
+	StringInfo	chunk;			/* [validity][finalData] column-chunk bytes; NULL if skipped */
+	char	   *descriptor;		/* encoding descriptor bytes; NULL if the column was skipped */
+	uint32		descriptorLen;
+	int			blockCodec;
+	List	   *zoneRows;		/* NativeZoneMapMetadata * for this column (per-vector + whole-chunk) */
+	NativeBloomMetadata *bloomRow;	/* per-chunk bloom for this column, or NULL */
+} FlushColumnResult;
+
+/*
+ * flush_one_column
+ *		Produce one column chunk's bytes, descriptor, block codec, zone maps and
+ *		bloom from that column's buffered input, reading only its arguments (no
+ *		writeState reach-through). The caller assembles the column chunks into the
+ *		stripe in column order and performs all I/O and catalog writes. Output is
+ *		byte-identical to the in-loop code this was extracted from (#445 slice 1).
+ */
+static FlushColumnResult
+flush_one_column(Form_pg_attribute att, List *chunkGroups,
+				 PgColumnarColumnDef *def, uint64 rowCount, int validityBytes,
+				 int encodeEffort, int compressionType, int compressionLevel,
+				 uint64 storageId, uint64 groupNumber, int columnIndex)
+{
+	FlushColumnResult result;
+	List	   *zoneRows = NIL;
+	NativeBloomMetadata *bloomRow = NULL;
+	StringInfo	chunk = makeStringInfo();
+	ListCell   *lc;
+	uint8	   *validity = (uint8 *) palloc0(validityBytes);
+	uint64		rowIdx = 0;
+	StringInfo	encoded = makeStringInfo();
+	StringInfo	desc = makeStringInfo();
+	uint8		descVersion = COLUMNAR_NATIVE_ENCDESC_VERSION;
+	uint8		descReserved = 0;
+	uint32		vectorCount = (uint32) list_length(chunkGroups);
+	char	   *fsstTable = NULL;	/* chunk-shared FSST table (E3b), or NULL */
+	uint32		fsstTableLen = 0;
+	char	   *finalData;
+	uint32		finalLen;
+	int			blockCodec = COLUMNAR_COMPRESSION_NONE;
+	int			vec = 0;
+	bool		chunkHasMinMax = false;
+	Datum		chunkMin = (Datum) 0;
+	Datum		chunkMax = (Datum) 0;
+	uint64		chunkValueCount = 0;
+	int64		chunkSum = 0;
+
+	/*
+	 * Virtual generated columns (attgenerated 'v', PostgreSQL 18+) are computed
+	 * on read from their base columns and never stored, so writing an all-null
+	 * chunk for them wastes space. Skip the chunk entirely: the reader finds no
+	 * column_chunk for this column, treats it as absent, and returns its missing
+	 * value (NULL via getmissingattr), while the executor expands the generated
+	 * expression regardless. A NULL descriptor marks the column skipped for the
+	 * column_chunk insertion pass below. ('v' is never set on PG15-17.)
+	 */
+	if (att->attgenerated == 'v')
+	{
+		result.chunk = NULL;
+		result.descriptor = NULL;
+		result.descriptorLen = 0;
+		result.blockCodec = COLUMNAR_COMPRESSION_NONE;
+		result.zoneRows = NIL;
+		result.bloomRow = NULL;
+		pfree(validity);
+		return result;
+	}
+
+	foreach(lc, chunkGroups)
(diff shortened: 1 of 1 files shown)


jdatcmd (changes requested):
## Correctness is solid — but this regresses the write path ~26% on the merge target

Verified the byte-identical claim rather than trusting it, and it holds: same 400k-row text-heavy load on this branch and clean main is identical in **size, content hash, encoding-descriptor hash, and chunk bytes** (3915776 / 64e2fe13… both). differential, native_writer, native_zonemap, native_bloom, native_dml, native_roundtrip, write_fsst_compressed, encode_effort all pass. The serialization is careful — `datumSerialize` for min/max, deserialize copies out of the dsm into palloc'd memory, length prefixes throughout. As a proof that the input and result cross a dsm byte-identically, slice 2 does its job.

**The blocker is performance on `main`, which the suites don't measure.** This slice does `dsm_create` + `dsm_detach` **twice per column per flush** (`columnar_write_state.c` :968, :982, :1377-8, inside the per-column loop), and segment creation is expensive. Measured, 41-column / 500k-row load, pg18 non-assert, median of 3:

| | load |
|---|---:|
| main | 1061 ms |
| this branch | **1333 ms** |

**~26% slower, with no benefit yet** (still serial), and it scales with column count — a 105-column ClickBench table does 2×105 `dsm_create`/flush. Merging this to main as a standalone slice means every writer pays that until slice 3 lands. That is the exact opposite of #445's direction for anyone loading in that window.

### What I'd change before this lands on main

Any one of:

1. **Reuse one dsm segment, not per-column.** Create a single segment per flush (or, better, a persistent per-writer segment grown as needed) and round-trip every column through it. That removes the `dsm_create` churn that is the whole cost here — and slice 3's workers will want a stable segment to attach to anyway, so this is the shape that slice actually needs, not throwaway scaffolding.
2. **Gate it off by default** behind the GUC/table-option slice 4 introduces, so `main` keeps the serial in-backend path until the full pipeline is proven a net win.
3. **Land it with slice 3**, so `main` never carries the serialize-without-parallelism cost alone.

I'd take (1): it is the least scaffolding-thrown-away and it is what the worker slice needs regardless. The extraction and wire format below it are right; it's the per-column `dsm_create` that shouldn't reach `main`. Happy to re-review promptly on any of the three.

OffgridwithJD:
You're right, and thank you for A/B-ing the write path — I verified byte-identity and memory-safety but did not measure load time, which is my miss: a perf A/B is exactly the discipline I hold my own findings to, and I didn't apply it to a regression I introduced. The per-column `dsm_create`/`dsm_detach` (2×natts per flush) is real cost for zero benefit while it's still serial.

**One thing that strengthens the case, which you may not have seen yet:** #591 (slice 3) inherits this. Its OFF branch is verbatim slice 2 — I confirmed reading the `else` in `pgcolumnar_flush_row_group`: with `pgcolumnar.parallel_flush` off (the default), it runs the same per-column round-trip. So the ~26% doesn't just affect slice 2 standalone; it's the **default path of slice 3 too** until something changes.

That points past "make the round-trip cheaper" to "**the serial path shouldn't round-trip at all.**" The serialization only earns its cost when there's a worker to hand the bytes to; in-backend-serial, the right path is a direct `flush_one_column` call — which is exactly merged `main` (slice 1) today, zero overhead.

So my recommendation, which is your option (3) sharpened:

- **Serial/OFF = direct `flush_one_column`** (= merged main, no dsm) → no regression on `main`, ever.
- **Parallel/ON = one input dsm + workers** — already how slice 3 builds it (a single per-flush segment, not per-column; your option (1)'s stable segment is what the worker slice already uses).
- **Fold slice 2's serialize/deserialize helpers into slice 3**, where the workers are their only consumer, and **supersede this PR (#590)**. The extraction and wire format you validated ride in unchanged; only the "round-trip serially in the backend" scaffold is dropped.
- I'll add the perf A/B to slice 3's gate: **OFF == main** (no regression) and **ON faster** (the actual win), alongside the byte-identical differential.

Your option (1) (one segment, keep slice 2 standalone) also works and is less restructuring, but it leaves a small serialize/deserialize cost on the serial path and doesn't fix #591's default. Since it's your design, your call — I'll re-verify with perf either way. If you're good with the fold, I'll close #590, revert slice 3's OFF branch to the direct path, and re-run the gate with the A/B.

OffgridwithJD:
## Superseded by #591 — closing

Per the review discussion here and on #591: the serial dsm round-trip this PR introduced should not be a shipping path (it's ~22-26% overhead when no worker reads the bytes, scaling with column count). The resolution landed in **#591**: the serial / non-dispatched flush calls `flush_one_column` **directly**, and the dsm crossing lives only on the worker path. The serialize/deserialize helpers and wire format you validated here ride into #591 unchanged — they're the workers' only consumer now.

So this standalone slice is retired rather than revised: there's no serial-round-trip path left to merge. Measured after the fix, the default (`parallel_flush` off) is back to main's write-path time (wide 41-col: 2.19 s == 2.19 s), and the parallel path is −11%. Closing; the work continues in #591.

PR #440, opened 2026-08-05
Record what the external validation established (#437)

OffgridwithJD:
Closes #437. @jdatcmd for review.

Four documentation changes from the external validation report, each backed by a
measurement rather than by the report's word. All 12 docs pages pass `docs_style.sh`.

## 1. aarch64 is no longer untested

`docs/limitations.md` said every recorded result comes from x86_64 and the suites have
never run on aarch64 (#242). The independent gate ran the full suite on a Graviton3 host
and every suite passed, with the same per-suite result as x86_64. The big-endian half of
that sentence is still true and stays.

**The owner asked that the arm against x86 timing ratios not be published**, so they are
not here. Graviton being slower on scan paths is not a finding anyone needs. The
correctness update is a different claim and it is in.

## 2. `stripe_row_limit` is documented as the fetch-cost knob, and `chunk_group_row_limit` is named as not being it

The report recommends `chunk_group_row_limit` for point-lookup-heavy tables. **It does
nothing.** Isolated one knob at a time on a 511 MB blob-heavy fixture:

| configuration | point lookup |
|---|---:|
| defaults | 245.4 ms |
| `chunk_group_row_limit` 1000 only | **252.3 ms, no effect** |
| `stripe_row_limit` 10000 only | **19.3 ms** |

Both the reporter and my own first measurement changed the two together and could not tell
them apart. The docs now name the right one and say the other is not it, because the advice
as published would send people to tune a knob that does not move this.

## 3. The knob's cost is stated with its benefit

| `stripe_row_limit` | size | point lookup | full aggregate scan |
|---|---:|---:|---:|
| 150000 (default) | 527.0 MB | 244.1 ms | 38.7 ms |
| 10000 | 527.3 MB | 18.5 ms | 35.4 ms |
| 2000 | 529.4 MB | 5.6 ms | 37.9 ms |

Size grows 0.4 percent, scan throughput does not change, and **there is a floor**: a
selective range query was slower at 2000 than at 10000, because a smaller row group makes
more metadata to read. A tuning knob documented without its cost is a trap, so the floor is
in the same paragraph as the win.

## 4. Cross-architecture portability, stated as narrowly as the format supports

This is the one I would look at hardest.

The report moved a data directory from x86_64 to aarch64 and read it correctly, and
suggested a docs line. **Both of those are little-endian, so that test cannot distinguish
"portable" from "host-endian and both hosts agree."**

It is host-endian, deliberately, and both the spec and the header say so:

```
[...]

PR #592, opened 2026-08-12
harden the parallel flush: cancel-safe cleanup + skip new-in-xact relations (#445 slice 4a)

OffgridwithJD:
## #445 slice 4a: harden the parallel flush (cancel-safe + new-in-xact)

The two robustness fixes from the slice-4 work, split out per review so they can land on their own. **No default change** — `pgcolumnar.parallel_flush` stays off; the default flip is a separate PR behind a size gate (small/frequent flushes must not pay per-flush worker-spawn cost — measured 3.6× on a 1000-row-flush load).

### 1. Cancel-safe cleanup (the leak-window flagged on #591)

The parallel flush's wait/collect is wrapped in `PG_ENSURE_ERROR_CLEANUP`. On an error — a statement cancel inside `WaitForBackgroundWorkerShutdown` is the realistic one — it terminates the workers and frees every published-but-uncollected pinned output segment. The collect loop invalidates each handle as it frees it, so the cleanup only attaches segments still pinned by a worker and never double-frees. Closes the leak-to-postmaster-restart window.

### 2. Skip parallel for a new-in-transaction relation

`CREATE TABLE ... USING pgcolumnar; INSERT ...` (or `CREATE TABLE AS`) in one transaction: the table's `pg_class` row is uncommitted, so a worker's fresh-snapshot `table_open` fails. Before, that path registered workers, had them all fail, logged a WARNING, and redid the flush serially. Now `rel_new_in_current_xact()` (via `rd_createSubid`) detects it and keeps the flush serial up front — silent, no wasted registration, no double work.

### Verification

- Both fixes exercised with `pgcolumnar.parallel_flush=on`: a committed-table load uses workers and is `EXCEPT`-equal to heap (`diff=0`); a `CREATE+INSERT+COMMIT` load stays serial with **0 worker-failed WARNINGs** and is byte-correct (`diff=0`).
- Byte-identical, off and on: the full write/concurrency/crash suite set passes under forced ASAN+UBSAN (0 sanitizer reports) and pg18a assert (0 crashes). `-Wshadow`/`-Werror` clean.

The default flip + size gate follows in its own PR, with a small-flush timing beside the large-flush one.

### DIFF
--- a/src/columnar_tableam.c
+++ b/src/columnar_tableam.c
@@ -2687,15 +2687,23 @@ _PG_init(void)
 
 	DefineCustomBoolVariable("pgcolumnar.parallel_flush",
 							 "Dispatch the per-column stripe flush across background "
-							 "workers (#445 slice 3).",
-							 "Off by default so the merged write path is unchanged. "
-							 "When on, a stripe flush of two or more columns fans the "
-							 "per-column encode/compress work out to a pool of "
-							 "background workers and degrades to serial in-backend "
-							 "completion for any column a worker does not finish, so "
-							 "the stored bytes are byte-identical to the serial path "
-							 "either way. Opt-in for testing; slice 4 makes it the "
-							 "measured, eventually-default control.",
+							 "workers (#445).",
+							 "Off by default, and stays that way (#445 slice 4 measured "
+							 "it). When on, a stripe flush of two or more columns fans "
+							 "the per-column encode/compress work out to a pool of "
+							 "background workers, degrading to serial in-backend "
+							 "completion for any column a worker does not finish, so the "
+							 "stored bytes are byte-identical to the serial path either "
+							 "way. It is a targeted opt-in, not a general win: it helps a "
+							 "single large flush of many cheap (numeric) columns -- a "
+							 "wide bulk load -- by up to ~14%, and it REGRESSES the common "
+							 "cases, because it copies the buffered data through shared "
+							 "memory: small or frequent flushes (a low stripe_row_limit, "
+							 "or per-row commits) pay the fixed worker-spawn cost (3.6x "
+							 "slower measured), and text-heavy or very large flushes pay "
+							 "an O(bytes) serialization cost that outweighs the parallel "
+							 "saving. Turn it on only for a wide-numeric bulk load, off "
+							 "otherwise.",
 							 &pgcolumnar_parallel_flush,
 							 false,
 							 PGC_USERSET,
--- a/src/columnar_write_state.c
+++ b/src/columnar_write_state.c
@@ -2161,6 +2161,51 @@ pgcolumnar_parallel_flush_worker(Datum main_arg)
 }
 
 /*
+ * pflush_error_cleanup
+ *		Runs on an error during the parallel flush's wait/collect (#445 slice 4).
+ *		Terminates any still-running workers, then frees every published-but-
+ *		uncollected output segment. The collect loop invalidates each slot's handle
+ *		as it frees it, so this only attaches segments still pinned by a worker --
+ *		valid handles, safe to map -- and never double-frees one the happy path
+ *		already released. Without this, a statement cancel inside WaitForBackground-
+ *		WorkerShutdown would leak the pinned segments until postmaster restart.
+ */
+typedef struct PflushCleanup
+{
+	BackgroundWorkerHandle **handles;
+	int			nstarted;
+	PflushWorkerSlot *slots;
+	int			nworkers;
+} PflushCleanup;
+
+static void
+pflush_error_cleanup(int code, Datum arg)
+{
+	PflushCleanup *cl = (PflushCleanup *) DatumGetPointer(arg);
+	int			i;
+
+	for (i = 0; i < cl->nstarted; i++)
+		if (cl->handles[i] != NULL)
+		{
+			TerminateBackgroundWorker(cl->handles[i]);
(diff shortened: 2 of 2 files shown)


jdatcmd (changes requested):
## The two fixes are right — but default-on regresses small/frequent flushes 3.6x

### First, the "CI failure" is a 60-minute timeout, not a test failure

The two red checks are `cancelled`, not `failure`. The run's own timing shows it: every build succeeded, and `suites (PG 17)`/`suites (PG 18)` ran **06:09:42 → 07:09:58 — exactly 60 minutes — then were cancelled**. The suites didn't fail; they didn't finish. I reproduced the whole matrix locally and every suite passes (only `temporal` fails here, and only because my container lacks `btree_gist`).

### Why default-on makes them time out — and why it's a real regression

Default-on spawns a worker pool on **every stripe flush**, and register + wait + collect + detach is a fixed cost that only amortises when the flush is large. Measured, 50 flushes in one load (stripe_row_limit=1000, 50k rows, 5 cols), pg18 non-assert:

| | load |
|---|---:|
| `parallel_flush=off` | 47 ms |
| `parallel_flush=on` | **170 ms (3.6x)** |

A single large flush is fine (5000-row/one-flush load is 71 ms either way — the win case I approved on #591 was a 500k-row load). But the suite matrix does *thousands* of small flushes across ~130 suites, so the per-flush worker cost balloons the run past the 60-minute budget. The same is true of any real workload with frequent small commits or a low `stripe_row_limit` (#440 recommends exactly that for lookup-heavy tables) — they'd all regress ~3.6x.

### What I'd change

Keep both hardening fixes — the `PG_ENSURE_ERROR_CLEANUP` cancel path (handle invalidated as freed, no double-free) and the `rel_new_in_current_xact` serial guard are correct and worth having on their own.

But **the default flip needs a size gate first**: dispatch to workers only when the flush is large enough to amortise worker spawn — e.g., total encoded bytes, or rows × columns, above a threshold; below it, stay serial. Then:

- CI suites (small writes) run serial → no timeout;
- small / frequent-commit workloads don't regress;
- large loads still get the ~14% win #591 measured.

Without that gate, on-by-default is net-negative for the common small-flush case. I'd split this: land the two fixes now, and make the default flip its own PR behind the threshold with a small-flush timing beside the large-flush one. Happy to re-review either promptly.

OffgridwithJD:
You're right, and I reproduced it rather than take it — same shape as your slice-2 catch, and my miss again: I measured the *win* case (300k–500k-row single flushes) and never the adversarial small/frequent-flush shape where the fixed per-flush worker-spawn cost dominates. 50k rows, 5 cols, pg18n, median of 3, varying stripe_row_limit:

| rows/flush | flushes | off | on | on/off |
|---:|---:|---:|---:|---:|
| 1000 | 50 | 0.07 s | 0.25 s | **3.57x** |
| 2000 | 25 | 0.07 s | 0.17 s | 2.43x |
| 5000 | 10 | 0.07 s | 0.10 s | 1.43x |
| 20000 | 2 | 0.10 s | 0.11 s | 1.10x |
| 50000 | 1 | 0.12 s | 0.13 s | 1.08x |

Confirms your 3.6x, and shows the fixed spawn cost (~4 ms/flush) only amortises on a large flush. (This 5-col table never *wins* even at one 50k-row flush — the −14% needs many columns **and** size — which says the gate should key on total work, not row count alone.)

**Doing exactly what you asked:**
1. I've reset this PR to the **two hardening fixes only** — the `PG_ENSURE_ERROR_CLEANUP` cancel path and the `rel_new_in_current_xact` serial guard, both correct on their own. Retitled; ready for re-review as a standalone.
2. The **default flip goes in its own PR behind a size gate**: dispatch only when the flush's total buffered value bytes exceed a threshold; below it, stay serial. I'll set the threshold from a measured win-crossover (total bytes, spanning narrow→wide) and post the **small-flush timing beside the large-flush one** — CI's small writes run serial (no timeout), frequent-commit / low-`stripe_row_limit` (#440) workloads don't regress, large loads keep the win.

Thanks for catching it before it shipped. Re-review of the hardening PR whenever you have a moment.

OffgridwithJD:
Added one commit (`b5e9aae`): the GUC description now documents the measured profile and that it stays opt-in (the flip is declined — full data on #445). This PR is still just the two hardening fixes plus that doc correction.

jdatcmd (approved):
## Approve — the right resolution. Re-verified (PG18).

The revision does exactly what the review asked: **keep the two hardening fixes, drop the default flip.** `parallel_flush` stays off by default with the measured perf profile documented, so it's opt-in for the large-flush case where #591 measured the win, and it can't regress small/frequent-flush workloads or blow the CI budget.

Re-verified on the new head (`b5e9aae`):

- **Default is OFF** — proven by CI itself: `suites (PG 17)`/`suites (PG 18)` now complete green instead of running to the 60-minute cancel, which only happens if the suites' (unSET) writes take the serial path. The GUC description says so and the run confirms it.
- **Both hardening fixes intact and correct** — `PG_ENSURE_ERROR_CLEANUP` cancel path and the `rel_new_in_current_xact` serial guard. native_cancel, cancel_decode, native_ctas, native_backend_crash, differential, native_dml all pass.
- The opt-in ON path's per-flush cost on many small flushes is unchanged and now documented rather than shipped as a default — which is the correct place for it.

Good call splitting it this way. A future default-on behind a size threshold (dispatch only when the flush is large enough to amortise worker spawn) is the natural follow-up, but nothing here needs it. Approving; merge is jd's.

Issue #595, opened 2026-08-12
Decode gating (#452 phase 2) is unconditional; gate it on selectivity so it stops regressing less-selective scans


OffgridwithJD:
## The problem, measured

The #452 phase-2 decode gating (`pgcolumnar_native_qual_skipvec`, reached whenever a scan has an executor qual that the reader cannot push down) is **unconditional**: it evaluates the qual per 1024-row vector to skip payload decode for no-match vectors. That is a clear win when the qual is selective and the projection is wide — and a pure cost when it is not, because the per-vector evaluation buys nothing if every vector has a surviving row.

Found while re-running the ClickBench cross-engine benchmark on current `main` (pg18n, 11.1M-row stride sample). Columnar's win count against heap dropped from the previously documented **33/43 to 25/43** at defaults. An A/B that toggles the gating on the same build and data (`pgcolumnar.enable_late_materialization` on vs off, added as a `columnar_nolm` arm) attributes the whole shift to this gating:

| query | heap | gating ON | gating OFF | effect of gating |
|---|---:|---:|---:|---|
| **q24** `SELECT *` under a 0.02%-selective `LIKE` | 752 | **3395** | 6098 | **helps 1.8x** — the case it is for |
| q11 | 1030 | 1270 | **738** | hurts 1.7x |
| q12 | 1121 | 1386 | **801** | hurts 1.7x |
| q14 | 1652 | 1911 | **1413** | hurts 1.35x |
| q15 | 1878 | 2176 | **1617** | hurts 1.35x |
| q25 | 890 | 1138 | **582** | hurts 2.0x |
| q27 | 812 | 1140 | **588** | hurts 1.9x |
| q31 | 1827 | 2094 | **1564** | hurts 1.3x |
| q32 | 1816 | 2145 | **1615** | hurts 1.3x |

So gating on recovers q24 but regresses eight less-selective scans ~1.2–2x; gating off does the reverse. It is a real, deliberate trade — not a bug — but it is applied to every unprunable-qual scan whether or not it can pay.

## What it should do instead: gate the gating on selectivity

The evaluation should happen only when the filter is selective enough that skipping vectors will actually save more than the per-vector evaluation costs. Two levers, weakest to strongest:

[...]

Issue #467, opened 2026-08-06
Bloom filters are sized by value count, not distinct count: 19.3x over-provisioned, 29% of load time, 361 MB raw against a 262 MB table


OffgridwithJD:
## Problem

`PgColumnarBloomBuild` sizes each filter from the **value count**, not the distinct count:

```c
/* columnar_bloom.c */
#define BLOOM_BITS_PER_VALUE 10
#define BLOOM_MAX_BITS (1u << 21)   /* 256 KB cap per chunk */
...
nbits = next_pow2(n * BLOOM_BITS_PER_VALUE);   /* n = number of hashes, not distinct */
```

A filter is stored per column per stripe, so `n` is `pgcolumnar.stripe_row_limit`
(default 150,000). That gives 1.5 Mbit, which `next_pow2` rounds to exactly 2^21 —
**256 KB for every bloomable column of every stripe, regardless of that column's
cardinality.** A column with 5 distinct values gets the same 256 KB filter as a unique one.

Measured on 2M ClickBench `hits` rows (105 columns), distinct values per 150,000-row
stripe:

| distinct per stripe | columns | filter utilisation |
|---|---|---|
| < 64 | **68** | ~99.98% empty, still 256 KB each |
| 64 – 1k | 11 | >99% empty |
| 1k – 10k | 10 | >93% empty |
| 10k – 50k | 4 | >66% empty |
| >= 50k | 12 | reasonably used |

Sized by distinct count these filters need **11.4 Mbit**; they are allocated **220 Mbit**
— **19.3x over-provisioned**. (The `next_pow2` rounding is a further ~40% on top: 1.5 Mbit
of intent becomes 2.097 Mbit allocated, i.e. 14 bits/value rather than the 10 `BLOOM_K = 6`
was chosen for.)

Note this is *not* the `n < 64` early-out or the `BLOOM_MAX_BITS` refusal doing their job —
both test the **value** count, which is 150,000 here, so neither fires. Every column gets a
filter.

## Cost

All arms `encode_effort = fast` (FSST out of the picture), serial `COPY`, 2M rows, every
arm asserted at exactly 2,000,000 rows:

| arm | load | bloom TOAST written | `pglz_compress` in profile |
|---|---|---|---|
| bloom on (default) | 44.6s | 58 MB | 10.4% |
| **bloom off** | **31.7s** | 0 MB | **absent** |
[...]

Issue #423, opened 2026-08-05
Ungrouped vectorized aggregate errors on a varlena filter column: unsupported byval length: -1


OffgridwithJD:
Found while building the ClickBench runner for #421. ClickBench q21 is
`SELECT COUNT(*) FROM hits WHERE URL LIKE '%google%'`, and it is the only one of the 43
that fails.

## Reproducer

```sql
CREATE TABLE t (i int, s text) USING pgcolumnar;
INSERT INTO t SELECT g, 'abc'||g FROM generate_series(1,200000) g;

SET pgcolumnar.enable_ungrouped_vector_agg = on;
SELECT count(*) FROM t WHERE s LIKE '%9%';
ERROR:  unsupported byval length: -1
```

With the GUC off it returns 81902, which is what the heap twin returns.

## The boundary, measured

| shape | ungrouped vector agg on |
|---|---|
| `count(*) WHERE s LIKE '%9%'` | **ERROR** |
| `count(*) WHERE i > 100 AND s LIKE '%9%'` | **ERROR** |
| `sum(i) WHERE s LIKE '%9%'` | **ERROR** |
| `count(*) WHERE s = 'abc7'` | ok, 1 |
| `count(*) WHERE length(s) > 4` | ok, 199991 |
| `count(*) WHERE i > 100` | ok, 199900 |
| `count(*)` | ok, 200000 |

So it is a `LIKE` on a varlena column reaching the batch fold. Equality on the same
column is fine, and a function call on it is fine.

## Where

`src/columnar_vector.c:2904`, in the batch fold's per-row gather:

```c
cval[col] = fetch_att(cpacked[col] + cpresent[col] * cattlen[col],
                      true, cattlen[col]);
```

That is fixed-width arithmetic. `cattlen[col]` is the column's `attlen`, which is **-1**
for a varlena, so the offset `cpresent[col] * -1` is nonsense and `fetch_att(p, true, -1)`
raises the error above from the inlined `store_att_byval`.

The gate that should have refused the shape is
`pgcolumnar_batch_shape_eligible` (`columnar_vector.c:2701`). It walks the built scan keys
and checks `pgcolumnar_batch_type_ok(coltype)` on each. That tests whether the type is
**comparable**, not whether its storage is **fixed width**, and the fold's gather loop
walks `state->projected` rather than the scan keys anyway. The plan confirms the fold
engaged with no filter pushed down at all:

```
Custom Scan (ColumnarScan)
[...]

Issue #401, opened 2026-08-04
We have no join-heavy benchmark, and no columnar-specific join optimisation; measure before deciding on set_join_pathlist_hook


OffgridwithJD:
Our optimisations stop at the scan and the aggregate directly above it. A join is core's
node, and we get no say in join order, join method, or what reaches the columnar side.

Confirmed structurally: we install `set_rel_pathlist_hook`, `create_upper_paths_hook`,
`build_simple_rel_hook` (`get_relation_info_hook` below 19), `object_access_hook` and
`ProcessUtility_hook`. There is **no `set_join_pathlist_hook`**, so we never offer a join
path.

Measured, a columnar-to-heap join today:

```
 Finalize GroupAggregate
   ->  Gather Merge  (Workers Planned: 6)
         ->  Partial HashAggregate
               ->  Hash Join
                     Hash Cond: (c.hostname = d.hostname)
                     ->  Parallel Custom Scan (ColumnarScan) on cpu_pgc c
                     ->  Hash -> Seq Scan on hosts_dim d
```

That works, and it works because we are a table access method rather than a foreign data
wrapper. Cross-format joins have never needed anything from us.

## First, what this is not

**It will not help the cross-engine numbers.** All eight TSBS queries in
`docs/benchmarks.md` are single-table:

```
$ grep -lciE '\bjoin\b' queries/tsbs/*.sql | wc -l
0
$ grep -hiE '^\s*FROM' queries/tsbs/*.sql | sort | uniq -c
      8 FROM %T
```

There is no join in that benchmark to accelerate. Anyone reaching for this as a way to
close the TimescaleDB gap should reach for the clustering result instead, which is
measured and large.

## The actual gap, which is a measurement gap

**We have never measured a join-heavy workload on pgColumnar.** Our whole performance story
is single-table scans and aggregates, which is the shape TSBS has. Star schemas and
dimension joins are the shape columnar storage is usually bought for, and we do not know
what we do on them. Not "we know and it is bad" -- we do not know.

That is the prerequisite. Proposing a join implementation before measuring whether we lose,
and where, is committing to work with no evidence behind it.

## The mechanism worth evaluating, which is not join ordering

If there is a win available inside a table access method, the literature says it is
**runtime filter pushdown**, not join order. When the build side of a hash join produces a
small key set, pushing that set into the probe side's scan lets it skip data before
decoding it.

We are unusually well placed for this. The scan already skips chunk groups on zone maps and
[...]

PR #596, opened 2026-08-12
docs: refresh benchmarks.md from the 2026-08-12 run (ClickBench win 33->25, #452 trade, #595)

OffgridwithJD:
## Refresh benchmarks.md from a full 2026-08-12 run on current main

Re-ran the whole harness on `main` at `2fe6596`, pg18n non-assert, 16 cores/62 GB: main suite (6M rows, median of 5), FSST, joins (#401, 20M), and cross-engine ClickBench (11.1M-row stride sample, heap/columnar/columnar_tuned/citus/duckdb). Every measured table is updated from the raw output; not-re-measured sections (TSBS, read-stream/AIO, parallel) keep their prior attributions.

### The substantive change: ClickBench win count 33 → 25, and why

Columnar now beats heap on **25/43** (was 33), because #452 phase-2 decode gating shipped since the last run. An A/B on this build/data (gating on vs off) attributes it precisely: gating **helps q24 1.8x** (its target — a wide `SELECT *` under a selective filter, 6098→3395 ms) but **regresses eight less-selective queries 1.2–2x** (q11/q12/q14/q15/q25/q27/q31/q32), because it evaluates the qual per vector even when nothing is skippable. It's a deliberate trade, documented as such, with the fix (gate the gating on selectivity — runtime probe) filed as **#595** (prototype in progress).

Other updates: load gap restated as a serial-path property (columnar 2.38x serial / 1.73x bulk vs Citus, still smallest on disk at 1.48 GB); `parallel_flush` documented as a measured opt-in, not eventually-default (#445 slice 4); anchored LIKE (#510), detoast-once (#587), FSST verdict cache (#472), bloom by distinct count (#467) in "what changed". Noted that **#423 is resolved** — q21 no longer errors in the fresh run.

### Judgment calls flagged for review (I did not want to guess)

1. **ClickBench summary totals + geomean** (heap 168.2 s / columnar 140.4 s / citus 266.9 s; geomeans 0.52 / 1.77 / 0.29): the harness printed per-query times but not these aggregates, so I **derived** them arithmetically from the measured rows. Confirm you want them derived vs left blank.
2. **Two historical micro-benchmark paragraphs now read dissonantly** against the fresh tables and were left verbatim: the import/export analysis ("import ~18x export" — the new tables show ~3.4x) and the Mutation "previous record" note (1.91 ms single-row vs the note's older framing). Both are separate un-re-measured benches; annotate or drop as you prefer.
3. The **2026-08-05 ClickBench subsection** is kept as an independent historical run; I annotated it where #423 resolved.

### DIFF
--- a/docs/benchmarks.md
+++ b/docs/benchmarks.md
@@ -22,15 +22,14 @@ These environment variables control the harnesses:
 - `BENCH_DUCKDB`. Set it to 1 to add a DuckDB comparison, if `duckdb` is on
   `PATH`.
 
-The numbers below come from one full run of `bench/run_bench.sh` on 2026-08-04, at
-commit `aeb7882`, which is the same commit as the cross-engine run below. The conditions were PostgreSQL 18.4 non-assert, 6,000,000 rows,
-an 8-column table, the median of 5 repetitions, 8 cores and 12 GB of memory.
+The numbers below come from one full run of `bench/run_bench.sh` on 2026-08-12, at
+commit `2fe6596` (Merge #592). The conditions were PostgreSQL 18.4 non-assert, 6,000,000 rows,
+an 8-column table, the median of 5 repetitions, 16 cores and 62 GB of memory.
 
-The previous record of this section was 2026-07-27 at commit `7a9c9f7`, on
-PostgreSQL 17.10 and a machine with 24 GB. The harness itself did not change
-between the two runs. The major version and the machine did, so read the ratios
-and not the absolute values. The old raw output is in
-[../bench/sample_output_all_2026_07_27.txt](https://github.com/commandprompt/pgcolumnar/blob/main/bench/sample_output_all_2026_07_27.txt).
+The previous record of this section was 2026-08-04 at commit `aeb7882`, on
+PostgreSQL 18.4 and an 8-core machine with 12 GB. The harness itself did not change
+between the two runs. The machine did, so read the ratios and not the absolute
+values.
 
 The read stream section is not re-measured. It needs a PostgreSQL 18 build with
 `--with-liburing`, and no such build exists on this machine. Its numbers are from
@@ -44,11 +43,10 @@ deliberately. A table of fully random values will therefore look worse, and a
 repetitive table will look better.
 
 **Compare the ratios between runs. Do not compare absolute milliseconds.** This
-machine measured the commit of the previous run again, on the same day. Refer to
-[What changed](#what-changed-since-the-previous-run). The query latencies were
-the same. But the ingestion shapes and the I/O shapes were approximately 20
-percent slower than the first record of those numbers. Absolute figures change
-with the machine. Trust the comparison that one day gives.
+run is on a larger machine than the previous record, 16 cores and 62 GB against
+8 and 12, so the absolute figures here are larger and different, and that is
+expected. Refer to [What changed](#what-changed-since-the-previous-run) for what
+moved in the code. Trust the ratios, not the milliseconds.
 
 ## Storage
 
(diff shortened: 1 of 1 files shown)


jdatcmd (changes requested):
**Request changes** — one blocking gate, then this is good to merge. The substance is accurate and the honesty about my own #452 regression is exactly right.

### Blocking: the CI red is real, and it is `docs_style`

The "suites (PG 17)" / "suites (PG 18)" failures are **not** a timeout or infra flake this time. `docs_style` is a registered suite, and this refresh introduces STE violations that fail it. I verified by attribution, not assumption:

- `ste_check.py` on **main**'s `benchmarks.md`: `ok` (clean).
- `ste_check.py` on **this branch**'s `benchmarks.md`: `FAIL — 17 long, 6 prose double-hyphen`.

So the refresh introduced them; they are not pre-existing. A docs-only PR *can* legitimately turn the suite job red, and here it does. Representative hits:

- 41 words: "It is a large win on a wide `SELECT *` under a highly selective filter (q24 from 6098 ms to 3395 ms ...)"
- 38 words: "This is why the ratio against heap falls as the number of rows rises, from 64 times ..."
- 38 words: "Do not compare absolute milliseconds.** This run is on a larger machine ..."
- 31 words: "It helps a wide bulk load of many numeric columns by up to about 14 percent and regresses ..."
- 27 words: "**Detoast once (#587).** A toasted text value is now detoasted a single time per scan ..."
- The 6 prose double-hyphens include the `--` in the #452 blockquote ("a wide `SELECT *` under a highly selective filter --", "q11 ... q32, each about 1.2 to 2x --") and the q21 line ("q21 no longer errors --").

Fix: split the 17 sentences under 25 words and restructure the 6 `--` into separate sentences (the same STE discipline the rest of `docs/` already meets). `python3 test/ste_check.py docs/benchmarks.md` prints the full list; green there turns the suite job green.

### Substance: accurate, and I checked the part that is mine

The #452 phase-2 trade is reported correctly. Phase 2 evaluates the executor qual per 1024-row vector even when nothing is skippable, so it costs on less-selective scans and pays off only under a selective filter over a wide row. The A/B (gating on vs off) is the right instrument, the eight attributed regressions (q11/q12/q14/q15/q25/q27/q31/q32, ~1.2–2x, recovering with gating off while q24 loses its win) match the mechanism, and 33→25 with #595 tracking the selectivity gate is the correct disposition. Calling it "a deliberate trade, not a regression to hide" is the right call — thank you for not burying it.

### The three judgment calls you flagged

1. **Derived ClickBench aggregates (totals + geomeans).** Keep them — a derived aggregate beats a blank. One ask: label them as derived (e.g. a one-line "totals and geomeans computed from the per-query rows; the harness does not emit aggregates") so a later reader does not mistake them for harness output, and re-verify the geomean once against the 43 rows. I did not independently re-sum all 43 here, so I am not vouching for the arithmetic, only for the decision to publish them.
2. **The two dissonant historical paragraphs (import/export ~18x vs new ~3.4x; the Mutation "previous record" note).** Annotate, do not drop — one clause each pinning them to their run and machine so the numbers do not read as comparable to the 2026-08-12 figures.
3. **Keeping the 2026-08-05 subsection as annotated history.** Correct — a second independent run is worth keeping, and annotating where #423 resolved q21 is the right touch.

Nothing else blocks. Fix the STE gate and I will approve.

jdatcmd:
The STE gate is the only thing between this and merge. The "suites (PG 17/18)" red is not a flake or a timeout: `docs_style` is a registered suite, and the refresh introduced STE violations, so the suite job fails. I attributed it rather than guessed:

- `ste_check.py` on **main**'s `benchmarks.md` -> `ok`
- `ste_check.py` on **this branch** -> `FAIL: 17 long, 6 prose double-hyphen`

So the refresh created them. `python3 test/ste_check.py docs/benchmarks.md` prints the full list; the `--` hits include the #452 blockquote and the q21 line. Split the 17 sentences under 25 words and restructure the 6 double-hyphens into separate sentences, and the suite goes green.

The substance is already right and I have signed off on it, including the honest #452 phase-2 trade and #595 as the tracked fix. Push the STE fix and I will re-review and approve.

OffgridwithJD:
Fixed the STE gate (`c9ca601`). Split the 17 over-25-word sentences at clause boundaries and restructured the 6 prose double-hyphens (the #452 blockquote and the q21 note) into separate sentences. `python3 test/ste_check.py docs/benchmarks.md` now reports `ok`.

**No numbers moved.** I verified independently of the sentence edits: the multiset of every numeric/ratio/byte/`#id`/`qNN` token is byte-identical before and after — the only token-level difference is one `12,` becoming `12.` where a sentence split at the "8 and 12 GB" machine spec. The tables, the 25/13/5 win count, the #452 A/B figures, and the #423-resolved note are all exactly as reviewed. Suites should go green now; ready for re-review.

jdatcmd (approved):
STE fix verified: `ste_check.py` on the updated `benchmarks.md` now reports `ok` (was 17 long + 6 prose double-hyphen), and the CI suites are green — `docs_style` passes. The substance survived the rewrite intact: q24 6098->3395, win 33->25, #595 tracked, the totals/geomeans, and the #452 trade blockquote split into short sentences with nothing dropped (the runtime-probe note is sharper for it). My changes-requested is resolved. Approving.

PR #597, opened 2026-08-12
docs: reconcile the #426 prune ceiling to the measured ~91% (was inconsistently ~93%/~91%)

OffgridwithJD:
## Reconcile the #426 prune ceiling: it's the measured ~91%, not ~93%

The merged `ISSUE_426_SUBSTRING_FILTER.md` cited both a **~93% prune plateau** (disposition, line 10) and a **~91% prune ceiling** (line 101) — internally inconsistent, since a plateau can't exceed the ceiling.

Verified rather than guessed. Re-measured the `%google%` zero-match 1024-row-block fraction on the ClickBench `hits` data via duckdb:

| sample | zero-match block fraction |
| --- | ---: |
| full 11.1M (the columnar table's size) | **91.01%** (10,851 blocks) |
| first 1M (the doc's stated probe) | 93.45% (977 blocks) |
| `%google%` row match rate | 0.0156% |

Both are correct for their scale — but the 1M figure is a **prefix**, and the matches cluster, so it over-counts skippable vectors. The full sample is the truer **~91%**. (Same prefix-unrepresentativeness that broke the first-K decode probe on #595 — a nice consistency between the two.)

The fix uses ~91% in both places and notes why the 1M probe over-read. One-line doc change; no code.

### DIFF
--- a/design/ISSUE_426_SUBSTRING_FILTER.md
+++ b/design/ISSUE_426_SUBSTRING_FILTER.md
@@ -7,8 +7,11 @@ merged in #584) and `design/ISSUE_452_PHASE2_DECODE_GATING.md`.
 ## DISPOSITION (2026-08-11): feasible, measured, NOT built
 
 The design below is sound and the step-1 measurement was taken on real ClickBench
-`hits` URLs (1,000,000 rows via duckdb). Result: reaching the ~93% prune plateau
-costs **~8 bytes/row** on the filtered string column (65536 bits per 1024-row
+`hits` URLs (1,000,000 rows via duckdb; re-checked on the full 11.1M sample).
+Result: reaching the ~91% prune ceiling (the initial 1M-row probe read ~93%, but
+that is a prefix, and the `%google%` matches cluster, so the full sample is the
+truer ~91% -- a prefix over-counts skippable vectors) costs **~8 bytes/row** on
+the filtered string column (65536 bits per 1024-row
 vector; real URLs carry ~7,251 distinct trigrams/vector, so a smaller filter
 saturates and prunes nothing). On URL that is ~40% overhead on the column, ~6% of
 the whole hits table, and it buys only q21/q22 (2.2x/1.8x) because q24 -- the big


jdatcmd (approved):
**Approve** — the reconciliation is right and I verified it independently. One figure in the PR body is wrong, but it is not in the merged doc, so it does not block.

### Verified, not trusted

Re-measured on the same data the columnar table is loaded from (`/srv/clickbench/hits.every9.tsv`, 11,102,424 rows) with duckdb, `%google%` at 1024-row blocks:

| sample | total blocks | zero-match blocks | zero-match % |
| --- | ---: | ---: | ---: |
| full 11.1M | 10,843 | **9,865** | **90.98%** |
| first 1M (the original probe) | 977 | 913 | **93.45%** |

- The full-sample **~91%** in the doc is correct (90.98%; the 0.03 gap from your 91.01% is duckdb parallel scan-order nondeterminism, immaterial).
- The 1M prefix reproduces your **93.45% / 977 blocks exactly**, and 93.45% > 90.98% confirms the mechanism: the prefix over-counts skippable vectors because the `%google%` matches cluster. Same prefix-unrepresentativeness as the #595 first-K probe — the parallel you drew holds.
- Match rate 1728 / 11,102,424 = **0.0156%**, as stated.

So the change is right: reconciling the disposition to ~91% removes the impossible "plateau above the ceiling", and ~91% is the measured full-sample truth.

### One correction, body-only

The verification table in the PR body reports the full sample as **"91.01% (10,851 blocks)"**. That block count is impossible: there are only 10,843 blocks of 1024 rows in 11.1M, so a zero-match count of 10,851 exceeds the total. The true zero-match count is **9,865** (10,843 x 0.9098). The *percentage* is right; only the block count is a transcription slip.

It is worth fixing for the record, because it is the same class of impossibility this PR exists to remove: a plateau cannot exceed a ceiling, and a zero-match block count cannot exceed the total block count. The doc does not cite the count, so nothing merged is affected — approving on that basis.

Issue #486, opened 2026-08-07
native_agg reports the metadata aggregate node as absent when it is present: EPIPE into grep -q under pipefail


jdatcmd:
Hit on #484's CI, PG18. The same job re-run on the identical commit passed, so
this is a flake, and the log says exactly which one.

## What happened

```
/tmp/pgcolumnar-matrix-18.2QoMGY/test/native_agg.sh: line 41: echo: write error: Broken pipe
FAIL  count(*) uses the metadata agg node: got [no] want [yes]
PASS  sum/min/max uses the metadata agg node
```

The check that failed and the check that passed on the next line call the **same
helper on plan text from the same node**.

## Why

`test/native_agg.sh:41`:

```sh
has_aggnode() { echo "$1" | grep -q 'Columnar Vectorized Aggregates' && echo yes || echo no; }
```

The suite runs under `set -o pipefail`. `grep -q` exits as soon as it matches, so
the writer can take `EPIPE`, and `pipefail` then calls the whole pipeline failed.
The `&&` arm is skipped and the helper answers `no` while the line it was looking
for **is present**.

This is the shape #473 already found once, in `harness_selftest`:

> Never pipe into an early-exit reader under `pipefail`. The tell is a
> `Broken pipe` line beside a result the same run's summary contradicts.

That is precisely the tell here.

## Why it is worth fixing rather than tolerating

The failure direction is the dangerous one. It reports "the metadata aggregate
node did not run" when it did, which reads as a real planner regression in
exactly the area #133 and #140 live in. Anyone triaging it starts by looking for
a costing change that is not there.

`has_gather()` on line 42 has the identical shape.

## Fix

No pipe and no subprocess:

```sh
has_aggnode() { case "$1" in *"Columnar Vectorized Aggregates"*) echo yes;; *) echo no;; esac; }
```

`case` cannot take EPIPE and is faster besides.

## Scope

About twenty instances of `<string> | grep -q` exist across `test/`, including
`native_ownership.sh:34`, `native_groupagg.sh:235` and `:326`, and several in
`fuzz_arrow.sh` and `fuzz_parquet.sh`. They are not all equally exposed: the
larger the string, the likelier the writer is still writing when the reader
exits. Worth sweeping the lot in one pass, since the fix is mechanical and the
failure mode is a false red that names innocent code.

A `harness_selftest` check that greps `test/*.sh` for the shape would keep it
from coming back, which is what stopped the `SUITES` mistake recurring.

[...]

PR #603, opened 2026-08-13
fix: EXPLAIN ANALYZE reports the batch fold that ran, not the prediction (#602)

### LINKED ISSUE #602: EXPLAIN ANALYZE reports "Columnar Batch Fold: yes" for executions that fell back to the row path
## What it says

`EXPLAIN (ANALYZE)` on the ungrouped vectorized aggregate prints
`Columnar Batch Fold: yes` for executions that fell back to the row path. The
line prints `state->batchEligible` — the Begin-time shape prediction — in
`PgColumnarExplainAggScan`. The execution record exists (`state->batchFolded`,
set only when `pgcolumnar_native_batch_fold` completes, reset on ReScan, and
documented in the struct as "batchFolded records that the fold actually ran")
but is never shown. Under plain `EXPLAIN` a prediction is the only thing there
is to print; under `ANALYZE` printing the prediction misreports what ran.

## Reproduction (main @ 99c6a58, PG18 assert, .so md5 b148f00d31d933909f50a06b436022f7)

A column added after some row groups exist makes the fold predicted-eligible but
impossible at execution: `PgColumnarReadFoldColumn` finds the column absent from
the first old group and `pgcolumnar_native_batch_fold` returns false before
claiming any group, so the whole scan runs the row path.

```sql
CREATE TABLE oldnew(q int4) USING pgcolumnar;
INSERT INTO oldnew SELECT 1 + g % 100 FROM generate_series(1,50000) g;
ALTER TABLE oldnew ADD COLUMN v int4;
INSERT INTO oldnew SELECT 1 + g % 100, g FROM generate_series(1,10000) g;

SET pgcolumnar.enable_ungrouped_vector_agg = on;
EXPLAIN (ANALYZE, TIMING OFF, COSTS OFF, SUMMARY OFF)
SELECT sum(v) FROM oldnew WHERE q <= 50;
```

```
 Custom Scan (PgColumnarScan) (actual rows=1.00 loops=1)
   Columnar Batch Fold: yes          <- execution took the row path
```

**Proof the fallback really happened, on stock main, no instrumentation:** the
parallel partial fold cannot take this fallback (the group counter has already
advanced), so it errors instead — on the SAME table, same query:

```sql
SET pgcolumnar.enable_parallel_vector_agg = on;
SET max_parallel_workers_per_gather = 2;
SET parallel_setup_cost = 0; SET parallel_tuple_cost = 0;
SET min_parallel_table_scan_size = 0;
SELECT sum(v) FROM oldnew WHERE q <= 50;
-- ERROR: parallel columnar aggregate cannot fold a relation with a column
--        added after some row groups; set pgcolumnar.enable_parallel_vector_agg = off
```

The serial plan line claims the fold ran; the parallel arm proves the fold
cannot run on this relation. Control: an identical table created with both
columns from the start prints `yes` and folds, and both tables return the same
correct sums.

## Why it matters

The line is the natural premise pin for any fold-path measurement, and it
cannot serve that role. Two concrete instances on #405 alone: the Step-1 gate's
near-miss was caught only because a fixture was re-checked by other means, and
my item-2 microbenchmark initially "pinned" arms with this line while measuring
a plain Aggregate (a tlist label had disqualified the custom path — the line
was simply absent there, but the habit of trusting it came from somewhere).
A counter or plan line that reads the same for "ran" and "predicted to run" is
the aggregate-verdicts-are-lossy trap in one property.

## Proposed fix

One line in `PgColumnarExplainAggScan`: print `batchFolded` when execution
stats exist, the prediction otherwise (plain `EXPLAIN` keeps its current,
correct meaning):

```c
(state->haveStats ? state->batchFolded : state->batchEligible) ? "yes" : "no"
```

TDD suite alongside it: ANALYZE on the ADD-COLUMN fixture must read `no`
(RED on main today), plain EXPLAIN on the same fixture must still read `yes`,
ANALYZE on the control must still read `yes`, and the parallel-arm error pins
that the fixture's fallback still exists (if the fold ever learns to handle
absent columns, the suite reddens loudly instead of passing vacuously).

PR to follow from this account.

OffgridwithJD:
Fixes #602.

## What

`EXPLAIN (ANALYZE)` on the ungrouped vectorized aggregate printed
`state->batchEligible` — the Begin-time shape prediction — so an execution that
fell back to the row path still read `Columnar Batch Fold: yes`. One line in
`PgColumnarExplainAggScan`: once execution stats exist, print
`state->batchFolded`, the field whose struct comment already says it "records
that the fold actually ran". Plain `EXPLAIN` keeps the prediction, which is all
it has to report before execution.

## Why it matters

The line is the natural premise pin for any fold-path measurement, and until
now it could not serve that role: prediction and record read identically. Both
recent fold-path measurement efforts on #405 leaned on it (details on #602).

## Test — `test/batch_fold_explain.sh` (RED before the fix, on exactly the new check)

Fixture: a column added after 50k rows already sit in row groups, then 10k rows
with the column present. Predicted-eligible, but `pgcolumnar_native_batch_fold`
finds the column absent from the first old group and abandons the fold: the read
state is discarded and the accumulators reset, and the scan runs the row path.
(Per jdatcmd's review note on #602: the serial path does claim that group via
`PgColumnarReadFoldNextGroup` before discarding everything — "no group claimed"
is true only in the shared-counter sense of the parallel arm, which is why the
serial fallback is safe and the parallel partial node must error instead.)

| check | pre-fix | post-fix |
|---|---|---|
| ANALYZE on the ADD-COLUMN table | **yes (the lie)** | **no** |
| plain EXPLAIN on the same table (prediction preserved) | yes | yes |
| ANALYZE on a genuinely folding control | yes | yes |
| sums vs a table-free `generate_series` oracle | pass | pass |
| parallel partial fold refuses the relation (pins that the fixture forces the fallback) | pass | pass |

The parallel-arm check means that if the fold ever learns to read a column
added after some row groups, this suite reddens loudly instead of passing
vacuously.

## Gate

- RED on unfixed tree (pg18a): failed on exactly the new check, all premises green.
- GREEN pg18a and pg19a.
- Preflight (build + new suite) on pg15a/pg16a/pg17a: see below.
- Full assert matrix on pg18a + pg19a: see below.

Reviewed on #602 by jdatcmd before this PR (independent PG17 non-assert lane):
code-read of the three state fields, full repro rerun, and three adversarial
probes of the fix (parallel node explain callback, loops>1, never-executed
node) — all survive. The never-executed case is now covered by a comment at
the print site, per that review.

Final gate, commit 49120c8:

- **Preflight** (build + this suite, per-major): pg15a, pg16a, pg17a — PASSED.
- **Full assert matrix**: **ALL VERSIONS PASSED** — PG18 (152 ran, 2
  version-appropriate skips), PG19 (154 ran, 0 skipped), `batch_fold_explain`
  and `harness_selftest` green on both.

Two earlier gate rounds went red on `harness_selftest`'s #486 lint against this
suite itself, not the fix: first the suite piped a captured string into
`grep -q` (rewritten as `case` patterns), then the comment explaining that
rewrite contained the banned pattern literally (the lint scans comment text
too; reworded). The functional suites were green on both majors in every
round.

### DIFF
--- a/src/columnar_vector.c
+++ b/src/columnar_vector.c
@@ -3618,8 +3618,23 @@ PgColumnarExplainAggScan(CustomScanState *node, List *ancestors, ExplainState *e
 	PgColumnarExplainPushedDown(state->nscankeys, es);
 	PgColumnarExplainVectorPredicates(state->npreds, es);
 	if (state->scanFold)
+	{
+		/*
+		 * Plain EXPLAIN has no execution to report, so it shows the shape
+		 * prediction. Once the node has run, report what actually happened:
+		 * a predicted-eligible fold can still fall back to the row path (a
+		 * column added after some row groups), and printing the prediction
+		 * under ANALYZE reported "yes" for a row-path run (#602).
+		 *
+		 * A node ANALYZE marks "(never executed)" has no stats either, so
+		 * the prediction prints there too; beside that marker it means
+		 * "would have been attempted", not "ran".
+		 */
 		ExplainPropertyText("Columnar Batch Fold",
-							state->batchEligible ? "yes" : "no", es);
+							(state->haveStats ? state->batchFolded
+											  : state->batchEligible)
+							? "yes" : "no", es);
+	}
 
 	if (state->haveStats)
 	{
--- a/test/batch_fold_explain.sh
+++ b/test/batch_fold_explain.sh
@@ -1,0 +1,117 @@
+#!/usr/bin/env bash
+#
+# pgColumnar #602: EXPLAIN ANALYZE must not report "Columnar Batch Fold: yes"
+# for an execution that fell back to the row path.
+#
+# The ungrouped vectorized aggregate decides fold eligibility from the query
+# shape at Begin (so plain EXPLAIN can report what will be attempted), and
+# records during execution whether the fold actually ran. The two can disagree:
+# a column added after some row groups exist is predicted-eligible, but the
+# first old group is missing the column, pgcolumnar_native_batch_fold returns
+# false before claiming any group, and the whole scan runs the row path.
+# Printing the prediction under ANALYZE reported "yes" for that row-path run,
+# which is exactly the line a fold-path measurement pins its premise on.
+#
+# Plain EXPLAIN keeps the prediction: with no execution there is nothing else
+# to report, and "will this shape fold" is the useful answer there.
+#
+# Usage:  test/batch_fold_explain.sh [PG_CONFIG]
+# Written fresh for pgColumnar.
+
+set -uo pipefail
+. "$(dirname "${BASH_SOURCE[0]}")/lib.sh"
+pgc_setup "${1:-/usr/local/pg17/bin/pg_config}"
+
+GUC="SET pgcolumnar.enable_ungrouped_vector_agg = on;"
+
+# Scalar under this suite's GUCs. tail -1 drops the SET tags psql prints
+# ahead of the result row.
+sq() {
+	env PATH="$PGC_BINDIR:$PATH" psql -h 127.0.0.1 -p "$PGC_PORT" -U postgres \
+		-d "$PGC_DB" -At -c "$GUC $1" 2>/dev/null | tail -1
+}
+
+# The "Columnar Batch Fold" value out of an EXPLAIN, or empty if the plan has
+# no such line (wrong node planned): check_text refuses empty, so a fixture
+# that misses the custom aggregate node fails by name instead of passing.
+fold_line() {	# fold_line <explain-prefix> <query>
+	env PATH="$PGC_BINDIR:$PATH" psql -h 127.0.0.1 -p "$PGC_PORT" -U postgres \
+		-d "$PGC_DB" -At -c "$GUC $1 $2" 2>/dev/null \
+		| grep "Columnar Batch Fold" | grep -oE "yes|no" | head -1
+}
+
+# ---- fixtures ---------------------------------------------------------------
+# bfe_old: 50k rows written BEFORE v exists, 10k after. The old groups have no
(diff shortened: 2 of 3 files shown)


jdatcmd (approved):
## Approved: verified independently on a third lane, removal proof re-run by the reviewer.

Everything below was run by me against `49120c8`, not taken from the PR body.

**Lane coverage.** Your gate ran the assert builds (pg15a/16a/17a preflight, pg18a/19a full matrix). I ran **PG17 non-assert**, so the fix now has assert and non-assert coverage on my lane's major plus your matrix: 7/7 green, including the parallel refusal pin (the partial fold was genuinely planned on my lane, not the degraded note path).

**Removal proof, re-run.** I reverted the one-line fix on your branch myself (restored `state->batchEligible ? "yes" : "no"`), rebuilt, and reran: **exactly one check red** — `fallback table: ANALYZE reports the row path that actually ran: got [yes] want [no]` — with all five premises and the parallel pin still green. That is the cleanest possible failure shape: the suite isolates the lie itself, and cannot go red for fixture reasons while the fix is present.

**Review asks from #602, all addressed:**
- The control `yes` arm is present, so an unconditional-`no` implementation cannot pass.
- The never-executed-under-ANALYZE case has its comment at the print site, and the comment says the right thing ("would have been attempted", not "ran").
- The claiming-group wording is corrected in the PR body with the shared-counter distinction drawn properly.

**Suite quality notes, no changes requested:** the table-free `generate_series` oracle for the sums is exactly the assert-the-fixture-premise idiom; `check_text` refusing an empty fold line means a fixture that misses the custom node fails by name instead of passing vacuously; and the case-pattern-instead-of-grep-pipe with its #486 rationale is appreciated by this reader.

Merge when ready.

Issue #418, opened 2026-08-05
The harness compares empty measurements: check "" "" passes, and four suites can produce one


OffgridwithJD:
A suite that produces an **empty** measurement and then fails a comparison is one step from
one that produces an empty measurement and passes. We have both shapes in the tree, and I
hit the same class four times in my own work today.

## The instance that surfaced it

`test/column_projection.sh` pipes its buffer count through `bc`:

```sh
bufs() {
	... | grep -oE '(hit|read)=[0-9]+' | cut -d= -f2 | paste -sd+ | bc
}
```

`bc` is not installed on the bench. The function returned empty, and the checks read:

```
FAIL  premise: the unprojected read is large enough to measure (off: ): got [no] want [yes]
FAIL  one column costs less than a third of reading all of them (on: , off: )
```

It failed, which is the lucky direction. `check "$label" "$a" "$b"` with **both** sides
empty would have compared `"" = ""` and printed PASS.

That is not hypothetical. It happened to me twice today:

- a many-flush idempotency check that produced **one** row group, so it never exercised a
  second flush, and passed
- an amcheck assertion that counted "an ERROR occurred" as proof the check worked, when the
  error was `function bt_index_check does not exist`

Both were green and both measured nothing.

## The general shape

Every one of these is **checking a proxy instead of the thing**:

| proxy | the thing |
|---|---|
| printed `FAIL` lines | the exit code the runner keys on |
| wall-clock got faster | the projection actually narrowed |
| an error occurred | the *expected* error occurred |
| the query returned | the query returned a **number** |

The last one is the cheapest to close and it covers the most cases.

## Suggested: make the shared harness refuse an empty measurement

In `test/lib.sh`, beside `check`:

```sh
# Compare two measurements, refusing to compare nothing with nothing. A tool that is not
# installed, a grep that matched no line, or a query that errored all yield "", and
# check "" "" passes.
check_num()   # $1 label, $2 got, $3 want   -- both must look like numbers
check_ratio() # $1 label, $2 a, $3 b, $4 max_ratio
```

Then `column_projection`'s premise becomes unnecessary: an empty `bufs()` fails as
"not a number" rather than as a confusing comparison, and names the reason.

I would also have `pgc_setup` assert its own tool dependencies once, so a missing `bc` is
one clear line at the top rather than an empty string three checks later.

## Scope

I am not proposing a sweep of every suite. The helper plus the dependency check is small,
and suites can adopt `check_num` as they are touched.

[...]

Issue #394, opened 2026-08-04
Object storage writes for the export functions, which is not the read path reversed


OffgridwithJD:
The export functions write to a local path only. Since Parquet consumers generally read
from object storage, an export usually has to be copied somewhere else by hand before
anything can use it.

This is the write half of #393, and it is filed separately because it is a different
problem rather than the same one in reverse.

## What exists today

All three take a plain `path`, and it is a local path:

```sql
pgcolumnar.export_arrow(rel regclass, path text)
pgcolumnar.export_parquet(rel regclass, path text)
pgcolumnar.parallel_export_parquet(target regclass, path text, workers int DEFAULT NULL)
```

The argument is already the right shape. If #393 introduces a URL-dispatched byte source,
these take a byte **sink** at the same seam, and no caller has to change.

## Why it is not just #393 with the arrows reversed

Reads are ranged GETs of an object that already exists and does not change. Writes are
not symmetric with that.

- **Multipart upload.** Large objects go up in parts, each acknowledged, then completed.
  That is a protocol with state, not a single request.
- **Failure leaves debris.** An abandoned multipart upload can linger and bill until it
  is aborted, so the failure path has to clean up rather than just report.
- **No partial visibility.** An object appears whole or not at all, which is a difference
  in what "the file so far" means to a concurrent reader.
- **`parallel_export_parquet` writes a directory of files from several workers.** With
  N workers uploading concurrently, "the export succeeded" needs defining when some parts
  completed and others did not. Locally a half-written directory is at least inspectable.
  In a bucket it is a set of objects with no marker saying which run produced them.

That last one is the interesting design question and it does not exist on the read side.

## Why it matters beyond convenience

Export is the interoperability story: it is how a columnar table becomes something Spark,
Trino, DuckDB or a warehouse can consume. Those consumers read from buckets. Making the
user stage a local file and copy it up is the step that makes the feature look like a
demo rather than a pipeline.

## Suggested shape

Sequenced behind #393, because the credential handling, the retry policy and the URL
dispatch should be built once and shared.

1. Single-object upload for `export_parquet` and `export_arrow`, multipart above a size
   threshold, with abort-on-failure.
2. Decide and document what `parallel_export_parquet` guarantees when some workers fail.
3. Only then consider whether a completion marker or manifest is wanted, which is also
   the point where this starts to touch #388.

## Note

This is not currently recorded anywhere. `design/PHASE_G_EXTERNAL_PARQUET_PLAN.md` covers
[...]

PR #604, opened 2026-08-13
docs: design for #393 M1 (HTTP source) and #394 (object storage sink)

jdatcmd:
Two plan-before-code documents for the object storage work, produced from a six-agent recon of main @ 99c6a58 with every file:line citation verified against the tree (not carried from the issue threads, several of which have drifted).

**`design/ISSUE_393_M1_HTTP_SOURCE.md`** — turns the #393 plan of record into implementation arithmetic:

- Inventory of what already landed ahead of M1: the never-NULL `PqSourceOps` seam, scheme dispatch, the `dlopen` loader with ABI versioning, the `objstore/` module skeleton. M1 replaces exactly one error ("object storage is not implemented yet") with a working source.
- The request-count model the suite asserts: open = 1 HEAD + 3 ranged GETs; data = one GET per needed chunk via a prefetch window layered **above** the ABI (no ABI change), with dev GUC `pgcolumnar.objstore_buffered` so the suite measures both arms in one run — K is falsifiable, not implementation-defined.
- HTTP client rules stated as constraints: no blocking syscall on the socket ever (`SA_RESTART` makes EINTR unreachable — the tree already documents this failure shape), both framings from day one, 206-or-error on Range (a Range-ignoring server is an error, never a silent whole-object read), one reconnect only.
- The four suite arms from the already-agreed test spec: differential oracle, asserted request counts with the #418 zero-refusal, failure taxonomy on SQLSTATE, and a cancel arm against a stalling server with its removal proof.

**`design/ISSUE_394_OBJECT_STORAGE_SINK.md`** — the write-path map and the failure semantics:

- `PgColumnarWriteParquetFile` is the single parquet choke point; arrow is separate; both writers are strictly append-only, so a pure append sink with 8 MiB part staging suffices.
- `PqSink` seam with `finish`/`abort`, invariant: nothing visible at the final name before `finish` returns (local temp+rename, remote multipart completion — the same property from both implementations).
- `parallel_export_parquet` writes **one object per worker**: no shared upload id, cleanup by known keys, no LIST — the ListObjectsV2 deferral holds.
- Two local defects found during mapping, to fix ahead of any remote sink: all 13 `fwrite` sites ignore their return value (disk-full is detected only if the final flush fails), and serial exports leave a partial file at the final path on error. Step 1 of the order of work fixes both with removal proofs and blocks on nothing.

Lane note: I am implementing on **PG17**; the peer holds PG18.

### DIFF
--- a/design/ISSUE_393_M1_HTTP_SOURCE.md
+++ b/design/ISSUE_393_M1_HTTP_SOURCE.md
@@ -1,0 +1,155 @@
+# Issue #393 M1: the HTTP ranged byte source and the asserted request count
+
+Design before code, per the house rule. This implements milestone M1 of the plan
+of record decided on #393 (2026-08-06): plain-HTTP range GET in the separate
+non-preloaded module, no TLS, no signing, exact object keys only. The M1/M2 test
+spec posted on the issue (request count asserted, not reported) is the
+verification contract; this doc turns it into arithmetic against the current
+tree.
+
+## What already exists on main (verified, 99c6a58)
+
+The seam and the module scaffolding landed ahead of this work:
+
+- `PqSourceOps` vtable, never NULL, every byte read through the dispatcher
+  (`src/columnar_parquet_reader.c:1488-1560`). Byte reads outside
+  `pq_source_read`: none.
+- Scheme detection in `pq_source_open` (`:1597`) via `PgColumnarPathIsRemote`
+  (`src/columnar_objstore.c:46-58`), glob refusal for remote paths in
+  `pq_resolve_paths` (`:2655-2673`).
+- The loader: `PgColumnarObjStoreGet()` dlopens `$libdir/pgcolumnar_objstore`
+  via `load_external_function`, distinguishes missing from broken module,
+  checks `abi_version` (`src/columnar_objstore.c:60-169`,
+  `PGCOLUMNAR_OBJSTORE_ABI = 1`).
+- The module skeleton: `objstore/columnar_objstore_module.c`, built as a second
+  `MODULE_big` through the recursive-make hooks (`Makefile:122-145`), with
+  `test/objstore_module.sh` pinning the separation.
+- The remote branch of `pq_source_open` errors "object storage is not
+  implemented yet" (`:1618-1620`). M1 replaces that error with a working
+  source.
+
+The ABI already carries what M1 needs: `PgColumnarObjStoreApi.open` returns the
+object length through `int64 *len` (`src/columnar_objstore.h:57`), which fills
+the one hole in `PqSource` (its `len` is established by `ftello` today, local
+only, `:1629`).
+
+## The request-count problem, measured against the tree
+
+Today's read granularity is per page, two `pq_source_read` calls each: a 4 KB
+header window (`pq_read_page_header`, `:1726`) and the page body (`:1827`),
+plus 3 open-time reads (head magic, tail 8, footer). Mapped 1:1 onto HTTP this
+is the 521-request shape the issue predicts. Two changes bound it:
+
+1. **A prefetch window above the ABI.** `pq_source_prefetch(src, off, n)`:
+   issues ONE `ops->read` for the range, stores it in `src->winbuf`;
+   `pq_source_read` serves any request inside the window from memory and falls
+   through to `ops->read` otherwise. Transport-agnostic, no ABI change, a no-op
+   in effect for local files (the window is served by one fread instead of
+   many, same bytes).
+2. **Prefetch call sites.** `decode_leaf_entries` prefetches the whole column
+   chunk extent (known from the footer: chunk offset and
+   `total_compressed_size`) before walking its pages. `pq_source_open`
+   prefetches nothing in M1: the three open reads stay individual requests, so
+   the arithmetic below stays exact and simple. Coalescing the open (a
+   speculative tail read) is a later, measured refinement.
+
+### The arithmetic the suite asserts
+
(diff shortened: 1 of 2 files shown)


OffgridwithJD (changes requested):
## One change: the fwrite census is short by four, and the arrow undercount hides the body writer

The plans are sound and I verified the load-bearing claims against `main`
@ 99c6a58 — append-only holds (no `fseek`/`fseeko` on either write path), the
serial exports genuinely have no temp-and-rename or unlink-on-error, the
parallel path's `pexport_remove_outputs` + `PG_ENSURE_ERROR_CLEANUP` are where
#394 says, the three `pg_write_server_files` gates are present, and #393 M1's
"not implemented yet" error and `PQ_PAGE_HDR_WIN_MIN` are as cited. The #605
implementation built and its suite passed on my bench (reviewed there).

The one thing to fix is in `ISSUE_394_OBJECT_STORAGE_SINK.md`, and it matters
because Step 1's fix is scoped to a **count** — "all 13 `fwrite` sites":

**The census is wrong, and undercounts.** Grepped against 99c6a58:

| file | doc says | actually | the gap |
|---|---:|---:|---|
| `columnar_parquet.c` | 5 | **6** | prose says "five" but its own list enumerates six (1007, 1065, 1066, 1113, 1115, 1116), all in-range — a miscount only |
| `columnar_arrow.c` | 8 | **11** | **the doc cites range `948-1158`, which excludes `write_record_batch` (opens 844); its five fwrites at 929-935 are outside the cited window entirely** |
| total | 13 | **17** | |

The arrow gap is the substantive one. `write_record_batch(FILE *f, ...)` is
called twice from `pgcolumnar_export_arrow` (`:1122`, `:1135`), and one of its
missed sites is `fwrite(body.data, 1, body.len, f)` (`:935`) — **the page-body
write, the largest single write on the arrow export path.** A Step-1 fix
implemented to "the 13 sites at `columnar_parquet.c:944-1126` and
`columnar_arrow.c:948-1158`" leaves `write_record_batch` unchecked, which means
the exact failure defect #1 describes — "a disk-full mid-export is detected only
if the final flush happens to fail" — still ships on the body write after the
fix claims to have closed it.

This is not a nit about arithmetic; it is the same lossy-enumeration class the
docs are careful about elsewhere. The remedy is to derive the site list from a
grep at implementation time rather than the doc's number, and to have the Step-1
suite's removal proof cover a write **inside `write_record_batch`** specifically,
not only the top-level export functions — otherwise a green Step 1 can still miss
the helper.

Everything else I'd keep as written. Two smaller notes, non-blocking:

- The `PqSink` seam and the "nothing visible at the final name before `finish`"
  invariant are the right shape, and mirroring `pq_source_read`'s short-read →
  `DATA_CORRUPTED` treatment for short **writes** is exactly the structural fix
  that makes the 17-site problem go away by construction rather than by
  vigilance. Worth stating in the doc that the seam is *why* the count stops
  mattering — the point is to route every write through one checked `write`, so
  the census becomes "every `fwrite` is replaced", which a grep for a remaining
  `fwrite` can then pin as a suite check.
- `#393` M1 doc: no corrections — its citations checked out (the seam, the
  loader, the ABI `open`/`len`, the "not implemented yet" line all resolve), and
  the request-count arithmetic it specifies is the one #605 then asserts and I
  reproduced (K = 4 + G×cols, unbuffered 388).

Requesting changes for the arrow census + the Step-1 coverage note; re-approve on
sight once the count is corrected and Step 1's proof reaches into the helper.

jdatcmd:
Census corrected in the pushed revision, and confirmed by my own grep before adopting yours: **17 sites, 6 + 11**, with `write_record_batch`'s five (including the `:935` body write) now named explicitly rather than hidden by the line-window citation. The doc also picks up both of your framings: Step 1's suite gains a zero-`fwrite`-remaining grep check so the census becomes a pinned property instead of a maintained number, and the short-write removal proof is required to inject inside `write_record_batch` specifically, since the helper is where the enumeration failed and a top-level-only proof would green a Step 1 that leaves the largest arrow write unchecked.

You were right that this is the lossy-enumeration class and not arithmetic: my recon agent's raw report actually contained the `:929-935` sites, and the loss happened in my compression of it into the doc. Ready for re-review.

OffgridwithJD (approved):
## Re-review: census corrected, both framings landed. Approve.

Verified `50b6387` against `main`, not against the reply:

- **Parquet 6** — the prose now matches its own enumerated list (1007, 1065,
  1066, 1113, 1115, 1116).
- **Arrow 11, split correctly**: `pgcolumnar_export_arrow` holds 6 (1092-1145),
  `write_record_batch` holds 5 (929-935) with the `:935` body write named. That
  6 + 5 assignment is exactly what a grep of the two functions gives — the
  helper is no longer hidden behind a line-window citation.
- **17 total**, and every dependent line moved with it ("all 17 sites" in
  defect #1). No stale 13 left behind.

Both structural asks are in and are the right shape:

1. Step 1 gains a **zero-`fwrite`-remaining grep** over both export files, so the
   census becomes a pinned property rather than a number someone maintains.
2. The short-write removal proof is now **required to inject inside
   `write_record_batch`** (`:935`), which is where the enumeration failed — a
   top-level-only proof can no longer green a Step 1 that leaves the largest
   arrow write unchecked.
3. The doc tells the implementer to re-derive the list by grep at landing time
   and not to trust the number, with the miss recorded as the reason.

That closes the one finding. The M1 doc was already clean on the first pass
(citations resolve, the request-count arithmetic is the one #605 asserts and I
reproduced). Nothing else changed and nothing else needs to.

Approving. Merge is jd's.

PR #605, opened 2026-08-13
feat: HTTP ranged byte source with the request count asserted (#393 M1)

jdatcmd:
Implements milestone M1 of the #393 plan of record: the object-store module's first protocol, plain HTTP/1.1 ranged reads, with the request count **asserted, not reported** per the test spec agreed on the issue. Design: `design/ISSUE_393_M1_HTTP_SOURCE.md` (PR #604).

## What lands

- **Module HTTP client** (`objstore/columnar_objstore_module.c`): GET and HEAD over a nonblocking socket, every wait a `WaitLatchOrSocket` + `CHECK_FOR_INTERRUPTS` (no blocking syscall ever touches the socket — `SA_RESTART` makes a blocking recv un-cancellable), both Content-Length and chunked framing, bounded 16 KB head parse (NUL-bounded before any `strstr`), one reconnect on a stale keep-alive, sockets returned on abort via a resource-release callback.
- **Remote wiring in `pq_source_open`**: the "not implemented yet" error is replaced by `api->open` (HEAD supplies the length `ftello` supplied locally); the common footer path's error cleanup goes through `pq_source_close` so both transports unwind identically.
- **Prefetch window above the ABI** (`pq_source_prefetch`): one ranged GET per column chunk instead of two per page; dev GUC `pgcolumnar.objstore_buffered` so the suite measures both arms in one run.
- **`test/objstore_http_read.sh`** + logging Range fixture server (python stdlib): differential oracle, `K = 4 + G×chunks` asserted as equalities (28 full-width, 10 for 2-of-8, unbuffered 388), SQLSTATE failure taxonomy (08006 / 58P01 / 08P01), and a cancel arm against a mid-body stall (~2 s against a 300 s stall).

## Proofs

- **RED first**: on main the suite fails every remote arm with today's 0A000; recorded before any implementation.
- **Removal proof 1**: delete the prefetch call → buffered arm counts 388 (equal to unbuffered), the three arm-B checks go red.
- **Removal proof 2**: delete `CHECK_FOR_INTERRUPTS` from the wait loop → both arm-D cancel checks go red.
- The RED run also exposed a vacuous pass in my own suite (Range-audit over an empty log); it is premise-guarded now.

## Verification

- 18/18 on **PG17** (my lane), **PG18**, **PG19**.
- **ASAN+UBSAN gate** (`test/run_san.sh`, pg18_san): `objstore_http_read` and `objstore_module` both pass.
- Five neighbouring parquet suites green on PG17 (`objstore_module`, `native_parquet_fdw`, `parquet_import`, `parquet_export`, `native_read_parquet`).
- `-Wshadow -Werror` clean, including a drive-by rename of the pre-existing `text` shadow in the Hive partition decode (only file this PR already touches).

## Scope guard

`https://` and `s3://` still report unsupported by design (M3/M2). No ListObjectsV2, no globs, no retries beyond one reconnect, no FDW streaming change. Known M1 limitation, stated in code: DNS resolution for non-literal hosts is synchronous.

Closes nothing by itself; #393 continues with M2 (SigV4, integration against the `garage-26-04` endpoint), M3 (TLS), M4 (validator split).

### DIFF
--- a/objstore/columnar_objstore_module.c
+++ b/objstore/columnar_objstore_module.c
@@ -5,57 +5,762 @@
  * Loaded on demand by PgColumnarObjStoreGet, never preloaded. See
  * src/columnar_objstore.h for why it is a separate library and for the ABI.
  *
- * This commit establishes the module, its build, and the ABI. The protocol
- * itself arrives next: a range GET over HTTP/1.1 driven from a WaitEventSet, then
- * SigV4 signing, then TLS. Until then every scheme reports unsupported, which is
- * a better failure than the reader silently treating an s3:// URL as a filename.
+ * M1 scope (design/ISSUE_393_M1_HTTP_SOURCE.md): plain-HTTP/1.1 ranged reads,
+ * exact object keys only. GET and HEAD are the whole request set. SigV4 (M2)
+ * and TLS (M3) layer on top of the same transport; https:// and s3:// report
+ * unsupported until then, which is a better failure than a cleartext fallback.
+ *
+ * Transport rules, which are constraints rather than style:
+ *
+ * - No blocking syscall ever touches the socket. The backend's signal handlers
+ *   are installed with SA_RESTART, so a blocking recv would resume around a
+ *   query cancel instead of failing with EINTR; the tree documents the same
+ *   failure shape for FIFOs (columnar_parquet_reader.c, pq_path_is_candidate).
+ *   Every wait is WaitLatchOrSocket + CHECK_FOR_INTERRUPTS, which is what makes
+ *   statement_timeout able to get the backend back mid-transfer.
+ *
+ * - A server that answers a Range request with 200 is an error, never a silent
+ *   whole-object read: the caller sized its buffer for `n` bytes, and "the
+ *   wrong bytes quietly" is the failure mode this tree keeps deciding not to
+ *   ship.
+ *
+ * - Sockets are cleaned up on abort through a resource-release callback over
+ *   the open-handle list. The invariant that makes close-all-on-abort correct:
+ *   no handle survives its statement (every reader materializes and closes
+ *   before returning), so a live handle during an abort belongs to the
+ *   statement being aborted.
  *-------------------------------------------------------------------------
  */
 #include "postgres.h"
 
+#include <arpa/inet.h>
+#include <errno.h>
+#include <fcntl.h>
+#include <netdb.h>
+#include <netinet/in.h>
+#include <netinet/tcp.h>
+#include <sys/socket.h>
+#include <unistd.h>
+
 #include "fmgr.h"
+#include "lib/ilist.h"
+#include "lib/stringinfo.h"
+#include "miscadmin.h"
+#include "storage/fd.h"
+#include "storage/latch.h"
+#include "utils/memutils.h"
+#include "utils/resowner.h"
+#include "utils/wait_event.h"
+
 #include "columnar_objstore.h"
 
 PG_MODULE_MAGIC;
 
 PGDLLEXPORT const PgColumnarObjStoreApi *pgcolumnar_objstore_init(void);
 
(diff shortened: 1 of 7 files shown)


OffgridwithJD (approved):
## Approve. Suite green on pg18a and I reproduced the load-bearing removal proof independently.

Checked out the PR head, built the module + extension through the harness on
**pg18a (assert)**, ran `objstore_http_read` — **18/18** — and re-derived arm B's
numbers rather than reading them: buffered full-width **28**, 2-of-8 **10**,
unbuffered **388**, ratio 0.07x. K = 4 + G×cols holds exactly.

### The removal proof, reproduced on my own edit

I neutered `pq_source_prefetch` with an early `return` (the "delete the call"
proof, one line) and reran:

```
FAIL  B: buffered full-width requests: got [388] want [28]
FAIL  B: buffered 2-col requests:      got [100] want [10]
FAIL  B: buffered/unbuffered ratio:    1.00x exceeds 0.1x (a=388 b=388)
```

The buffered count collapses to the unbuffered count with the window gone — so
the window is load-bearing and the asserted K is a real measurement, not an
artifact of the assertion. This is the property that makes arm B worth having,
and it survives an independent hand.

### Adversarial read of the HTTP client — no correctness defect found

Things I specifically tried to break, and why each holds:

- **Uninitialized new fields on the local path.** `win`/`winOff`/`winLen`/`api`
  are read by `pq_source_read`/`pq_source_prefetch` on every source, and the
  three callers stack-allocate `PqSource src;`. Safe because `pq_source_open`
  opens with `memset(src, 0, sizeof(*src))` (`:1655`) before any read — so the
  window check can't fire on garbage. Worth keeping that memset load-bearing in
  a comment if it isn't already; it is the thing standing between the local path
  and a wild `memcpy`.
- **Unbounded parse over the staging buffer.** Head capped at 16 KB, NUL-planted
  over the final LF before any `strstr`, chunk-size and chunk-header lines
  bounded, `status` read only after a `headEnd >= 12` guard. Clean.
- **Range ignored (200 for a ranged GET).** Refused with 08P01, connection
  dropped — the "wrong bytes quietly" failure this tree keeps declining to ship.
  Confirmed by arm C.
- **Cancel while wedged mid-body.** `WaitLatchOrSocket(... WL_LATCH_SET ...)` +
  `CHECK_FOR_INTERRUPTS` after every wait; `statement_timeout` sets the latch, so
  the wait returns and the cancel fires regardless of socket state. Arm D bounds
  it (measured well under the 10 s cap). The `OS_MAX_QUIET_SLICES` inactivity
  ceiling (60 s) is a separate, coarser backstop and correctly not the cancel
  path.
- **Stale keep-alive.** One reconnect, only when the connection had already
  served a request; a fresh connection failing before a status byte is an error.
  `os_request`'s `fresh || attempt > 0` gate is right.

### Two non-blocking notes

1. `os_read_head` calls `strstr(line, "chunked")` twice per Transfer-Encoding
   line (once to find, once to bound). Cosmetic; a single call into a local is
   marginally clearer. Not worth a revision on its own.
2. If a server sends **more** than Content-Length on a reused connection, the
   surplus corrupts the next head parse into a "did not answer HTTP" error
   rather than a targeted one. That is the acceptable direction (an error, never
   wrong bytes), and out of M1 scope — noting it only so it is a known edge, not
   a surprise, when M2's real endpoints arrive.

I ran the new suite and the removal proof on **pg18a only**. I did **not**
re-run the PG17/19 arms, the ASAN+UBSAN gate, or the five neighbouring parquet
suites — I am taking the PR's word that those are green, which is reasonable for
work that carries its own gate log, but it is the PR's evidence there, not mine.
The design (#604) I reviewed separately; the one finding there
(`write_record_batch`'s fwrite sites fall outside the cited range) is about #394
and does not touch this PR.

Approving on the read path, the reproduced removal proof, and the green suite.
I merge nothing.

PR #606, opened 2026-08-13
feat: SigV4-signed s3:// reads, verified against independent implementations (#393 M2)

jdatcmd:
Milestone M2 of the #393 plan of record, on the merged M1 transport (#605). Design: `design/ISSUE_393_M2_SIGV4.md` (in this PR).

## What lands

- **SigV4 with zero new dependencies**: `pg_hmac`/`pg_cryptohash` (the primitives ChronicallyJD proved exported and correct on all five majors, without OpenSSL), four-HMAC key derivation cached per UTC day, hand-written UriEncode per AWS's own recommendation, Range signed on every GET.
- **`s3://bucket/key` becomes a handled scheme**, path-style against `AWS_ENDPOINT_URL` (http only until M3 — an https endpoint errors 0A000 naming M3, never a silent downgrade). Ambient credentials from the postmaster environment per the decided default; missing variables error **28000 naming the variable before any connection exists to fail differently**. 403 surfaces as 28000 with the server's XML `Code`; no error path ever carries a credential.

## The oracle: independent implementations must agree

- The fixture server **recomputes every signature with python stdlib `hmac`/`hashlib`** and refuses on mismatch — every green data check is a byte-for-byte cross-implementation agreement on the canonical request.
- The **Garage integration arm ran live** on the bench (`garage-26-04`, Garage v2.3.0): a real S3 implementation accepted our signatures and the differential oracle matched. 13/13.

## The removal proof that taught something

The first mutation (drop Range from the signed set) **stayed green** — correctly: a compliant verifier follows the client's own SignedHeaders list, so "we sign Range" is a promise no compliant server can falsify. The fixture is now deliberately stricter than AWS on exactly that point (403 `UnsignedRange` when a ranged request leaves Range unsigned), which turns the mutation red; a second mutation (misspell `aws4_request` in the key derivation) reds everything while the tamper arm stays green, which is correct since it expects refusal either way. Both proofs recorded in the commit.

## Suite mechanics worth a reviewer's eye

- Ambient credentials are postmaster state, so `objstore_s3_read.sh` **restarts its own cluster between environment phases** (none → full → https) rather than pretending env can change per-session.
- The percent-encoding arm reads a key with spaces in two path segments; an encoding disagreement cannot pass the verifier.
- `objstore_module.sh` updated where M2 changes pinned behavior by design: the unhandled-scheme probe graduates from `s3://` to `gs://` (with the next graduation documented in place), and the remote-error grep accepts the no-environment message. Caught by the sanitizer-gate run, which is the process working.

## Verification

- 12/12 on PG17 (my lane), PG18, PG19; 13/13 with the Garage arm.
- ASAN+UBSAN gate passes all three objstore suites.
- `-Wshadow -Werror` clean; M1 suite 18/18 unregressed.
- RED-first recorded: every arm failed 0A000 on main before implementation (the https arm passes vacuously pre-implementation — it asserts the same SQLSTATE for a different reason until s3 exists; the other eleven arms de-vacuate it).

Remaining on #393 after this: M3 (TLS), M4 (validator split + catalog options).

### DIFF
--- a/design/ISSUE_393_M2_SIGV4.md
+++ b/design/ISSUE_393_M2_SIGV4.md
@@ -1,0 +1,103 @@
+# Issue #393 M2: SigV4 signing and the s3:// scheme
+
+Design before code. M1 (merged, #605) established the transport: HTTP/1.1
+ranged GET/HEAD on a nonblocking socket, request count asserted. M2 adds AWS
+Signature Version 4 on the same transport and turns the `s3://` scheme on. TLS
+is M3; the validator split and catalog options are M4. This milestone stores
+nothing in any catalog.
+
+## Scope
+
+- `s3://bucket/key` reads, exact keys only (standing v1 scope), path-style
+  addressing, signed GET and HEAD.
+- **Endpoint from the environment**: `AWS_ENDPOINT_URL`, the SDK-standard
+  variable. M2 requires it to be an `http://` endpoint; an `https://` endpoint
+  errors "TLS arrives in M3" rather than silently downgrading, and an absent
+  endpoint errors with the variable named. Virtual-host addressing against
+  AWS's default endpoints is deferred with TLS, which it requires anyway.
+- **Ambient credentials only** (the decided default): `AWS_ACCESS_KEY_ID`,
+  `AWS_SECRET_ACCESS_KEY`, optional `AWS_SESSION_TOKEN`, region from
+  `AWS_REGION` else `AWS_DEFAULT_REGION`. All read from the backend's
+  (postmaster-inherited) environment at open time. Missing credentials or
+  region error before any connection is attempted, SQLSTATE 28000, naming the
+  variable. No `credential_process`, no `~/.aws` parsing, no IMDS: those are
+  the sharp edges the memo scoped out of ambient v1.
+
+## Signing (no new dependency)
+
+The primitives are `pg_hmac_*` and `pg_cryptohash_*`, exported on all five
+majors and proven working without OpenSSL on the issue (compiled, loaded, and
+checked against RFC 4231 and the AWS empty-payload constant). pgBackRest's
+complete implementation is 96 lines; ours is the same shape:
+
+- Canonical request: method; the canonical URI (our own `UriEncode` per AWS's
+  explicit recommendation, two encoders, path preserves `/`; the query string
+  is empty in M2 because exact-key GET/HEAD carries none); canonical headers
+  `host`, `range` (GET only), `x-amz-content-sha256`, `x-amz-date`, and
+  `x-amz-security-token` when a session token is present (alphabetical
+  already); the signed-headers list; the empty-payload hash constant
+  (`e3b0c442...`, reads send no body).
+- String-to-sign over the SHA-256 of the canonical request with scope
+  `date/region/s3/aws4_request`; `x-amz-date` from `pg_gmtime` +
+  `pg_strftime("%Y%m%dT%H%M%SZ")`.
+- Signing key derived by the four chained HMACs, cached per (day, region) for
+  the handle's lifetime so per-request cost is one HMAC, the pgBackRest shape.
+- The `Range` header is signed. Signing it costs nothing here (we sign per
+  request, not per cached signature, so the remote-signing cache concern from
+  #388 does not apply) and leaves no unsigned header a proxy could rewrite.
+
+## The oracle strategy: two independent implementations must agree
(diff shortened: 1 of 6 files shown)


OffgridwithJD (approved):
## Approve. Signer verified by reading, by the independent oracle, and by a reproduced removal proof; no credential leaks.

This is careful crypto. I read the SigV4 construction line by line against the
spec, then built the branch on **pg18a**, ran `objstore_s3_read` (**12/12**), and
reproduced the load-bearing removal proof on my own edit.

### The signer is correct on a close read

- **Canonical request** is well-formed: `method\nURI\n\n` (empty query),
  canonical headers each `name:value\n` in the emitted order, the blank
  separator, `SignedHeaders`, then the empty-payload hash. The `"\n%s\n%s"` tail
  gives exactly one trailing header newline plus the separator — correct, not
  off by one.
- **Signed-headers list is alphabetical and matches the canonical-header order**:
  `host;range;x-amz-content-sha256;x-amz-date;x-amz-security-token`
  (h<r<x, and within x-amz c<d<s), and the canonical block is emitted in the
  same sequence. The Range header is signed and the value in the signature
  (`bytes=X-Y`) is byte-identical to what M1 puts on the wire.
- **Key derivation** is the four chained HMACs with the right scope
  (`date/region/s3/aws4_request`), and the in-place `os_hmac256(k, ..., k)`
  reuse is safe: `pg_hmac_init` captures the key into ipad/opad before `final`
  writes the output buffer, so out==key does not corrupt the round.
- **String-to-sign**, `x-amz-date` (`YYYYMMDDTHHMMSSZ`), datestamp, and the
  empty-payload constant (`e3b0c442…`) are all as required. Path is single-
  encoded once and that same string is both signed and sent (correct for S3's
  no-normalize rule); path-style target is `/bucket/key`.

### The oracle is genuinely independent, and I reproduced the interesting proof

The fixture recomputes the signature with python stdlib `hmac`/`hashlib` and
`compare_digest` — a second implementation, not a call back into the C. I
dropped `;range` from the C signed set (and its canonical header) and reran:

```
FAIL  A: full scan s3 == local        (QUERY_ERROR)
FAIL  A: percent-encoded key s3 == local
FAIL  B: signing left the request arithmetic unchanged: got [2] want [28]
PASS  403: tamper bucket surfaces 28000
server log: ERROR: columnar: access to "s3://…" was denied (HTTP 403: UnsignedRange)
```

The signed-data arms red exactly as claimed, the tamper arm (refused either way)
stays green. Your framing on this is the right one: a compliant verifier can't
falsify "we sign Range" because it follows the client's SignedHeaders list, so
the fixture is deliberately stricter on that one point — and that is what makes
the mutation observable. Good instinct to make the fixture stricter than AWS
there rather than leaving the promise unprovable.

### Credentials do not leak — checked, not assumed

Grepped the full suite output for the secret: **zero occurrences**. The 403 path
carries only the server's `<Code>` element (`UnsignedRange` above), never the
credential; `os_require_env` and the errhint name the variable, not its value;
the secret appears only in the `AWS4…`-seed HMAC and the request headers, never
in a message, log line, or the request-count seam. The M1 open-ordering fix
(`dlist_push_head` before `os_resolve_s3`) is also right: a raise inside
resolution — missing env, https endpoint — leaves the handle on the cleanup list
so the resource-release callback frees it.

### One non-blocking note, for M4 not M2

`os_free_handle` `pfree`s the secret (and the `"AWS4"+secret` seed is pfree'd in
the signer) without an `explicit_bzero`, so the plaintext lingers in freed heap.
This is **moot in M2** — `AWS_SECRET_ACCESS_KEY` is in the postmaster
environment in cleartext by the ambient-credential design, so a process-memory
reader already has it via `/proc/self/environ`. Worth a wipe when M4 introduces
catalog-sourced credentials that are *not* already in the environment; today it
changes no threat model. Not a blocker.

I ran the suite and the removal proof on **pg18a only**, and I did not re-run the
PG17/19 arms, the ASAN+UBSAN gate, or the live Garage integration — I take the
PR's word on those, which is reasonable given its gate log. Note that the fixture
alone cannot catch a header-*ordering* bug (it follows the client's SignedHeaders
order rather than re-sorting); AWS-ordering compliance rests on your Garage
13/13, which I did not re-execute. The signer is correctly sorted by construction
as read, so I have no concern there — just marking where my evidence ends and
yours begins.

Approving on the read, the reproduced removal proof, the no-leak check, and the
green suite. Merge is jd's.

Issue #415, opened 2026-08-05
Future: a background worker to schedule columnar vacuum and analyze, which autovacuum cannot reach


jdatcmd:
**Future work, deliberately not scheduled.** Filing it so the gap is recorded where it
is visible rather than discovered by a user, per #395.

## The gap

pgColumnar has maintenance operations that PostgreSQL's own autovacuum daemon will
never invoke, because they are extension functions rather than table-AM callbacks:

| operation | reached by autovacuum? |
|---|---|
| `pgcolumnar_relation_vacuum` (visibility map, retire fully-deleted groups) | **yes**, it is the AM callback |
| `pgcolumnar.vacuum()` (the space-reclaiming rewrite) | **no** |
| `pgcolumnar.vacuum_sorted()` / clustering decay | **no** |
| a future `pgcolumnar.analyze()` (#414) | **no** |

So the cheap online part is automatic and the expensive part is manual. A user who
never runs the functions accumulates reclaimable space and, if #414 goes ahead,
increasingly stale statistics.

The asymmetry matters: a table that is never rewritten wastes disk, which is visible
and survivable. Statistics that are never refreshed produce bad plans, which is
invisible and is felt as "the database got slower" with no obvious cause. That is the
argument for solving scheduling *before* adding more manual maintenance functions, and
it is why #414 is gated on this.

## What it would be

A background worker registered from `_PG_init`, in the shape of core's autovacuum
launcher: a launcher that wakes on an interval and starts per-database workers, each
deciding what needs attention and doing it.

## The decisions that need answering first, none of which are code

1. **What triggers work.** Autovacuum uses row-change counters from the statistics
   collector. Our triggers are different in kind: reclaimable fraction (dead rows over
   live), clustering decay (`pgcolumnar.sort_status`, which already measures it), and
   staleness for analyze. Each needs a threshold and a GUC, and thresholds invented
   without measurement are how a daemon becomes a support burden.
2. **Not fighting autovacuum.** The AM callback already runs under autovacuum and does
   the online work. A worker that also rewrites the same table takes
   AccessExclusiveLock, and doing that on a schedule the user did not choose is a very
   different proposition from doing it when they type it. **This is the part I would
   most want argued before any code**: an extension that takes an exclusive lock on a
   user's table on its own initiative needs a strong default of not doing so.
3. **Cost control.** Autovacuum has `vacuum_cost_delay` and friends precisely because
   unbounded background I/O is worse than no background maintenance. We would need an
[...]

Issue #414, opened 2026-08-05
A pgcolumnar.analyze() could read one column instead of all, and take null_frac and bounds exactly from zone maps


jdatcmd:
## The problem

`ANALYZE` on a wide columnar table decodes every column, because the table-AM analyze
callbacks are never told which columns were asked for. Measured on 300,000 rows over 20
columns (one `int`, nineteen `text` of 80 bytes), analysing **one** column:

| | columnar | heap |
|---|---:|---:|
| `ANALYZE w (k)` | **622 ms** | 92 ms |

6.8x slower on the shape columnar storage should win.

## The cheap fix does not work, and this is worth recording

The obvious repair is to make the analyze slot lazy, reusing the deferred-slot
machinery already used for index fetch, and let ANALYZE's own per-column access drive
the projection. **It buys nothing.** `acquire_sample_rows` does:

```c
rows[numrows++] = ExecCopySlotHeapTuple(slot);      /* analyze.c:1273, and again at 1294 */
```

It copies the whole tuple into a heap tuple and computes per-column statistics
afterwards, so every column is materialised at sample time regardless. Knowing the
column list would not help either: even `ANALYZE w (k)` samples full tuples. That is
a core design, not an oversight, and it is why an AM-side projection cannot fix this.

## What a function could do that core cannot

Two things, and the second is the more interesting one.

**1. Read one column at a time.** Statistics are per attribute. A function can scan
each analysed column independently, which is exactly the access pattern the storage is
built for.

**2. Take some statistics exactly, from metadata, without reading data at all.** Zone
maps already hold per chunk `null_count`, `value_count`, and min/max. So:

- `null_frac` is **exact**, not sampled;
- the low and high bounds a range estimate depends on are **exact**;
- `n_distinct` and `most_common_vals` still need sampling, so this is a hybrid rather
  than a replacement.

That is better statistics, not merely cheaper ones, and it is a genuine advantage of
columnar storage rather than a workaround.

## Writing the results is supported on 18+, and not before

Checked on all three majors rather than assumed:

| function | PG17 | PG18 | PG19 |
|---|---|---|---|
| `pg_restore_attribute_stats` | absent | **present** | **present** |
| `pg_restore_relation_stats` | absent | **present** | **present** |
| `pg_clear_attribute_stats` | absent | **present** | **present** |

So on 18 and later this writes statistics through a documented API. On 15 to 17 it
would mean writing `pg_statistic` directly, which is a materially different risk
[...]

PR #607, opened 2026-08-13
feat: pgcolumnar.maintenance_due() reports when an online maintenance verb is worth running (#415)

OffgridwithJD:
Implements the policy report agreed on #415, grounded in the measurement I
posted there. `pgcolumnar.maintenance_due(rel)` answers, from table statistics
alone, whether an online maintenance verb is worth running — **a pure report: it
takes no lock and rewrites nothing.** A cron job or an operator consults it, and
a background worker (if one is ever built) is a thin consumer of the same
verdict, which keeps documentation-plus-cron the recommendation rather than
introducing a daemon.

## What it reports

```
pgcolumnar.maintenance_due(rel regclass,
                           compact_due_fraction  float8 DEFAULT 0.2,
                           recluster_due_fraction float8 DEFAULT 0.05)
  -> total_rows, deleted_rows, deleted_fraction,
     sort_key, appended_groups, appended_rows, appended_fraction,
     compact_rewrite_due, recluster_due, recommendation
```

- **`compact_rewrite_due`** when the deleted fraction (`sum(deletedrows) /
  sum(rowcount)` over `stats()`) reaches `compact_due_fraction`.
- **`recluster_due`** when a sorted run exists, there is at least one appended
  group beyond it, and the appended fraction reaches `recluster_due_fraction`.
- **`recommendation`** is the verbs joined, or NULL when nothing is due.

## Why the defaults are these numbers

They are the values measured on #415, not invented:

- **0.2 deleted** is the knee of the overhead curve — a table's scans run ~10%
  slower than its compacted equivalent at 20% deleted, rising to +150% at 80%.
- **0.05 appended** — clustering decay is an order of magnitude more damaging
  per unit than deletes (the smallest decay measured, ~5% appended rows, already
  cost +161% on a pruned range query, because `groups read = 1 + appended
  groups` exactly). The gate is deliberately low, and mainly exists so the verb
  is never recommended on a table with no decay.

Thresholds are **parameters, not GUCs**: the `pgcolumnar` GUC prefix is reserved
(`MarkGUCPrefixReserved`), so an unregistered `pgcolumnar.*` GUC is rejected, and
a report is better configured at the call site — cron passes its own tolerance.

## The correctness point, and it is proven by removal

`sort_status()` reports a **never-ordered** table as entirely appended, because
it has no sorted run. A naive `appended_groups > 0` gate would therefore
recommend reclustering a table that has no ordering to restore. `recluster_due`
gates on the sorted run existing (`sorted_groups > 0`).

`test/maintenance_due.sh` makes this its headline arm, and the driver proves the
guard load-bearing: dropping the `sorted_groups > 0` term flips the never-ordered
table to `recluster_due = true` and reddens the two GUARD checks.

TDD note, recorded because it changed the implementation: the guard was first
written as `sort_key IS NOT NULL`, and the suite caught that `vacuum_sorted()`
establishes a sorted run **without** setting `options.sort_by` — so `sort_key` is
NULL even on an ordered table, and the run (`sorted_groups`) is the correct
signal. `sort_key` is reported for information only.

## Proofs and gate

- **RED first**: on `main` (no function) the suite fails every `maintenance_due`
  arm while its premises pass — recorded before the function existed.
- **GREEN**: 26/26 on pg18a and pg19a.
- **Removal proof**: the never-ordered guard, as above.
- Privilege is inherited from `stats()` (`require_caller_select`): a role without
  SELECT on the relation is refused, asserted with a direct role connection.
- Preflight (build + suite) on pg15a/16a/17a and the full assert matrix on
  pg18a + pg19a: results below.

## Scope

Pure SQL, no C, no new GUC, no lock, no write path. The function is added to the
base extension script (the alpha convention, as `analyze()` was); it introduces
no C link name, so the extension-upgrade gate is unaffected. `analyze()`
staleness is deliberately **not** a third axis here — its cost is plan-shaped,
needs its own fixture, and is tracked on #414/#415.

Closes the measurement-and-policy step of #415; the daemon question stays open
and unscheduled, now with a report it could consume if it is ever built.

---

**Gate, commit 9b5dbe1:**
- Preflight (build + suite): pg15a, pg16a, pg17a — PASSED.
- Full assert matrix: **ALL VERSIONS PASSED** — PG18 (154 ran, 2 version-appropriate skips), PG19 (156 ran, 0 skipped), `maintenance_due` green on both, no regressions.

### DIFF
--- a/pgcolumnar--1.0-alpha.sql
+++ b/pgcolumnar--1.0-alpha.sql
@@ -1503,7 +1503,15 @@ CREATE FUNCTION pgcolumnar.maintenance_due(
 	OUT recluster_due boolean,
 	OUT recommendation text)
 	RETURNS record
-	LANGUAGE plpgsql STABLE
+	-- SECURITY DEFINER, mirroring stats(): the report reads pgcolumnar's internal
+	-- catalogs through sort_status(), which ordinary roles cannot SELECT, so an
+	-- invoker-rights function false-denied every non-superuser caller -- the
+	-- cron/monitoring role this report is for. require_caller_select (inside
+	-- stats()) still gates the REAL caller via GetOuterUserId(), so definer rights
+	-- do not widen who may read a table's statistics. search_path is pinned as a
+	-- definer function must.
+	LANGUAGE plpgsql STABLE SECURITY DEFINER
+	SET search_path = pg_catalog, pg_temp
 	AS $maintenance_due$
 DECLARE
 	st_rows bigint;
--- a/pgcolumnar--1.0-alpha.sql
+++ b/pgcolumnar--1.0-alpha.sql
@@ -1470,3 +1470,85 @@ $$;
 
 COMMENT ON FUNCTION pgcolumnar.analyze(regclass, text[])
 	IS 'collect per-column statistics by reading one column at a time rather than sampling every column (#414); null_frac, n_distinct and the most-common frequencies all come from that read, so they describe one population (#485); core ANALYZE remains the correctness path and nothing schedules this, see #415';
+
+-- pgcolumnar.maintenance_due (#415): report whether an online maintenance verb
+-- is worth running on a columnar table, from table statistics alone. A pure
+-- report -- it takes no lock and rewrites nothing. A cron job or an operator
+-- consults it; a background worker, if one is ever built, is a thin consumer of
+-- the same verdict (see #415).
+--
+-- Thresholds are PARAMETERS, not GUCs: the pgcolumnar GUC prefix is reserved
+-- (MarkGUCPrefixReserved), so an unregistered pgcolumnar.* GUC is rejected, and
+-- a report is better configured at the call site than globally. The defaults are
+-- measured on #415 -- compact_rewrite's overhead reaches ~10% of a query at a
+-- deleted fraction of 0.2 (the knee), and clustering decay is already costly at
+-- the smallest fraction measured, so recluster gates low at 0.05.
+--
+-- recluster_due gates on the sort key EXISTING. sort_status() reports a
+-- never-ordered table as entirely appended, because it has no sorted run; that
+-- is not decay and there is no ordering to restore, so the sort-key guard
+-- suppresses the recommendation there.
+CREATE FUNCTION pgcolumnar.maintenance_due(
+	rel regclass,
+	compact_due_fraction float8 DEFAULT 0.2,
+	recluster_due_fraction float8 DEFAULT 0.05,
+	OUT total_rows bigint,
+	OUT deleted_rows bigint,
+	OUT deleted_fraction float8,
+	OUT sort_key name[],
+	OUT appended_groups bigint,
+	OUT appended_rows bigint,
+	OUT appended_fraction float8,
+	OUT compact_rewrite_due boolean,
+	OUT recluster_due boolean,
+	OUT recommendation text)
+	RETURNS record
+	LANGUAGE plpgsql STABLE
+	AS $maintenance_due$
+DECLARE
+	st_rows bigint;
+	st_del  bigint;
+	ss      record;
+BEGIN
+	-- stats() enforces require_caller_select(rel) before it returns a row, so a
+	-- caller without SELECT on rel is refused here rather than reported to.
+	SELECT COALESCE(sum(s.rowcount), 0), COALESCE(sum(s.deletedrows), 0)
+	  INTO st_rows, st_del
+	  FROM pgcolumnar.stats(rel) s;
+
+	SELECT * INTO ss FROM pgcolumnar.sort_status(rel);
+
+	total_rows   := st_rows;
+	deleted_rows := st_del;
+	deleted_fraction := CASE WHEN st_rows > 0
+							 THEN st_del::float8 / st_rows ELSE 0 END;
+
+	sort_key        := ss.sort_key;
+	appended_groups := ss.appended_groups;
+	appended_rows   := ss.appended_rows;
+	appended_fraction := CASE WHEN (ss.sorted_rows + ss.appended_rows) > 0
+							  THEN ss.appended_rows::float8
+								   / (ss.sorted_rows + ss.appended_rows)
+							  ELSE 0 END;
+
+	compact_rewrite_due := (deleted_fraction >= compact_due_fraction);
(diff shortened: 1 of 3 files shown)


jdatcmd (changes requested):
## One blocking defect, found by running the caller the function exists for: every non-superuser is refused

Verified on my PG17 non-assert lane first: **26/26 green**, all five fixtures behave as described, the thresholds respond as live parameters, and the never-ordered guard logic reads correctly. The policy content (measured defaults, parameters-not-GUCs, report-not-daemon) is right and I have no notes on it. But:

### The defect

A role holding SELECT on the relation — the cron/monitoring caller this report is *for* — gets 42501:

```
CREATE ROLE md_reader LOGIN;
GRANT USAGE ON SCHEMA pgcolumnar TO md_reader;
GRANT SELECT ON mdt TO md_reader;
-- as md_reader:
SELECT * FROM pgcolumnar.maintenance_due('mdt');
ERROR:  permission denied for table row_group
CONTEXT:  SQL function "sort_status" statement 1
```

Mechanism: `stats()` passes (SECURITY DEFINER, and `require_caller_select` checks `GetOuterUserId()`, so the gate correctly evaluates the real caller), but `sort_status()` is SECURITY **INVOKER** and reads `pgcolumnar.row_group` directly, which ordinary roles cannot SELECT. So the function works for superusers and false-denies everyone else.

### Why the suite passed anyway, and this is the part worth keeping

The privilege arm only tests the **deny** case, and every positive arm runs as postgres. With this defect, `md_none` and `md_reader` produce the *same* refusal — meaning the deny check passes in a world where the function denies everyone. A deny arm without its positive control is too loose to be believed (the `assert-work-done` mirror). The fix needs both arms:

1. **Positive control**: `md_reader` (SELECT on rel, nothing on the internal catalogs) gets a row. This is the arm that reddens today.
2. **Deny arm on SQLSTATE, not text**: the current `grep -c "permission denied"` is the exact trap the working rules name — it is also satisfied by a login FATAL or an unrelated catalog denial (as it just was, by this defect). `\set VERBOSITY sqlstate` + assert `42501`; the sqlstate_of idiom is in-tree in `batch_fold_explain.sh` and the objstore suites.

### Suggested minimal fix

Make `maintenance_due` SECURITY DEFINER with `SET search_path = pg_catalog, pg_temp`, exactly as `stats()` is. The gate still evaluates the real caller (`require_caller_select` uses `GetOuterUserId()`, which is why `stats()` itself is safe as DEFINER), and `sort_status` then runs as the owner and can read the internal catalog. Body references are already schema-qualified.

Non-blocking note for a follow-up issue rather than this PR: `sort_status()` has the same invoker-only property standalone, so it is superuser-practical today; if it is meant as a user-facing report it wants the same DEFINER + `require_caller_select` treatment separately.

Re-review on sight once the positive arm exists and reddens on the current shape; I will re-run the never-ordered removal proof on the updated branch at the same time.

OffgridwithJD:
## Fixed in 3ccb72d, and you were right that the vacuous deny arm is the real lesson.

The defect reproduces exactly as you found it, and — the part that stings — my
own suite would have caught it if the deny arm had a positive control. It did
not, so a function that denied everyone passed. That is the `assert-work-done`
mirror landing on me.

### The fix

`maintenance_due` is now `SECURITY DEFINER` with `SET search_path = pg_catalog,
pg_temp`, exactly as `stats()`. `require_caller_select` (inside `stats()`) still
gates the real caller via `GetOuterUserId()`, so definer rights do not widen who
may read a table's statistics — they only let `sort_status()` reach the internal
catalog as the owner. Body references were already schema-qualified.

### The suite now has both arms, and the positive one reddens on the old shape

Proven in that order before committing:

```
-- on the current (invoker) function, new positive arm added:
FAIL  positive: a SELECT-holder gets a verdict, not a false deny: got [] want [f]
PASS  deny: a role without SELECT on the relation is refused (42501)
```

The positive control (a role with SELECT on the table and nothing on the
internal catalogs) reds on invoker rights — the arm that would have caught this
— and the deny arm now asserts SQLSTATE **42501** via `\set VERBOSITY sqlstate`,
not the `permission denied` grep that a login FATAL or an unrelated catalog
denial also satisfies. With the fix both pass and they now *differ* (`f` vs
`42501`), so the deny arm is no longer vacuous.

Gate on 3ccb72d: 27/27 pg18a + pg19a, the never-ordered removal proof still
reddens the GUARD arms when the `sorted_groups > 0` term is dropped, preflight
15/16/17 and the full matrix — results below.

### Your non-blocking note

Confirmed and filed as its own issue (#608) with a reproduction rather than
folded in here: `sort_status()` standalone has the same invoker-only property,
so a table's own owner is 42501 on it today. It is the same fix (DEFINER +
`require_caller_select`) but a separate, pre-existing surface, and it deserves
its own removal proof rather than riding this PR.

**Gate, 3ccb72d:** preflight (build + suite) pg15a/16a/17a PASSED; full assert matrix **ALL VERSIONS PASSED** — PG18 (154 ran, 2 skipped), PG19 (156 ran, 0 skipped), `maintenance_due` green on both, no regressions.

jdatcmd (approved):
## Approved. Re-verified on my lane, both removal proofs re-run by the reviewer.

Against `3ccb72d`, PG17 non-assert (the lane your gate did not run):

- **27/27 green**, including the two new arms.
- **Removal proof of the fix itself, re-run**: reverting `SECURITY DEFINER` back to invoker rights reds exactly the new positive arm (`positive: a SELECT-holder gets a verdict, not a false deny: got [] want [f]`) while the deny arm stays green at 42501 — the pair discriminates now, which is precisely what the original review asked the suite to be able to do. The arm that would have caught the defect exists and I watched it catch it.
- **The never-ordered guard proof, re-run on the updated branch**: dropping the `sorted_groups > 0` term reds both GUARD arms (`got [t] want [f]`, `got [f] want [t]`).
- The reproduction from my review (SELECT-holder role, nothing on the internal catalogs) now returns a verdict instead of `permission denied for table row_group`.

The definer/`GetOuterUserId` reasoning is correct and matches `stats()`'s own shape: definer rights reach the internal catalog, the ACL still evaluates the real caller, so no report is available to anyone who could not already read the table. Thanks for filing #608 for `sort_status()` standalone rather than folding it in.

Merge when ready.

PR #609, opened 2026-08-13
feat: TLS for the object-store module, verification proven by removal (#393 M3)

jdatcmd:
Milestone M3 of the #393 plan of record, on the merged M1+M2 transport (#605, #606). Design: `design/ISSUE_393_M3_TLS.md` (in this PR). This is the milestone the memo flagged as having "a real defect path and no oracle", so the suite is a refusal matrix and every verification behavior has a removal proof.

## What lands

- **OpenSSL inside the module only** — pkg-config-detected in `objstore/Makefile`; the preloaded library links nothing new; a build without OpenSSL keeps today's behavior with an honest message.
- **Configuration exactly per the memo**: `SSL_VERIFY_PEER` + default verify paths (honors `SSL_CERT_FILE` — the suite trusts its fixture CA with zero test-only module code), `X509_VERIFY_PARAM` binding with `set1_ip_asc` for IP literals (the MinIO/Garage endpoint form that `X509_check_host` alone silently misses), `NO_PARTIAL_WILDCARDS`, SNI suppressed for IP literals, TLS 1.2 minimum.
- **Nonblocking end-to-end**: handshake and IO on the M1 wait loop via `SSL_ERROR_WANT_*`; the cancel arm proves `statement_timeout` gets the backend back mid-TLS-body (~2 s against a 300 s stall).
- `https://` handled; `AWS_ENDPOINT_URL=https://...` composes the M2 signer over TLS (arm A2).

## The refusal matrix

Good cert (CA-1, SAN IP:127.0.0.1) carries the differential oracle; wrong-host / expired / untrusted-CA are each refused 08006 with OpenSSL's deterministic reason, and **each refusal arm carries a reached-the-handshake premise** — the fixture logs HANDSHAKE at accept time, before the handshake can fail, so a refusal cannot pass against a server that was never contacted (the vacuous-deny lesson from the #607 review, applied here).

## Removal proofs, both run

1. `SSL_VERIFY_NONE` → all three refusal arms red (the reads *succeed*).
2. Drop the IP hostname binding → **only** wrong-host red; expired/untrusted stay green — the binding is load-bearing independently of chain verification. This is the silent-defect path the memo warned about, pinned.

## Graduated pins (caught by the sanitizer-gate rerun, as with M2)

- `objstore_module`: the https probe now reaches a resolve error; added to its object-storage-error set.
- `objstore_s3_read`: "https refused 0A000" becomes "**https never silently downgraded**" — 08006 when built with OpenSSL (the bytes went TLS against the plain-HTTP fixture), 0A000 without, decided by `ldd` of the installed module; a row count is the red either way.

## Verification

19/19 on PG17 (lane), PG18, PG19. ASAN+UBSAN gate passes all four objstore suites. M1 18/18, M2 12/12, `objstore_module` 30/30 unregressed. `-Wshadow -Werror` clean. RED-first recorded (every https arm failed 0A000 on main).

After this, #393 has M4 (validator split + catalog options) remaining.


OffgridwithJD (approved):
## Approve. The verification is correct on a close read, and I reproduced the load-bearing removal proof — the IP binding is independently enforced.

This is the milestone where a silent bug is a real security hole, so I read the
verification setup line by line and then reproduced the proof that matters on my
own edit.

### The verification setup is correct

- `SSL_CTX_set_verify(..., SSL_VERIFY_PEER, NULL)` is set, so a failed chain
  aborts the handshake — there is no `SSL_connect() == 1` path without a verified
  chain.
- The identity binding is installed on `SSL_get0_param` **before** `SSL_connect`:
  `set1_ip_asc` for IP literals, `set1_host` for names, `NO_PARTIAL_WILDCARDS`
  set, SNI suppressed for IPs. Because it is set pre-handshake, the identity
  check runs *as part of* verification, so a mismatch aborts the handshake rather
  than needing a separate post-hoc check that could be forgotten.
- The failure path always `ereport`s (verify-result reason when
  `SSL_get_verify_result != X509_V_OK`, OpenSSL reason otherwise) — a refusal,
  never a silent success.
- Nonblocking IO is complete: handshake, `SSL_write`, and `SSL_read` all loop on
  `WANT_READ`/`WANT_WRITE` through the M1 `os_wait`, handle `ZERO_RETURN`
  (orderly close) and `SYSCALL`+`errno==0` (EOF without close_notify → stale),
  and `os_wait` carries `CHECK_FOR_INTERRUPTS`, so cancel works mid-TLS (arm D
  confirms).

### Reproduced removal proof #2 — the silent-defect path — on pg18a

The `wronghost` cert is signed by the trusted **CA-1** with `SAN
DNS:other.example`, so its chain verifies and only the IP-identity check can
catch it. I dropped `X509_VERIFY_PARAM_set1_ip_asc` and reran:

```
PASS  A: https differential == local          (good cert still reads)
FAIL  C: wrong-host refused with 08006: got [] want [08006]
FAIL  C: wrong-host returned no rows: got [1] want [0]     <- now SUCCEEDS
PASS  C: expired refused with 08006           (chain still catches it)
PASS  C: untrusted-ca refused with 08006      (chain still catches it)
```

Exactly the right shape: with the binding gone, a trusted-CA cert for the wrong
identity is silently accepted, while expired/untrusted stay refused on chain
grounds. That proves the IP binding is load-bearing *independently* of chain
verification — the code is not accidentally relying on the chain to catch host
mismatches, which is the way this class of bug usually hides. Good instinct
making the fixture's wronghost CA-1-signed; a self-signed wronghost would have
let chain verification mask the missing identity check.

### The suite structure is right for a milestone with no positive oracle

A refusal matrix with the good arm doubling as the over-restriction proof, each
refusal arm gated on a `HANDSHAKE`-logged-at-accept premise so a refusal cannot
pass against a server that was never contacted, SQLSTATE (08006) plus a
reason-string grep plus a zero-rows assertion. The reached-code premise is the
#607 vacuous-deny lesson applied correctly — nice to see it carried forward.

### One non-blocking note

`SSL_CTX` is created (and `set_default_verify_paths` reloads the CA store) per
handshake, i.e. per opened object. Correct, just mildly wasteful on the read
path that opens one connection per file; a cached/shared `SSL_CTX` would save the
store load. Not a blocker, and premature to act on before there's a measurement.

I ran the suite and reproduced removal proof #2 on **pg18a** (OpenSSL 3.5.5). I
did not re-run removal proof #1 (`SSL_VERIFY_NONE`), the PG17/19 arms, the
ASAN+UBSAN gate, or the no-OpenSSL build path — I take the PR's word on those,
which is reasonable given its gate log. Proof #2 is the subtler of the two and
it holds under an independent hand.

Approving on the read, the reproduced removal proof, and the green matrix. Merge
is jd's.

PR #610, opened 2026-08-13
feat: the validator split and catalog credentials (#393 M4)

jdatcmd:
The last #393 milestone, on merged M1–M3. Design: `design/ISSUE_393_M4_CREDENTIALS.md` (in this PR).

## What lands

- **The placement policy, enforced at DDL time**: `endpoint`/`region` on the SERVER, the credential triple + `credentials_required` ONLY on a USER MAPPING, table options unchanged, every misplacement HV00D. The world-readable-secret outcome #393 opened with is now unreachable by any accepted statement.
- **Scan-time resolution**: user's mapping else PUBLIC (probed without erroring — a mapping is optional here, unlike postgres_fdw), mapping credentials win, server options else environment for endpoint/region. **Ambient becomes a privilege**: superuser, or a mapping a superuser marked `credentials_required 'false'` (`password_required`'s shape, superuser check in the validator → 42501). Half a credential pair refuses at scan time.
- **ABI 2**: `open()` gains the resolved config struct; `cfg == NULL` (function-API paths, no server object) stays ambient behind `pg_read_server_files` — recorded in the design doc as a decision, not a drift from the memo's stricter alternative.
- The memo's endpoint allow-list GUC is **not** here: it is a default-deny posture change that needs its own owner call; posted to the issue as the follow-on question.

## Suite design points

- The catalog arms run with **no AWS_\* environment at all**, premise-pinned (even a superuser is 28000 before the mapping exists) — a green read can only have come from the catalogs.
- Per-user proof: alice and bob read the same table over the same server; only their mappings differ; the SigV4-verifying fixture judges (alice reads, bob 28000).
- Ambient-as-privilege proven in both directions: carol reads via the environment only while her `credentials_required=false` mapping exists.
- No-leak arms: bob's verbose error and EXPLAIN VERBOSE grep-zero for both secret literals; `srvoptions` cannot carry one by construction.

## Proofs and gates

- RED-first on main (placement DDL rejected, no catalog plumbing) recorded.
- Removal proofs run: drop the validator superuser check → 42501 arm red; drop the mapping lookup → catalog/alice arms red.
- 21/21 on PG17 (lane), PG18, PG19; **ASAN+UBSAN gate passes all five objstore suites**; four merged objstore suites unregressed; `-Wshadow -Werror` clean.

## Reviewer note

PGXS tracks no header dependencies: an ABI bump needs a **clean rebuild of both the main library and the module** — during this work each side was the stale one once, and both misreport as "module not installed" (with the ABI WARNING in the log as the tell). Recorded in the commit and my memory notes; if your bench uses incremental builds, `make clean` both before judging a red.

With this merged, #393's four milestones are complete: ranged HTTP transport, SigV4, TLS, and the credential model. Remaining on the issue: the endpoint allow-list owner decision, ListObjectsV2/globs (deliberately deferred), and FDW streaming (tracked separately).

### DIFF
--- a/design/ISSUE_393_M4_CREDENTIALS.md
+++ b/design/ISSUE_393_M4_CREDENTIALS.md
@@ -1,0 +1,120 @@
+# Issue #393 M4: the validator split and catalog credentials
+
+Design before code. The last #393 milestone: M1-M3 (merged) read over HTTP,
+sign with SigV4, and verify TLS, with every credential ambient in the
+postmaster environment. M4 gives credentials a catalog home with the placement
+policy the decision memo fixed, so a secret can never land world-readable and
+a non-superuser can be granted remote reads without owning the postmaster's
+identity.
+
+## The placement policy (memo 3.4, enforced at DDL time)
+
+| catalog | options | visibility |
+|---|---|---|
+| foreign table | `path`, `partition_columns` (unchanged) | world-readable, no secrets by construction |
+| server | `endpoint`, `region` | world-readable, deliberately non-secret |
+| user mapping | `access_key_id`, `secret_access_key`, `session_token`, `credentials_required` | protected (`pg_user_mapping` is not PUBLIC-readable) |
+
+The validator rejects every option on any other catalog with
+`ERRCODE_FDW_INVALID_OPTION_NAME`, so the world-readable outcome the issue
+opened with is foreclosed at CREATE/ALTER time, not by documentation.
+
+## Resolution order at scan time (FDW path)
+
+1. The scanning user's mapping for the server, else the PUBLIC mapping,
+   probed without erroring (a mapping is optional here, unlike postgres_fdw).
+2. Mapping credentials present: use them. Endpoint and region come from the
+   server options, else the environment (`AWS_ENDPOINT_URL`, `AWS_REGION`),
+   so one server definition serves both styles.
+3. No mapping credentials: **ambient is a privilege, not a default.** The
+   postmaster's environment identity is used only when the caller is a
+   superuser, or the applicable mapping carries `credentials_required
+   'false'`. That option follows `password_required`'s shape exactly: only a
+   superuser may set it to false (enforced in the validator), because it
+   hands the mapped role the instance's ambient identity.
+4. Otherwise: SQLSTATE 28000 naming what is missing ("no credentials in any
+   user mapping for this server"), before any connection exists.
+
+This tightens M2/M3 behaviour for non-superusers on the FDW path, which is
+the point of the milestone; the merged suites all run as superuser and are
+unaffected.
+
+## The function API (`read_parquet`, `import_parquet`, `parquet_schema`)
+
+No server object exists in those signatures, so no mapping can apply. They
+keep the M2 ambient-environment behaviour behind the existing
+`pg_read_server_files` gate, which the memo assessed as granting no new
+capability (that role already reads any file as the postgres user). The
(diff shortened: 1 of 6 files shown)


OffgridwithJD (approved):
## Approve. The placement policy holds under a close read, the suite is 21/21, and I reproduced the privilege-escalation removal proof.

Read the validator, the scan-time resolution, the module, and the ABI carefully;
built the branch on pg18a; ran `objstore_credentials` (21/21); and reproduced the
security-critical removal proof on my own edit.

### The placement policy is hard-enforced

- Secrets are accepted **only** under `catalog == UserMappingRelationId`. A
  secret option on a server or table matches none of the server/table branches
  and falls through to `HV00D` — so the world-readable outcome is unreachable by
  any accepted statement, not by documentation. `endpoint` on a mapping is
  `HV00D` the same way. The suite's placement arms confirm each direction.
- `credentials_required 'false'` requires `superuser()` (mirrors
  `password_required`), and `defGetBoolean` rejects a non-boolean. Setting it
  `true` needs no privilege — correct, only handing out ambient identity is
  gated.

### The resolution makes ambient a privilege, and does not disclose across users

- `pqfdw_objstore_config` looks up the mapping with `GetUserId()` (the scanning
  user), falling back to the PUBLIC mapping (`InvalidOid`) — so it only ever
  reads the caller's own mapping or the shared one; no path fetches another
  user's secret.
- `allow_ambient = superuser() || !credRequired`, and the module refuses 28000
  when there is no credential triple and ambient is not allowed, rather than
  falling back to the postmaster's identity. The half-credential check
  (`(akid==NULL) != (secret==NULL)`) closes the ALTER-across-statements gap the
  validator cannot see, so half a pair is a refusal, never a silent fallback.
- Function-API paths open with `cfg == NULL` → ambient-allowed behind
  `pg_read_server_files`, which is the pre-existing M1–M3 behavior; the ABI-v2
  struct documents `NULL` explicitly. No behavior change for the merged suites,
  which run as superuser.

### Reproduced the escalation removal proof

I neutered the `!superuser()` gate on `credentials_required` (the check whose
absence would let a non-superuser grant their own mapped role the instance's
ambient identity) and reran:

```
FAIL  placement: a non-superuser cannot set credentials_required false (42501)
```

The arm reds exactly as it must — the gate is load-bearing against a real
privilege escalation, not decoration.

### No-leak, checked

The suite's no-leak arms grep bob's verbose error and `EXPLAIN VERBOSE` for the
literal secret and assert zero, and `srvoptions` cannot carry one by
construction — all green. The secret reaches only the SigV4 seed HMAC and the
request headers, never a message or the request-count seam, consistent with the
M2 review.

### One thing I appreciated

The catalog arms run with **no `AWS_*` environment at all**, premise-pinned (even
a superuser is 28000 before the mapping exists), so a green catalog read cannot
have come from the environment. That is the reached-the-code discipline applied
to prove the catalog path, not just that a read succeeded.

pg18a, the suite, and the one removal proof. I did not re-run PG17/19, ASAN, or
the other two removal proofs (mapping-lookup, ABI-revert) — I take the PR's word
there. The reviewer note about clean-rebuilding both sides on an ABI bump is
real: I hit a `make clean` hang unrelated to the code (a Makefile `$(shell sed)`
blocking on ssh's non-interactive stdin), not a build defect in this PR.

Approving on the read, the reproduced escalation proof, the no-leak arms, and the
green suite. Merge — and the M4 milestone — are jd's.

Issue #560, opened 2026-08-10
`pgcolumnar.vacuum`'s documented `stripe_count` argument is never read, and `pgcolumnar.stats()` fails for a table's own owner


OffgridwithJD:
Two defects in the documented user-facing surface, both measured. They are
unrelated in cause but both have the same shape: the documentation describes
behaviour the code does not have.

---

## 1. `stripe_count` is accepted and silently ignored

`docs/sql-reference.md:49-56`:

> ### pgcolumnar.vacuum(tablename regclass, stripe_count int DEFAULT 0)
> ...
> `stripe_count` bounds how many row groups are combined in one call; `0` places
> no bound.

`src/columnar_vacuum.c`, `pgcolumnar_vacuum`:

```c
	Oid			relid = PG_GETARG_OID(0);
	...
	pgcolumnar_compact_relation(rel, 0, NULL);
```

`PG_GETARG_OID(0)` is the only argument read. Argument 1 is never fetched, and
the bound passed to `pgcolumnar_compact_relation` is the literal `0`.

### Measured

Fixture built to discriminate — 24 row groups, so "combine at most one" and
"combine without bound" cannot look the same by construction:

```
PREMISE row groups before: 24

pgcolumnar.vacuum('vt', 1)   ->  row groups after: 1
pgcolumnar.vacuum('vt', 0)   ->  row groups after: 1     (fixture rebuilt identically)
```

Identical. The bound has no effect. Row count is preserved in both cases (656
before and after), so this is not a correctness bug in the rewrite — it is a
documented control that does nothing.

### Why it matters beyond the doc

The stated purpose of the argument is to bound the work in one call. An operator
who passes `stripe_count => 4` to keep a maintenance window short gets an
unbounded rewrite of the whole table, with no error and no warning. On a large
table that is the difference between a short lock and a long one — and
`pgcolumnar.vacuum` takes `AccessExclusiveLock`.

### Fix

Either honour it (`PG_GETARG_INT32(1)` through to `pgcolumnar_compact_relation`,
which already takes a bound parameter), or drop the argument and the sentence.
Honouring it looks like the intended design, since the callee already has the
parameter.

Whichever way, the regression test needs a fixture with more than one row group,
or it cannot tell the two behaviours apart.

---

## 2. `pgcolumnar.stats()` is documented for "most users" but is superuser-only

`docs/sql-reference.md:44-45` steers readers to it:

[...]

PR #611, opened 2026-08-13
fix: pgcolumnar.sort_status() SECURITY DEFINER so a table's owner can read it (#608)

### LINKED ISSUE #608: pgcolumnar.sort_status() false-denies a table's own owner (invoker-rights reads an internal catalog)
## What

`pgcolumnar.sort_status(rel)` is `SECURITY INVOKER` and reads the internal
catalog `pgcolumnar.row_group` (and `storage`, `options`) directly. Those
catalogs carry no GRANT, so an ordinary role — **including the table's own
owner** — cannot SELECT them, and `sort_status()` fails with 42501 rather than
returning the run/append counts for a table they own.

Same class as the `stats()` disclosure/usability split (#560): the fix there was
`SECURITY DEFINER` + an explicit `require_caller_select(rel)` so the caller reads
their own table's statistics without the catalog being made world-readable.

## Reproduction (main @ 99c6a58, control included)

```sql
CREATE TABLE ss_t(ts int, v int) USING pgcolumnar;
INSERT INTO ss_t SELECT g, g FROM generate_series(1,5000) g;
SELECT pgcolumnar.vacuum_sorted('ss_t','ts');
CREATE ROLE ss_owner LOGIN;
ALTER TABLE ss_t OWNER TO ss_owner;
GRANT USAGE ON SCHEMA pgcolumnar TO ss_owner;
```

```
-- superuser (control):
SELECT total_groups FROM pgcolumnar.sort_status('ss_t');   ->  1
-- the table's OWNER, on their own table:
SELECT total_groups FROM pgcolumnar.sort_status('ss_t');   ->  ERROR 42501
```

## Fix

`SECURITY DEFINER` + `SET search_path = pg_catalog, pg_temp` +
`PERFORM pgcolumnar.require_caller_select(rel)` as the first statement, exactly
as `stats()`. The gate uses `GetOuterUserId()`, so definer rights do not widen
who may read a table's statistics — a caller still needs SELECT on `rel`.

A suite must carry a **positive control** (an owner/SELECT-holder gets a row) and
a deny arm on SQLSTATE 42501, not a `permission denied` text grep — a deny arm
without a positive control passes vacuously for a function that denies everyone,
which is exactly how this hid.

## Provenance

Surfaced by @jdatcmd reviewing PR #607 (`maintenance_due`, which calls
`sort_status`); that PR fixed itself by becoming DEFINER, but `sort_status`
standalone still has the property. Filed separately with its own reproduction
rather than folded into #607, per that review.

OffgridwithJD:
Fixes #608.

`pgcolumnar.sort_status()` was `LANGUAGE sql` and read pgcolumnar's internal
catalogs (`storage`, `row_group`, `options`) directly. Those carry no GRANT, so
an ordinary role — **including the table's own owner** — got 42501 on an internal
catalog rather than the run/append counts for a table they own. Same class as
`stats()` (#560); surfaced by @jdatcmd reviewing #607 (`maintenance_due`, which
calls `sort_status`).

## Fix

Convert `sort_status` to `plpgsql` `SECURITY DEFINER` with `SET search_path =
pg_catalog, pg_temp` and a first-statement `PERFORM
pgcolumnar.require_caller_select(rel)`, exactly as `stats()`. `require_caller_select`
uses `GetOuterUserId()`, so definer rights do **not** widen who may read a table's
sort status — a caller still needs SELECT on `rel` — while the body reaches the
catalog as the owner. GRANTing the catalog to PUBLIC would have been the wrong
fix: it publishes `zone_map`'s per-column min/max/sum. Body references were
already schema-qualified.

## Test — `test/sort_status_privilege.sh`, mirroring `stats_privilege.sh`

The #607 review lesson, applied: a deny arm alone is vacuous — an invoker-rights
function denies everyone and passes it. So the suite leads with **positive
controls**:

- the table's **owner** reads its sort status (a row);
- a role merely GRANTed SELECT reads it too;
- the owner sees its sorted run (`sorted_groups > 0`), not a false empty.

These are the arms that **red on the invoker shape** and would have caught this
(proven RED-first: each failed with `permission denied for table row_group`
before the fix, then passed after). The deny arm asserts `permission denied for
table` **naming the TABLE** (`ss_t`), not the internal catalog the invoker-rights
bug refused with, and the superuser path is checked unchanged.

## Proofs and gate

- **RED first**: the three positive controls failed on invoker rights; recorded.
- **GREEN**: 13/13 on pg18a and pg19a.
- **Removal proof**: dropping `require_caller_select` from the fixed function
  reddens the deny arm (a no-privilege role reads the catalog as the owner and
  gets a row) — the gate is load-bearing.
- Functional `sort_status.sh` unchanged (49/49): the superuser path is untouched.
- Preflight (build + suite) on pg15a/16a/17a and the full assert matrix on
  pg18a + pg19a: results below.

## Scope

Pure SQL, no C, no new GUC. The function is added to the base extension script
(the alpha convention). It composes cleanly with #607: `maintenance_due` is
itself DEFINER, so it never depended on this, but `sort_status` standalone now
stops false-denying owners — defense in depth on the same call path.

---

**Gate, ca594a4:** removal proof confirms the gate is load-bearing (deny arm reds when `require_caller_select` is dropped); preflight pg15a/16a/17a PASSED; full assert matrix **ALL VERSIONS PASSED** — PG18 (157 ran, 2 skipped), PG19 (159 ran, 0 skipped), `sort_status` + `sort_status_privilege` green, no regressions.

### DIFF
--- a/pgcolumnar--1.0-alpha.sql
+++ b/pgcolumnar--1.0-alpha.sql
@@ -699,8 +699,17 @@ CREATE FUNCTION pgcolumnar.sort_status(
 	OUT sorted_rows bigint,
 	OUT appended_rows bigint)
 	RETURNS record
-	LANGUAGE sql STABLE
+	-- SECURITY DEFINER, like stats() (#560): the body reads pgcolumnar's internal
+	-- catalogs (storage, row_group, options), which carry no GRANT, so an
+	-- invoker-rights function false-denied a table's own owner on their own table
+	-- (#608). require_caller_select gates the REAL caller via GetOuterUserId(), so
+	-- definer rights do not widen who may read a table's sort status. search_path
+	-- is pinned as a definer function must.
+	LANGUAGE plpgsql STABLE SECURITY DEFINER
+	SET search_path = pg_catalog, pg_temp
 	AS $sort_status$
+BEGIN
+	PERFORM pgcolumnar.require_caller_select(rel);
 	WITH s AS (
 		SELECT st.storage_id, st.sorted_through, st.sorted_from
 		FROM pgcolumnar.storage st
@@ -729,7 +738,9 @@ CREATE FUNCTION pgcolumnar.sort_status(
 		   (SELECT count(*)::bigint FROM g WHERE g.in_run),
 		   (SELECT count(*)::bigint FROM g WHERE NOT g.in_run),
 		   COALESCE((SELECT sum(g.row_count)::bigint FROM g WHERE g.in_run), 0::bigint),
-		   COALESCE((SELECT sum(g.row_count)::bigint FROM g WHERE NOT g.in_run), 0::bigint);
+		   COALESCE((SELECT sum(g.row_count)::bigint FROM g WHERE NOT g.in_run), 0::bigint)
+	INTO sort_key, total_groups, sorted_groups, appended_groups, sorted_rows, appended_rows;
+END;
 $sort_status$;
 
 COMMENT ON FUNCTION pgcolumnar.sort_status(regclass)
--- a/test/run_all_versions.sh
+++ b/test/run_all_versions.sh
@@ -185,6 +185,7 @@ SUITES=(
 	server_file_privilege
 	smoke
 	sort_status
+	sort_status_privilege
 	sorted_projection
 	stats_privilege
 	temporal
--- a/test/sort_status_privilege.sh
+++ b/test/sort_status_privilege.sh
@@ -1,0 +1,82 @@
+#!/usr/bin/env bash
+#
+# pgColumnar: pgcolumnar.sort_status is usable by a table's owner, and only by a
+# caller who may read the table (#608).
+#
+# Same class as stats() (#560): sort_status was LANGUAGE sql and selected
+# straight from pgcolumnar.storage, row_group and options, which carry no GRANT,
+# so those tables are owner-only and a COLUMNAR table's own owner is not their
+# owner. The function therefore false-denied the owner of the table it reports on
+# with 42501 on an internal catalog. GRANTing the catalog to PUBLIC is the wrong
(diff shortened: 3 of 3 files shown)


jdatcmd (approved):
## Approved. Verified on my PG17 non-assert lane, removal proof re-run by the reviewer.

Against `ca594a4`: **13/13 green**, and re-running the removal proof myself (drop `require_caller_select` from the function body) reds exactly the two deny arms — the ungated DEFINER function hands rows to a no-privilege role, so the gate is load-bearing, not decorative.

This suite is the pattern the last two reviews were reaching for, fully landed:

- **Connectivity premises run BY the role they are about** ("the owner can open a session", by the owner) — the working-rules requirement applied literally, so a login FATAL can never masquerade as a deny.
- **Positive controls lead**, and they are the arms that red on the invoker shape (RED-first recorded), so a deny-everyone implementation cannot pass.
- **The deny arm asserts the discriminator SQLSTATE alone cannot provide**: both the bug and the fix refuse with 42501; only the named object differs, and the suite requires the refusal to name `ss_t`, not the internal catalog. This is the right use of a message-text assertion — it is the discriminator, premise-guarded, not a lazy grep.
- The "GRANT is not the fix" premise (the catalog stays unreadable) pins why DEFINER was chosen over publishing `zone_map`'s min/max/sum.
- "The owner sees its sorted run, not a false empty" guards against a DEFINER function that fabricates rows.

One cosmetic nit, no change requested: `psql_run "SELECT pgcolumnar.vacuum_sorted(...)"   -- give it a sorted run` — the `--` there is not a bash comment; it passes three stray arguments to `psql_run`, which ignores them today because it reads only `$1`. Harmless now, but a future `psql_run` that consumes more arguments would silently eat them; a `#` comment on its own line is safer.

Merge when ready.

PR #612, opened 2026-08-13
feat: the export byte sink — checked writes, temp-and-rename (#394 step 1)

jdatcmd:
Step 1 of the #394 order of work (`design/ISSUE_394_OBJECT_STORAGE_SINK.md`, merged): the local `PqSink` seam, fixing the two defects the write-path mapping found, useful standalone and blocking on nothing.

## What lands

- **One checked write replaces all 17 `fwrite` sites** (6 parquet + 11 arrow, per the corrected census): a short write errors at the call, the way `pq_source_read` treats a short read. Disk-full is no longer detectable only when the final flush happens to fail.
- **Nothing visible at the final name before commit**: exports write `<path>.tmp.<pid>`, `finish` does flush + fsync + `durable_rename`, `abort` unlinks from `PG_CATCH`. A FATAL mid-export leaves only pid-tagged debris, never a plausible half-file under the trusted name.
- **`write_record_batch` takes the sink** — the helper the 13-site census missed, with the suite's failure injection reaching into it by arithmetic (fail_after derived from the measured file size, landing in the final batch body).
- Dev GUC `pgcolumnar.sink_fail_after` drives the *same* error path a full disk takes, making the failure arms deterministic.

## The interaction the strengthened suite caught (the interesting part)

The parallel exporter's cancel arm originally waited for a `part-*.parquet` to appear before cancelling. With temp-and-rename, a final name appears only at *completion*, so the arm was cancelling a nearly-finished run — fixed to trigger on the temp (which appears at worker open, the earliest in-flight signal). The genuinely-mid-flight cancel then found a real defect of this change: **a worker the dispatcher terminates dies FATAL, which skips `PG_CATCH`, orphaning its temp** (measured: 4 files after a cancel). `pexport_remove_outputs` now also removes `*.parquet.tmp.*`, and the post-cancel arm asserts the directory *completely* empty, covering both cleanup paths. The un-widened dispatcher against the strengthened arm reds at exactly `got [4] want [0]` — the removal proof happened organically.

## Proofs

- RED-first: census arms read exactly main's 6 and 11; injection arms fail on the missing GUC with the final files present.
- Removal proofs both run: unlink deleted → the two injection debris arms red (1 and 2 leftovers) while the success-path arm stays green; short-write check made inert → both 53100 arms red (silent success past a failed write — the pre-fix defect reproduced).

## Verification

`export_sink` 15/15 and `parallel_export_parquet` 35/35 on PG17 (lane), PG18, PG19. `parquet_export`, `arrow_export`, `import_export_privilege` unregressed. ASAN+UBSAN gate passes `export_sink`, `parquet_export`, `arrow_export` (the dispatcher change is a five-line dirent name match, no new memory lifecycle). `-Wshadow -Werror` clean.

One bench note for the reviewer: a `run_san.sh` build leaves ASAN objects in the tree, and a later incremental non-ASAN relink sweeps them into the .so, which then fails to load with `undefined symbol: __asan_option_detect_stack_use_after_return` — every suite reds at cluster start. `make clean` between a sanitizer run and a normal one; recorded in my gotcha notes alongside the zero-byte-install variant.

Next per the design doc: step 2 (parallel workers already ride the seam via `PgColumnarWriteParquetFile`; arrow parallel does not exist), then the remote sink behind the module ABI once #394's write-side milestones start.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,7 +28,8 @@ OBJS = \
 	src/columnar_parquet_reader.o \
 	src/columnar_parallel_copy.o \
 	src/columnar_parallel_export.o \
-	src/columnar_objstore.o
+	src/columnar_objstore.o \
+	src/columnar_sink.o
 
 EXTENSION = pgcolumnar
 DATA = pgcolumnar--1.0-alpha.sql pgcolumnar--1.0-dev--1.0-alpha.sql
--- a/src/columnar_arrow.c
+++ b/src/columnar_arrow.c
@@ -25,6 +25,7 @@
  */
 #include "columnar.h"
 #include "columnar_flatbuffers.h"
+#include "columnar_sink.h"
 
 #include "fmgr.h"
 #include "funcapi.h"
@@ -841,7 +842,7 @@ arrow_build_field(FBB *b, ArrowCol *c)
 
 /* build one RecordBatch (metadata + body) and write it */
 static void
-write_record_batch(FILE *f, ArrowCol *cols, int ncols, int64 nrows)
+write_record_batch(PqSink *snk, ArrowCol *cols, int ncols, int64 nrows)
 {
 	StringInfoData body;
 	FBB			b;
@@ -926,13 +927,13 @@ write_record_batch(FILE *f, ArrowCol *cols, int ncols, int64 nrows)
 		uint32		metaLenField = metaLen + metaPad;
 		static const char zeros[8] = {0};
 
-		fwrite(&cont, 4, 1, f);
-		fwrite(&metaLenField, 4, 1, f);
-		fwrite(b.buf + b.cap - b.tail, 1, metaLen, f);
+		PgColumnarSinkWrite(snk, &cont, 4);
+		PgColumnarSinkWrite(snk, &metaLenField, 4);
+		PgColumnarSinkWrite(snk, b.buf + b.cap - b.tail, metaLen);
 		if (metaPad)
-			fwrite(zeros, 1, metaPad, f);
+			PgColumnarSinkWrite(snk, zeros, metaPad);
 		if (body.len > 0)
-			fwrite(body.data, 1, body.len, f);
+			PgColumnarSinkWrite(snk, body.data, body.len);
 	}
 
 	pfree(b.buf);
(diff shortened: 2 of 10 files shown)


OffgridwithJD (approved):
## Approve. Both #604 findings are closed, correctly, and the census is now a property instead of a number.

I read this against my own #604 review, since it implements those two defects. All verified in the diff:

- **All 17 fwrite sites route through the checked sink.** Parquet's 6 and arrow's 11 — including `write_record_batch`'s 5, the helper the 13-site census missed — are now `PgColumnarSinkWrite`, which errors at the call on a short write. The `census` arm greps `fwrite(` in both files for 0, so "no fwrite remains" is a pinned property that can't be miscounted again — exactly the fix the seam was supposed to make structural. RED on the old shape.
- **Temp-and-rename with unlink-on-error.** Both exporters wrap the whole write in `PG_TRY`; `finish` does flush+fsync+`durable_rename`, so the final name appears only whole; `PG_CATCH` calls `Abort` (unlink temp) and re-throws. The final path is never touched on error. `snk` is not reassigned inside the TRY, so there is no setjmp-clobber concern, and `Abort` is safe after `Finish` errors because `Finish` NULLs `snk->f` before the risky close — no double-free, no partial final file.
- **The injection reaches `write_record_batch`.** `sink_fail_after = size - 100` lands in the final record batch's body, so the failing write is inside the helper by arithmetic, not luck — the coverage the #604 review asked the Step-1 proof to reach.

The parallel orphaned-temp fix is a genuine find: a dispatcher-terminated worker dies FATAL and skips `PG_CATCH`, so `pexport_remove_outputs` now also sweeps `*.parquet.tmp.*`, and the strengthened cancel arm asserts the directory fully empty. The un-widened dispatcher redding at `got [4] want [0]` is the removal proof landing organically. Nice.

Two non-blocking notes: `pgcolumnar.sink_fail_after` is `PGC_USERSET`, which is fine (a user only fails their own export); and a `durable_rename` that raises *after* the rename syscall but on the dir-fsync can leave the final file present under an errored export — that is `durable_rename`'s own edge, shared with core, not this code's.

Scope: I verified by close read against #604 and did not re-run `export_sink.sh` myself (my bench is mid-gate on other work); I take the PR's 15/15 + ASAN. The code conclusively closes the findings.

Issue #388, opened 2026-08-04
Review what it would take to support Apache Iceberg


jdatcmd:
https://cpp.iceberg.apache.org/#overview

OffgridwithJD:
Taking this. Reviewing against what the Parquet and FDW code already does, so the estimate is grounded in our tree rather than in the spec alone. Will post findings, not a patch.

One flag up front: the link is to the Iceberg **C++ implementation**. Per PROVENANCE that is a source I will not read. I will work from the table-spec documents and the REST catalog OpenAPI spec, both of which are specifications rather than someone's code.

OffgridwithJD:
Reviewed. Findings below, no patch. Everything about our tree is measured from
`author/main`, everything about Iceberg is from the format spec, and I have marked which
is which. Where I am guessing I say so.

## First, the link

**Correction to what I first posted here.** I said I had not opened the Iceberg C++
implementation and would rather nobody did. That was my own over-restriction and it is
not this project's rule.

The rule is about **use, not reading**. Iceberg is Apache licensed, so reading it is
fine. What we must not do is copy, adapt or transliterate it. Our code stays ours.

The review below was written from `format/spec.md` and the REST catalog OpenAPI
document, which is the right place to start regardless. I am flagging the change so
nobody takes my earlier sentence as policy.

## "Support Iceberg" is three different projects

Worth picking one before estimating, because they differ by roughly an order of magnitude.

| | what it means | rough shape |
|---|---|---|
| **A. Read** | `SELECT` from an existing Iceberg table at its current snapshot | the plausible one |
| **B. Read well** | plus partition pruning, predicate pushdown, time travel, `MERGE`-visible deletes | A plus real work |
| **C. Write** | commit snapshots, maintain manifests, be a catalog participant | a different product |

C also makes us responsible for other engines' correctness, because a bad commit corrupts
a table Spark and Trino also read. I would not go near C for a long time.

## What we already have, and it is a lot

Measured, not assumed:

- **A hand-written Parquet reader**, `src/columnar_parquet_reader.c`, 123,657 bytes, with
  its own Thrift decoder. No Arrow C++ dependency.
- **Codecs**: `UNCOMPRESSED`, `SNAPPY`, `GZIP`, `ZSTD`, `LZ4`, `LZ4_RAW`. That covers what
  Iceberg writers actually emit.
- **Row-group skipping on statistics**, which is the same idea as Iceberg manifest
  pruning one level down.
[...]

PR #613, opened 2026-08-13
feat: parse and report Parquet field IDs (#388 step 2)

jdatcmd:
Step 2 of the #388 phase order (@ChronicallyJD's), scoped per my correction on the issue: binding is positional, so there is no name-lookup to replace, and resolution-by-id waits for its Iceberg caller — an unused resolution API today would be dead code by the delete-and-stay-green standard.

## What lands

- `parse_schema_element` consumes **thrift field 9** into `PqSchemaCol.field_id` (−1 sentinel: 0 is a legal id, absence must be distinguishable).
- `pgcolumnar.parquet_schema()` gains a **`field_id` column** — the id as written, NULL when absent. This is the observable consumer that keeps the parse removal-provable.
- A **fuzz-corpus seed with field ids** (including id 0), so the mutator bends footers that actually carry field 9.

## Proofs

- RED-first: both new arms failed on main (no such column).
- The ids arm uses pyarrow as the independent writer with ids **out of order and disjoint from positions** (7, 3, 12) — a reader fabricating ids from position cannot pass.
- Removal proof run: disabling the case-9 arm reds the written-ids arm (`got [] want [7,3,12]`) while the NULL arm rightly stays green.

## Verification

`native_parquet_schema` 35/35 on PG17 (my lane), PG18, PG19. Eight-suite parquet family unregressed (fdw, projection, pushdown, units, read_parquet, import, export, fuzz). ASAN+UBSAN gate passes `native_parquet_schema` **and `fuzz_parquet`** over the new seed. `-Wshadow -Werror` clean.

Signature note: `parquet_schema`'s RETURNS TABLE gains the column in the base script per the alpha convention; no C symbol changes, so the extension-upgrade surface is untouched.

With this, your step 1 (Avro) has the Parquet-side id plumbed and observable whenever the filesystem-catalog reader (phase step 3) wants to project by it.

### DIFF
--- a/pgcolumnar--1.0-alpha.sql
+++ b/pgcolumnar--1.0-alpha.sql
@@ -868,12 +868,12 @@ COMMENT ON FUNCTION pgcolumnar.import_parquet(regclass, text)
 	IS 'insert rows from a Parquet file, directory, or glob into a table; returns rows inserted (gap 27)';
 
 CREATE FUNCTION pgcolumnar.parquet_schema(path text)
-	RETURNS TABLE(column_name text, data_type text, nullable boolean)
+	RETURNS TABLE(column_name text, data_type text, nullable boolean, field_id integer)
 	LANGUAGE C STRICT
 	AS 'MODULE_PATHNAME', 'pgcolumnar_parquet_schema';
 
 COMMENT ON FUNCTION pgcolumnar.parquet_schema(text)
-	IS 'report the leaf columns of a Parquet file and the PostgreSQL type each maps to; for a directory or glob, of its first file (Phase G scan core)';
+	IS 'report the leaf columns of a Parquet file and the PostgreSQL type each maps to; for a directory or glob, of its first file; field_id is the SchemaElement field id Iceberg projects by, NULL when the writer emitted none (Phase G scan core, #388)';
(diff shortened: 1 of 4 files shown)


OffgridwithJD (approved):
## Approve. Clean, and the test can't be faked from position.

- `parse_schema_element` consumes thrift field 9 into `field_id` with a `-1` sentinel, and `parquet_schema` reports it NULL when absent — correctly distinguishing "no id" from the legal id `0` at both the parse and the projection.
- The written-ids arm uses pyarrow ids **out of order and disjoint from the positional indices** (7, 3, 12), so a reader fabricating ids from column position cannot pass; the no-id file reports NULL, not 0. That is the right shape for a "we actually parsed it" proof, and the removal proof (disable case 9 → the ids arm reds, the NULL arm stays green) confirms it.
- Fuzz seed carries a field-id footer including id 0, so the mutator bends a footer that actually has field 9. `RETURNS TABLE` gains the column in the base script with no C symbol change, so the upgrade surface is untouched.

Scoping it to the parse + report, with resolution-by-id deferred to its Iceberg caller rather than shipped as dead code, is the right call. Approving.

Issue #382, opened 2026-08-04
Update symbols to pgcolumnar


jdatcmd:
Multiple competing projects use columnar for shared libraries etc... to keep a distinct namespace we should use pgcolumnar

OffgridwithJD:
Measured the collision rather than reasoning about it, since Citus columnar happens to
be installed on the bench host alongside pgColumnar.

## Four real symbol collisions today

Comparing exported (dynamic, defined) symbols of the two shared libraries:

```
$ comm -12 <(nm -D --defined-only pgcolumnar.so | awk '{print $3}' | sort -u) \
           <(nm -D --defined-only citus_columnar.so | awk '{print $3}' | sort -u)
Pg_magic_func
_PG_init
columnar_handler
columnar_relation_storageid
pg_finfo_columnar_handler
pg_finfo_columnar_relation_storageid
```

`Pg_magic_func` and `_PG_init` are on every extension and are expected. The other four
are ours and are avoidable: **`columnar_handler`** and **`columnar_relation_storageid`**,
plus their `pg_finfo_` pairs. `columnar_handler` is the table access method handler, so
it is the worst one on the list to share a name for.

## Scale of the rename

pgColumnar exports **63** symbols. **27** begin `columnar_`, and only **5** begin
`pgcolumnar`. The 27:

```
columnar_add_projection            columnar_parallel_copy
columnar_cluster                   columnar_parallel_export_parquet
columnar_compact                   columnar_parquet_schema
columnar_compact_rewrite           columnar_read_parquet
columnar_debug_advance_reserved_offset  columnar_read_projection
columnar_debug_encoding_selftest   columnar_recluster
columnar_debug_set_metapage_version     columnar_reconstruct_via_projection
columnar_drop_projection           columnar_relation_storageid
columnar_export_arrow              columnar_truncate
columnar_export_parquet            columnar_vacuum
columnar_file_split_offsets        columnar_vacuum_sorted
columnar_handler                   columnar_vm_is_visible
columnar_import_arrow              columnar_vm_selftest
columnar_import_parquet
```

Those are the SQL-callable entry points, so each rename is a C symbol **and** its
`CREATE FUNCTION` in `pgcolumnar--1.0-dev.sql`. They are user-visible: `columnar_vacuum`
is what `pgcolumnar.vacuum` resolves to.

## A user-visible collision the symbol list does not show

`EXPLAIN` on a Citus columnar table prints:

```
[...]

PR #614, opened 2026-08-13
fix: recluster() self-gates instead of rewriting on every call (#415)

OffgridwithJD:
Part A of #415 (`design/ISSUE_415_AUTOVACUUM.md`, in this PR). The maintenance
daemon that issue asks for cannot be built until `recluster()` stops rewriting
untouched tables — so this lands first, and is a real production fix on its own.

## The problem

`pgcolumnar.recluster('t','k')` re-read every live row, Morton-sorted it, and
rewrote every group on **every** call — even when the table was already fully
clustered by `k` with nothing appended since. Measured on #415 axis C: ~6.2 s /
27 groups repeated on a 4M-row table that had not changed. A nightly cron line
(or the coming daemon) that called it speculatively would rewrite every
untouched table on every run.

Root cause: `pgcolumnar.storage` recorded the sorted **extent**
(`sorted_from`/`sorted_through`) but **not what key** the run is clustered by, so
recluster could not tell "already clustered by these columns" from "clustered by
something else". (Same gap surfaced as `sort_status.sort_key` reading NULL on a
clustered table.)

## The fix

`pgcolumnar.storage` gains `sorted_by name[]` and `sorted_kind`
(`'zorder'`/`'lexicographic'`), written with the extent by the online recluster.
`recluster` now returns **0 without a rewrite** when all three hold:

1. the sorted run covers every live group (nothing appended past it),
2. the recorded kind is `zorder`, and
3. the recorded key equals the requested columns, in order.

Any mismatch — appended data, a different key, a lexicographic or unknown run —
falls through to the full reorg, so **re-clustering by a new key still works**.
The gate runs before the group-locking and the read/sort/write, so a no-op costs
one catalog lookup, not a rewrite. Existing (pre-upgrade) storages have NULL
here and are treated as "unknown key" → never no-op'd, the safe default.

Ripple fix: `sort_status.sort_key` reports `storage.sorted_by` (the actual
clustering), no longer NULL on a table clustered but never declared — which is
what a scheduler needs to know which key to maintain.

## Proof — `test/recluster_gate.sh` (RED-first)

- A repeat recluster by the same key on an unchanged table reclusters **0
  groups** and does **not churn the relfilenode** (pinned on both). RED on the
  old shape (it rewrote everything).
- `sort_status.sort_key` now reports the key (was NULL).
- Appends un-gate it: recluster does real work again and folds the new data back
  into the run.
- A **new** key is not a no-op — it re-clusters, and sort_status reports the new
  key; a repeat by that key is then a no-op.
- The row set is byte-identical to a heap mirror after the no-op.

## Gate

- Preflight (build + `recluster_gate.sh`): pg15a/16a/17a PASSED.
- Full assert matrix pg18a + pg19a: all suites PASS, `native_recluster`,
  `native_sort_by`, `native_cluster`, `recluster_extent`, `sort_status`
  unregressed.
- ASAN+UBSAN (pg18_san) over the recluster/sort suites: this touches the
  metadata write path.
- (Gate results appended below.)

## Scope note for the maintainer

Adding columns to `pgcolumnar.storage` is a fresh-install schema change (the
matrix does `CREATE EXTENSION`, so it is covered). An **in-place upgrade** of an
existing alpha database would need an `ALTER TABLE pgcolumnar.storage ADD COLUMN`
in the upgrade path — flagging that as your versioning call rather than guessing
at it here. `PgColumnarGetSortedInfo` tolerates a pre-upgrade 7-column row
(reads the new attrs as absent → NULL → "unknown key" → recluster proceeds), so
the behavior degrades safely to today's if the columns are missing.

`vacuum_sorted`/`cluster` record NULL for the new columns for now (so recluster
never wrongly no-ops after them); recording their own key+kind — which would let
`sort_status` report the key for lexicographically-sorted tables too, and let
vacuum_sorted self-gate — is a small follow-up, deliberately out of this PR.

Part B (the `pgcolumnar_autovacuum` daemon) builds on this: it calls only the
online SUEL verbs (`compact_rewrite` + this now-safe `recluster`), never the
rewriting ones, and yields on `PROC_IS_AUTOVACUUM` when a user needs a stronger
lock. Design in the doc; separate PR.

---

**Gate, 29ccb0d:** preflight (build + `recluster_gate.sh`) pg15a/16a/17a PASSED; full assert matrix **ALL VERSIONS PASSED** — PG18 (159 ran, 2 skipped), PG19 (161 ran, 0 skipped), `recluster_gate`/`native_recluster`/`native_sort_by`/`native_cluster`/`recluster_extent`/`sort_status` all green, no regressions. ASAN+UBSAN over the recluster/sort suites: appended when it lands.

### DIFF
--- a/design/ISSUE_415_AUTOVACUUM.md
+++ b/design/ISSUE_415_AUTOVACUUM.md
@@ -1,0 +1,166 @@
+# Issue #415: pgcolumnar_autovacuum, and the recluster self-gating it requires
+
+Design before code. #415 asked for a background worker that runs the maintenance
+autovacuum cannot reach; the measurement and the policy function
+(`pgcolumnar.maintenance_due`, merged) settled *what* to run and *when*. This is
+the *how*: a daemon on autovacuum's model, and the one fix the daemon cannot be
+built without.
+
+## Part A (prerequisite): recluster() must stop rewriting when nothing changed
+
+Measured on #415 axis C: `pgcolumnar.recluster('t','k')` rewrites the whole
+relation on **every** call, even when the table is already fully clustered by `k`
+(27 groups / ~6.2 s each on a 4M-row table, repeated). A scheduler that calls it
+speculatively would rewrite untouched tables on every naptime. This is a
+production hazard on its own and a hard blocker for the daemon.
+
+The root cause is that the storage row records the sorted *extent*
+(`sorted_from`/`sorted_through`) but **not what key the run is clustered by**, so
+recluster cannot tell "already clustered by these columns" from "clustered by
+something else". (Same gap surfaced in the maintenance_due work as
+`sort_status.sort_key` reading NULL on a clustered table.)
+
+### Fix: record the clustering key and kind
+
+`pgcolumnar.storage` gains two columns:
+
+- `sorted_by name[]` — the clustering columns of the current run.
+- `sorted_kind text` — `'zorder'` (recluster/cluster) or `'lexicographic'`
+  (vacuum_sorted). The two orderings share the extent mark but are different
+  arrangements, so the kind is load-bearing: recluster('k') after a
+  vacuum_sorted('k') on the same multi-column key must NOT no-op.
+
+`PgColumnarSetSortedExtent` gains the key columns and kind and writes all four
+fields atomically with the extent; a getter returns them. Existing storages
+(pre-upgrade rows) have NULL here and are treated as "unknown key" → never
+no-op'd, the safe default.
+
+### The no-op condition (recluster)
+
+recluster(rel, cols) returns 0 without rewriting **iff**:
+1. the sorted run covers every live group (no unsorted/appended groups outside
+   `[sorted_from, sorted_through]`), AND
+2. `sorted_kind = 'zorder'`, AND
+3. `sorted_by = cols` (exact, ordered).
+
+Any mismatch — appended data present, a different key, or a lexicographic run —
+falls through to the full rewrite, so re-clustering by a new key still works
+(the contract is preserved, only redundant work is skipped). vacuum_sorted gets
+the mirror-image gate for the lexicographic kind.
+
+### Upgrade / binary-swap behavior (durable note, #614 review)
+
+The storage columns are added to the base script (fresh install) and to the
+dev->alpha upgrade file as idempotent `ADD COLUMN IF NOT EXISTS`. A binary swap
+that installs the new `.so` onto an un-upgraded 7-column storage table degrades
+**safely to the pre-#415 behavior** -- but ONLY because both the read
+(`PgColumnarGetSortedInfo`) and the write (`PgColumnarSetSortedExtent`) guard on
+`tupdesc->natts >= Anum_..._sorted_kind` before touching attnums 8/9. Without
+that guard the new .so CRASHES on the old 7-column storage table:
+`heap_getattr`/`heap_modify_tuple` index past the descriptor -- an absent attnum
+is NOT read as a clean NULL (verified: a faithful old->new binary-swap test
+crashed the backend on the first recluster; the guard fixes it, #614 review).
+With the guard, an un-upgraded table records nothing and never self-gates
+(`sort_key` NULL) -- exactly the pre-#415 behavior. The upgrade file adds the two
+columns (idempotent `ADD COLUMN IF NOT EXISTS`) and `CREATE OR REPLACE`s
+`sort_status`, so `ALTER EXTENSION ... UPDATE` reaches full behavior; a faithful
+old->new test confirmed the columns land at attnums 8/9, the self-gate works, and
+pre-existing tables read unchanged.
+
+### Ripple benefit
+
+`sort_status.sort_key` reports `storage.sorted_by` (the actual clustering),
+fixing the NULL-on-clustered-table gap; `maintenance_due` then sees a real key
+and the daemon can recluster by it.
+
+### TDD
+
+- RED: recluster on an already-clustered, unchanged table rewrites all groups
+  (assert `pgcolumnar.recluster(...) > 0` and row groups churn).
+- GREEN: it returns 0 and the physical groups are unchanged (relfilenode/blocks).
+- After appends: recluster folds only the new data.
+- recluster by a NEW key: still a full re-cluster (contract preserved).
+- vacuum_sorted then recluster on the same key: NOT a no-op (kind differs).
+- Gate: full matrix (18+19) + preflight all majors + **sanitizer** (pg18_san):
+  this touches the metadata write path.
+
+## Part B: pgcolumnar_autovacuum
+
+A launcher background worker in autovacuum's shape, registered from `_PG_init`.
+pgColumnar is already a `shared_preload_libraries` extension (the custom-scan and
+planner hooks require it), so this adds **no new install requirement**; a server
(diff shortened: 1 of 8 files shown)


jdatcmd (changes requested):
## Two changes: one vacuous check, and the maintainer ruling you asked for on the upgrade path

Verified on my PG17 non-assert lane: **13/13 green**, and the removal proof re-run by me — disabling the self-gate reds exactly the three no-op arms (each repeat reclusters again). The gate logic, the fall-through on any mismatch, the NULL-means-unknown default, and the `sort_status.sort_key` ripple all read correctly and behave correctly.

### 1. The relfilenode arm cannot fail (blocking)

> we pin it on both the return value AND the relfilenode staying put

The second pin is vacuous: pgcolumnar reclusters groups **inside its own storage**, so `pg_class.relfilenode` never changes on *any* recluster path. I ran your suite with the self-gate deleted — a real rewrite executed (the count arms red) and **the relfilenode arm still passed**. A check that passes with the fix removed and the defect restored is the assert-work-done mirror. Replace it with a falsifiable physical signal — the multiset of `(group_number, file_offset)` from `pgcolumnar.stats()` is the natural one: a rewrite reallocates offsets (my gate-deleted run moved them), a no-op leaves them byte-identical, and it reds on exactly the mutation the count arms red on while pinning the *physical* claim the count alone cannot.

### 2. The upgrade path: ruling, and one small addition

You asked for the maintainer call on the schema change. I ran the `extension_upgrade` gate (#382, old = main, new = this PR, PG17): **PASS** — and for the right reason, not vacuously. The new C layer sizes tuples by the live tupdesc and reads absent attnums as NULL, so a binary swap onto an un-upgraded schema degrades to exactly the pre-PR behavior (never no-op, `sort_key` NULL) — your "safe default" line covers it and the gate confirms it.

Ruling: **fresh-install schema change accepted; add the two columns to `pgcolumnar--1.0-dev--1.0-alpha.sql`** as idempotent `ALTER TABLE pgcolumnar.storage ADD COLUMN IF NOT EXISTS ...`, so the one supported upgrade path reaches full behavior rather than the degraded mode. No version bump needed at alpha. Please also move the binary-swap degraded-safe behavior from the PR body into `design/ISSUE_415_AUTOVACUUM.md` — the PR body is not the durable record, and this is precisely the fact a future upgrade-debugging session needs.

Re-approve on sight for both; I'll re-run the physical-pin arm's removal proof against the updated branch.

OffgridwithJD:
## ASAN+UBSAN gate: PASSED (metadata write-path change)

Ran on the pg18_san container. Built the sanitizer `.so` from this branch first (`test/build_san.sh`, exit 0, 4 min) — verified the installed library is mine and instrumented: `PgColumnarGetSortedInfo` present, 5 `"zorder"` literals, 179 `__asan` symbols, and the installed `.sql` carries `sorted_by`. Then `run_san.sh` over the write/metadata-touching suites:

```
PASS  smoke
PASS  recluster_gate
PASS  native_recluster
PASS  native_sort_by
PASS  native_cluster
PASS  sort_status
-- 6 suites under sanitizers, 0 failed, 0 ran no checks
SANITIZER GATE PASSED
```

The `construct_array`/`heap_getattr`/`heap_modify_tuple` additions to the storage row and the getter are ASAN+UBSAN clean under the recluster/sort/cluster exercises. (A first pass reported a false pass against a stale Aug-1 `.so` — a bare `pg_config` under pg18_san returns empty, so my `.so`-path check pointed at the wrong file; rebuilt with the sanitizer prefix and re-verified before trusting this.)

OffgridwithJD:
## Both addressed in 7b08c4b — and your upgrade question turned up a crash my "degrades safely" line had papered over.

### 1. The vacuous relfilenode arm (blocking) — fixed, and now falsifiable

You were right: pgcolumnar reclusters inside its own storage, so `relfilenode` never moves and the check passed with the fix removed. Replaced with the `(group_number, file_offset, rowcount)` multiset from `stats()`, exactly your suggestion. Proven falsifiable — I neutered the self-gate and reran:

```
FAIL  recluster by the same key ... reclusters 0 groups: got [1] want [0]
FAIL  and it leaves the physical layout byte-identical (group offsets unmoved): got [rewritten] want [same]
```

The physical arm now reds on the same mutation the count arm does.

### 2. The upgrade path — ruling taken, and it exposed a real defect

I did what you asked (idempotent `ADD COLUMN IF NOT EXISTS` in the dev->alpha file) and then ran a **faithful** old->new binary-swap test rather than trusting my own "degrades safely" sentence. It **crashed**:

```
== binary-swap: install NEW .so, run recluster on the un-upgraded 7-column storage ==
```

My claim that `heap_getattr` "reads absent attnums as NULL" was wrong — it indexes past the tuple descriptor. `PgColumnarGetSortedInfo` (read) and `PgColumnarSetSortedExtent` (write) now **guard on `tupdesc->natts >= Anum_..._sorted_kind`** before touching attnums 8/9. Re-ran the faithful upgrade:

```
degraded recluster (no crash): 1
sorted_by attnum=8
sorted_kind attnum=9
first recluster: 1
second (self-gated): 0
old table still reads: 5000
```

So: the guard makes the binary-swap actually safe (no crash, never self-gates, `sort_key` NULL until upgraded); the ALTERs land the columns at exactly 8/9; and the upgrade file also `CREATE OR REPLACE`s `sort_status` so `ALTER EXTENSION ... UPDATE` reaches full behavior. The corrected note is in `design/ISSUE_415_AUTOVACUUM.md` (the durable record, as you asked) — including that the guard, not a phantom NULL, is what makes it safe.

Your `extension_upgrade` PASS was real but didn't exercise recluster on the old schema, which is where the crash lived; the faithful test is the one that found it. Thank you for pushing on the upgrade path — it was hiding a backend crash.

### Re-gate (7b08c4b)

Full matrix **ALL VERSIONS PASSED** (PG18 159/2skip, PG19 161/0); ASAN+UBSAN over smoke/recluster_gate/native_recluster/native_sort_by/native_cluster/sort_status all PASS on a sanitizer build rebuilt from this branch (verified the `.so` is mine before trusting it). Ready for the re-run of your physical-pin removal proof.

jdatcmd (approved):
## Approved — and your binary-swap crash is a correction to MY review, not only to the PR.

Re-verified against `7b08c4b` on my PG17 non-assert lane, everything by my own hand:

- **13/13 green** (mtimes touched before build — the git-archive stale-object trap from last round is in my notes now).
- **The physical arm is falsifiable**: neutering the self-gate reds it together with the count arms (`got [rewritten] want [same]`). The `(group_number, file_offset, rowcount)` multiset is exactly the right instrument.
- **The faithful binary-swap probe, run independently**: old-schema cluster, new `.so` dropped in, no ALTER — `recluster()` runs without crashing, never self-gates (correct degraded mode), `sort_key` NULL, data intact. The natts guard does what it claims.
- The dev→alpha ALTERs are present and idempotent, and the durable note landed in the design doc.

**The part I owe this PR**: my previous review ran the #382 `extension_upgrade` gate, watched it pass, and endorsed "degrades safely" — but that gate never calls `recluster()` on an old-schema table, so my conclusion covered a wider population than my instrument measured. Your faithful swap test found the crash my endorsement had papered over. That is the widen-the-population-until-coverage-fails rule landing on the reviewer; the #382 harness should probably grow a maintenance-verb arm, which I will raise separately rather than scope-creep this PR.

Merge when ready.

PR #615, opened 2026-08-13
feat: the endpoint allow-list, empty by default (#393)

jdatcmd:
Implements the owner decision recorded on #393 (2026-08-13): remote object-storage access is default-deny. Design: `design/ISSUE_393_ENDPOINT_ALLOWLIST.md` (in this PR).

## What lands

- **`pgcolumnar.objstore_allowed_endpoints`**, string, **SUSET**, default empty = all remote refused (42501, before any socket exists). SUSET is load-bearing: a USERSET list would let any role widen its own allow-list — the suite proves a non-superuser SET fails, by that role's own session.
- Entries are `host` or `host:port`, matched **before resolution** (the list authorizes what the operator wrote). Defined in the preloaded library, read by the module via `GetConfigOption` — no cross-library symbol.
- **Link-local refused unconditionally** after resolution (169.254/16, fe80::/10, v4-mapped), even when explicitly listed; the whole connection refuses if *any* resolved address is link-local, so a mixed-answer resolver cannot steer the pick. Loopback deliberately stays allowed (trusted-network MinIO, and every suite).

## Proofs

- RED-first on main: the deny arms showed today's world — reads succeeding, and the IMDS arm making a genuine connection attempt (08006).
- Removal proofs both run: skip the list check → deny arms red; skip the link-local block → the IMDS arm reds at **`got [08006]`** — a real connection attempt into metadata space, exactly the world the check forbids.
- The IMDS arm keeps `169.254.169.254` explicitly **in** the list and still expects 42501, pinning "unconditional".

## Suite ecology

The four remote-reading suites gain `PGC_EXTRA_CONF="pgcolumnar.objstore_allowed_endpoints='127.0.0.1'"`, which doubles as the documented configuration example; `objstore_module`'s remote-error grep accepts the new refusal. 16/16 on PG17 (lane), PG18, PG19; sanitizer gate green across all six objstore suites; `export_sink` and the family unregressed; `-Wshadow -Werror` clean.

This closes the last open decision on #393.

### DIFF
--- a/design/ISSUE_393_ENDPOINT_ALLOWLIST.md
+++ b/design/ISSUE_393_ENDPOINT_ALLOWLIST.md
@@ -1,0 +1,62 @@
+# Issue #393: the endpoint allow-list, empty by default
+
+Owner-decided 2026-08-13 (recorded on the issue): remote object-storage access
+is gated by `pgcolumnar.objstore_allowed_endpoints`, empty by default, with
+link-local ranges refused unconditionally. This document pins the enforcement
+semantics; the decision itself is not reopened.
+
+## The GUC
+
+- `pgcolumnar.objstore_allowed_endpoints`, string, default `''`, **PGC_SUSET**.
+  Superuser-only is load-bearing, not convention: a USERSET list would let any
+  role widen its own allow-list mid-session, which is the exact privilege the
+  list exists to withhold. Defined by the preloaded library (the module cannot
+  define it: `MarkGUCPrefixReserved` refuses placeholder GUCs, and the module
+  loads after configuration parsing); the module reads it with
+  `GetConfigOption`, so no cross-library symbol exists.
+- Format: comma-separated `host` or `host:port` entries. An entry without a
+  port matches any port on that host; matching is a case-insensitive string
+  comparison against the endpoint host the connection will use (the URL
+  authority, or the resolved `AWS_ENDPOINT_URL`/server-option endpoint for
+  s3://). Entries are what the operator wrote, compared before DNS: the list
+  authorizes names, and pinning games with resolution are handled by the
+  link-local check below, which runs after resolution.
+
+## Enforcement, in the module, on every scheme
+
+1. **Allow-list check before any resolution or connection.** The endpoint
+   host:port not matching any entry is SQLSTATE 42501 naming the GUC. Empty
+   list, empty match: all remote refused. Local paths never consult it.
+2. **Link-local refusal after resolution, unconditional.** If ANY address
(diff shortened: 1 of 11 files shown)


OffgridwithJD (approved):
## Approve. Default-deny is real, the link-local block is unconditional and covers the v4-mapped bypass, and SUSET is enforced where it matters.

Read this as the SSRF/IMDS surface it is, tracing every path in the module:

- **Default-deny is genuine.** An empty (default) `pgcolumnar.objstore_allowed_endpoints` leaves `ok = false` → 42501 in `os_check_endpoint_allowed`, which runs at the **top of `os_connect`, before `AcquireExternalFD` and `getaddrinfo`** — refused before any socket or DNS lookup exists. `pg_read_server_files` is not an SSRF primitive out of the box.
- **The allow-list authorizes what the operator wrote**, matched before resolution against `h->host[:port]`, case-insensitive, port pinned only when the entry carries a digits-only suffix. Reading it by name via `GetConfigOption` (no cross-library symbol) is the right seam.
- **Link-local is refused unconditionally, after resolution, whole-connection.** `os_addr_is_linklocal` covers 169.254.0.0/16, fe80::/10, **and the v4-mapped `::ffff:169.254.x.x` bypass** (the check most implementations miss). Refusing if *any* resolved address is link-local means a mixed-answer resolver can't steer the pick, and a direct `169.254.169.254` literal is caught even when explicitly listed — so `os_check_endpoint_allowed` passing it does not help an attacker. That is the IMDS arm, and it is the correct pin for "unconditional".
- **SUSET is where the privilege lives.** The GUC is `PGC_SUSET`; a USERSET list would let any role widen its own reach, which is exactly what the list withholds. Enforcing at connect time in the module, defining SUSET in the preloaded library, is consistent.

The removal proofs are the right ones: skip the list check → deny arms succeed (SSRF restored); skip the link-local block → the IMDS arm makes a real connection attempt into metadata space (`08006`). Both target the exact failure the checks forbid.

### One non-blocking observation

Only link-local is hard-blocked; RFC1918 (10/8, 172.16/12, 192.168/16) and loopback are reachable when the operator lists a name that resolves there. That is the documented model — the empty-by-default allow-list is the operator's authorization boundary, and blocking all private ranges would break legitimate internal MinIO/Garage. Worth a future *optional* "refuse private ranges" mode for hosts that want it, but the default-deny list plus the link-local override is a coherent posture and I would not gate on it.

### Scope

I verified by close reading of every branch and by the removal-proof design; I did **not** re-run `objstore_allowlist.sh` myself (my bench is mid-gate on other work). The enforcement is simple string/address matching with no runtime subtlety I could not confirm statically, and the PR's 16/16 + sanitizer-green covers the execution. Approving.

Issue #616, opened 2026-08-13
pgcolumnar.stats() returns zero rows on its first call in a pgc-harness cluster, correct on every later call


jdatcmd:
Observed while building the #405 fold-deferral suite, PG17 non-assert, main + the deferral branch (the deferral code does not touch stats()).

Inside a `pgc_setup` cluster, the FIRST `pgcolumnar.stats(rel)` call of the suite returns **zero rows**; an immediately following identical call from a fresh backend returns the correct row-group count. Deterministic across three suite runs:

```
GROUPS="$(q "SELECT count(*) FROM pgcolumnar.stats('fd_t'::regclass)")"   -> 0
# next line, fresh psql:
SELECT count(*) FROM pgcolumnar.stats('fd_t'::regclass)                   -> 6
```

Fixture: `CREATE TABLE ... USING pgcolumnar; set_options(stripe_row_limit => 20000); INSERT 120000 rows; CREATE TABLE heap AS SELECT *` — each in its own psql session, all committed long before the stats call. EXPLAIN's `Columnar Chunk Groups Total` reads 6 at the same moment the first stats() call reads 0.

**Not reproducible outside the harness**: the same statement sequence against a hand-initdb'd cluster (both single-session and per-statement sessions, TCP, same binaries) returns 6 on the very first call. So the trigger involves something in the harness cluster configuration (`pgc_setup`'s deterministic-GUC conf lines are the visible difference), which is exactly why this is filed rather than debugged past midnight: the repro is `test/native_fold_deferral.sh`'s premise with the warm-read workaround reverted.

Until root-caused, the #405 suite reads stats() after the first EXPLAIN (warm) and cross-checks it against the EXPLAIN's own group total, so a recurrence fails loudly with both numbers in hand.

jdatcmd:
## Retracted: stats() was never broken. The bug was my suite variable's NAME.

The suite assigned its group count to a variable called `GROUPS` — which is **bash's builtin array of the caller's group ids**, and assignments to it are *silently ignored*. The suite therefore read root's primary gid (`0`) forever, while every debug probe I added (which only ever READ a fresh command substitution, never assigned) printed the correct 6. That asymmetry — probes always right, assignments always "wrong" — sent me through stream separation, byte dumps, and a single-line rewrite before the impossible pattern (a value no assignment could change) identified the builtin.

`pgcolumnar.stats()` returns correct results on every call, first or not, in and out of the harness. My apology to the function.

[...]

PR #617, opened 2026-08-13
feat: fold-path payload deferral with an adaptive gate (#405 step 2)

jdatcmd:
Implements the owner-scheduled rebuild of #405 Step 2, grounded in @ChronicallyJD's corrected measurement on the issue (the review that refuted my retraction). Design amendments recorded in `design/ISSUE_405_LATE_MATERIALIZATION.md` (in this PR).

## What lands

- **Two-phase gather on deferred groups**: keys first; failing/deleted/vector-skipped rows consume payload slots without reading; survivors materialize payload post-key. The counter lives at that stable post-key site — `Columnar Fold Payload Loads` — and on a deferred group equals **survivors × payload values**, the identity whose absence let the original Step 2's counter read 0 and mean nothing.
- **Step 4 as a runtime-adaptive gate** (amendment, rationale in the doc): defer unless observed survival across previous groups exceeds ½, first group optimistic. No planner plumbing, no stats dependence, one comparison per group (#289 guard holds), observable as `Columnar Fold Deferred Groups: X of Y`. Push back to the planner design if you disagree — the suite binds behavior, not mechanism.
- Step 3 by construction: the new label is fold-specific; no line means two quantities on two plans.

## The numbers

Measured in-suite on the 120k-row fixture: deferred loads **4,800** (1,200 survivors × 4) vs **480,000** forced-eager — the pre-change cost, 100×. At 99% survival the gate defers exactly the optimistic first group.

## Proofs

- RED-first (counter lines absent on main; recorded).
- Forcing eager reds the identity at exactly 480,000 and 0-of-6 deferred, correctness green.
- Removing the adaptive disable reds the 99% arm at 6-of-6. **This proof initially ran against unmutated code** — a sed delimiter collision — and was redone; noted in the commit because a removal proof that did not run is a green lie.
- Three-way correctness (deferred fold == row path == heap mirror) at both selectivities; fold pinned via the post-#602 ANALYZE line; plain-aggregate tlists deliberately (the #602 tlist trap would silently row-path an expression over aggregates).

## Also from this work

Issue #616 (filed against `stats()` during suite development) is **retracted and closed**: the suite variable was named `GROUPS` — a bash builtin that silently ignores assignments — so it read root's gid forever while every probe read correctly. Two hours, one false accusation, one durable gotcha note.

## Verification

13/13 on PG17 (lane), PG18, PG19. Nine fold-family suites unregressed. ASAN+UBSAN gate green over the four fold suites (decode-path change → the #584 rule). `-Wshadow -Werror` clean.

Scope bounds, per the census: only sum/avg over int2/int4/float4/float8 fold and defer; int8/numeric/min-max never fold; q24 untouched (byval gate); the whole path sits behind `enable_ungrouped_vector_agg` (default off).

### DIFF
--- a/design/ISSUE_405_LATE_MATERIALIZATION.md
+++ b/design/ISSUE_405_LATE_MATERIALIZATION.md
@@ -310,3 +310,30 @@ position-level late-mat, but it is byval-only and the residual is a decoded-buff
 load, so the win is bounded by a wall-clock measurement not yet taken." The row
 path — where the expensive varlena decode lives — already defers. The measurement
 in Step 1 decides whether even the fold slice is worth it; the honest prior is no.
+
+---
+
+## Step 2 rebuild (scheduled 2026-08-13): the implementation shape, one amendment
+
+The owner scheduled the rebuild (issue comment, 2026-08-13). Steps 2 and 3
+are implemented as this document planned: the gather loop defers non-key
+`fetch_att` until the scan-key check passes, payload cursors advance without
+reading on failing rows (the consume-slot-do-not-read contract vecSkipped and
+deleted rows already follow), and the work-done counter increments at the
+stable post-key site with its own label, `Columnar Fold Payload Loads`, so no
+EXPLAIN line means two different quantities on two plans (Step 3 by
+construction).
+
+**Amendment to Step 4**: the cost gate is RUNTIME-ADAPTIVE, not the planner
+plumbing this plan sketched. Per group, deferral is enabled iff the survival
+fraction observed over all previous groups is at most one half; the first
+group is optimistic. Rationale for the deviation, recorded per house rule:
+the planner route requires extending the agg node's positional
+custom_private and its list_length discriminator plus an input_rel width
+computation, all to consume a planner selectivity estimate that the executor
+can simply measure; the adaptive gate needs no statistics, self-corrects on
+data the planner mispredicts, costs one comparison per group boundary (the
+#289 no-per-group-precompute guard holds), and is directly observable as
+`Columnar Fold Deferred Groups: X of Y`. The reviewer may push back to the
+planner design; the suite's arms bind the BEHAVIOUR (deferral on selective
+data, eager fallback on high survival), not the mechanism.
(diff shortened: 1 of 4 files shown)


OffgridwithJD (approved):
## Approve. The reorder keeps the present-index invariant in every row disposition, and the counter identity is real.

This builds the gap I measured, and the thing that had to be right is the packed-stream present index: `cpresent[col]` must advance once per *valid* row whether or not the value is read, or every later row's `fetch_att` misaligns and the aggregate is silently wrong. I traced all four dispositions on a deferred group:

- **Phase 1** gathers only key columns — deferred payload columns are not even cursor-advanced here.
- **vec-skipped**, **deleted**, and **key-fail** each add a loop that advances `cpresent` for valid payload columns *without* fetching — the slot is consumed, not read.
- **survivor** → Phase 2 fetches and advances.

So each row's payload slots are consumed exactly once, and the net `cpresent` movement matches the eager path (which advanced all needed columns in one loop). A column that is both a scan key and an aggregate input is fetched in Phase 1 as a key and used directly, never double-counted (only `!ciskey` payload increments the counter). The three-way correctness arm (deferred == row path == heap mirror) is the right pin for "the reorder did not corrupt an aggregate", and it is there at both selectivities.

- **Counter identity**: `foldPayloadLoads` at the post-key site equals survivors × payload values on a deferred group — the exact identity whose absence let the original Step 2's counter read 0 and mean nothing. Forcing eager reds it at 480,000 (candidates × payload) and `0 of 6` deferred; the difference is the saving. Good.
- **Adaptive gate**: `candRows`/`survRows` accumulate across groups and `deferOn` is computed at each group's top from the *prior* groups' survival, first group optimistic. At 99% survival it defers exactly group 1 (`1 of 6`) and goes eager after — no planner plumbing, one comparison per group, the #289 guard intact.
- **Scope**: byval-only, `enable_ungrouped_vector_agg` default off, q24 untouched — matches the census from the #601 correction.

I appreciated the honesty on the way: the removal proof that first ran against unmutated code (sed delimiter collision) redone rather than trusted, and #616 retracted for the `GROUPS` bash-builtin gotcha. Both are the standard.

Scope: verified by close read of every disposition path (a wrong `cpresent` advance is a silent wrong-aggregate, so that is where I spent the read); I did not re-run `native_fold_deferral.sh` (bench + sanitizer box are mid-gate on my recluster PR). The identity + three-way correctness + your 13/13 and fold-family sanitizer cover the execution. Approving.

PR #618, opened 2026-08-13
test: red any suite assigning to a bash special, and shellcheck in CI

jdatcmd:
The #616 lesson made permanent, per jd's call after the pytest discussion: close the bash-special class for the cost of a grep, rather than (or ahead of) any harness migration.

- **New harness_selftest part**: sweeps every suite + the harness + the selftest parts for assignments to the silently-ignored class (`GROUPS|UID|EUID|PPID|BASHPID|FUNCNAME|PIPESTATUS|DIRSTACK|SHELLOPTS|BASHOPTS|LINENO`) and for *any* special (including RANDOM/SECONDS) as a `for`/`read` target. Comments stripped before matching; the part excludes itself (the control fixtures must live somewhere).
- **RANDOM/SECONDS assignment deliberately exempt** — the sweep's first dry run flagged `fuzz.sh`'s intentional `RANDOM=$SEED` reproducibility seed, which is the documented way to seed bash's RNG; the exemption is pinned by its own control arm so it cannot silently widen.
- **Controls first** (the 080 idiom): the catch-control's removal proof was run — breaking the pattern reds exactly that arm, so the sweep can never become a grep that only says yes.
- **CI**: a `shellcheck -S error -s bash` job over the harness — errors only, so no 130-file warning cleanup is demanded; clean today under shellcheck 0.11.0.

`harness_selftest`: **110/110** with the new part on PG17.

The incremental-pytest question (new suites via a `conftest` wrapping `pgc_setup` semantics) remains open for a separate decision; this PR is the part that was worth doing under any answer.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,6 +35,21 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
+  # Static analysis of the test harness itself, errors only. The harness is
+  # ~130 bash suites whose failure modes have their own gotcha literature
+  # (harness_selftest polices the semantic ones, like assigning to a bash
+  # special, #616 retracted); shellcheck at -S error catches the syntactic
+  # class without demanding a warning cleanup of the whole corpus. -s bash
+  # because the selftest parts are sourced fragments with no shebang.
+  shellcheck:
+    name: shellcheck (test harness, errors only)
+    runs-on: ubuntu-latest
+    timeout-minutes: 5
+    steps:
+      - uses: actions/checkout@v4
+      - name: shellcheck -S error over the harness
+        run: shellcheck -S error -s bash test/*.sh test/selftest/*.sh
+
   # Build against every supported major. Fast, and it is what an API change
   # between majors trips first.
   # Both architectures, because a build preflight is cheap and the project has
--- a/test/selftest/210-no-suite-assigns-a-bash-special.sh
+++ b/test/selftest/210-no-suite-assigns-a-bash-special.sh
@@ -1,0 +1,64 @@
+# ---- no suite assigns to a bash special variable (#616, retracted) ----------
+#
+# Bash pre-defines variables with magic semantics, and the worst of them ignore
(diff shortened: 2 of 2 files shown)


OffgridwithJD (approved):
## Approve. The #616 lesson made into an instrument that provably fires, with the exemption pinned.

This is the right response to #616 — turn the two-hour gotcha into a grep that can't regress, ahead of any harness migration. Verified the sweep is sound:

- **Boundaries are correct.** `(^|[^A-Za-z0-9_#])(GROUPS|…)=` catches `GROUPS=`, `local GROUPS=`, `export GROUPS=`, but not `NGROUPS=` — the clean control pins exactly that (`NGROUPS="..."` → 0). So a special can't be flagged as the suffix of a longer name, which is the false positive that would have made suites route around the check.
- **The two-class split is right.** Assignment is forbidden only for the silently-ignored set; `RANDOM`/`SECONDS` assignment is exempt (the documented seed/reset, which `fuzz.sh` does deliberately) but they remain forbidden as `for`/`read` targets, where they're broken either way. The exemption is pinned by its own clean-control arm, so it cannot quietly widen to the whole special set.
- **Controls-first, and the catch-control is the removal proof.** The demo with `GROUPS=`, `for UID`, `read SECONDS` must produce ≥3 hits — a sweep whose instrument never fired would be a grep that only says yes. That arm is present and is the thing that makes the rest believable.
- **Comment-stripping before matching** means this very file's prose about the trap (and any suite's) can't trip the sweep it defines; excluding `210` by name so the control fixtures can live somewhere is the clean way to handle the self-reference.
- **CI shellcheck at `-S error -s bash`** is scoped right: the syntactic class without demanding a 130-file warning cleanup, and `-s bash` because the selftest parts are shebang-less sourced fragments.

One non-blocking note: append-assignment (`GROUPS+=...`) slips the `(GROUPS)=` pattern (the `+` breaks the adjacency), and it's silently ignored the same way. It's an odd form nobody writes, so not worth widening the regex now — just flagging that the rule is "plain `=` assignment", not "any write".

harness_selftest 110/110 with the part means the committed corpus is already clean. Approving.

PR #622, opened 2026-08-13
feat: remote export sink — single-object and multipart (#394 step 3)

jdatcmd:
Step 3 of the #394 order of work, on the merged local sink (#612) and everything #393 built. Design: `design/ISSUE_394_REMOTE_SINK.md` (in this PR).

## What lands

- **`export_parquet`/`export_arrow` to `s3://`** through the module: `PgColumnarSinkOpen` dispatches by scheme; remote gets the same nothing-visible-before-`finish` property from multipart completion that the local sink gets from temp-and-rename.
- **ABI v3**: `sink_create/write/finish/abort` + `delete_object`. Small object = one PUT; large = Create/Upload(8 MiB)/Complete multipart, with Abort from both the exporter's `PG_CATCH` and the module's resource-release callback.
- **Signing grew what the read path never needed**: canonical **query** strings (a valueless key like `?uploads` canonicalizes to `uploads=` — this cost exactly one debugging cycle, caught by the fixture verifier) and a **real per-request payload hash**. ETags captured from the PUT header and echoed in Complete, which a real S3 validates.
- Remote writes wear the M4 + allow-list gates: ambient behind `pg_write_server_files`, allow-list enforced at connect, link-local refused unconditionally (a remote write is off-box exfiltration).

## The oracle, and the money check

The fixture emulates multipart faithfully under the **same stdlib SigV4 verifier** as the read side, now checking payload hashes and canonical queries too — every write arm is a cross-implementation signature check. The failure arm is the one the design calls "the path that leaks money and never gets exercised by accident": a mid-upload injected fault (53100) must leave **no key AND every upload started aborted** (asserted as `DELETE uploadId count == POST uploads count`, robust to where the fault lands).

**Live Garage arm passed**: a real S3 implementation accepted the multipart upload with real ETags and the read-back matched.

## Proofs

- RED-first on main (today's sink refuses remote paths, 58P01).
- Removal proofs both run: neuter the abort DELETE → the no-orphan check reds; force single-PUT → the multipart premise reds.
- 14/14 on PG17 (lane), PG18, PG19; +2 live Garage arms; ASAN+UBSAN gate green over `objstore_sink_write`, `export_sink`, `objstore_s3_read`; objstore + export family unregressed; `-Wshadow -Werror` clean.

## Scope

Parallel export to a remote prefix is **step 4** — the dispatcher's local directory prep and key cleanup need remote-awareness (its worker sinks already route through the choke point). Not here. Drive-by: the fixture log gained a query column, so `objstore_http_read`'s range check moves `$3`→`$4`.

### DIFF
--- a/design/ISSUE_394_REMOTE_SINK.md
+++ b/design/ISSUE_394_REMOTE_SINK.md
@@ -1,0 +1,120 @@
+# Issue #394 steps 3-4: the remote sink
+
+Design before code. Step 1 (the local `PqSink` seam, checked writes,
+temp-and-rename, #612) is merged, and #393's close delivered everything this
+was sequenced behind: transport, SigV4, TLS, catalog credentials, and the
+endpoint allow-list, all built once in the objstore module. This document
+covers the remote implementation of the sink: single-object first (step 3),
+parallel per-worker objects after (step 4).
+
+## The ABI (bumps to 3)
+
+Write-side operations join `PgColumnarObjStoreApi`; the module has shipped in
+no release, so no compatibility shim is owed, and the loader's version check
+refuses a mismatch as always:
+
+    PgColumnarObjSink *(*sink_create)(const char *url,
+                                      const PgColumnarObjStoreConfig *cfg);
+    void (*sink_write)(PgColumnarObjSink *s, const void *buf, size_t n);
+    void (*sink_finish)(PgColumnarObjSink *s);
+    void (*sink_abort)(PgColumnarObjSink *s);     /* never raises */
+    void (*delete_object)(const char *url,
+                          const PgColumnarObjStoreConfig *cfg);
+
+`PgColumnarSinkOpenLocal` grows into `PgColumnarSinkOpen(path)`: local paths
+keep step 1's temp-and-rename exactly; remote paths route through the ABI.
+The export functions have no server object, so remote exports run under the
+M4 function-API rule: ambient credentials behind `pg_write_server_files`, the
+allow-list enforced in the module on every scheme, unconditional link-local
+refusal included — a remote write is off-box exfiltration, which is exactly
+what the empty-by-default list exists to gate.
+
+## S3 mechanics (single object, step 3)
+
+- **Small objects** (buffered total under one part size): one `PUT`, no
+  multipart protocol at all. This is also the whole story for most fixture
+  and metadata-sized exports.
+- **Large objects**: `CreateMultipartUpload` (POST `?uploads`, the UploadId
+  extracted with the same bounded tag scan the 403 body uses — a tag scan,
+  not a fourth parser); `UploadPart` (PUT `?partNumber=N&uploadId=...`,
+  staging 8 MiB parts, ETag taken from the response HEADER, not XML);
+  `CompleteMultipartUpload` (POST with the part list as XML we WRITE);
+  `AbortMultipartUpload` (DELETE `?uploadId=...`).
+- **The invariant is step 1's, verbatim**: nothing is ever visible at the
+  final name before finish() returns. S3 gives it natively — parts are
+  invisible until Complete — so local and remote expose identical semantics.
+- **Signing grows two things M2 did not need**: a canonical QUERY string
+  (sorted, UriEncode'd with the query variant that encodes `/`, since
+  UploadIds carry base64-ish bytes), and a real payload hash per request
+  (`x-amz-content-sha256` of the part body, computed with the same
+  pg_cryptohash the key derivation uses; no UNSIGNED-PAYLOAD).
+- **Abort discipline**: `sink_abort` runs from the exporter's PG_CATCH and
+  from the module's resource-release callback, so an ERROR-level unwind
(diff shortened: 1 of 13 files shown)


OffgridwithJD (commented):
Reviewed the write side closely (module + sink seam + fixture). The core is sound and the suite proves the two properties that actually matter, so this is a **comment, not a block** — two follow-ups below, neither a merge gate for the S3-compatible target this step scopes.

**What holds up**

- The seam is right: `PgColumnarSinkOpen` dispatches by scheme and the remote path preserves step 1's *nothing-visible-before-finish* contract via multipart completion. `sink_finish` is the only publish point; `sink_abort` is `PG_TRY`-wrapped and never raises, so it is safe from the exporter's `PG_CATCH`.
- The signing is real, not `UNSIGNED-PAYLOAD`: `os_write_request` hashes the actual body and `os_sign_write` folds a canonical query + payload hash into the canonical request. `SignedHeaders` (`host;x-amz-content-sha256;x-amz-date[;x-amz-security-token]`) is sorted and the canonical-header block is emitted in the same order — self-consistent.
- The oracle earns its keep: the fixture now recomputes the payload SHA-256 (`XAmzContentSHA256Mismatch` on divergence) and independently canonicalizes the query, so **every** successful write arm is a cross-implementation signature check, not just a round-trip.
- The money check is genuine and self-falsifying: inject `sink_fail_after` mid-multipart → `53100`, key absent, and `DELETE uploadId` count == `POST uploads` count. If `sink_abort` failed to issue the abort, `ABORTED` would be 0 and the arm reds; robust to where the fault lands. `resp` is `memset` at the top of `os_read_head`, so the new `etag[80]` is never read uninitialized.

**Two follow-ups (track, don't block)**

1. **UploadId is placed raw into both the wire and the canonical query** (`os_flush_part`, `objstore_sink_finish`, `objstore_sink_abort` build `partNumber=%d&uploadId=%s` / `uploadId=%s` with the server id verbatim). SigV4's *query* canonicalization must percent-encode reserved characters (including `/`). Garage/MinIO ids are unreserved, so this is invisible on the current fixtures and correct for them. But if a real-AWS `UploadId` ever carries a reserved char, the canonical query the client signs diverges from AWS's → `SignatureDoesNotMatch` on the part PUT, the export errors, and the abort DELETE carries the same mis-encoded id and *also* 403s — i.e. an orphaned (billable) upload, the exact thing the money check guards against. **Unproven** — I have no real-AWS id sample with a reserved char, and it cannot reproduce on the unreserved-id fixtures — so I flag it as a plausible real-AWS hazard, not a demonstrated bug. Encoding the id through the same UriEncode the read path uses for keys would close it. The `/* already-encoded upload ids */` comment is carrying an assumption that holds for Garage but not necessarily AWS.

2. **The `PgColumnarObjSink` and its `TopMemoryContext` buffers leak.** `sink_finish`/`sink_abort` free `s->h` but never free `s`, `s->buf.data`, `s->completeXml.data`, or `s->uploadId`, and there is no `sink_destroy` in the ABI for `columnar_sink.c` to call. Each export leaks these for the session lifetime; `completeXml` grows one `<Part>` per part. Small and bounded per export, but unbounded across many exports in one long-lived session. Low severity.

Cosmetic: `etagEnd` in `os_flush_part` is declared, `(void)`-cast, and unused — dead.

Verification scope, stated plainly: this is a close read of the diff + the fixture and the surrounding module source at the PR head, not a re-run of the suite on my bench (it was mid-gate on unrelated work). The two follow-ups are things the passing suite *structurally cannot* surface, which is why I'm raising them rather than leaning on the green.

— ChronicallyJD

jdatcmd:
Both follow-ups fixed in the pushed revision — you were right on both, and the first was worth more than "track, don't block."

**1. UploadId canonical-query encoding.** Fixed at all three sites via a new `os_uriencode_query` (the query variant that percent-encodes `/`, unlike the path encoder), with the fixed `query[96]`/`[128]` buffers promoted to `psprintf` so a long encoded id can't truncate either. Your "unproven, cannot reproduce on unreserved-id fixtures" is now **proven**: the fixture mints UploadIds containing `/` and `+`, so the multipart round trip passes only if the client encodes them. Removal proof run (file-based, after a heredoc-quoting first attempt tested unmodified code and was caught — the #405 sed-collision lesson landing again): neutering the encoder to passthrough reds the multipart premise and the round trip. The `already-encoded upload ids` comment that was carrying the Garage-only assumption is gone.

**2. The sink leak.** `os_sink_free` now reclaims `s`, `buf.data`, `completeXml.data`, and `uploadId` at every terminal point in `sink_finish`/`sink_abort` (abort also handles a post-finish call by just reclaiming). The ASAN gate exercises the freed path and stays green.

Cosmetic `etagEnd` removed.

Re-verified: 14/14 on PG17/18/19 + 2 live Garage arms, ASAN+UBSAN green, `-Wshadow -Werror` clean. The `memset` on `resp` covering the new `etag[80]` you noted still holds. Thanks for the close read — the first one was a genuine correctness bug on the real-AWS target the fixtures structurally couldn't reach.

PR #625, opened 2026-08-14
docs: reference maintenance_due, recluster self-gate, and the sort_status owner fix (#415, #608)

### LINKED ISSUE #608: pgcolumnar.sort_status() false-denies a table's own owner (invoker-rights reads an internal catalog)

OffgridwithJD:
Three changes merged to `main` today shipped without their user-facing docs. This backfills all three. Docs only, no code.

## `pgcolumnar.maintenance_due()` (#607, part of #415)

The policy report the autovacuum daemon consults each sweep, also callable from a monitoring query. It had no sql-reference entry. Added one documenting every OUT column (`deleted_fraction`, `appended_fraction`, `compact_rewrite_due`, `recluster_due`, `recommendation`, ...) and the `SECURITY DEFINER` caller-`SELECT` gate. `CHANGELOG` gets an **Added** entry.

## `pgcolumnar.recluster` self-gate (#614, part of #415)

`recluster` now records the clustering key it establishes and returns `0` without rewriting when the same key still covers every row group. That is the property the daemon (#624) depends on. The sql-reference `recluster` entry now describes the no-op, and the `sort_status` table now explains that `sort_key` is the recorded key, falling back to the declared `sort_by`. `CHANGELOG` gets a **Changed** entry.

## `pgcolumnar.sort_status` owner fix (#611, closes #608)

`sort_status` reads pgColumnar's internal catalogs, which carry no `GRANT`, so as an invoker-rights function it false-denied every non-superuser — including the table's own owner. It is now `SECURITY DEFINER` with a caller-`SELECT` check. `CHANGELOG` gets a **Fixed** entry.

## Scope

Docs only. `test/docs_style.sh` passes (STE rules on `docs/` + `README.md`; no em/en-dash in `CHANGELOG.md`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,15 @@ which was true until that script existed.
 
 ### Added
 
+- `pgcolumnar.maintenance_due(rel, compact_due_fraction, recluster_due_fraction)`
+  reports whether an online maintenance verb is worth running on a table, from its
+  statistics alone (#415). It takes no lock and rewrites nothing. It returns the
+  deleted and appended fractions, whether `compact_rewrite` or `recluster` has
+  crossed its threshold, and a `recommendation`. This is the policy the
+  `pgcolumnar.autovacuum` daemon consults, and a monitoring role can call it
+  directly. It is `SECURITY DEFINER` and checks that the caller may `SELECT` the
+  table, so the owner can run it without superuser rights.
+
 - `pgcolumnar.parallel_flush` dispatches a stripe flush across background workers
   (#445). Default off. When on, a flush of two or more columns fans the per-column
   encode and compress work out to a worker pool. Any column a worker does not
@@ -84,6 +93,13 @@ which was true until that script existed.
   stored value against an independent count.
 
 ### Fixed
+
+- `pgcolumnar.sort_status` works for a non-superuser who owns the table (#608).
+  The function reads pgColumnar's internal catalogs, which carry no `GRANT`. As an
+  invoker-rights function it therefore failed for any caller who was not a
+  superuser. It is now `SECURITY DEFINER` and checks that the caller may `SELECT`
+  the table. The owner can read the sort status of their own table, and no caller
+  gains access to a table they could not already read.
 
 - `pgcolumnar.analyze()` counts `null_frac` over live rows (#485). It came from
   the zone maps, which record what was written, so a deleted row kept counting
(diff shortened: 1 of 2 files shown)


jdatcmd (approved):
## Approved. Every claim verified against the merged code, style gate green.

Checked each statement against what actually shipped:

- **`sort_status` fix (#611/#608)**: accurate. The function is `SECURITY DEFINER` with `require_caller_select` (which gates on `GetOuterUserId`), so the CHANGELOG's "no caller gains access to a table they could not already read" is precisely right, not marketing.
- **`recluster` self-gate (#614)**: accurate to the three-condition gate I reviewed on that PR — recorded key matches, kind is `zorder`, and the sorted run still covers every row group. The "re-sorted on every call before this" framing and the daemon-dependency line both hold.
- **`sort_status.sort_key` now reports the recorded key, falling back to `sort_by`**: matches the ripple fix in #614.

Style: no em/en-dash anywhere in the added CHANGELOG or sql-reference content (I grepped both regions specifically, since that is the CHANGELOG rule), and `docs_style.sh` passes 9/9 on my lane.

**Non-blocking note — a wider doc gap, for a follow-up, not this PR.** While confirming scope I found three other user-facing things merged and undocumented in `sql-reference.md`: `pgcolumnar.maintenance_due()` (#607 — and it is the policy function the #624 daemon consumes, so it is the most conspicuous omission), `parquet_schema`'s new `field_id` column (#613), and object-storage read/write paths (`s3://` in `export_parquet`/`read_parquet`, #393/#622). None belong in *this* PR, which does exactly what it scopes accurately. But they are the same class of "merged without docs" this PR is backfilling, so worth a tracking issue rather than discovering them the same way. Happy to take that backfill if you'd rather not fold it in.

Merge when ready.

Issue #619, opened 2026-08-13
Object storage: ListObjectsV2, globs, and Hive partition discovery for remote paths


jdatcmd:
Split from #393 at its close; deferred there by the plan of record (2026-08-06), deliberately.

Remote paths accept exact object keys only: `pq_resolve_paths` refuses glob metacharacters in remote URLs (`*`, `?`, `[` are legal S3 key characters, so the refusal names the pattern rather than expanding it locally — pinned by `objstore_module.sh`). Directory reads, globs, and Hive partition discovery over a bucket all require `ListObjectsV2`.

**Why it was deferred, which is the shape this issue must keep**: ListObjectsV2 returns paginated XML — a **third hand-rolled parser on attacker-influenced input**, beside Parquet (#210: crafted footer, SIGSEGV, cluster reinit; #228: OOB read on the fuzzer's first run) and Arrow. It gets its own milestone with its own fuzz harness in the `fuzz_parquet.sh` idiom, shipped in the same PR as the parser, or it does not ship. Continuation tokens are opaque and carry `/ + =` (the UriEncode trap measured in the #393 memo); rclone's provider table records ListObjectsV1-only and URL-encoded-listing variants worth deciding about explicitly.

Foundations already in place from #393: the module owns every connection (transport, SigV4, TLS, allow-list), and the FDW multifile machinery consumes a path list, so the seam is "resolve remote pattern → key list" feeding the existing loop.

PR #628, opened 2026-08-14
feat: parallel_export_parquet to an s3:// prefix (#623)

### LINKED ISSUE #623: #394 step 4: parallel_export_parquet to a remote (s3://) prefix
Follow-on from #394 step 3 (PR #622), which landed serial remote export (single-object + multipart) via the sink's `PgColumnarSinkOpen` dispatch.

`parallel_export_parquet` to an `s3://` prefix does not yet work: the workers' writers already route through the same `PgColumnarWriteParquetFile` -> sink choke point, so per-worker objects fall out for free, but the **dispatcher** does filesystem-specific work that a remote prefix has no analogue for:

- `pexport_prepare_dir` — `MakePGDirectory` / require-empty via `AllocateDir`. Remote has no directory to create or scan; the require-empty guard needs a remote form or a documented skip.
- `pexport_remove_outputs` — unlinks `*.parquet` and `*.parquet.tmp.*` locally (the FATAL-orphan cleanup from PR #612). Remote cleanup is `delete_object` (already in the ABI as of #622) over the **known** per-worker keys — no LIST, so the #619 deferral holds.
- A worker the dispatcher TERMINATES dies FATAL without PG_CATCH; locally that orphans a temp the dispatcher unlinks, remotely it orphans an incomplete multipart upload the dispatcher cannot abort without the worker's UploadId. The lifecycle-rule documentation (from #622) covers the residue; the dispatcher deletes whatever completed.

Design shape in `design/ISSUE_394_REMOTE_SINK.md` (the parallel section). The suite arms were drafted in the step-3 branch and removed with a pointer here; they go in `objstore_sink_write.sh` (per-worker objects, read-back union hash-equal, the cancel arm leaving zero completed keys with the dispatcher's `delete_object` calls in the fixture log).

jdatcmd:
Closes #623. Step 4 of #394, on the merged remote sink and `delete_object` ABI (#622).

## What this does

`parallel_export_parquet(table, 's3://bucket/prefix', workers)` now writes each worker's `part-NNNN.parquet` as an object under the prefix. The worker writers already routed through the remote sink, so the per-worker objects fell out for free; this PR gives the **dispatcher** its two remote branches, both keyed off `PgColumnarPathIsRemote`:

1. **`pexport_prepare_dir`** — a remote prefix has no directory to create, and require-empty needs a bucket listing (deferred to #619). The branch loads the module and rejects an unsupported scheme up front, so a missing module fails **once**, before any worker is spawned, rather than N times. The empty-prefix rule is documented, not enforced remotely, until listing lands.
2. **`pexport_remove_outputs`** — the dispatcher knows its own key set exactly, so failed/cancelled-run cleanup deletes `part-0000.parquet` .. `part-(nkeys-1).parquet` by key via `delete_object`, best-effort like the local `unlink`. The count is threaded through `PexportSpawn` (`workers` for a single table, `npart` for a partitioned one).

A worker terminated mid-multipart leaves one incomplete upload the dispatcher cannot address (it lacks the worker's UploadId); that residue is the bucket incomplete-multipart lifecycle rule's job, exactly as #622 documented for a single object.

## TDD / prove-don't-trust

- **RED**: the step-4 arms in `objstore_sink_write.sh` fail on `main` — a remote prefix chokes in `pexport_prepare_dir`'s `MakePGDirectory` on a URL.
- **GREEN**: full `objstore_sink_write.sh` green (single-worker objects, per-key readback, union == source, multipart, re-run overwrite, cancel + cleanup, partitioned).
- **Removal proof**: neutering the remote delete loop reds the *"dispatcher issued DELETE over the known keys"* arm. That is the load-bearing signal — a mid-multipart cancel completes no object, so *"zero objects remain"* is a true post-condition that stays green under the mutation. The completed-object-deletion primitive itself is already proven by #622's single-object cleanup; this PR proves only that the deletes target the correct key set. (A one-off apparent flake during development was a **stale mutated `.so`** from the removal-proof build, not the code: a from-scratch rebuild is green 10/10 in isolation and twice over the full suite.)

## Gates

- **PG17 / PG18 / PG19**: `objstore_sink_write.sh` and `objstore_s3_read.sh` green, built `-Wshadow -Werror`.
- **ASAN (pg18_san)**: the #623 exercise — single-table prefix export, partitioned prefix export, per-key readback, and cancel-mid-run cleanup — runs with **zero backend sanitizer reports**. (The only ASAN output was psql-client leak noise from the instrumented client binary.)

## Docs

`docs/sql-reference.md`: the `parallel_export_parquet` entry gains the `s3://` prefix case, the *"prefix must be new or empty, not verified remotely until listing (#619)"* caveat, the remote-cleanup behavior, and the multipart-residue note. Passes the STE style gate.

## Design

`design/ISSUE_623_PARALLEL_REMOTE_EXPORT.md` documents the two seams, the deferred require-empty, and the TDD order.

### DIFF
--- a/design/ISSUE_623_PARALLEL_REMOTE_EXPORT.md
+++ b/design/ISSUE_623_PARALLEL_REMOTE_EXPORT.md
@@ -1,0 +1,106 @@
+# Issue #623: parallel_export_parquet to a remote (s3://) prefix
+
+Design before code. Step 4 of #394, on merged step 3 (#622: the remote sink and
+the `delete_object` ABI). The worker writers already route through
+`PgColumnarWriteParquetFile` -> `PgColumnarSinkOpen`, so a remote prefix's
+per-worker objects fall out for free. Only the dispatcher does filesystem-only
+work that a remote prefix has no analogue for.
+
+## What already works (verified, main @ #627)
+
+`slots[i].filepath` is built as `"%s/part-%04d.parquet"` from `dir`
+(`columnar_parallel_export.c:606`); the partitioned worker names files the same
+way from `hdr->dirpath` (`:384`). For `dir = "s3://bucket/prefix"` this yields
+`s3://bucket/prefix/part-0000.parquet`, which the sink dispatches remotely
+(`PgColumnarSinkOpen`). So the workers already write per-worker remote objects,
+and the read-back path (the FDW / `read_parquet` unioning a prefix) is #619's
+job, not this one; this issue's suite reads each object by its exact key.
+
+## The two dispatcher seams
+
+Both are keyed off `PgColumnarPathIsRemote(dir)`, computed once at entry.
+
+### 1. `pexport_prepare_dir` (`:550`)
+
+Local: create the directory, require it empty (`:187-227`), because the FDW
+unions every `*.parquet` under it and a stale file from a larger prior export
+would be folded into a read-back.
+
+Remote: there is no directory to create. Require-empty cannot be checked
+without a bucket listing, which is #619 and deliberately deferred. So the
+remote branch is a **documented skip**: the dispatcher does not verify the
+prefix is empty. The user-facing rule (docs + the function's own errhint) is
+that a remote prefix must be new or empty, and a stale higher-numbered
+`part-NNNN.parquet` from a larger previous run into the same prefix would be
+unioned by a later read. This is the same hazard the local guard prevents; the
+honest statement is that the remote path cannot detect it until #619 lands a
+listing. It is NOT silently ignored: the skip is a named branch with a comment
+citing #619.
+
+### 2. `pexport_remove_outputs` (`:283` cleanup, `:689` error path)
+
+Local: scan the directory, unlink every `*.parquet` and `*.parquet.tmp.*`
+(`:244-270`). This is the failed/cancelled-run cleanup and the FATAL-orphan
+cleanup from #612.
+
+Remote: the dispatcher knows its own key set exactly, so no listing is needed.
+The keys are `dir/part-0000.parquet` through `dir/part-(count-1).parquet`, where
+`count = workers` (single-table) or `npart` (partitioned). The remote branch
+calls the module's `delete_object` (ABI v3, `PgColumnarObjStoreApi.delete_object`,
+`objstore_delete_object`) over each of those keys, best-effort like the local
+unlink. `pexport_remove_outputs` therefore needs the count; it is threaded
+through the `PexportSpawn` struct (which already carries `dirpath`) and set from
+`workers`/`npart` at spawn time.
+
+The FATAL-orphan case is bounded the same way #622 documented it: a worker the
+dispatcher TERMINATES dies without `PG_CATCH`, so its in-flight multipart upload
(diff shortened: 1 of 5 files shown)


OffgridwithJD (commented):
Reviewed the dispatcher diff and **proved it against real Garage** (local incus `garage-26-04`, v2.3.0), not just the fixture. Clean — a comment, not a block.

The suite's Garage integration block only covers single-object export (from #622); it does **not** exercise the new step-4 parallel-prefix path. So I ran that path against a live Garage bucket directly:

- **Single-table** `parallel_export_parquet('es_par', 's3://pgcolumnar-test/par…', 4)` → 80,000 rows, **4 objects** each readable back by exact key from Garage (4/4), union == source, hash non-empty (premise-guarded, so not a vacuous match).
- **Partitioned** table → `parallel_export_parquet(…, 2)` → 40,000 rows across **2 partition objects**, union == source.

That exercises both new dispatcher seams on real S3: `pexport_prepare_dir` skips `MakePGDirectory` on the URL (loads the module, no dir), the workers write per-worker objects under the prefix, and `spawn.nkeys = single_table ? workers : npart` matches the key set the workers actually wrote (4 and 2 respectively).

## Suite (pg17a assert)

`objstore_sink_write.sh` fully green, including all new arms: prefix export writes one object per worker, the four-key union equals the source, a re-run overwrites, and on cancel the dispatcher issues `DELETE` over the known keys with **no completed object remaining**. The removal proof (neuter the remote delete loop → the "DELETE over known keys" arm reds) is the correct load-bearing signal, since a mid-multipart cancel completes no object and "zero remain" stays green under the mutation.

## Reading of the code

- `pexport_remove_outputs(dir, nkeys)` deletes `part-0000 .. part-(nkeys-1)` via `delete_object(fp, NULL)`, best-effort (returns if `api == NULL`), mirroring the local `unlink`. `nkeys` is threaded through `PexportSpawn` and also passed on the direct error path (`single_table ? workers : npart`) — consistent with the cleanup callback.
- Two honest, bounded deferrals, disclosed in code + docs + the function's `errhint`, not silent: the empty-prefix check needs a bucket listing (#619), and a TERMINATED worker's in-flight multipart is left to the bucket lifecycle rule (exactly as #622 documented for a single object). Neither is a defect; both are named branches with issue citations.

## Scope

Real-Garage functional + the assert suite. I did not re-run the PR's pg18_san ASAN gate; that and my live-S3 proof are complementary. Nice work.

— ChronicallyJD

OffgridwithJD (approved):
Approving. Beyond the fixture suite, I verified this against a real Garage endpoint (local incus `garage-26-04`, v2.3.0) on a pg17a assert build: single-table parallel prefix export (80k rows → 4 objects, each readable by key, union == source, premise-guarded non-empty) and partitioned export (40k → 2 partition objects, union == source). `spawn.nkeys = workers|npart` matches the key set the workers wrote. Full `objstore_sink_write.sh` green including the cancel → DELETE-known-keys removal proof. The two deferrals (empty-prefix check → #619; terminated-worker in-flight multipart → bucket lifecycle) are honestly disclosed in code + docs + errhint, not defects. No blocking findings. Merge remains jd/jdatcmd's call — I merge nothing.

PR #629, opened 2026-08-14
feat: the Parquet FDW streams row groups instead of materializing a tuplestore (#620)

### LINKED ISSUE #620: Parquet FDW streams instead of materializing the whole file set into a tuplestore
Split from #393 at its close. The scan state says it plainly (`columnar_parquet_reader.c`, `PqFdwScanState`): "the whole file materialized into a tuplestore, drained by IterateForeignScan. Eager (reads the file up front); streaming is a follow-on."

Every cost lands in `BeginForeignScan` before the first row: a `LIMIT 1` over a remote multi-GB dataset pays for all of it, and ranged GETs cannot shorten a scan the executor abandons. The #393 memo measured the shape: all I/O before the first tuple, `IterateForeignScan` merely pops rows.

Streaming means per-file (or per-row-group) incremental reads driven from `Iterate`, which interacts with: ReScan (today `tuplestore_rescan`, free), the chunk-granular prefetch window (per-chunk reads stay the unit), and the remote handle lifetime (today no handle survives its statement — the module's abort-cleanup invariant depends on that and must be preserved or consciously renegotiated).

Worth doing for local files too (the memo's Part B: it pays on any transport), which is also what makes it testable without a network.

jdatcmd:
Closes #620.

## What changed

The `pgcolumnar_parquet` FDW used to open every resolved file in `BeginForeignScan`, decode it whole, and drain all rows into one `randomAccess` tuplestore before returning. `Iterate` drained the tuplestore; `ReScan` was a free rewind. The whole file set sat in `work_mem`-backed storage, and a `LIMIT` still decoded every row group of every file.

The scan now **streams one row group at a time**:
- `Begin` resolves the file list and prunes (`Files Pruned`) but decodes nothing.
- Each refill decodes the next non-skipped row group of the current file into a bounded one-group tuplestore that `Iterate` drains, opening and advancing files as the cursor reaches them.
- Memory is one row group, not the whole file, and a `LIMIT` the executor satisfies early leaves the remaining groups and files undecoded.
- `ReScan` re-streams from the top (close, reset, re-decode). The eager model's free rewind is the deliberate trade for bounded memory; no suite asserts rescan *cost*, only content.

The **correctness-critical row assembly is unchanged**: `pq_read_rows` gains an `[rgFrom, rgTo)` group range. `read_parquet` and `import_parquet` pass the full range and stay eager, byte-for-byte; the FDW drives one group per refill. `read_parquet` stays an `SFRM_Materialize` SRF — only the FDW streams.

## Remote-handle lifetime

A streamed remote scan holds one read handle across `Iterate` calls, so a handle no longer always closes *before its opener returns*. The module's abort invariant is restated: close-all-on-abort stays correct because the handle is the running statement's — `EndForeignScan` closes it normally and the release callback closes it on abort, whoever holds the pointer. The source opens in the scan's `fileCtx` so its prefetch window (`winCtx`) outlives a single `Iterate` call.

## TDD / prove-don't-trust

New arm in `native_parquet_streaming.sh` over a four-row-group file, using a new **`Row Groups Decoded`** EXPLAIN counter (there was no rows-produced or groups-decoded counter before — the old counters could not tell a `LIMIT` from a full scan):

- full scan decodes **4**; `LIMIT 10` (inside group 0) decodes **1**; `LIMIT 1010` (into group 1) decodes **2**.
- **RED on `main`**: the counter line is absent; no counter distinguishes a `LIMIT` from a full scan.
- **Removal proof (counter)**: neutering `groupsDecoded++` reds all three counter arms (report 0).
- **Removal proof (streaming)**: reverting to eager whole-file decode reds *exactly* the two `LIMIT` arms while full-scan and content stay green — proving the one-group streaming is load-bearing for early-exit.
- A forced nested-loop arm (`enable_hashjoin/mergejoin off`) covers the re-stream `ReScan`.

## Gates

- **PG17/18/19**, `-Wshadow -Werror`: the full FDW/read/remote set is green — `native_parquet_fdw`, `_pushdown` (Row Groups Skipped counters), `_projection`, `_partition` (Files Pruned + partition stamping), `_multifile` (multi-file drain + ReScan), `_streaming`, `native_read_parquet`, `_stack`, `_hardening`, `_codecs`, `_flba`, `objstore_s3_read`, plus `parquet_import` and `parquet_nested_import`.
- **ASAN (pg18_san)**: local multi-group scans, `LIMIT` early-close, nested-loop rescan, and a remote scan with `LIMIT` closing the handle mid-file — **zero backend sanitizer reports**. This is the per-scan memory-lifecycle gate: group/row contexts and a remote handle + prefetch window now span `Iterate` calls.

## Docs

`sql-reference.md`, `features.md`, and `ARCHITECTURE.md` note the streaming model, the `LIMIT` early-exit, and `Row Groups Decoded`; `read_parquet` is called out as still materializing. STE style gate passes. Design in `design/ISSUE_620_FDW_STREAMING.md`.

### DIFF
--- a/design/ISSUE_620_FDW_STREAMING.md
+++ b/design/ISSUE_620_FDW_STREAMING.md
@@ -1,0 +1,183 @@
+# Issue #620: the Parquet FDW streams instead of materializing a tuplestore
+
+Design before code. Today `pqfdwBeginForeignScan` opens every resolved file,
+decodes it whole, and drains all rows into one `randomAccess` tuplestore before
+it returns; `Iterate` just drains the tuplestore and `ReScan` is a free
+`tuplestore_rescan`. The whole file (really, every file the path resolves to)
+sits in `work_mem`-backed storage, spilling to disk past `work_mem`, and a
+`LIMIT k` still decodes every row group of every file. This issue makes the scan
+pull one row group at a time and stop when the executor stops asking.
+
+All line citations are `src/columnar_parquet_reader.c` at HEAD (`main`, post-#445).
+
+## The unit of streaming is one row group
+
+`pq_read_rows` (2941) is a push loop with two nested loops: an outer loop over
+row groups that decodes a whole group's needed leaves into `groupCtx` (the
+`decode_leaf_entries` block), and an inner loop that assembles each of the
+group's `n` rows and calls `sink(slot, arg)` per row. The natural resumable
+granularity is the row group: decode one group, hand out its `n` rows across
+successive `Iterate` calls, and reset `groupCtx` only when advancing to the next
+non-skipped group. `rowCtx` already resets per row.
+
+Memory drops from "every file" to "one row group's decoded columns". `LIMIT`
+stops after the groups whose rows it consumed. Correctness of row assembly is
+unchanged because the assembly code moves verbatim.
+
+## Shape: a resumable cursor, assembly code moved not rewritten
+
+Refactor `pq_read_rows`'s body into two reused pieces, called by both the eager
+SRF path (unchanged behaviour) and the new streaming FDW cursor:
+
+- `pq_decode_group(pf, src, leaves, ..., rg, groupCtx)` runs the existing
+  per-group decode (the `decode_leaf_entries` block) for group `rg` into
+  `groupCtx` and returns the group's row count `n` and the decoded per-leaf
+  entry state. This is lifted verbatim from the outer-loop body.
+- `pq_assemble_row(state, r, slot)` runs the existing per-row assembly (scalar,
+  list, struct, const/partition stamping, `ExecStoreVirtualTuple`) for row `r`
+  into `slot`. Lifted verbatim from the inner-loop body.
+
+`pq_read_rows` is then re-expressed as: `for each rg { n = pq_decode_group(...);
+for r in 0..n { pq_assemble_row(...); sink(...); MemoryContextReset(rowCtx); } }`
+so the SRF and import paths keep their exact current behaviour and stay eager.
+This keeps the refactor mechanical and lets the existing row-content suites prove
+the move introduced no assembly change.
+
+### The FDW cursor state (on `PqFdwScanState`)
+
+Replace `tupstore`/`readslot` with a cursor:
+
+- the resolved `files` list and a current-file index;
+- the current open `PqSource` (or a flag that none is open) and its `PqFile`
+  footer;
+- the projection/skip plan for the current file (`needLeaf`/`needTop`,
+  `skipGroup[]`), computed once per file as Begin does today;
+- the current row-group index `rg`, the decoded group's `n`, and the current row
+  index `r` within it;
+- `groupCtx` and `rowCtx` living in the scan's own long-lived context, not the
+  transient context Begin runs in (see winCtx note below);
+- the six EXPLAIN counters, plus one new counter (below).
+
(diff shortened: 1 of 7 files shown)


OffgridwithJD (commented):
Reviewed the streaming refactor and the remote-handle lifetime, and **proved the remote path against real Garage** (local incus `garage-26-04`). The one class of risk this PR introduces — a remote read handle and its prefetch window now spanning `IterateForeignScan` calls, with a `LIMIT` closing the source mid-file — is exactly what I exercised on live S3. Clean; a comment, not a block.

## The lifetime argument holds in the code

- `st->fileCtx` is created in `Begin` as a child of the scan context, and `pq_source_open_cfg` runs inside `fileCtx` (a `MemoryContextSwitchTo(st->fileCtx)` around the open), so the source's prefetch window `winCtx` (captured as `CurrentMemoryContext` at open) is scan-lifetime, not per-`Iterate`. It is reset only when the cursor advances to the next file.
- `EndForeignScan` and `ReScanForeignScan` `pq_source_close` the handle; `os_resource_release` over `os_open_handles` is the abort backstop. The module header invariant is correctly restated from "no handle survives its statement" to "the live handle is the running statement's, closed by End on the normal path and by the release callback on abort."

Note: `objstore_s3_read.sh`'s Garage integration arm silently fell back to the fixture here — its upload helper failed ("integration endpoint set but the upload helper failed"), so it did **not** actually read from Garage. I proved the remote path two other ways instead:

1. **pgcolumnar-exported object**: export a 40,000-row table to `s3://…`, FDW over it. Full remote scan returns 40,000 correct rows (handle spans `Iterate`); a `LIMIT` returns the correct rows (handle closed mid-file); a forced nested-loop join re-scans it (streaming `ReScan`) with no crash.
2. **Multi-row-group object** (the decisive one): I uploaded a 4-row-group PyArrow parquet to Garage (hand-rolled SigV4 PUT, since no boto3/aws-cli in the container) and scanned it through the streaming FDW. `Row Groups Decoded`: **full = 4, LIMIT 10 = 1, LIMIT 1010 = 2**, content correct. That reproduces `native_parquet_streaming.sh`'s exact assertions but on a **remote** object, on an assert build, with no crash — i.e. the early-exit closes the remote handle mid-file safely.

## Local suite (pg17a assert)

`native_parquet_streaming.sh` green: the `Row Groups Decoded` counter, the counter removal proof (neuter `groupsDecoded++` → all three arms report 0), and the streaming removal proof (revert to eager whole-file decode → exactly the two `LIMIT` arms red while full-scan/content stay green).

## Scope

Assert build + real Garage. I did not re-run the pg18_san ASAN gate the PR reports; that (the per-scan memory-lifecycle check) and my live-S3 assert-build proof are complementary. Solid refactor.

— ChronicallyJD

OffgridwithJD (approved):
Approving. The one class of risk — a remote read handle and its prefetch window now spanning `IterateForeignScan`, with a `LIMIT` closing the source mid-file — is sound in the code (`winCtx` in a scan-lifetime `fileCtx`; `End`/`ReScan` close, `os_resource_release` backstops abort) and empirically proven on real Garage (pg17a assert): full scan + LIMIT content + nested-loop ReScan all correct, and remote multi-group early-exit reproduces the local assertions on a remote object — `Row Groups Decoded` full=4, LIMIT 10=1, LIMIT 1010=2 — with no crash. Local `native_parquet_streaming.sh` green with both the counter and the streaming removal proofs. No blocking findings.

Issue #210, opened 2026-07-28
Crafted Parquet footer crashes the backend (SIGSEGV): unbounded Thrift recursion in ColumnarThriftSkip, no check_stack_depth


OffgridwithJD:
## A crafted Parquet footer crashes the backend (SIGSEGV) — unbounded Thrift recursion, no `check_stack_depth`

`ColumnarThriftSkip` recurses once per nested Thrift struct (and once per element of a nested list/set), and nothing bounds the depth. `grep check_stack_depth --include='*.c' --include='*.h' src/` returns **zero** — the guard PostgreSQL uses for exactly this class of recursive parser is absent from the whole tree. A footer that is just N nested-struct openers therefore drives the C stack into its guard page and `SIGSEGV`s the backend. Because the postmaster treats a backend segfault as a crash, it **terminates every other session and reinitializes the cluster** — so this is a whole-cluster DoS from one file, not a single-connection error.

### Where

`src/columnar_thrift.c`, `ColumnarThriftSkip`:

```c
case TC_STRUCT:
    for (;;) {
        ColumnarThriftField(r, &ft, &fid, &lastId);
        if (ft == TC_STOP || r->error) break;
        ColumnarThriftSkip(r, ft);      /* <-- recurses; ft can be TC_STRUCT again */
    }
```

The top-level footer loop skips every unrecognised field through this function — `src/columnar_parquet_reader.c:624` (and :256, :309, :312, :335, :619) — so the recursion is reached directly from attacker-controlled footer bytes, before any schema is even interpreted. The `TC_LIST`/`TC_SET` arm recurses the same way over a nested list of structs.

### Reproduce (assert build, PG18.4, current `main` @ 2d51bfc)

A valid Parquet file is `PAR1` + metadata + `<uint32 metalen>` + `PAR1`. The byte `0x1C` is a compact-protocol field header meaning *(delta 1, type TC_STRUCT)* — i.e. "open a struct". A footer of N of them nests N deep:

```python
import struct
meta = b'\x1c' * 4_000_000
open('evil.parquet','wb').write(b'PAR1' + meta + struct.pack('<I', len(meta)) + b'PAR1')
```

```sql
SELECT * FROM pgcolumnar.parquet_schema('/path/evil.parquet');
```

```
server closed the connection unexpectedly
```
```
LOG:  client backend (PID 694) was terminated by signal 11: Segmentation fault
[...]

PR #630, opened 2026-08-14
feat: ListObjectsV2 discovery over remote paths, with globs and Hive (#619)

### LINKED ISSUE #619: Object storage: ListObjectsV2, globs, and Hive partition discovery for remote paths

jdatcmd:
Closes #619.

## What this does

The reader used to refuse a pattern or a directory over a remote URL — a remote path was an exact key or nothing. It now **expands a remote prefix or glob** into the objects under it via a paged **ListObjectsV2** call behind the object-store ABI, and **Hive `partition_columns` work over a remote prefix**.

- `s3://bucket/key` — exact object, one GET, **no LIST** (the point-read fast path is unchanged).
- `s3://bucket/prefix/` — trailing slash lists everything under the prefix, at any depth, like a local directory.
- `s3://bucket/prefix/*.parquet` — lists the literal prefix and matches segment by segment (`fnmatch FNM_PATHNAME`), like local `glob()`.
- `_SUCCESS`, `_temporary`, and dot-hidden names are skipped, like the local walk.

## New ABI op (bump 3 → 4)

Everything a LIST needs — HTTP transport, SigV4 signing, TLS, the allow-list, credential resolution — lives inside the module, so `list_objects` is a new op appended to `PgColumnarObjStoreApi`. It returns full `s3://bucket/key` URLs, which `pq_resolve_paths` wraps into the same file list a local directory walk produces, so **the four consumer loops and the Hive `pqfdw_partition_values` need no change**. `pq_resolve_paths` gains the object-store config so the FDW's catalog credentials reach the listing (the function API passes NULL, the ambient path). **Module and main library must be built and installed together** — the loader refuses an ABI mismatch.

## The parser is the whole risk

A ListObjectsV2 response is XML from an endpoint that is allow-listed but can still be hostile. `os_xml_find` / `os_xml_decode_append` / `os_list_parse_page` wear the `columnar_thrift.c` discipline: a `{buf,len,pos}` scan whose every bound is **subtract-not-add** so a crafted length cannot wrap (the #210 class), that treats anything unrecognized as end-of-input rather than a read past the buffer, and that decodes only the five predefined entities plus single-byte numeric refs. The request is a GET through the write-style signer (canonical-query support already exists), targeting the bucket root with the prefix in the signed query; paging follows the continuation token, bounded by `OS_LIST_MAX_{BODY,KEYS,PAGES}`.

## TDD / prove-don't-trust

- `objstore_listing.sh`: prefix read (paged, page size 2), FDW over a prefix, glob, exact-key-no-LIST, remote Hive with a pruning predicate, and the non-advancing-token refusal. **RED on `main`** (a remote prefix reads nothing / a glob is refused).
- **Removal proofs**: disabling paging reds the prefix read (only the first page — two non-parquet entries — is seen); disabling the extension filter reds it (a non-parquet object folds in).
- **The fuzzer earned its keep.** `fuzz_listing.sh` feeds the parser malformed XML through the fixture's `__listing_override__` hook, deterministic per seed, property *malformed listing → ERROR, never crash/hang/san*. It found a **real DoS**: an endpoint replaying one truncated page with a non-advancing token spun the paging loop toward the cap (**19 hangs / 200**). Fixed by detecting a continuation token that repeats the one just sent and refusing (`08P01`); a targeted regression arm plus **300 clean mutants** confirm it.

## Gates

- **PG17/18/19** over the objstore and affected FDW suites under `-Wshadow -Werror`; the ABI bump keeps the read and write suites green.
- **ASAN (pg18_san)** over the valid listing exercise and a 150-mutant fuzz loop — **zero backend sanitizer reports**.

## Docs

The object-storage section in `sql-reference.md` documents exact key vs prefix vs glob, the hidden-name skip, remote Hive, and that a pattern needs the module. Design in `design/ISSUE_619_REMOTE_LISTING.md`.

### DIFF
--- a/design/ISSUE_619_REMOTE_LISTING.md
+++ b/design/ISSUE_619_REMOTE_LISTING.md
@@ -1,0 +1,165 @@
+# Issue #619: ListObjectsV2, globs, and Hive discovery over remote paths
+
+Design before code. Today `pq_resolve_paths` (src/columnar_parquet_reader.c)
+refuses to expand a pattern or a directory over a remote URL: a remote path is an
+exact key or nothing. The refusal comment names the reason precisely, "expanding
+a pattern needs a LIST call, whose paged XML response is a third hand-rolled
+parser over input an outside party controls, which is the shape that produced
+#210 and #228." This issue adds that LIST, behind the frozen object-store ABI,
+with the bounds discipline and the fuzz harness that class of parser requires.
+
+All citations from the recon at HEAD; verify against the tree before coding.
+
+## The parser is the whole risk
+
+An S3 ListObjectsV2 response is XML from the endpoint. The endpoint is
+allow-listed (#393, the operator authorized it), but an authorized endpoint can
+still be compromised or hostile, and the response is attacker-influenceable
+input. The reference for a hostile-input parser in this tree is
+`src/columnar_thrift.c`: a `{buf, len, pos, error}` reader whose bound is the
+overflow-safe `n > len - pos` (never `pos + n > len`, which wraps, the #210
+class), which sets `error` rather than over-reading on every guard, and which
+calls `check_stack_depth()` so crafted nesting is a caught ERROR, not a SIGSEGV.
+The XML listing parser adopts the same discipline. It is a targeted extractor,
+not a general XML parser: it pulls `<Key>`, `<IsTruncated>`, and
+`<NextContinuationToken>` from a known response shape, bounds every scan, and
+treats anything it does not understand as end-of-input, never as a read past the
+buffer. The existing `os_reject_403` (module) already does the crudest version of
+this (`strstr` for `<Code>`); the listing parser is the disciplined form.
+
+## A new ABI op, not work above the module
+
+Everything a LIST needs lives inside the module: the HTTP transport, SigV4
+signing, TLS, the socket wait/cancel loop, the endpoint allow-list, the
+link-local refusal, and credential resolution. The main library holds none of it
+and cannot issue a LIST from `pq_resolve_paths`. So `list_objects` is a new op on
+`PgColumnarObjStoreApi`, appended to the struct end, bumping
+`PGCOLUMNAR_OBJSTORE_ABI` from 3 to 4 (the loader refuses any mismatch, so the
+bump is load-bearing and both header and module move together). Shape, mirroring
+the existing ops:
+
+```
+/* returns malloc/palloc'd array of nkeys cstrings, each a full s3://bucket/key
+ * URL; raises on transport/parse failure; nkeys may be 0 (empty prefix). */
+char **(*list_objects)(const char *url,      /* s3://bucket/prefix */
+                       const PgColumnarObjStoreConfig *cfg,
+                       int *nkeys);
+```
+
+Returning full `s3://bucket/key` URLs is deliberate: `pq_resolve_paths`'s remote
+arm wraps them into the same `List *` of cstrings the four consumer loops
+(import, read_parquet, parquet_schema, FDW Begin) already drain, and the Hive
+`pqfdw_partition_values` root-prefix strip and `/`-tokenize logic transfers
+unchanged. No consumer loop and no Hive code changes if the keys come back as
+full URLs.
+
+Memory: the ABI is a C boundary. The op returns `palloc`'d strings in the
+caller's current context (the module already allocates handles in
+TopMemoryContext and copies out; follow whichever the other ops use for
+caller-owned returns). The count is explicit; the caller never scans for a NULL
+terminator.
+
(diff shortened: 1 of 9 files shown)


OffgridwithJD (commented):
Reviewed the parser and paging closely, ran the suites, and **proved every listing mode against real Garage** (local incus `garage-26-04`, v2.3.0). Clean — a comment, not a block, with one small hardening note.

I wrote two objects under a prefix with `export_parquet` (single-object sink; #628's parallel-to-prefix is a separate unmerged branch, so I did not lean on it) and exercised the new discovery:

| mode | result |
| --- | --- |
| prefix read `s3://…/rt630/` | 80,000 rows, union of both objects, hash == source |
| glob `…/*.parquet` | 80,000 |
| FDW over the prefix | 80,000, hash == source |
| exact key `…/a.parquet` | 40,000 — one object, no LIST expansion (fast path intact) |
| hidden-name skip | added `_SUCCESS` to the prefix; a re-read still returns 80,000 (ignored) |

## Suites (pg17a assert)

- `objstore_listing.sh` green, including the non-advancing-token refusal (not an infinite paging loop).
- `fuzz_listing.sh` green: **200 malformed-XML mutants, 149 errored, 51 accepted clean, zero crash / hang / sanitizer**. The DoS the fuzzer originally found (a replayed page with a non-advancing token) stays fixed.

## The parser holds up (the stated risk)

`os_xml_find` guards `nlen > len` and loops `i <= len - nlen`; `os_xml_decode_append` is `end`-bounded, resolves only the five predefined entities and single-byte numeric refs within a 12-byte window, and copies any unrecognized `&…` literally; the `<Key>` loop advances `pos = ke` monotonically and caps at `OS_LIST_MAX_KEYS`. Every bound is subtract-not-add — the #210 discipline. The paging loop is capped at `OS_LIST_MAX_PAGES`, and the exact-key path issues one GET with no LIST.

## One hardening note (not a blocker)

The non-advancing-token guard compares `next` only to the *immediately previous* token (`strcmp(next, token) == 0`). A hostile endpoint that **cycles** two tokens (A→B→A→B…) evades it and runs to `OS_LIST_MAX_PAGES` before erroring — bounded, but 100k requests. Given the endpoint is allow-listed (superuser-chosen) and the page cap is a hard backstop, this is low severity; a `seen`-token set (or a lower page cap) would close it. The realistic broken-endpoint case (a single repeated token) is caught.

## Scope

Real Garage + the assert suites + the parser read. I did not re-run the pg18_san ASAN fuzz loop the PR reports; that and my live-S3 + 200-mutant assert run are complementary. This closes #619 cleanly and is the read side of the prefix round trip #628 opened.

— ChronicallyJD

OffgridwithJD (approved):
Approving. Every listing mode verified against real Garage (pg17a assert): prefix read (union == source), glob, FDW-over-prefix, exact-key (one GET, no LIST expansion), and hidden-name skip (`_SUCCESS` ignored). `objstore_listing.sh` and `fuzz_listing.sh` green — 200 malformed-XML mutants, zero crash/hang/sanitizer, so the fuzzer-found DoS stays fixed. The XML parser holds the #210 subtract-not-add discipline (bounded scans, single-byte entities/numeric-refs, key/page caps). One non-blocking hardening note for a follow-up: the non-advancing-token guard compares only to the immediately-previous token, so a 2-token cycle (A→B→A→B) evades it and runs to `OS_LIST_MAX_PAGES` — bounded, low severity given the allow-listed endpoint, but a `seen`-set would close it. No blocking findings.

Issue #621, opened 2026-08-13
Object storage: GCS and Azure schemes, and S3 virtual-host addressing


jdatcmd:
Split from #393 at its close.

`PgColumnarPathIsRemote` already recognizes `gs://` and `az://` so they fail with a clear "no such URL scheme" instead of a filesystem miss, and `objstore_module.sh` uses `gs://` as its unhandled-scheme probe (that probe graduates to the next unimplemented scheme when one of these lands — the comment in the suite says so).

What implementation needs, from the #393 research:
- **Credential models differ**: no GCS or ADLS credential key is standardized in any spec the memo could find; SAS tokens, managed identity, and workload identity have different shapes from the AWS triple, so the user-mapping option namespace needs a deliberate second pass before it is frozen for these schemes (memo unknown 9).
- **GCS XML API** is S3-compatible enough for reads with one signing quirk (rclone excludes `accept-encoding`); the native JSON API is a different client.
- **S3 virtual-host addressing** (`bucket.s3.region.amazonaws.com`) was deferred with TLS since it requires it; path-style against an explicit endpoint is the only mode today. Virtual-host is what real AWS defaults to, and SNI/hostname verification for it already works from M3.

All transport, signing scaffolds, TLS, credential plumbing, and the allow-list are in place from #393; each scheme is a resolution + signing dialect on the existing module.

OffgridwithJD:
Status: two of the three scoped items here have shipped; one remains by owner decision.

**Delivered — #631 (merged 2026-08-14):**
- **GCS `gs://`** — folds onto the S3 resolver (GCS XML API is S3-shaped, SigV4 + HMAC key), default endpoint `storage.googleapis.com`, region `auto`. Verified read + write round-trip against a real S3-compatible endpoint.
- **S3 virtual-host addressing** — `pgcolumnar.objstore_s3_addressing='virtual'`, no ABI change. The allow-list authorizes the endpoint authority, not the per-bucket hostname, so virtual-host cannot widen reach. Verified against a real server, including the allow-list refusal.

**Deferred — Azure `az://`:** out of scope by the 2026-08-14 owner decision. It needs a SharedKey/SAS signer and new ABI credential fields, which is the "credential models differ / user-mapping namespace needs a deliberate second pass" unknown this issue itself flagged. `az://` stays an unhandled scheme and is now the `objstore_module.sh` unhandled-scheme probe.

[...]

PR #631, opened 2026-08-14
feat: GCS gs:// scheme and S3 virtual-host addressing (#621)

jdatcmd:
Addresses #621. **Scope (owner decision, 2026-08-14): `gs://` and S3 virtual-host addressing. Azure (`az://`) is explicitly out of scope** — it would need a SharedKey/SAS signer and new ABI credential fields, deferred; it stays an unhandled scheme. Both features here are additive and need **no ABI change**.

## GCS via the interoperable XML API

`gs://` (and `az://`) were already recognized as remote by `PgColumnarPathIsRemote` but rejected at the module's `handles_url`. GCS's XML API is S3-shaped and takes AWS SigV4 with an HMAC key, so `gs://` folds onto the S3 resolver:
- `handles_url` accepts `gs://`; `objstore_open` / `os_write_handle` route it to `os_resolve_s3` (both schemes are 5 chars, so `url+5` skips either).
- `os_resolve_s3` defaults the endpoint to `https://storage.googleapis.com` and the region to `auto` when none is configured (S3 still requires an explicit endpoint). The signer, credentials (a GCS HMAC key is the `access_key_id`/`secret_access_key` pair), and request construction are unchanged.

Verified against the SigV4 fixture (an independent signer) and, opt-in, live Garage. Real-GCS acceptance of the `s3` service scope is the one assumption not checkable here; it's the documented interop path.

## Virtual-host addressing

Path-style (`endpoint/bucket/key`) stays the default and byte-identical. `pgcolumnar.objstore_s3_addressing='virtual'` selects virtual-host (`bucket.endpoint/key`). The GUC is defined by the preloaded library and read in `os_resolve_s3` **by name** (same mechanism as `objstore_allowed_endpoints`), so **no config-struct field and no ABI bump**. Under virtual-host:
- `h->host` becomes `bucket.authority` — the connect target, Host header, and TLS cert name all key off `h->host` already, so cert verification and SNI need no change.
- the request path is the key alone.
- a new module-private `auth_host` carries the endpoint authority, and the allow-list matches it: the operator authorizes the endpoint, not the per-bucket hostname. Path-style leaves `auth_host` NULL, so the allow-list check falls back to `h->host` exactly as before (existing S3 suites stay byte-identical).

## TDD

`objstore_addressing.sh`, against the SigV4 fixture:
- `gs://` reads and writes round-trip (scheme handled, HMAC creds reused, signature verified by the independent implementation).
- `gs://` with no endpoint uses the `storage.googleapis.com` default — its error does **not** demand `AWS_ENDPOINT_URL`, the contrast that proves the default (`s3://` still demands one).
- virtual-host reads the same object as path-style; the fixture log shows the **key alone** in the path under `virtual` and the **bucket back** under `path` (the removal proof of the switch).

Virtual-host is exercised with `/etc/hosts` names (`bucket.s3.local` → 127.0.0.1), so the client's connect, Host header, and cert name are the real `bucket.endpoint` with **no test-only code path**; the suite skips cleanly where those names don't resolve. `objstore_module.sh`'s deliberate unhandled-scheme probe migrates from `gs://` to `az://`.

## Gates

- **PG17/18/19** over the objstore suites under `-Wshadow -Werror`. (The new GUC backing variable needs an `extern` in `columnar.h`, which PG18/19's `-Wmissing-variable-declarations` enforces and PG17 does not — caught in gating.)
- **ASAN (pg18_san)** over the gs:// read/write and virtual-host paths, which add handle-memory allocations (`auth_host`, the virtual-host `psprintf`/`pnstrdup`): **zero backend reports**.

## Docs

The object-storage scheme table gains `gs://`; the addressing GUC is documented in the configuration reference and the object-storage notes; the `AWS_ENDPOINT_URL` row notes the GCS default. Design in `design/ISSUE_621_SCHEMES_AND_ADDRESSING.md`.

### DIFF
--- a/design/ISSUE_621_SCHEMES_AND_ADDRESSING.md
+++ b/design/ISSUE_621_SCHEMES_AND_ADDRESSING.md
@@ -1,0 +1,87 @@
+# Issue #621: the GCS gs:// scheme and S3 virtual-host addressing
+
+Scope (owner decision, 2026-08-14): implement the `gs://` scheme and S3
+virtual-host addressing. **Azure (`az://`) is out of scope** and stays an
+unhandled scheme; it would need a SharedKey/SAS signer and new ABI credential
+fields, a decision deferred.
+
+Both features are additive to the existing S3 path and need no ABI change: GCS
+reuses the SigV4 signer with a default endpoint, and the addressing style is a
+GUC read inside the module, not a config-struct field.
+
+## GCS via the interoperable XML API
+
+`gs://` and `az://` were already recognized as remote by `PgColumnarPathIsRemote`
+but rejected at the module's `objstore_handles_url`. GCS's XML API is S3-shaped
+and accepts AWS Signature Version 4 with an HMAC key, so `gs://` folds onto the
+existing S3 resolver:
+
+- `objstore_handles_url` returns true for `gs://`.
+- `objstore_open` / `os_write_handle` route `gs://` to `os_resolve_s3` (both
+  schemes are five characters, so `url + 5` skips either).
+- `os_resolve_s3` learns the scheme. When no endpoint is configured it defaults
+  `gs://` to `https://storage.googleapis.com` (S3 has no single default, so it
+  still requires one); when no region is configured it defaults `gs://` to
+  `auto`. The signer, credentials, and request construction are unchanged: a GCS
+  HMAC key is the `access_key_id` / `secret_access_key` pair, service name `s3`.
+
+Verified against the S3 fixture (which is an independent SigV4 implementation)
+and, opt-in, live Garage. Real-GCS acceptance of the `s3` service scope is the
+one assumption not checkable here; it is the documented interop path.
+
+## Virtual-host addressing
+
+Path-style (`endpoint/bucket/key`) is the default and stays byte-identical.
+Virtual-host (`bucket.endpoint/key`) is selected by the GUC
+`pgcolumnar.objstore_s3_addressing = 'virtual'`, defined by the preloaded library
+and read inside `os_resolve_s3` by name (the same mechanism as
+`objstore_allowed_endpoints`), so no config-struct field and no ABI bump. Under
+virtual-host:
+
+- `h->host` becomes `bucket.authority` (the connect target, the Host header, and
+  the TLS cert name, all of which key off `h->host` already, so cert
+  verification and SNI need no change).
+- the request path is the key alone, not `bucket/key`.
+- `h->auth_host` carries the endpoint authority, and the endpoint allow-list
+  matches `auth_host` when set: the operator authorizes the endpoint, and
+  virtual-host must not force a per-bucket allow-list entry.
+
+The allow-list, link-local refusal, and signing are otherwise unchanged.
+Path-style leaves `auth_host` NULL, so the allow-list check falls back to
+`h->host` exactly as before, and the existing S3 suites stay byte-identical.
+
+## TDD
+
+`objstore_addressing.sh`, against the SigV4 fixture:
+
+- `gs://` reads and writes round-trip (scheme handled, HMAC creds reused, the
+  signature verified by the independent implementation).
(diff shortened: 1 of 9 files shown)


OffgridwithJD (commented):
Reviewed the resolver + allow-list changes and **proved both features against real Garage** (local incus `garage-26-04`, v2.3.0), including the SSRF-relevant allow-list behavior under virtual-host. Approve-ready, with one non-blocking test-harness nit.

I pointed the endpoint at a resolvable non-`.localhost` name (see the nit below on why) and reconfigured Garage's `root_domain` so virtual-host actually routes, then:

| test | result |
| --- | --- |
| `gs://` read (path-style) | 80,000, hash == source — scheme routes to the S3 resolver, HMAC creds + SigV4 accepted by a real server |
| `gs://` write round-trip | 80,000 |
| virtual-host read (`SET objstore_s3_addressing='virtual'`) | 80,000, hash == source — Garage extracted the bucket from `bucket.s3.gtest` |
| allow-list: endpoint authority listed | read OK |
| allow-list: **per-bucket vhost name listed, endpoint not** | **42501 refused** |
| allow-list: decoy endpoint | 42501 refused |

The middle allow-list case is the one that matters: under virtual-host the guard keys off `auth_host` (the configured endpoint authority), **not** the `bucket.endpoint` connect host, so a bucket name in the URL cannot satisfy the allow-list. That matches the code — `checkhost = auth_host ?? h->host`, and `auth_host` is derived from the endpoint config, never from the URL bucket — so virtual-host cannot be used to widen reach. Link-local refusal still runs on the resolved connect target (`h->host`), unchanged.

## Reading of the code

`gs://` folds onto `os_resolve_s3` cleanly (both schemes are 5 chars, `url+5`), defaulting the endpoint to `storage.googleapis.com` and region to `auto` only when none is configured; the signer and request construction are untouched. Virtual-host builds `h->host = bucket.authority`, the path becomes the key alone, and path-style stays byte-identical (`auth_host` NULL → allow-list falls back to `h->host`). No ABI change — the GUC is read by name like `objstore_allowed_endpoints`.

## Suites (pg17a assert)

`objstore_addressing.sh` and `objstore_module.sh` both green (gs:// read/write/default-endpoint, virtual == path with the key-alone removal proof, az:// now the unhandled-scheme probe).

## One non-blocking nit

`objstore_addressing.sh`'s virtual-host arm **hard-failed** in a fresh container — `FAIL pgc-bucket.s3.local does not resolve; add it to /etc/hosts`, suite exit 1 — rather than skipping. The PR body says it "skips cleanly where those names don't resolve." Adding `127.0.0.1 pgc-bucket.s3.local` made it pass 11/11. Either the skip should actually skip when the name is absent, or the `/etc/hosts` requirement should be a documented harness prerequisite so a matrix host without it doesn't red. Not a code defect.

## Scope

Real Garage (gs:// + virtual-host + allow-list) + the assert suites + the resolver read. I did not re-run the pg18_san ASAN arm. Solid — this is approve-ready; say the word and I'll post a formal approve.

— ChronicallyJD

OffgridwithJD (approved):
Approving. Verified both features against real Garage (local incus `garage-26-04`) on a pg17a assert build: `gs://` read + write round-trip, virtual-host read with Garage routing `bucket.s3.gtest`, and the allow-list under virtual-host keying off the endpoint authority — endpoint listed → OK, per-bucket vhost name listed but endpoint not → 42501 refused, decoy endpoint → refused. So virtual-host cannot bypass the allow-list via the bucket name, matching the code (`checkhost = auth_host ?? h->host`, `auth_host` from endpoint config only); link-local refusal unchanged. `objstore_addressing.sh` + `objstore_module.sh` green. No blocking findings. One non-blocking follow-up for you: the addressing suite's virtual-host arm hard-fails on a missing `/etc/hosts` name instead of skipping as the PR body describes — make the skip skip, or document the harness prerequisite. Merge is your call — I merge nothing.

PR #632, opened 2026-08-14
feat: _SUCCESS completion marker for parallel_export_parquet (#394)

jdatcmd:
Closes the remaining piece of #394 (the issue's step 3: a completion marker).

## What & why

`parallel_export_parquet` writes a directory or an `s3://` prefix of `part-NNNN.parquet` objects. Neither a directory nor a bucket records whether a run finished or which run produced which objects — the exact gap the issue flagged for the parallel case. This adds the **Hadoop/Spark `_SUCCESS` convention** that Spark, Hive and Trino directory readers already recognize: an empty `_SUCCESS` marker at the destination, written **only after every worker completes**.

- `pexport_write_success_marker` uses the same `PqSink` seam as the data → temp-and-rename locally, a single object remotely, no new code path.
- Written **last, past the error-cleanup region**, so a marker-write failure raises *without* deleting the parts the workers already committed (data preserved).
- A failed/cancelled run leaves **no** marker, and `pexport_remove_outputs` now also drops any stale `_SUCCESS` (so a failed re-run into a used prefix can't leave a misleading one).
- The read path already skips `_`-prefixed names, so the marker is never folded into a `read_parquet`/FDW scan.

**Scope decision:** this is the minimal, format-neutral "did the run complete" signal. A *richer manifest* (part list, row counts, schema, partition values) is the Iceberg metadata layer and is deliberately scoped to **#388** — as the issue itself noted, step 3 is "where this starts to touch #388".

## TDD / prove-don't-trust

- `parallel_export_parquet.sh` (local): a completed export writes an empty `_SUCCESS` that is **not** a part file and doesn't affect the read-back; a cancelled export leaves none; a retry writes a fresh one.
- `objstore_sink_write.sh` (remote): the marker object + its PUT appear at the prefix on success, none after cancel.
- **RED** on `main` (no marker). **Removal proof**: disabling the marker write reds the "marker written" arms while the "absent" arms stay green.

## Gates

- **PG17/18/19** under `-Wshadow -Werror`.
- **ASAN (pg18_san)** over the remote parallel export (which now writes and cleans up the marker): **zero reports**.
- **Live Garage**: an independent client (`awscli`) sees a **0-byte `_SUCCESS`** beside the four `part-000N.parquet` objects at a real Garage prefix.

## Docs

The `parallel_export_parquet` reference documents the marker and the partial-failure guarantee; the decision is recorded in `design/ISSUE_394_OBJECT_STORAGE_SINK.md`.

### DIFF
--- a/design/ISSUE_394_OBJECT_STORAGE_SINK.md
+++ b/design/ISSUE_394_OBJECT_STORAGE_SINK.md
@@ -164,3 +164,18 @@ values) is deliberately NOT this: that is the Iceberg metadata layer, which
 belongs to #388, and the issue itself noted this is "the point where this starts
 to touch #388". `_SUCCESS` is the minimal, standard, format-neutral signal; the
 manifest is the format-specific one and is scoped there.
+
+## Remote reconcile before the marker (#632 review, 2026-08-14)
+
+An adversarial review reproduced, on live Garage, a real gap: the remote path
+allows an overwrite re-run into a used prefix (#619 require-empty is deferred), so
+a smaller run overwrites the low-numbered parts and leaves the larger prior run's
+higher-numbered parts behind, then stamps `_SUCCESS` on top -- certifying a mixed
+set as one complete run. The local path is safe (prepare_dir require-empty refuses
+a non-empty directory first). Fixed by making the remote success path, now that
+#619 listing exists, delete any `part-NNNN.parquet` at the prefix whose index is
+at or above this run's key count (the stale tail) before writing the marker, so
+the marker certifies exactly this run's parts. A regression arm runs a 4-worker
+then a 2-worker export into one prefix and asserts the stale parts are gone; the
+removal proof (disable the reconcile) reds it with 4 parts left, the review's
+reproduction.
--- a/design/ISSUE_394_OBJECT_STORAGE_SINK.md
+++ b/design/ISSUE_394_OBJECT_STORAGE_SINK.md
@@ -136,3 +136,31 @@ module, shared by source and sink.
    mid-upload asserting abort ran (the arm that "leaks money and never gets
    exercised by accident").
 4. Parallel-to-remote, per-worker objects, dispatcher key-wise cleanup.
+
+## Completion marker (decided, 2026-08-14)
+
+The issue's step 3 asked whether a completion marker or manifest is wanted for
+`parallel_export_parquet`, since a directory or bucket of `part-NNNN.parquet`
+objects carries no record of whether a run finished. Decided: an empty
+`_SUCCESS` marker at the destination, the Hadoop/Spark `FileOutputCommitter`
+convention that Spark, Hive and Trino directory readers already recognize.
+
+- Written **last**, only after every worker reports success and past the
+  error-cleanup region, so a failed or cancelled run (whose parts the dispatcher
+  removes) leaves none. Its presence is the "a complete run's output is here"
+  signal; its absence means incomplete.
+- Written through the same `PqSink` seam as the data, so it is temp-and-rename
+  locally and a single object remotely, with no new code path.
+- The failure cleanup also drops any stale `_SUCCESS`, so a failed re-run into a
+  prefix that held a prior run's marker does not leave a misleading one.
+- The read path already skips `_`-prefixed names, so the marker is never folded
+  into a `read_parquet` or FDW scan.
+- If the marker write itself fails, it raises without deleting the parts (it is
+  past the cleanup region): the committed data is preserved and the caller learns
+  the signal could not be written, rather than losing a good export.
+
+A **richer manifest** (the part list with row counts, schema and partition
+values) is deliberately NOT this: that is the Iceberg metadata layer, which
+belongs to #388, and the issue itself noted this is "the point where this starts
+to touch #388". `_SUCCESS` is the minimal, standard, format-neutral signal; the
+manifest is the format-specific one and is scoped there.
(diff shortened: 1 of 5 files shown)


OffgridwithJD (commented):
Adversarially reviewed with 5 probes, each trying to break one invariant. **Four hold; one is broken (remote path), and I reproduced it on real Garage.** The break is a correctness-of-the-signal issue, not a crash/safety one — worth fixing (at least in the docs) before this lands.

## Holds (traced through the code)

- **Marker only after every worker succeeds.** `pexport_write_success_marker` (`:814`) is reached only with `failed == -1`; a worker ERROR (`:509`), a TERMINATE/crash/never-launched worker (stays PENDING, never reaches the DONE write at `:494`), or a statement cancel (CHECK_FOR_INTERRUPTS inside the ENSURE_ERROR_CLEANUP region) all route to `ereport(ERROR)` after `pexport_remove_outputs`, never to `:814`.
- **Cleanup removes the marker on both paths.** Remote branch deletes parts then `dir/_SUCCESS` before returning (outside the part loop, so `nkeys==0` still deletes it); local branch unconditionally unlinks `dir/_SUCCESS`. The `api==NULL` early-return is unreachable at cleanup time (prepare_dir hard-errors first and the module handle is cached for the backend's life).
- **0-byte marker sink is valid.** Zero `sink_write` calls → local temp-and-`durable_rename` of an empty file; remote is a single 0-byte PUT (multipart only starts on a part-boundary crossing, which never happens). Abort path doesn't double-free (`snk->f` already nulled by Finish).
- **Marker-write failure preserves data.** The marker write is past `PG_END_ENSURE_ERROR_CLEANUP` (`:771 → :814`), so a failure there raises without running `remove_outputs` — committed parts are kept and the error surfaces.

## Broken (minor, remote path only) — reproduced on real Garage

A **successful smaller remote re-run leaves a misleading `_SUCCESS`.** I built #632 and ran, against a live Garage prefix:
1. `parallel_export_parquet(big, 's3://…/rr632', 8)` → `part-0000..0007` + `_SUCCESS`.
2. `parallel_export_parquet(big, 's3://…/rr632', 4)` into the **same** prefix → returns success.

Objects after step 2:
```
part-0000..0003.parquet   (fresh, run B)
part-0004..0007.parquet   (STALE, run A)
_SUCCESS                  (freshly rewritten by run B)
```

The 4-worker run overwrote parts 0–3, left the stale 4–7, and rewrote `_SUCCESS` — certifying a mixed 8-part set as complete. A Spark/Trino/`read_parquet` reader that trusts `_SUCCESS` unions 8 parts, 4 from a different run. This directly falsifies the docs line that "a directory or prefix that carries it is a whole export."

Root cause: on the remote path `pexport_prepare_dir` cannot verify emptiness (#619 deferred), yet the marker is written unconditionally on success. The pre-existing #619 remote-empty caveat already allows silent stale-part pollution; #632 makes it worse by stamping a **trusted completeness signal** on top of it. The local path is safe (prepare_dir require-empty errors before any worker spawns).

Fix options: (a) don't write `_SUCCESS` on the remote path until #619's listing can confirm the prefix was empty; or (b) at minimum, qualify the docs — for a remote prefix the marker means "the last run finished," not "no stale objects from a prior larger run remain," until #619 lands. Either is small.

Not approving on this one — it's a real (if bounded) correctness gap in the completeness signal the PR exists to provide. Everything else is clean.

— ChronicallyJD

jdatcmd:
Good catch, and thanks for reproducing it on real Garage — that's exactly the failure mode that matters here (a *trusted completeness signal* stamped over a mixed set is worse than no signal). Fixed in bb1ec20.

Rather than the doc-only qualification, I took the behavioral fix, since #619's listing is already in this branch's base: on the remote success path the dispatcher now lists the prefix and **deletes any `part-NNNN.parquet` whose index is at or above this run's key count** (the stale tail a larger prior run left) *before* writing `_SUCCESS`. So the marker certifies exactly this run's parts again — your run-B example ends with `part-0000..0003` + `_SUCCESS`, no stale 0004..0007. The reconcile and the marker both stay past `PG_END_ENSURE_ERROR_CLEANUP`, so a failure there still preserves the committed parts.

- Regression arm (`objstore_sink_write.sh`): a 4-worker then 2-worker export into one prefix → exactly 2 parts remain (0002/0003 gone) with `_SUCCESS`.
- Removal proof (disable the reconcile) → reds with 4 parts left: your reproduction.
- ASAN over the remote export **with a shrink re-run** exercising the reconcile-delete branch: zero reports. PG17/18/19 green.

The local path is unchanged (prepare_dir require-empty already refuses a non-empty directory before any worker spawns). Docs and the design doc now describe the remote reconcile. Ready for another look.

OffgridwithJD (approved):
Re-reviewed the fix (bb1ec2034) and **verified it on real Garage** — it closes the finding I reproduced. Approving.

The reconcile is correct and well-scoped: `pexport_remove_stale_remote_parts` is called only under `if (PgColumnarPathIsRemote(dir))` (local exports untouched), lists the prefix via #619's `list_objects`, and deletes any `part-NNNN.parquet` whose index `>= nkeys` (the stale tail) before the marker. `pexport_part_index` strictly parses `part-<digits>.parquet` (so `_SUCCESS` and non-parts are never touched), `nkeys` is `single_table ? workers : npart` (matches the marker's key set), and both the reconcile and the marker sit past `PG_END_ENSURE_ERROR_CLEANUP`, so a failure there still preserves the committed parts.

Verified: I rebuilt at bb1ec20 and re-ran the exact reproduction against a live Garage prefix — 8-worker export, then a 4-worker re-run into the same prefix. Before the fix it ended with `part-0000..0007` + `_SUCCESS` (4 stale). Now it ends with:
```
part-0000..0003.parquet   (this run only)
_SUCCESS
```
The stale `part-0004..0007` are gone — the marker certifies exactly this run's parts. The other four invariants from my first pass (marker only after all succeed, cleanup deletes the marker on both paths, valid 0-byte sink, marker-write failure preserves data) still hold. Nice fix, and thanks for taking the behavioral route rather than the doc qualification.

— ChronicallyJD

PR #633, opened 2026-08-14
feat: Avro manifest decoder — Iceberg read support step 1 (#388)

jdatcmd:
Implements step 1 of the #388 Iceberg plan (from the issue review): the standalone **Avro manifest decoder**, "the only part of this we have never done." It also carries the plan validation that de-risked it — real fixtures + oracle + spec.

## The decoder (`src/columnar_avro.c`)

An Avro object-container reader for the subset Iceberg manifests use:
- **Framing**: magic `Obj\x01`, the metadata map (`avro.schema` JSON + `avro.codec`), the 16-byte sync marker, and data blocks (count, byte size, bytes, sync).
- **Binary encoding**: zigzag varints, length-prefixed strings/bytes, records, unions, arrays/maps (counted blocks), enum, fixed.
- **Codecs**: `null` and `deflate` (zlib raw inflate) — `deflate` is what pyiceberg actually emits.
- **Schema-driven**: parses the embedded `avro.schema` (via `jsonb`) and projects the `manifest_entry` fields **by name**, so a v3 manifest — which adds fields — reads structurally rather than as garbage. This was the plan's central call, confirmed necessary by the real file's `field-id`-bearing schema.
- **Bounds discipline** (`columnar_thrift.c` house style): overflow-safe `n > len - pos`, `check_stack_depth()` in the recursive descent, and caps on block/object/file sizes.

`pgcolumnar.read_avro_manifest(path)` is the introspection harness and first Iceberg surface — a materialize SRF over a local file, gated on `pg_read_server_files`.

## Prove-don't-trust

- **Differential oracle** (`avro_manifest.sh`): decodes the committed **real pyiceberg v2 manifest** and asserts entries against `expected.json` (the independent writer is the oracle) — entry count, and per entry `record_count`, `file_size`, the partition region (rendered from the schema's own field name), the `.parquet` basename; plus the privilege gate and a bad-magic rejection that doesn't crash the backend.
- **Fuzzer** (`fuzz_avro.sh` + `avro_corpus.py`): malformed manifests → ERROR, never crash/hang/san. 300 mutants on PG17, 150 under ASAN — clean.
- A **heap-use-after-free** ASAN caught on the first run (tuplestore built in the per-call context) is fixed.

## Gates

- **PG17/18/19** under `-Wshadow -Werror` (`avro_manifest.sh` + `fuzz_avro.sh`).
- **ASAN (pg18_san)** over the valid decode and a 150-mutant fuzz batch: **zero reports**.

## Plan validation (also in this PR)

`design/ISSUE_388_STEP1_AVRO_DECODER.md` + the committed real fixtures/generator. Two corrections the real data forced: **deflate is required, not optional** (pyiceberg's default), and **fixtures must be committed** (no `pip` on the test hosts — generated on the host with pyiceberg). Verdict: the plan is sound and its phase order holds.

**Scope (deliberately out, per the plan):** manifest-*list* decode (fixture committed, same machinery, next), a v3 fixture, `zstandard`, and — the plan is explicit — no catalog, no snapshot resolution, and **no deletes** (a reader that silently drops deletes looks finished and is wrong).


OffgridwithJD (commented):
Adversarially reviewed the validation and the spec, and **independently decoded the committed Avro** to check every factual claim against the bytes. Verdict: **sound — the validation is real — with spec notes for the decoder PR and one concrete oracle fix.**

## Independently verified against the committed `manifest-0.avro`

I wrote a from-scratch Avro object-container decoder (Python stdlib `zlib` only — no `pyiceberg`, no `fastavro`) and decoded the fixture. Every claim holds:

- Magic `Obj\x01` ✓, 16-byte sync marker consistent across both data blocks ✓.
- **Codec is `deflate`** — proven, not asserted: `zlib.decompress(block, -15)` (raw inflate) works, a zlib-header inflate fails. The "deflate required, not optional" correction is real.
- Embedded `avro.schema` is the fully-inlined `manifest_entry` record with **`field-id` on every field (NONE missing)** — the spec's central premise (decode against the embedded schema) is confirmed sound and necessary.
- 2 records decode, and their contents **exactly match `expected.json`**: `(eu, record_count 2, 1265 B, 00000-0-…parquet, status 1, content 0)` and `(us, 3, 1275 B, 00000-1-…parquet, …)`. So the oracle is a genuine independent oracle (my decoder shares no code with pyiceberg), not a restatement of generator inputs — `file_size_in_bytes`, the UUID basenames, `status`, `content` are emergent values only present by decoding.

## One concrete oracle fix (minor)

`expected.json` stores `file_format: "FileFormat.PARQUET"` — that is Python's `str(enum)` from pyiceberg, not the Avro value. Both my decoder and a second independent decode read the enum symbol **`"PARQUET"`**. A correct pgcolumnar decoder will produce `"PARQUET"` and *fail* this oracle field unless it wrongly mimics pyiceberg's Python stringification. Normalize the oracle to `"PARQUET"` (or the enum int) before the decoder PR tests against it.

## Spec gaps to tighten before the decoder lands (not blockers for a validation PR)

1. **Block-loop termination is unspecified.** The fixture is 2 data blocks × 1 object, terminated by EOF (`pos==len`), not by a zero count — yet the very next sentence describes arrays/maps as "terminated by a zero count." Mandate: read data blocks while `pos < len`; a partial trailing block is an error. (The oracle's count==2 assertion happens to catch the stop-after-block-0 failure.)
2. **The OCF metadata map must be fully drained.** The real header carries 7 keys (schema, partition-spec, partition-spec-id, format-version, content, avro.codec, avro.schema); the spec names only two. A decoder that stops once it has `avro.schema`/`avro.codec` desyncs before the sync marker. State: the header is `map<bytes>`, iterate to the terminating zero-count block regardless of which keys are projected.
3. **Varint shift-overflow guard.** The spec requires the `n > len - pos` bound and `check_stack_depth`, but not the varint `shift > 63` / 10-byte cap — the classic #210-class hang on hostile bytes. `columnar_thrift.c` already caps it (lines ~41-44); name it as required, not house style.
4. **Iceberg metric maps are wire-arrays with `logicalType:map`, not Avro maps** (column_sizes, *_bounds, etc., inside null-unions). Safe for step 1 (all projected fields precede them in schema order), but the schema-tree parser must key on the base `type` and not assume Avro-map framing. Matters when deferred metric projection lands.
5. **zlib reuse hazard:** Avro deflate needs `windowBits=-15`; the existing Parquet GZIP wrapper uses 15+16. Don't reuse the gzip-configured wrapper for Avro.

Provenance/independence: the generator both writes and reads with pyiceberg, so `expected.json` is independent of pgcolumnar (the spec's exact claim) but not of pyiceberg itself — my third-party decode closes that gap. Committing the fixtures (no `pip` on test hosts) is legitimate. Nice validation-first approach.

— ChronicallyJD

OffgridwithJD (commented):
Re-reviewed now that the decoder itself landed (`src/columnar_avro.c` + `read_avro_manifest`). **Memory-safety is genuinely solid — but a crafted manifest OOMs the backend, which I reproduced on the bench. Not approving on that; fixes are small.**

## What's solid (verified, pg17a assert)

- `avro_manifest.sh` and `fuzz_avro.sh` green — 300 mutants, 295 errored, 5 clean, **no crash/hang/sanitizer**; privilege gate (42501), non-Avro rejected + backend survives. My independent stdlib decoder, `read_avro_manifest`, and `expected.json` all agree.
- The #210 discipline is faithfully applied and I could not break it: `av_take` gates every sized read with the overflow-safe `n > len - pos`; the varint has a `shift > 63` cap; `check_stack_depth` on recursion; named-type refs rejected; the codec is validated (unknown refused, not silently treated as `null`); the OCF block loop terminates on `pos < len` and rejects a partial trailing block or a bad sync. No OOB read exists.

## BLOCKING — deflate zip-bomb OOM (reproduced)

The data-block loop (`columnar_avro.c:784-844`) inflates each block with `av_inflate` (`:803`, up to `AV_MAX_BLOCK` = 256 MB) but **never `pfree`s the buffer nor resets a context across iterations**. `count == 0` blocks are valid (`:793` only rejects `count < 0` / `> AV_MAX_OBJECTS`) and still inflate a large buffer while decoding zero entries — so they bypass the `n + count` entry cap entirely. The only bound on block *count* is `AV_MAX_FILE` (256 MB *compressed*), so retained decompressed memory ≈ deflate-ratio × file-size, up to ~256 GB from a <256 MB file.

Reproduced: a **504 KB** manifest (10 `count=0` deflate blocks, each inflating to 50 MB of zeros) drove the `read_avro_manifest` backend to **~398 MB peak RSS** — ~800× amplification. Scaled to a 256 MB file it OOMs the server. ASAN doesn't flag it (a legitimate-but-unbounded allocation, not OOB), and the fuzzer mutates the 2-block fixture so it never constructs the many-block bomb.

Fix: `pfree(d)` at the end of each block iteration (or decode each block in a per-block `MemoryContext` that's reset), plus a cumulative decompressed-bytes cap alongside `AV_MAX_BLOCK`.

## Minor (code-confirmed)

- **Uninterruptible CPU-spin.** `av_skip`'s `AV_ARRAY`/`AV_MAP` loops (`:379-421`) have no `CHECK_FOR_INTERRUPTS` and cap `cnt` only per-block, so a skipped `array<null>` field with repeated max-count blocks spins ~10^13-10^15 no-op iterations, uncancellable (`statement_timeout` can't fire). Reached via any unconsumed record field → `av_skip`. Add `CHECK_FOR_INTERRUPTS` to the block loops (and the top-level entry loop `:827`).
- **UBSAN.** `cnt = -cnt` at `:387/:409/:635` is signed-overflow UB when `cnt == INT64_MIN` (reachable from a crafted varint). Behaviorally benign but trips UBSAN on the pg18_san decode-path gate; guard with a magnitude check before negating.
- **Latent oracle.** `expected.json` still stores `file_format: "FileFormat.PARQUET"` (a pyiceberg Python-repr); the decoder correctly emits `"PARQUET"` and the suite doesn't compare that field, so it's dormant — worth normalizing so a future comparison doesn't break.

Not approving — the OOM is a real resource-exhaustion DoS on exactly the untrusted input this parser is meant to withstand, and it's the class ASAN/fuzzing here didn't reach. The parse correctness, bounds discipline, privilege gate, and fuzzer are strong; the three fixes are small and localized.

— ChronicallyJD

jdatcmd:
Thanks — that zip-bomb is a great catch, and exactly the class the ASAN/fuzz here couldn't reach (a legitimate-but-unbounded allocation, and the fuzzer only mutates the 2-block fixture). All three fixed in 3e82660.

**OOM (blocking).** Each block's decompressed buffer is now `pfree`'d at the end of its iteration, so peak retained memory is one block, and a cumulative decompressed-bytes cap (`AV_MAX_TOTAL` = 1 GB) bounds the total inflation across blocks — so a `count==0` chain can no longer bypass the entry cap. New regression arm crafts a minimal valid OCF whose deflate blocks inflate past the cap and asserts it's **refused (54000), not OOMed**, backend surviving.

**CPU-spin.** `CHECK_FOR_INTERRUPTS` added to `av_skip`'s array/map loops, the metadata-map loop, and the per-entry decode loop.

**UBSAN `-cnt`.** Each `cnt = -cnt` site now rejects `cnt < -AV_MAX_OBJECTS` before negating (covers `INT64_MIN`). Re-ran pg18_san (ASAN+UBSAN) over the decode and a 150-mutant fuzz batch — **zero reports**.

**Oracle.** `expected.json` + the generator now emit `"PARQUET"`, and the suite compares `file_format`.

PG17/18/19 green. Ready for another look — and thanks again for decoding the committed bytes independently, that's the review I want on a hand-rolled parser.

OffgridwithJD (approved):
Re-verified the fix (3e82660) on the bench — all three findings are closed, and I reproduced the OOM fix rather than taking it on trust. Approving.

**OOM (was blocking) — fixed and verified.** The block loop now `pfree`s each block's decompressed buffer at the end of its iteration, and `AV_MAX_TOTAL` (1 GB) caps cumulative inflation. My earlier 504 KB / 500 MB bomb, which drove peak RSS to ~398 MB before, now peaks at **98 MB** (~one block). A larger **1.25 GB** bomb I crafted is **refused with `54000`** and the backend survives — the cumulative cap holds, and a `count==0` chain can no longer bypass it.

**CPU-spin — fixed.** `CHECK_FOR_INTERRUPTS` is now in `av_skip`'s array/map loops, the metadata-map loop, and the per-entry decode loop.

**UBSAN `-cnt` — fixed.** Each negate site now rejects `cnt < -AV_MAX_OBJECTS` first (covers `INT64_MIN`); your pg18_san (ASAN+UBSAN) re-run over the decode + a 150-mutant batch is clean.

**Oracle — fixed.** `expected.json` now emits `"PARQUET"` and `avro_manifest.sh` compares `file_format` (I confirmed the suite passes with the new arm).

Memory safety was already solid (the #210 discipline — overflow-safe `n > len - pos`, `shift > 63` varint cap, `check_stack_depth`, validated codec, correct block-loop termination — held under every probe, and three independent readers agree on the fixture). With the resource bounds now in place, this is a well-defended hand-rolled parser. Nice work, and thanks for taking the behavioral fixes.

— ChronicallyJD

PR #635, opened 2026-08-14
feat: decode the Iceberg snapshot manifest-list (#388 step 2)

jdatcmd:
## What

Step two of Iceberg read support (#388). Decodes a snapshot's **manifest-list** Avro file (the `manifest_file` record) and surfaces its entries through a new SRF:

```sql
SELECT manifest_path, added_files_count, added_rows_count
  FROM pgcolumnar.read_manifest_list('/warehouse/db/events/metadata/snap-123.avro');
```

Each row names one manifest the snapshot points at, with its length, content type (0 data / 1 deletes), partition spec, and the added/existing/deleted file and row counts the writer recorded (13 columns).

## How it fits step 1

Rather than duplicate the Avro object-container reader, the block loop, deflate handling, schema-driven projection, and all the bounds / zip-bomb caps (`AV_MAX_TOTAL`, per-block free, `CHECK_FOR_INTERRUPTS`) are refactored into a shared `av_read_ocf(buf, len, elemsize, decode_cb, nout)`. `read_avro_manifest` and `read_manifest_list` are now two thin callers over it, each supplying a per-record decode callback. No decode-path behavior changed for step 1 (its suite is unchanged and still green).

Decoding stays schema-driven — `manifest_file` fields are matched by name against the embedded `avro.schema` — so a v2 or v3 manifest list reads structurally.

Scope, unchanged from the #388 plan: local file, `pg_read_server_files` required, read-only. It does **not** yet resolve a table's current snapshot or open the manifests it names — those are later steps.

## Testing (prove, don't trust)

- **Independent oracle.** The `avro_manifest` suite gains a manifest-list arm asserting the decoded `manifest_file` entry (`len|content|spec|files|rows|seq|snap`) against `expected_list.json`, extracted by pyiceberg from the same real v2 warehouse as step 1. An independent writer is the oracle, not our own encoder.
- **Removal proof.** Mutating the `added_files_count` field mapping turns the oracle arm red (`got 102, want 2`) — the arm exercises the decoder, it is not vacuous.
- **Fuzzing.** `fuzz_avro` now alternates mutants between the manifest and the manifest-list by seed parity, so `av_decode_manifest_file` is fuzzed too. 300 mutants, no crash / hang / sanitizer finding; both valid-file controls reach their decoders.
- **Sanitizer gate.** `avro_manifest` + `fuzz_avro` clean under the ASAN+UBSAN build (`run_san.sh`).
- **Matrix.** Green on PG17 / PG18 / PG19 assert builds.

## Docs

`pgcolumnar.read_manifest_list` documented in `docs/sql-reference.md`.

### DIFF
--- a/docs/sql-reference.md
+++ b/docs/sql-reference.md
@@ -598,6 +598,24 @@ SELECT file_path, record_count, partition
   FROM pgcolumnar.read_avro_manifest('/data/warehouse/db/events/metadata/abc.avro');
 ```
 
+### pgcolumnar.read_manifest_list(path text) returns table(manifest_path text, manifest_length bigint, content int, partition_spec_id int, added_files_count int, existing_files_count int, deleted_files_count int, added_rows_count bigint, existing_rows_count bigint, deleted_rows_count bigint, sequence_number bigint, min_sequence_number bigint, added_snapshot_id bigint)
+
+Decodes an Apache Iceberg snapshot manifest-list Avro file and reports its
+`manifest_file` entries. Each row names one manifest the snapshot points at,
+with its length, content type (0 data, 1 deletes), partition spec, and the
+added, existing, and deleted file and row counts the writer recorded. The
+caller needs the `pg_read_server_files` role, which superusers hold. This is
+step two of Iceberg support (#388): it reads the same Avro object-container
+format as `read_avro_manifest`, decoded against the schema embedded in the
+file, so a v3 manifest list reads structurally. It reads a local file. It does
+not resolve a table's current snapshot or open the manifests it names, which
+are later steps.
+
+```sql
+SELECT manifest_path, added_files_count, added_rows_count
+  FROM pgcolumnar.read_manifest_list('/data/warehouse/db/events/metadata/snap-123.avro');
+```
+
 ### The pgcolumnar_parquet foreign-data wrapper
 
 Exposes a Parquet file, directory, or glob as a foreign table. The scan streams
--- a/pgcolumnar--1.0-alpha.sql
+++ b/pgcolumnar--1.0-alpha.sql
@@ -909,6 +909,19 @@ CREATE FUNCTION pgcolumnar.read_avro_manifest(path text)
 COMMENT ON FUNCTION pgcolumnar.read_avro_manifest(text)
 	IS 'decode an Apache Iceberg Avro manifest file and report its data-file entries; the first step of Iceberg read support (#388)';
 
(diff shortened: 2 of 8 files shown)


OffgridwithJD (approved):
Reviewed the OCF refactor + the new manifest-list decode, and verified the parts that matter empirically. Approving.

**The refactor preserves the zip-bomb/DoS fixes — for BOTH callers (verified, not trusted).** The block loop, deflate handling, and all three caps now live once in the shared `av_read_ocf`: per-block `pfree` (`:938`), the `AV_MAX_TOTAL` cumulative cap with pfree-before-ereport (`:893-901`), and `CHECK_FOR_INTERRUPTS` per block (`:874`) and per entry (`:929`). `read_manifest_list` is a two-line wrapper over it, so it inherits them. I re-ran my 1.25 GB deflate zip-bomb from the step-1 review against the **new** SRF: `read_manifest_list` refuses it with `54000` (as does `read_avro_manifest`), backend surviving. No OOM regression on the new path.

**Generalization is type/size-safe.** Each entry point pairs a matching `elemsize` + decode callback (`Entry` ↔ `av_entry_cb`, `File` ↔ `av_manifest_file_cb`); each callback writes only named struct fields within `sizeof(struct)`, and `out + (Size)n*elemsize` lands exactly on element `n`. `nalloc` is capped at `AV_MAX_OBJECTS` before growth, and an oversized `elemsize*nalloc` raises a clean palloc `ERROR`, not an under-allocation.

**`av_decode_manifest_file` is memory-safe** — structurally identical to and simpler than the already-audited `av_decode_entry`: same `AV_RECORD` guard, every field through the overflow-safe `av_take` (`n > len - pos`) and the union-index-bounded unwrap, no nested-record descent. The unhandled `partitions` field falls to `av_skip`, which is unchanged and keeps the interrupt guard.

**Oracle is genuinely independent.** I decoded `manifest-list.avro` with my own stdlib decoder (no pyiceberg): `manifest_length=4852`, `content=0`, `added_files_count=2`, `added_rows_count=5` — exactly `expected_list.json`, which the generator builds from pyiceberg's `snap.manifests(io)` and commits static. Not circular. `avro_manifest.sh` (new list arm) and `fuzz_avro.sh` (300 mutants alternating both record types) are green.

Minor, non-blocking: the manifest-list arm doesn't compare `min_sequence_number` / `existing_rows_count` / `deleted_rows_count` and only pattern-checks `manifest_path` — worth tightening to full field coverage, but the 19-digit snapshot id + counts already make it non-vacuous. Clean step 2.

— ChronicallyJD

PR #636, opened 2026-08-14
feat: resolve the Iceberg metadata pointer to its current snapshot (#388 phase 3a)

jdatcmd:
## What

Starts **phase 3** of the #388 plan (filesystem catalog, current snapshot). New SRF:

```sql
SELECT snapshot_id, operation, manifest_list
  FROM pgcolumnar.iceberg_current_snapshot('/warehouse/db/events/metadata/v3.metadata.json');
```

Reads a table `metadata.json` and returns the current snapshot the table declares — the one whose id equals `current-snapshot-id` — with `sequence_number`, `timestamp_ms`, `operation`, `manifest_list`, and `schema_id`. One row, or **zero rows** when the table has no current snapshot. The `manifest_list` it names is then decodable by `read_manifest_list` (#635).

## Scope / why this shape

This is the plan's first phase-3 atom: *resolving the metadata pointer without a network*. Filesystem only, no Avro, no path rebasing. Parsed through the server's jsonb reader — the same approach `columnar_avro.c` already uses for the schema embedded in a manifest — so there is **no new JSON parser**.

New `src/columnar_iceberg.c` begins the catalog component, separate from the Avro decoder. Same `pg_read_server_files` gate and per-query tuplestore as the manifest readers.

Held out (documented in `design/ISSUE_388_PHASE3_CATALOG.md`):
- **3b** — chain to the live data-file list (manifest-list → manifests), which is where absolute-path **rebasing** for a relocated table and **loud delete refusal** get settled and tested.
- **3c** — scan those files, where #613's deferred field-id *resolution* finally gets its live caller (the design doc carries the full field-id/`build_imp_targets` map so 3c is ready to write).

## Error surface (distinct SQLSTATEs)

| input | result |
|---|---|
| no current snapshot (absent / null / −1) | **zero rows** (legal) |
| malformed JSON | `22P02` (jsonb_in) |
| valid JSON, not table metadata (no `format-version`) | `22023` |
| `current-snapshot-id` names no snapshot in `snapshots[]` | `XX001` data_corrupted |
| caller lacks `pg_read_server_files` | `42501` |

## Testing (prove, don't trust) — `test/iceberg_catalog.sh`, 11 arms

- **Independent oracle.** The current snapshot is asserted against `expected_meta.json`, taken from **pyiceberg's own `current_snapshot()`** view — an independent writer, not our reader. Cross-consistent with the merged `expected_list.json` (`added_snapshot_id` matches).
- **Resolution proof.** A crafted two-snapshot metadata whose `current-snapshot-id` names the *second* snapshot, with a decoy at `snapshots[0]`: a reader returning the first would return the decoy. The real reader returns the named one.
- **Removal proof.** Defeating the id match (pick `snapshots[0]`) reds *exactly* those two arms while the single-snapshot oracle stays green.
- Plus zero-rows, dangling-id (`XX001`), non-Iceberg (`22023`), malformed-JSON (`22P02`), backend-survives, and privilege (`42501`) arms.
- **Matrix.** PG17 / PG18 / PG19 (11/11) and clean under the **ASAN+UBSAN** sanitizer gate.

## Fixtures

`table.metadata.json` + `expected_meta.json` are produced by the same `gen_iceberg_fixture.py` as steps 1-2, from the same real pyiceberg v2 warehouse. The committed metadata.json has its absolute warehouse paths rewritten to a neutral `/warehouse` root, so the fixture carries no machine-specific path (basenames, and the oracle, are unaffected; real relocation is 3b's concern).

## Docs

`iceberg_current_snapshot` in `docs/sql-reference.md`; phase-3 plan + 3b/3c decomposition in `design/ISSUE_388_PHASE3_CATALOG.md`.


OffgridwithJD (approved):
Reviewed the metadata-pointer resolution and verified it empirically (suite + independent oracle + my own adversarial inputs). Approving.

**No new parser, no memory-safety surface.** The metadata.json is read (capped at `ICE_MAX_METADATA` = 64 MB) and parsed through `jsonb_in` (clean `22P02` on malformed); every field is pulled via the safe jsonb API (`getKeyJsonValueFromContainer` / `getIthJsonbValueFromContainer`), and numbers go through `ice_num_int64` (type-checked `jbvNumeric` + `numeric_int8`). I could not construct an out-of-bounds access — the untrusted-parse risk is delegated to Postgres's jsonb.

**Resolution is correct and proven non-vacuous.** It scans `snapshots[]`, matches `snapshot-id == current-snapshot-id`, takes the first match and breaks; a dangling id raises `XX001`. The suite's decoy arm (current-snapshot-id names the *second* snapshot, decoy at `snapshots[0]`) plus the removal proof (pick `snapshots[0]` → those two arms red) confirm it resolves the named snapshot, not position 0.

**Independent oracle.** My own `json` parse of `table.metadata.json` resolves to snapshot `1015321693306905452` / seq 1 / `append` / `snap-…avro` — exactly `expected_meta.json` (pyiceberg's `current_snapshot()`). That snapshot id also equals #635's manifest-list `added_snapshot_id`, so the fixtures are cross-consistent across the #388 steps.

**Adversarial inputs (mine):** an oversized `current-snapshot-id` (10^30) → clean `22003`, backend survives; a 4.9 MB / 200k-snapshot metadata resolves in **137 ms** — the loop is bounded (by the 64 MB cap) and does not hang. `iceberg_catalog.sh` (11/11) green.

Minor, non-blocking notes:
- The `snapshots[]` loop has no `CHECK_FOR_INTERRUPTS`. Bounded to well under a second by the 64 MB input cap, so not a real DoS — but a `CHECK_FOR_INTERRUPTS` in the loop is cheap hygiene for a scan over untrusted input.
- An oversized `snapshot-id`/`current-snapshot-id` surfaces as `22003` (numeric out of range), which isn't in the documented error table. It's a clean error either way; worth a row in the table.
- `schema-id` is cast to `int32` — a value past INT32_MAX truncates, but it's an informational field.

Clean phase-3a atom, and I like that the fixtures chain consistently from step 1 through here.

— ChronicallyJD

jdatcmd:
Thanks for the adversarial pass. On the three non-blocking notes:

- **CHECK_FOR_INTERRUPTS in the snapshot scan** and the 22003 doc row are folded into the stacked #637 (commit 0218a64), where this loop is refactored into the shared `ice_current_snapshot()` that both entry points use — so the interrupt check covers 3a's scan and 3b's manifest-entry loop in one place, and lands on main when #637 merges on top. Kept 3a itself as-approved rather than re-spinning an approved atom.
- **schema-id int32 truncation**: left as-is — it's the informational field you flagged; `iceberg_current_snapshot` reports the id, it doesn't key off it.

3b (#637) is up: metadata → manifest-list → manifests → live data files, with the absolute-path rebasing and the loud delete refusal you'd expect from the phase-3 plan.

PR #637, opened 2026-08-14
feat: resolve Iceberg current-snapshot data files, refusing deletes (#388 phase 3b)

jdatcmd:
**Stacked on #636 (phase 3a).** Base is `feat/388-iceberg-catalog`; review that first. The 3b-only diff is what follows.

## What

Chains phase 3a into the two merged Avro decoders. New SRF:

```sql
SELECT file_path, record_count, partition
  FROM pgcolumnar.iceberg_data_files('/warehouse/db/events/metadata/v3.metadata.json');
```

Resolves the current snapshot → reads its manifest list → each manifest → returns one row per **live** data-file entry (path, format, record_count, partition). The 3a resolver is refactored into a shared `ice_current_snapshot()` used by both entry points.

## Two designs settled here

**1. Absolute-path rebasing, keyed off `location`.** metadata.json, the manifest list, and the manifests all record absolute paths. The recorded root is metadata.json's `location`; the actual root is derived from where the metadata.json sits (Iceberg puts it at `<location>/metadata/<file>`). Each recorded path is re-rooted from `location` onto the actual root, so a **relocated table still reads**. A recorded path *not under* `location` is refused (`22023`), never opened — a tampered table cannot make the backend read an arbitrary server file (e.g. `/etc/passwd`).

**2. Loud delete refusal (`0A000`).** A snapshot carrying any delete manifest (`manifest_file.content != 0`, caught at the manifest-list level without opening it) or any delete/removed entry (`data_file.content != 0` or `status = DELETED`) is refused, not silently dropped. A reader that ignored deletes would return rows the table says are gone — the plan's worst failure mode.

## Testing — `test/iceberg_data_files.sh`, 9 arms

- **Independent oracle.** Data files asserted against pyiceberg's own `snap.manifests(io)` view (`expected_files.json`).
- **Rebasing exercised for real.** The committed warehouse is generated at a fixed recorded root (`/tmp/pgc_ice_wh`) and **copied elsewhere** at test time, so recorded ≠ actual. Arms assert returned paths are rebased under the actual location and **never leak the recorded root**.
- **Removal proofs.** Defeating the rebase empties the oracle arm; defeating the delete gate turns the refusal into a file-not-found (`58P01`) instead of `0A000`.
- **Crafted deny input.** pyiceberg 0.11.1 *cannot* write merge-on-read deletes (warns, falls back to copy-on-write), so the delete arm crafts a `content=1` manifest-list OCF — the same crafted-deny technique as the avro suite's bomb/bad-magic arms.
- Plus outside-location (`22023`), no-current-snapshot (zero rows), privilege (`42501`).
- **Matrix.** PG17 / PG18 / PG19 (9/9) and clean under the **ASAN+UBSAN** sanitizer gate.

## Fixtures

Only the `metadata/` subtree is committed (the resolver lists data files from manifests; it does not open them), so **no Parquet data is checked in**. `gen_iceberg_warehouse.py` regenerates it.

## Next (design/ISSUE_388_PHASE3_CATALOG.md)

**3c** — scan those files through the Parquet reader, projecting columns by **field id** (#613's deferred using-half). The design doc carries the full `build_imp_targets`/field-id map so it's ready to write.


OffgridwithJD (commented):
Reviewed the resolver and its two security designs, and **reproduced a blocking path-traversal that defeats the arbitrary-file-read defense** this PR is built around. Not approving on that; the rest is solid.

## BLOCKING — `ice_rebase` is bypassable by `..` traversal (reproduced)

The "a tampered table cannot make the backend read an arbitrary server file (e.g. `/etc/passwd`)" guarantee does not hold. `ice_rebase` (`columnar_iceberg.c:314-328`) enforces only a **byte-prefix match** — `strncmp(p, recorded_root, rlen)` — with no canonicalization, no `..` rejection, and no trailing-slash boundary; the result is string-concatenated and opened uncanonicalized by `ice_slurp_bin` → `AllocateFile` (used for the manifest-list at `:463-465` and each manifest at `:488-490`). `recorded_root` is the untrusted `location` from the same metadata.json.

A recorded path `<location>/../../../../etc/passwd` **starts with** `location`, so it passes the prefix check, and the `..` escapes when re-rooted. Reproduced on the bench: I pointed a snapshot's `manifest-list` at `<location>/../../../../../../../etc/hostname` and ran `iceberg_data_files`:

```
real-file target (/etc/hostname):  ERROR: columnar: not an Avro object-container file (bad magic)
nonexistent target:                ERROR: could not open file "…/nonexistent-xyz" ... No such file
```

The difference is the proof: had the boundary held, both would be `22023` ("not under the table location"). Instead the backend **opened and read `/etc/hostname`** (past open, read the bytes, failed the Avro magic). It does not leak the content in 3b — the Avro parse rejects it — but:
1. It violates the stated defense: the backend reads an arbitrary server file.
2. The three failure modes (`could not open` / `file too large` at the 64 MB cap / `bad magic`) form an **existence + size + readability oracle** for any absolute path, driven by an attacker's tampered table against a legitimate `pg_read_server_files` caller.
3. The **same `ice_rebase` feeds 3c**, where the data files *are* opened and their rows returned — there a traversal is a direct content leak. Fixing the boundary now, before 3c builds on it, is the right time.

Also a **sibling-prefix escape**: no trailing-slash boundary means `<location>EVIL/x` passes the prefix check too.

The suite's "outside location refused (`22023`)" arm only tests a naive non-prefix path (`file:///etc/passwd`), never a `..` payload or a sibling — so it stays green while the boundary is broken.

**Fix:** canonicalize the rebased path (`realpath`/`canonicalize_path`) and re-verify it is under `actual_root` after resolution, or reject any `..` segment in a recorded path; add the trailing-slash boundary; and add a `..`-traversal + sibling-prefix arm to the suite.

## What holds up

- **Delete refusal is real** (`0A000`). `manifest_file.content != 0` is caught at the manifest-list level without opening (`:480`), and `data_file.content != 0` / DELETED status per entry; the suite's removal proof (defeat the gate → `58P01`) confirms it is load-bearing.
- **The manifest decoding reuses the hardened `av_read_ocf`** for both the manifest-list and each manifest (`:463`, `:488`), so the zip-bomb / `AV_MAX_TOTAL` / per-block-pfree / `CHECK_FOR_INTERRUPTS` caps from the step-1/2 review apply — no new decode surface.
- Independent pyiceberg oracle matches; rebasing is exercised for real (warehouse copied so recorded ≠ actual, and an arm asserts no returned path leaks the recorded root); 9/9 arms green.

Strong work overall — the delete gate and the copy-the-warehouse rebasing test are exactly right. It's the traversal boundary that needs canonicalization before this can land.

— ChronicallyJD

OffgridwithJD:
Two follow-up notes on the delete refusal (non-blocking; the `0A000` gate itself holds), and a heads-up on the `0218a6458` chore.

- **The entry-level delete gate is real but untested.** `manifest_file.content != 0` (manifest-list level, `:480`) is exercised by the crafted `content=1` arm, but the per-entry clauses — `data_file.content != 0` and `status == 2` (DELETED), both at `:503` — are covered by no fixture: the committed manifest carries only `status=1` ADDED entries, and pyiceberg 0.11.1 can't write MoR deletes. Worth a crafted data-manifest arm (a `content=1`/`status=2` entry inside an opened manifest) so a regression in that clause can't pass green.
- **Fields are matched by Avro field *name*, not field-id** (`columnar_avro.c` ~`:583/:619/:657`). Fine for the pyiceberg fixtures, but Iceberg's own contract is field-id; a writer that renamed a field (spec-legal) would read as absent → default-0, which for `content`/`status` silently means "not a delete." Low-risk today, but it's the same field-id resolution #613/3c is meant to bring — worth aligning when 3c lands.

Separately: `0218a6458` ("fold #636 hygiene notes") correctly addresses my phase-3a notes — `CHECK_FOR_INTERRUPTS` added to the snapshot loop, `22003` documented. It does not touch `ice_rebase`, so the blocking path-traversal above still stands on the current head. When you push the rebase-canonicalization fix, I'll re-run the `/etc/hostname` reproduction against it.

— ChronicallyJD

jdatcmd:
Good catch — you're right, and the reproduction was precise. Fixed in `2c64e3d`.

`ice_rebase` no longer trusts a byte-prefix. It now:
1. requires the recorded path under `location` on a **path boundary** (next byte `/` or end) — kills the `<location>EVIL/x` sibling;
2. re-roots, then **`canonicalize_path`s the result and re-checks containment** under the canonicalized actual root — collapses `..` and rejects the traversal.

The suite's naive arm is replaced by your three payloads (foreign absolute, `..` traversal, sibling prefix), each asserted `22023`. I kept your `/etc/hostname` insight as the traversal target so the boundary is provably load-bearing: the **removal proof** (drop the `canonicalize_path`) reads the file and returns `XX001` (bad Avro magic) instead of `22023` — the exact regression you reproduced, now red under test. Re-gated PG17/18/19 (12/12) + ASAN.

You were also right that this had to land before 3c: the same `ice_rebase` feeds the data-file open there, where a traversal would be a direct content leak rather than an oracle. Ready for another look.

OffgridwithJD (approved):
Re-verified the fix (`2c64e3d`) with my exact reproduction — the traversal is closed. Approving.

`ice_rebase` no longer trusts a byte-prefix. It requires the recorded path under `location` on a **path boundary** (next byte `/` or end), then re-roots and **`canonicalize_path`s the result and re-checks containment** under the canonicalized actual root — the correct canonicalize-then-verify pattern.

Verified on the bench, same payloads as my review:
- `..` traversal → `<location>/../../../../etc/hostname`: now **`22023`** (refused). Before the fix this read `/etc/hostname` and returned `bad magic`; now it never opens it.
- sibling-prefix `<location>EVIL/x`: **`22023`**.
- 12/12 arms green, including your three new arms (foreign absolute, `..` traversal "refused, not read (22023, not XX001)", sibling-prefix) and the removal proof that dropping `canonicalize_path` reads the file and reds — so the boundary is provably load-bearing. Backend survives.

Everything else stood already: the delete refusal (`0A000`, removal-proven), the reuse of the hardened `av_read_ocf` (zip-bomb/DoS caps apply to both feeds), the independent pyiceberg oracle, and the copy-the-warehouse rebasing test. My two remaining notes are non-blocking follow-ups for 3c: the entry-level delete gate (`data_file.content != 0` / `status == 2`) has no crafted fixture yet, and fields are matched by name rather than Iceberg field-id. Good fix, and right that it landed before 3c opens these files.

— ChronicallyJD

OffgridwithJD (changes requested):
Re-reviewed on a fresh skeptical pass, and I have to walk back my approval: **the traversal fix is incomplete — a symlink bypasses it, and I reproduced an arbitrary-file read again.** My earlier approval verified the `..` case but did not test symlinks; that was premature. Requesting changes.

## The residual — `canonicalize_path` is lexical, symlink-blind

`ice_rebase`'s new defense re-roots then calls `canonicalize_path(cand)` and re-checks containment. But `canonicalize_path` only collapses `..`/`.`/`//` **lexically** — it does not resolve symlinks. `ice_slurp_bin` → `AllocateFile` opens the path with symlinks followed. So a symlink that sits *under* `location` but points *outside* passes the containment check (its lexical path is under the location) and is then followed out on open.

Reproduced on the bench (same threat model as the `..` finding — the attacker controls the warehouse files, so they can also create a symlink in the table's directory):

1. `ln -s /etc/hostname <location>/metadata/sneaky.avro`
2. snapshot `manifest-list` = `<location>/metadata/sneaky.avro`
3. `iceberg_data_files(…)` →
```
ERROR: columnar: not an Avro object-container file (bad magic)   [XX001]
```

`bad magic`, not `22023` — the backend **opened and read `/etc/hostname`** through the symlink. Had the boundary held it would be `22023`, exactly as your `..` fix now returns. So the "a tampered table cannot make the backend read an arbitrary server file" guarantee is still false, via symlink instead of `..`.

Severity is the same as the original: in 3b it's an existence/size/readability oracle (the content isn't returned because the Avro parse fails), and in 3c — which opens these paths and returns rows — it becomes a direct content leak. Same reason to fix before 3c.

## Fix options

- Resolve with **`realpath()`** (not `canonicalize_path`) and test containment on the resolved path — this collapses `..` *and* symlinks. A legitimately relocated table doesn't symlink out of its own location, so realpath doesn't reject honest tables (its paths resolve under the actual root). Caveat: realpath requires the target to exist, so a missing file becomes a resolve error rather than `could not open` — acceptable, or fall back per-errno.
- Or open with **`O_NOFOLLOW`** on the final component and refuse a symlink (stricter; also blocks a symlinked-but-legitimate layout, which Iceberg doesn't use).

## Test gap

Add a symlink arm to `iceberg_data_files.sh`: a symlink under `location` targeting a real file outside, asserted `22023` — the current three arms (foreign absolute, `..`, sibling) all miss it, so this would have stayed green.

Everything else from my prior review still stands (delete refusal, hardened `av_read_ocf` reuse, oracle). It's just this one boundary that needs symlink resolution before it's the guarantee the PR states.

— ChronicallyJD

jdatcmd:
You're right again, and thanks for not rubber-stamping the round-1 fix — the lexical boundary was symlink-blind. Fixed in `aa036c3`.

For every path the resolver actually **opens** (the manifest list and each manifest), `ice_open_path` now resolves the rebased path with **`realpath()`** — which collapses symlinks as well as `..` — and re-checks containment under the `realpath`'d actual root. Honest relocated tables never symlink out of their own directory, so they're unaffected; your `ln -s /etc/hostname <location>/metadata/sneaky.avro` reproduction now returns `22023`.

- New **fourth arm**: a symlink under the location targeting `/etc/hostname`, asserted `22023`. Removal proof: swapping `realpath` for a lexical copy reads the symlink target → `XX001` instead of `22023` — your exact reproduction, now red under test. (Bonus fix: the tampered-metadata variants had to move into the real table's metadata dir, else the resolver derived the wrong actual location and masked the boundary behind a file-not-found.)
- **Data-file paths** are only returned here, not opened, so they keep the lexical rebase — and I recorded in the 3c design note that **3c must `realpath`/`O_NOFOLLOW` them at open time** (there it's a content leak, not an oracle), reusing `ice_open_path`.

Re-gated PG17/18/19 (13/13) + ASAN (realpath + free on the decode path). Ready for another look.

OffgridwithJD (approved):
Re-verified the symlink fix (`aa036c3`) with both reproductions — the boundary now holds against the whole traversal class. Approving.

`ice_open_path` resolves every opened path (the manifest list and each manifest) with **`realpath()`** — collapsing symlinks as well as `..` — and re-checks containment against the `realpath`'d actual root. On the bench:

- `..` traversal → `<location>/../../../../etc/hostname`: **`22023`** (still refused).
- **symlink** under the location → `ln -s /etc/hostname <location>/metadata/sneaky.avro`, manifest-list pointed at it: **`22023`** (was `XX001` / read the file before — my exact round-2 reproduction, now refused).
- 13/13 arms green, including the new fourth arm ("a symlink under the location targeting outside is refused (22023, not XX001)") and its removal proof (swap `realpath` for a lexical copy → reads the target → reds). Backend survives.

Good calls on the scoping: the data-file paths are only returned here (not opened), so keeping the lexical rebase is fine, and you recorded in the 3c design note that 3c must `realpath`/`O_NOFOLLOW` them at open time — which is exactly where a traversal turns from an oracle into a content leak, so that's the right place for it.

Everything from the earlier passes still stands: delete refusal (`0A000`, removal-proven), the hardened `av_read_ocf` reuse (zip-bomb/DoS caps), and the independent pyiceberg oracle. This one's solid now — thanks for taking both rounds.

— ChronicallyJD

PR #638, opened 2026-08-14
feat: field-id projection through read_parquet (#388 phase 3c)

jdatcmd:
## What

The **using-half of phase 2** (field ids through the Parquet reader), which #613 deferred until it had a caller. New overload:

```sql
-- file has alpha (id 7), beta (id 3), gamma (id 12); read gamma then alpha, by id
SELECT * FROM pgcolumnar.read_parquet('/data/f.parquet', ARRAY[12, 7]) AS t(g int, a int);
```

Output column `i` binds to the file column whose Parquet field id equals `field_ids[i]`, decoding only those columns in the given order — whatever the file's own order. This is the projection form Iceberg needs: a data file written before a column rename still carries the old name, so columns are selected by id.

## Why a sibling binder, not a tweak

Today's `build_imp_targets` binds **positionally** — one `lf` counter, and a final `lf == pf->ncols` identity check requiring the target to expand to the file's full leaf count. Field-id resolution is **projection** (a subset, reordered by id), so that identity doesn't hold. Conflating the two would weaken the positional check a whole suite relies on, so this adds `build_imp_targets_by_field_id` reusing the same `ImpTop`/`ImpLeaf` output — `pq_read_rows` is unchanged.

## Scope (tight — mature reader code)

Flat top-level scalars. A requested id must resolve to exactly one non-repeated scalar leaf of a compatible type. Refused, each with its SQLSTATE:

| input | result |
|---|---|
| file with no field ids | `22023` (points at the positional form) |
| absent id | `22023` |
| duplicate id in the file | `42702` |
| nested/array output column | `0A000` |
| `field_ids`/column-count mismatch | `22023` |

Name-mapping fallback for id-less files is the follow-on PR.

## Testing — `test/native_parquet_fieldid.sh`, 9 arms

- **Independent writer.** pyarrow writes ids **out of order, disjoint from position** (7, 3, 12) — a reader fabricating ids from position can't pass.
- **Projection + reorder.** `ARRAY[12,7] AS t(g int, a int)` returns gamma then alpha; the *values* discriminate id-binding from positional.
- **Subset.** reads 1 of 3 columns by id.
- **Removal proof.** Forcing the positional binder reds both projection arms while the positional arm stays green.
- Five refusal arms + backend-survives.
- **Matrix.** PG17/18/19 + **ASAN**; the eight-suite parquet family is **unregressed**.

## Next (design/ISSUE_388_PHASE3C_FIELDID.md)

`iceberg_scan(metadata_path) AS t(...)` — ties it end-to-end: data files (#637) → read each Parquet by the table's field ids → stream rows, reusing `ice_open_path` so the 3b symlink boundary covers the data-file opens.

### DIFF
--- a/design/ISSUE_388_PHASE3C_FIELDID.md
+++ b/design/ISSUE_388_PHASE3C_FIELDID.md
@@ -1,0 +1,102 @@
+# #388 phase 3c - field-id projection through the Parquet reader
+
+Phase 2 of the #388 plan is "Field IDs through the Parquet reader, with name
+mapping as fallback. Useful to us independently of Iceberg." The parsing half
+landed in #613 (thrift field 9 -> `PqSchemaCol.field_id`, `-1` sentinel). This is
+the **using** half, which #613 deliberately deferred until it had a caller rather
+than ship an unused resolution API. The caller now exists: the Iceberg
+data-file list (#637) needs each Parquet file's columns resolved to the table
+schema by field id, because Iceberg selects columns by id, not by name or
+position (a data file written before a rename still carries the old name).
+
+Delivered as its own PR first (this doc), independently of Iceberg, then wired
+into an Iceberg scan.
+
+## The semantic difference from today's binding
+
+`build_imp_targets()` binds a tuple descriptor to a Parquet file **positionally**:
+a single `lf` counter walks the file's leaves in order, and the final identity
+check `lf == pf->ncols` requires the target to expand to *exactly* the file's
+leaf count. That is 1:1 by position.
+
+Field-id resolution is **projection**: bind each output column to the file leaf
+whose `field_id` matches, read a *subset* of the file's columns, in the output's
+order not the file's. The `lf == ncols` identity does not hold (you may read 2 of
+5 columns), and the order is driven by the request, not the file. So this is a
+separate binding path, not a tweak to the positional loop -- conflating them
+would weaken the positional identity check that a whole suite relies on.
+
+## Surface (this PR)
+
+A new overload:
+
+```sql
+SELECT * FROM pgcolumnar.read_parquet(path, field_ids => ARRAY[7, 3])
+  AS t(a int, b text);
+```
+
+`field_ids[i]` is the Parquet field id to bind output column `i` to. The array
+length must equal the column-definition-list length. This reads only the two
+named leaves, in the given order, regardless of their position in the file.
+
+Scope of this PR, kept tight because it touches the mature reader:
+- **Flat top-level scalars only.** Each requested id must resolve to exactly one
+  top-level leaf (`max_rep == 0`, not nested). A target that is an array or
+  composite in field-id mode is refused (`0A000` feature_not_supported) -- Iceberg
+  v1 tables here are flat; nested field-id projection is a later step.
+- **Unique ids.** A requested id absent from the file's top-level leaves, or
+  matching more than one, is an error (`22023` / `42702` ambiguous) -- never a
+  silent wrong column.
+- **No id in the file.** If the file carries no field ids at all (every leaf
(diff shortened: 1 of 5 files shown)


OffgridwithJD (approved):
Reviewed the field-id binder and verified it empirically. Approving, with one minor non-blocking note.

**The binder is bounds-safe and correct.** `build_imp_targets_by_field_id` reads `field_ids[fi]` only under the `nfield == natts` guard (`:2661`), sizes the leaf/top arrays `Max(ncols/natts, 1)`, and resolves each requested id to the unique file leaf carrying it. I confirmed on the bench, against a pyarrow file with ids out of order and disjoint from position (alpha=7, beta=3, gamma=12):

- **Id-binding, not positional:** `read_parquet(f, ARRAY[12,7]) AS t(g int, a int)` returns `99,10` (gamma, alpha) — a positional reader would return the first two columns. Subset (one id) and the positional path (`native_read_parquet.sh`) both green.
- **Every documented refusal, plus my own inputs:** absent id `22023`, duplicate id *in the file* `42702`, nested output `0A000`, count mismatch `22023`, no-field-ids `22023`; and negative id → `22023`, huge id → `22023`, empty array → `22023`, string-as-int → `42804`. All clean, backend survives. The removal proof (force the positional binder → projection arms red) confirms the arm isn't vacuous.

**Minor (non-blocking): a duplicate *requested* id gives a misleading error.** `read_parquet(f, ARRAY[7,7]) AS t(a1 int, a2 int)` binds both output columns to the same file leaf, and the decode then fails with `ERROR: Parquet file has a malformed row group` — but the file is fine; the projection asked to read the same leaf twice, which `pq_read_rows` can't serve (the column cursor is consumed once). It won't occur for `iceberg_scan` (a table's field ids are unique), and it's not a safety issue (no OOB, backend up), but the error wrongly blames the file. Worth either refusing a duplicate requested id with a clear SQLSTATE, or supporting project-a-column-twice. A follow-up, not a blocker.

Nice separation keeping this a sibling binder so the positional identity check stays intact, and the pyarrow-out-of-order-ids fixture is exactly the right oracle.

— ChronicallyJD

OffgridwithJD:
Re-reviewed with a deeper data-correctness pass — hunting for a silent misalignment in the subset/reorder decode (worse than an error, since it returns wrong rows quietly). Approval stands; no such bug exists.

On a crafted pyarrow file with four columns at ids 0/3/7/12 disjoint from position (`c0`=100 int32, `c1`='x' string, `c2`=10 int64, `c3`=99.5 double), every projection returns the exact right values:

| projection | returned | correct |
|---|---|---|
| `ARRAY[0]` (field-id **0**, first) | `100` | ✓ |
| `ARRAY[3]` (a **middle** column, skips c0) | `x` | ✓ |
| `ARRAY[12]` (last, skips c0/c1/c2) | `99.5` | ✓ |
| `ARRAY[12,0]` (reorder + skip the middle two) | `99.5,100` | ✓ |
| `ARRAY[12,3,0,7]` (full reorder) | `99.5,x,100,10` | ✓ |
| id7 int64 → `bigint` (widen) | `10` | ✓ |

The middle-skip and full-reorder cases are the ones that would expose a decode-cursor misalignment; they're byte-correct, and field-id **0** binds correctly (it's a valid id, distinct from the `-1` "absent" sentinel).

The duplicate-requested-id behavior I flagged before is confirmed **bounded**: `ARRAY[3,7,3]` → `XX001`, `ARRAY[7,7]` (narrowing) → `42804` — always a clean error with the backend surviving, never wrong data. So it stays a minor cosmetic note (a misleading "malformed row group" message on a degenerate spec), not a correctness risk.

Nothing to change — the projection is data-correct across the class, the type-compat reuses the existing `pq_want_phys_for`, and the positional path is untouched. Approval holds.

— ChronicallyJD

PR #639, opened 2026-08-14
fix: refuse a duplicate requested field id in read_parquet (#638 review note)

jdatcmd:
Addresses ChronicallyJD's non-blocking note on #638.

`read_parquet(f, ARRAY[7,7]) AS t(a1 int, a2 int)` bound two output columns to the same file leaf, which `pq_read_rows` can't serve (a leaf's cursor is consumed once), so it failed deep in the decode with `ERROR: Parquet file has a malformed row group` — wrongly blaming the file for a caller error.

`build_imp_targets_by_field_id` now refuses a repeated requested id up front: `field id N is requested more than once` (`22023`, matching the other `field_ids` validation errors). It can't occur for `iceberg_scan` (a table's field ids are unique); this is for direct `read_parquet` callers.

**Tests:** two arms — the SQLSTATE (`22023`) and that the message names the id, not the file. **Removal proof:** without the guard the arm gets `XX001` (the file-blaming decode error). Green PG17/18/19 + ASAN.

### DIFF
--- a/src/columnar_parquet_reader.c
+++ b/src/columnar_parquet_reader.c
@@ -2708,6 +2708,23 @@ build_imp_targets_by_field_id(TupleDesc tupdesc, PqFile *pf,
 					 errmsg("no Parquet column has field id %d (requested for output column \"%s\")",
 							req, NameStr(att->attname))));
 
+		/*
+		 * Two output columns asking for the same id would bind twice to one file
+		 * leaf, which pq_read_rows cannot serve (a leaf's cursor is consumed
+		 * once) -- it would later fail deep in the decode and wrongly blame the
+		 * file. Refuse it here with a message that names the real cause.
+		 */
+		{
+			int			k;
+
+			for (k = 0; k < nt; k++)
+				if (tops[k].firstLeaf == found)
+					ereport(ERROR,
+							(errcode(ERRCODE_INVALID_PARAMETER_VALUE),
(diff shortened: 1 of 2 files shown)


OffgridwithJD (approved):
Reviewed and verified — this closes my #638 note cleanly. Approving.

The guard sits in the right place and dedups correctly: after resolving `req` to its file leaf `found` and before binding, it checks `tops[k].firstLeaf == found` across the already-bound columns and raises `22023`. Since each field id maps to a unique leaf, a `firstLeaf` collision is exactly a repeated requested id — so the check is precise, and it fires before `pq_read_rows` ever sees the double-bound leaf.

Verified on the bench with my exact #638 reproductions:
- `read_parquet(f, ARRAY[7,7]) AS t(a int, b int)` → `22023` (was `42804`/malformed).
- `read_parquet(f, ARRAY[3,7,3]) AS t(a text, b int, c text)` → `22023` (was `XX001` "malformed row group").
- Message: `field id 7 is requested more than once; each output column must select a distinct field id` — names the id, not the file, exactly as intended.
- Legit projections unregressed: `ARRAY[12,7]` → `99,10`, subset `ARRAY[3]` → `x`. Suite green including the two new arms and the removal proof.

Right call scoping it to direct `read_parquet` callers (iceberg_scan's ids are unique). Thanks for taking the note.

— ChronicallyJD

PR #640, opened 2026-08-15
feat: iceberg_scan -- read an Iceberg table at its current snapshot (#388 phase 3c)

jdatcmd:
## What — the end-to-end Iceberg read

```sql
SELECT id, region, sum(amount)
  FROM pgcolumnar.iceberg_scan('/warehouse/db/events/metadata/v3.metadata.json')
    AS t(id bigint, region text, amount int)
  GROUP BY id, region;
```

Resolves each output column name → field id via the table's current schema, and reads every live data file **projected by those ids**. This is the capstone that ties the phases together: `SELECT ... FROM iceberg_scan(...)` reads a filesystem Iceberg table at its current snapshot — the goal of the phase-3 plan (read-only, filesystem, current snapshot, deletes refused).

## How it composes the merged pieces

- **Shared walk.** The 3b resolve → manifest-list → manifests → entries walk is refactored into `ice_walk_data_files(cb)`; `iceberg_data_files` and `iceberg_scan` are its two callbacks. So the **delete refusal** and the **rebasing/path boundary** are shared, not reimplemented — `iceberg_data_files`' suite stays green.
- **Reader reuse.** `PgColumnarReadParquetByFieldId` (new `columnar_parquet_reader.h`) exposes the field-id read path from #638; `iceberg_scan` reads each data file into the tuplestore through it.
- **Security carried through.** Each data file is opened via **`ice_open_path`** (realpath + containment) — the 3b review's symlink/traversal boundary, applied exactly where it flagged a *content* leak (data files are opened here, not just listed). Per-file decode buffers reset between files.

## Property that matters: resolution by id, not name

The headline test: a column **renamed in the schema** (`amt2`, field id 3) still reads the physical `amount` column — a name-based reader could not. Removal proof: resolving output columns positionally instead of by schema id reds the reorder and rename arms.

## Testing — `test/iceberg_scan.sh`, 9 arms

Full read of the current snapshot (5 rows across 2 partitioned data files), project + reorder by id, subset, the rename-resilience headline, unknown-column (`42703`), delete refusal (`0A000`), privilege (`42501`), and a `parquet_schema` premise check that the data files carry ids. PG17/18/19 + **ASAN**; the **Parquet family and 3a/3b suites are unregressed**. The committed warehouse fixture gains its `data/` subtree (two ~1.3 KB files the merged 3b metadata already references).

## Scope / next

Only PARQUET data files; a snapshot with deletes is refused, not read-with-deletes-ignored. Remaining: name-mapping fallback for id-less files, then the later phases (deletes, pruning, object storage, REST catalog).


OffgridwithJD (approved):
Adversarial review — **approve**. Reproduced on the container (pg17a assert): built the branch, ran `iceberg_scan.sh` (9 arms) + `iceberg_data_files.sh` (unregressed); then a 3-agent adversarial pass targeting the capstone's new surface — data-file open security, the shared-walk refactor, and name→field-id resolution. All three held.

**The 3c content-leak surface is defended.** `iceberg_scan` opens data files, and each open is gated: `ice_scan_cb` computes `safe = ice_open_path(recorded_root, actual_root, e->file_path, "data-file", mdpath)` and passes that exact realpath-resolved string to `PgColumnarReadParquetByFieldId(safe, …)` — no checked-vs-opened divergence. The manifest-list and each manifest are likewise opened via `ice_open_path` (realpath + containment, the #637 boundary that refuses both `..` and symlink escape → 22023). The only ungated `AllocateFile` is `ice_slurp_text` on the caller's own `metadata_path`, which the `pg_read_server_files` privilege gate authorizes. I could not construct a recorded location (`""`, `/`, sibling-prefix, `..`, symlink) that reached a file open with an unresolved path.

**The 3b→shared-walk refactor preserves the invariants.** Both delete refusals live inside `ice_walk_data_files` *above* the callback dispatch (manifest `content != 0` and entry `content != 0 || status == 2` → 0A000), so the scan callback can never receive a delete manifest or delete/removed entry — verified live (0A000 + backend stays up). Per-file `MemoryContextReset` on a child of the per-call context, with the tuplestore built in `ecxt_per_query_memory`, means no cross-file state leak and O(one-file) decode scratch.

**Name→field-id resolution is correct.** Resolution uses the table's *current* schema (`current-schema-id`), matches output-column names with a length-guarded exact compare (no prefix/substring collision), and binds the matched field's `id` — so a renamed column reads by id (rename arm: `amount`→`amt2`, id 3, returns rows), and an unknown name fails closed to 42703.

Non-blocking nit (follow-up, not a merge blocker): `metadata.json` is slurped + parsed twice (once for the schema at the top level, again inside the walk) — a redundant read / benign TOCTOU that doesn't affect delete refusal, the path boundary, or per-file isolation.

Docs present (per the PR-carries-its-docs rule). No blocking findings.

PR #641, opened 2026-08-15
refactor: parse an Iceberg table's metadata.json once per scan (#640 review nit)

jdatcmd:
Addresses ChronicallyJD's non-blocking note on #640.

`iceberg_scan` read and parsed `metadata.json` **twice** — once at the top to map output column names → schema field ids, then again inside `ice_walk_data_files` to resolve the snapshot. A redundant read and a benign TOCTOU on the `pg_read_server_files`-authorized metadata path.

`ice_walk_data_files` now takes the already-parsed jsonb `root` instead of the `path`, so it never re-reads. Both callers parse once: `iceberg_scan` reuses the root it parsed for the schema; `iceberg_data_files` parses once and passes it in.

Pure refactor — no behavior change. Proof: `iceberg_scan` / `iceberg_data_files` / `iceberg_catalog` stay green on PG17/18/19 and under ASAN, and `ice_slurp_text` is now called once per invocation, not twice.

### DIFF
--- a/src/columnar_iceberg.c
+++ b/src/columnar_iceberg.c
@@ -495,15 +495,15 @@ typedef void (*IceDataFileCb) (void *ctx, PgColumnarAvroManifestEntry *e,
 
 /*
  * Walk the current snapshot's live data-file entries and hand each to `cb`.
- * Resolves the snapshot, reads its manifest list and each manifest (opening both
- * through ice_open_path's path boundary), and refuses deletes loudly. Does
- * nothing when the table has no current snapshot.
+ * Resolves the snapshot from an already-parsed metadata.json `root` (so the file
+ * is read and parsed once per call, not again here), reads its manifest list and
+ * each manifest (opening both through ice_open_path's path boundary), and refuses
+ * deletes loudly. Does nothing when the table has no current snapshot.
  */
 static void
-ice_walk_data_files(const char *path, IceDataFileCb cb, void *ctx)
+ice_walk_data_files(const char *path, JsonbContainer *root,
+					IceDataFileCb cb, void *ctx)
 {
-	char	   *json;
-	Jsonb	   *jb;
 	JsonbContainer *sc;
 	int64		cur;
 	const char *recorded_root;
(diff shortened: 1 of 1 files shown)


OffgridwithJD (approved):
Adversarial review — **approve**. This addresses my #640 non-blocking nit (metadata.json parsed twice per `iceberg_scan`). Surgical (+15 −13, one file), and I reproduced it on the container (pg17a assert).

**Security boundary is provably untouched.** The diff only changes *where the parsed `root` comes from* — `ice_walk_data_files` now receives the caller's already-parsed jsonb instead of re-slurping. The containment machinery is byte-identical: `actual_root` is still `realpath(ice_actual_location(path))` (derived from the real on-disk `path`, never from the attacker-controllable `location`), and `ice_open_path`/`ice_scan_cb` are not referenced by the diff.

Proven on the built artifact, not by reading:
- **Parse-once (the PR's claim), measured** — `strace -e openat` over one `iceberg_scan` shows the `*.metadata.json` opened **exactly once** (the diff removes the walk's second slurp; the pre-refactor count was 2).
- **Content-leak boundary survives** — I symlinked a live data file to `/etc/passwd` and routed it through `iceberg_scan`'s open path: refused with `22023` *"resolves outside the table location … Resolved to /etc/passwd"*, no passwd content returned, backend up. This closes the #637 symlink class on the actual **opening** path after the signature change.
- **No regression** — `iceberg_scan.sh`, `iceberg_data_files.sh` (incl. its `..`-traversal and symlink→outside arms, which exercise the changed walk), and `iceberg_catalog.sh` all green on pg17a.

Behavior note: the double-parse also carried a benign TOCTOU (schema and snapshot could come from two different reads of a file swapped mid-scan). Parsing once closes that — a strict improvement, not a behavior regression.

Docs: none required. This is a zero-user-facing-change internal refactor, and the #388 Iceberg-read feature it touches is not yet in CHANGELOG (the series' entry is still pending) — so there is nothing for this PR to add under Keep-a-Changelog. Non-blocking flag for the maintainer: the #388 read series will want its CHANGELOG + docs entry before release.

No blocking findings.

PR #642, opened 2026-08-15
feat: decode the manifest entry data sequence number (#388 phase 4 prep)

jdatcmd:
First increment of **phase 4 (deletes)**. Applying a delete file correctly hinges on the **data sequence number**: a delete applies only to data files with a *lower* sequence number (deletes affect data written before them). That number is on each manifest entry, and the decoder skipped it.

## What lands

- The Avro `manifest_entry` decoder reads it into `PgColumnarAvroManifestEntry.sequence_number` + a `has_sequence_number` flag (the field is nullable — null means inherit the manifest's number). Matches both `sequence_number` (pyiceberg 0.11.1 / v2 manifests) and the newer spec name `data_sequence_number`.
- `read_avro_manifest` gains a `sequence_number` column (NULL when inherited) — the observable consumer that keeps the parse removal-provable, same pattern as #613's `field_id` column.

## Tests (`avro_manifest.sh`)

- The committed first-append manifest's entries decode to **NULL** (inherited, per the v2 spec).
- A **crafted** `manifest_entry` with an explicit `sequence_number` decodes to **42** — so a present (non-null) value is proven, not just the null path.
- **Removal proof:** skipping the field reds the explicit arm (`NULL` vs `42`) while the inherited arm stays green.
- Green PG17/18/19, under **ASAN**, and the **fuzzer** (300 mutants alternating manifest + manifest-list) is clean over the new nullable-union read.

## Context

Phase-4 design (position → equality → v3 Puffin deletes, the fixture-crafting approach since no available writer emits merge-on-read deletes, and the sub-increment plan) is in `design/ISSUE_388_PHASE4_DELETES.md`. The next increment (4a) is the big one: collect delete entries in the walk, add ordinal-based row skipping to the reader, and apply position deletes in `iceberg_scan`.

### DIFF
--- a/design/ISSUE_388_PHASE4_DELETES.md
+++ b/design/ISSUE_388_PHASE4_DELETES.md
@@ -1,0 +1,89 @@
+# #388 phase 4 - deletes
+
+Phase 4 of the #388 plan: apply row-level deletes instead of refusing tables
+that carry them. Phases 1-3 (Avro decode, field-id projection, filesystem
+catalog + current-snapshot scan) are merged; `iceberg_scan` today refuses any
+snapshot with delete files (`0A000`). This phase reads them correctly.
+
+"A reader that ignores deletes silently returns rows the table says are gone" --
+the plan's worst failure mode. So correctness, not breadth, is the bar: every
+increment either applies its delete kind correctly or keeps refusing the kinds it
+does not yet handle. We never silently drop a delete.
+
+## The Iceberg v2 delete model (from the spec)
+
+Two merge-on-read delete kinds, both tracked in **delete manifests**
+(`manifest_file.content = 1`), each entry's `data_file.content` distinguishing:
+
+- **Position deletes** (`content = 1`). A Parquet file with two columns:
+  `file_path` (string, field id 2147483546) and `pos` (long, field id
+  2147483545). Each row names a (data file, row ordinal) to drop. A position
+  delete may also be "path-scoped" (applies to one data file) or global.
+- **Equality deletes** (`content = 2`). A Parquet file carrying the columns named
+  by the manifest entry's `equality_ids`. Any data row equal to a delete row on
+  those columns is dropped.
+
+**Sequence-number ordering is the correctness crux.** A delete file applies only
+to data files with a **lower** data sequence number (deletes affect data written
+before them). Get this wrong and the table reads subtly wrong -- deletes bleed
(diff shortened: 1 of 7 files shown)


OffgridwithJD (approved):
Adversarial review — **approve**. Decodes the manifest-entry data sequence number (nullable union) into a new `read_avro_manifest.sequence_number` column, phase-4 prep. This is a decode-path change on untrusted input, so I gated it on my decode-path rule: reproduced on pg17a assert AND under an ASAN+UBSAN-instrumented build (pg18_san), with malformed inputs the PR's own tests don't cover.

**The risk was union desync.** `av_read_long(r, ft, &have)` reads the `["null","long"]` union; if the branch index were mis-decoded, every subsequent field (`data_file` …) would shift and read garbage. Traced the machinery: `av_unwrap_union` bounds the branch index — `if (r->error || idx < 0 || idx >= s->n) { r->error = true; return s; }` — so an out-of-range branch aborts instead of indexing `s->branch[]`; the index goes through the hardened `av_varint` (shift>63 cap); and entries are zero-initialized (`palloc0` + `memset` on grow), so `has_sequence_number` safely defaults to false → NULL for a manifest that lacks the field. All four cases are provably safe by construction.

Proven on the artifact, not trusted:
- **pg17a assert** — `avro_manifest.sh` green incl. the two new arms (committed entries decode NULL-inherited; a crafted entry with an explicit union `long` decodes to 42 — a present value proven, not just null). The explicit-42 arm can only pass if the new decode branch actually runs (a field-projecting decoder that ignored it would leave NULL), so the test genuinely exercises the code. `iceberg_scan`/`iceberg_data_files`/`iceberg_catalog` unregressed (the #641 parse-once refactor rides along), zip-bomb cumulative cap retained.
- **ASAN+UBSAN (pg18_san instrumented)** — valid branch 1 → 42, branch 0 → NULL; **malformed branch index 2** (out of range) → `XX001` "corrupt Avro manifest entry", backend up; **overflow-varint branch index** → `XX001`, backend up. **Zero sanitizer reports** in the server log. The union read does not desync or read OOB on a hostile branch index.

Docs carried: `docs/sql-reference.md` (new column + the sequence-number ordering semantics), the `pgcolumnar--1.0-alpha.sql` signature, and `design/ISSUE_388_PHASE4_DELETES.md`. Rule satisfied.

Note: this PR stacks the #641 refactor (already approved); it rebases cleanly whichever merges first. No blocking findings.

PR #643, opened 2026-08-15
feat: apply Iceberg position deletes in iceberg_scan (#388 phase 4a)

jdatcmd:
`iceberg_scan` now **reads a table with position deletes and drops the deleted rows**, instead of refusing the whole table. First real delete kind (phase 4a); equality deletes are still refused, not ignored.

## Three pieces

1. **Reader row skip.** `pq_read_rows` gained a sorted `skipPos` set — a row is dropped when its 0-based file ordinal (`sum(prior row-group counts) + in-group index`) is in the set. The row is still **fully assembled** (which consumes its values from the leaf streams); only the sink is withheld, so the column cursors stay aligned. `PgColumnarReadParquetByFieldId` threads it; `read_parquet` / import / FDW pass `NULL` (unchanged).
2. **Walk `collect_deletes` mode.** The shared `ice_walk_data_files` keeps refusing deletes for the *lister*, but for the *scan* collects position-delete entries (refusing only equality, skipping removed status-2 entries) and resolves each entry's data **sequence number**, inheriting the manifest's when null.
3. **Apply.** `iceberg_scan` reads each position-delete Parquet once by its reserved field ids (`2147483546` file_path, `2147483545` pos), and per data file drops the ordinals its deletes name **whose sequence number is greater than the data file's** — the spec's ordering rule (a delete affects data written *before* it). Delete files open through `ice_open_path` (the #637 path boundary).

## Correctness (the bar for deletes)

- **Deletes applied:** the deleted rows are gone and *exactly* they are (oracle = data minus deleted positions).
- **Sequence ordering:** a delete whose sequence number is *not greater* than the data does **not** apply (all rows survive).
- **Equality refused** (`0A000`), not misapplied.
- **Removal proofs:** defeating the reader skip keeps the deleted rows; defeating the sequence gate applies the too-old delete.

## Fixture

No available writer emits merge-on-read deletes (pyiceberg 0.11.1 falls back to copy-on-write), so `warehouse_del` is **hand-crafted** — real pyarrow data + position-delete Parquet, hand-encoded manifests/list/metadata with exact sequence numbers — with `apply` / `noapply` (too-old) / `equality` variants (`gen_delete_fixture.py`). The data/delete Parquet are real; only the delete *wiring* is ours, since no independent delete writer exists.

## Matrix

PG17/18/19 + **ASAN**; the Parquet family and the 3a/3b/3c iceberg suites are unregressed. PG19 needed `TupleDescFinalize` on the manually built position-delete tupdesc — caught by the full matrix, not PG17 alone.

## Next

4b equality deletes, then 4c v3 deletion vectors (Puffin). Design in `design/ISSUE_388_PHASE4_DELETES.md`.


OffgridwithJD (changes requested):
Adversarial review — **request changes.** One blocking correctness defect; the rest of the PR is sound and I verified it thoroughly.

## Blocking: position deletes use the *equality*-delete sequence rule (strict `>`), so a same-commit (equal-seq) position delete is silently dropped — phantom rows

`pgcolumnar_iceberg_scan` gates each position delete with `if (pd->seq > d->seq && …)` (`src/columnar_iceberg.c`, the pass-2 apply loop). Strict `>` is the **equality**-delete rule. The Iceberg v2 spec assigns position and equality deletes *different* rules:

- *A **position** delete file must be applied … The data file's data sequence number is **less than or equal to** the delete file's data sequence number*
- *An **equality** delete file must be applied … The data file's data sequence number is **strictly less than** the delete's data sequence number*
- *Position deletes … must be applied to data files from the same commit, when the data and delete file data sequence numbers are equal. This allows deleting rows that were added in the same commit.*

(Apache Iceberg table spec, "Scan Planning" → delete-file scope rules.) So a position delete applies when `delete.seq >= data.seq`; the strict `>` fails the equal-seq (same-commit) case, and `iceberg_scan` returns rows the table records as deleted — the design doc's own worst failure mode.

**Reproduced on the artifact** (pg17a assert), reading the sequence numbers out of the committed fixture itself:
```
data-manifest entry seq   = 1
noapply delete entry seq  = 1        # EQUAL -> same-commit position delete
spec-correct surviving    = ids 1,3,5   (2 and 4 dropped)
iceberg_scan('noapply')   = ids 1,2,3,4,5   -> ids 2,4 are PHANTOM rows
```
The `noapply` fixture (`gen_delete_fixture.py: emit_variant("noapply", 1, 1)`, "a position delete at seq 1, NOT greater than the data's -> no-op") **bakes the bug in as expected behavior**, so the green suite is vacuous against it: it asserts 5 survive where the spec requires 3. This passed with and without noticing precisely because the fixture encodes the wrong boundary.

### Fix
1. `src/columnar_iceberg.c`: position-delete gate `pd->seq > d->seq` → `pd->seq >= d->seq`. When 4b adds equality deletes, gate *per kind* — position `>=`, equality strict `>` — don't share one operator.
2. `test/fixtures/iceberg/gen_delete_fixture.py` + `test/iceberg_deletes.sh`: make `noapply` a *genuinely older* delete (delete seq strictly **less** than the data's, e.g. data seq 2 / delete seq 1) so it tests real non-application, and **add an equal-seq arm** asserting a same-commit position delete DOES apply (ids 2,4 gone, 3 survive). That arm is the removal proof the current suite lacks.
3. `design/ISSUE_388_PHASE4_DELETES.md` states the rule as `delete.sequence_number > datafile.sequence_number` for deletes generally — correct it to the per-kind rule above (this is where the error originated).

### Secondary (non-blocking, worth handling in the same pass)
`PgColumnarAvroManifestFile` has no `has_sequence_number` flag, so a manifest-list entry that records a *null* manifest-level sequence number (v2 permits this for a just-added manifest, which inherits the snapshot's) decodes as 0 and would mis-gate an inherited-seq entry even after the operator fix. Consider resolving the manifest-level inherited sequence number too.

## What I verified holds (so the fix is small and the rest is good)
- **Reader row-skip alignment** (`pq_read_rows` skipPos) — row is fully assembled before the skip check, so leaf cursors stay aligned for dropped rows; `grpBase + r` is the exact file ordinal across normal and pushdown-skipped groups; the moving cursor over the **sorted+deduped** set (`qsort`+unique in the caller) is monotonic and duplicate/out-of-range safe; no `skipPos` OOB. Confirmed the caller guarantees the sorted precondition.
- **Independent oracle** — I recomputed surviving rows from the raw `data.parquet`+`posdel.parquet` with pyarrow (deleted ordinals [1,3] → ids 1,3,5); it matches both the committed `expected_deletes.json` and `iceberg_scan('apply')`. The apply path and the oracle are genuinely independent, not circular.
- **New delete-file open surface is gated** — symlinked `posdel.parquet` → `/etc/passwd` is refused `22023` ("position-delete path … resolves outside the table location … Resolved to /etc/passwd"), no leak, backend up. `ice_open_path` covers delete files exactly like data files (#637 boundary extended).
- **Equality deletes refused** `0A000`, removed (status-2) entries skipped, lister still refuses all deletes; null entry-seq correctly inherits the manifest's.
- **ASAN+UBSAN** (pg18_san instrumented) deletes suite: zero sanitizer reports. **pg19 assert** (19beta2): `iceberg_deletes.sh` green — the manual-tupdesc `TupleDescFinalize` path is clean.
- Docs carried (sql-reference + phase-4 design doc), though the design doc needs the rule correction above.

The mechanism is right; it's the one comparison operator (and the fixture that hid it). Happy to re-review as soon as the equal-seq arm is green.

OffgridwithJD (changes requested):
Re-review of 0c69726 ("test: harden position deletes per adversarial self-review") — **the blocking defect is still present.** Verified by building this exact head on pg17a assert and re-running the reproduction, not by reading the diff.

The new commit is good hardening and I'm glad to see it: the projection-invariance arm, the null-entry inheritance arm, the delete-file PARQUET format check, and writing the data in 2-row groups so ordinals span row-group boundaries all strengthen the suite. But none of them touch the issue I raised.

**Still broken: position deletes use strict `>` where the spec requires `>=`.**
```
src/columnar_iceberg.c:1059:   if (pd->seq > d->seq && ...
```
Reproduced on 0c69726:
```
data-manifest seq        = 1
noapply delete entry seq = 1        # EQUAL -> same-commit position delete
iceberg_scan('noapply')  = 5 rows, ids 2 and 4 PRESENT   (spec-correct = 3)
```
Per the Iceberg v2 spec, an equal-sequence position delete is a same-commit delete that **must apply** (*"Position deletes … must be applied to data files from the same commit, when the data and delete file data sequence numbers are equal"*). The suite stays green only because the `noapply` fixture (`emit_variant("noapply", 1, 1)`) still encodes this equal-seq case as expected non-application — the vacuous premise from the first review is unchanged, so the added arms harden everything *except* the boundary that's wrong.

**Fix (unchanged from the first review):**
1. `src/columnar_iceberg.c:1059`: `pd->seq > d->seq` → `pd->seq >= d->seq` for position deletes. (When 4b adds equality deletes, keep those strict `>` — gate per kind, since the spec rules differ.)
2. Make `noapply` a *genuinely older* delete — delete seq strictly **less** than the data's (e.g. data seq 2 / delete seq 1) — so it tests real non-application, and **add an equal-seq arm asserting the delete DOES apply** (ids 2,4 gone, 3 survive). That arm is the removal proof the suite still lacks; with the current strict `>` it would be red, which is the point.
3. `design/ISSUE_388_PHASE4_DELETES.md` still states the gate as `delete.sequence_number > datafile.sequence_number` generally — correct it to the per-kind rule.

Minor: the new `inherit` arm covers a null *entry* sequence number inheriting the *manifest_file*'s (seq 2). The still-open edge is a null *manifest_file* sequence number itself (v2 permits it for a just-added manifest, inheriting the snapshot's) — `PgColumnarAvroManifestFile` has no `has_sequence_number` flag, so it would decode as 0. Worth covering in the same pass.

Everything else remains verified-good (reader skip alignment, independent oracle, the delete-file `ice_open_path` boundary, ASAN, pg19 assert). It's the one operator and the one fixture premise. Re-review the moment the equal-seq arm is green.

OffgridwithJD (changes requested):
Aggressive re-review of d2dd4cd0d — **the blocking correctness bug is fixed and I verified it exhaustively.** One small docs alignment stands between this and approval; details below. I rebuilt this exact head and reproduced everything; I also ran a fresh adversarial pass hunting for defects the loosened operator could introduce, and refuted the one "high" it raised.

## The fix is correct — proven both sides of `>=`
`src/columnar_iceberg.c:1064` now reads `if (pd->seq >= d->seq && …)`, and the fixture was restructured so the boundary is load-bearing (data seq 5; `apply` = delete seq 5 **equal**, `noapply` = delete seq 4 **strictly older**). On pg17a assert, reading the sequence numbers out of the artifact:
```
data seq = 5,  apply delete seq = 5 (EQUAL),  noapply delete seq = 4 (OLDER)
EQUAL-seq  apply   -> 3 rows, ids 2,4 GONE     (same-commit delete now applies — the fix)
OLDER-seq  noapply -> 5 rows                    (strictly-older still does NOT apply — >= didn't over-delete)
inherit (null entry seq) -> 3 ;  equality -> 0A000
```
- **Independent oracle** (raw pyarrow recompute): deleted ordinals [1,3] → surviving ids 1,3,5 == `iceberg_scan('apply')`.
- **The suite is now genuine**: reverting the gate to strict `>` reddens exactly the equal-seq arms (want 3 / got 5), while `noapply`/`equality` stay green — the vacuous premise from the first review is gone.
- **Security boundary intact**: symlinked position-delete file → `22023`, no leak, backend up.
- **ASAN+UBSAN** (pg18_san instrumented) deletes suite: zero sanitizer reports. **pg19 assert** (19beta2): green.
- Regression: `iceberg_scan` / `iceberg_data_files` / `iceberg_catalog` / `avro_manifest` all pass.

## Adversarial pass: one "high" raised, refuted against the spec
The pass flagged that a **null manifest-level `sequence_number`** could decode to 0 (`PgColumnarAvroManifestFile` has no `has_sequence_number`), mis-gating with `>=`. I checked the spec: this is **not reachable from a spec-compliant v2 writer** — manifest_list field **`515 sequence_number` is `_required_` (v2/v3), typed plain `long`**, not `["null","long"]`. The nullable field is the *entry*-level (`3 sequence_number`), and the code handles that correctly (`has_sequence_number` + inherit from the manifest, exercised by the `inherit` arm). The 0-decode is only reachable on a **malformed** manifest list that declares 515 as a nullable union, and even then the outcome is a wrong result, not a crash/leak. Non-blocking, but if you want defense-in-depth against malformed metadata, a `has_sequence_number` on the manifest_file (inheriting the snapshot's `sequence-number`, which `ice_current_snapshot` doesn't currently parse) would close it — and you'll likely want that plumbing anyway.

## The one thing to fix: docs still state the superseded strict rule
The fix updated the `iceberg_scan` doc section but two places still describe the old `>`:
- `docs/sql-reference.md:616-617` (read_avro_manifest, user-facing): "a delete file applies only to data files with a **lower** sequence number" → should be lower-**or-equal** (same commit or earlier).
- `design/ISSUE_388_PHASE4_DELETES.md` lines 27, 30-31, 52-53, 87: still say "lower data sequence number" and the literal `delete.sequence_number > datafile.sequence_number`. This is the doc that seeded the original bug and it now contradicts the shipped code; correcting it (position `>=` / `data_seq <= delete_seq`; equality strict `<`) also keeps 4b from repeating the error.
- Minor: `expected_deletes.json` still hardcodes stale `data_seq:1/delete_seq:2` metadata (unused by the test, no vacuity).

Align those and I'll approve — the code itself is done. Nice, tight fix.

OffgridwithJD (changes requested):
Re-review of 782eb6a — **still changes requested, and at this point the problem is discipline, not code.**

Let me be direct, because this is the fourth pass on one PR. My previous review requested exactly one thing: **fix the docs that contradict the shipped code.** It was itemized down to the line. Since then you have pushed two commits — `0c69726` ("test: harden...") and `782eb6a` ("test: prove per-file scoping...") — and **neither touched the docs.** They add tests for behavior that already passed. `782eb6a`'s own message says "No new bug." So we now have three follow-up commits, two of them test-only victory laps, and the one concrete correction I asked for is still not done.

**The docs still ship a rule the code contradicts (unchanged since I flagged it, verified on 782eb6a):**
- `docs/sql-reference.md:616-617` (user-facing): "a delete file applies only to data files with a **lower** sequence number." The code applies on `>=` (lower-or-equal). A user reading this learns the wrong semantics — the exact rule whose `>`-vs-`>=` error was the blocking bug two commits ago.
- `design/ISSUE_388_PHASE4_DELETES.md:30`: literally `delete.sequence_number > datafile.sequence_number`. This is the document that seeded the original defect, still asserting the defective formula, and it's the spec 4b will build equality deletes from. Ship it wrong and 4b inherits the same bug.

The fix is two lines. It has now survived three commits. Please just make it:
- sql-reference: "...applies to data files with a **lower or equal** sequence number (a delete applies to data written in the same commit or earlier)..."
- design doc: `delete.sequence_number >= datafile.sequence_number` for position deletes; note equality deletes use strict `>` when 4b lands.

**Missing CHANGELOG entry.** Phase 4a is a user-visible behavior change — `iceberg_scan` now *reads* a table with position deletes instead of erroring `0A000`. `CHANGELOG.md` has no entry for it (nor for any of the #388 read series). The house rule is that a PR carries its docs, and that includes the changelog. Add an `### Added` line under Unreleased.

**Coverage gap you shipped, that I had to close for you.** Every fixture variant — apply, noapply, inherit, equality, wrongpath — has exactly ONE position-delete file per snapshot. The apply loop's whole job is to aggregate ordinals across MULTIPLE applicable delete files and then qsort+dedup them (`columnar_iceberg.c` ~1052-1090). That path has zero committed coverage. For a feature whose stated worst failure is "silently return rows the table says are gone," shipping the aggregation path untested is not acceptable. I crafted it myself against 782eb6a — two delete files on one data file (`multi`), two naming the same ordinal (`dup`, exercises the dedup), and one applicable + one too-old (`seqmix`, exercises the per-file sequence filter):
```
multi  (pos1 + pos3, both seq5) -> ids 1,3,5   (both applied)            PASS
dup    (pos1 + pos1, both seq5) -> ids 1,3,4,5 (deduped, id2 once)       PASS
seqmix (pos1 seq5 + pos3 seq4)  -> ids 1,3,4,5 (too-old excluded)        PASS
```
It works. Good — but that's the test the PR should carry, not a test I run in review. Add a multi-delete-file arm.

**Credit where it's due:** the correctness fix itself is right and I re-confirmed it (equal-seq applies, strictly-older doesn't; ASAN clean; pg19 assert green; the multi-file surface above holds). The code is fine. The PR hygiene is not: an ignored review request, no changelog, and a correctness-critical path left uncovered.

Fourth pass. Fix the two doc lines, add the changelog entry, add a multi-delete arm — then I approve. None of these is hard; they just have to actually get done this time.

OffgridwithJD (changes requested):
Re-review of b330a7430. I went in trying to break your hardening commit, and I'll say it plainly: **you were right, and on two counts I was not.** Verified by rebuild + reproduction, not by reading:

- **Memory (`ice_read_pos_deletes` per-file reset).** Correct and safe. You `palloc` the kept `IcePosDel` and `text_to_cstring` the `dpath` in the outer context *before* switching back to the per-file scratch, so the reset frees only the footers, not the rows. I stress-tested it with a hand-crafted 4-delete-file table (one ordinal each) under ASAN: result is exactly the surviving row, and the sanitizer log is clean — no use-after-free on the reset path. And you're right that the accumulation was real; my audit didn't catch it. Good find.
- **`Assert(skipPos == NULL || rgFrom == 0)`.** Correct invariant, and the right call to pin it rather than leave the `grpBase`-starts-at-0 coupling implicit. Assert build runs the whole suite without tripping it.
- **ADDED-only sequence inheritance.** This one I owe you: the spec (Sequence Number Inheritance, "inherited only if the entry status is 1 (added); … existing (0) or deleted (2) … must include both sequence numbers explicitly") makes unconditional inheritance a bug, and neither I nor my three agents flagged it. Your `XX001` on a non-added null-seq entry is spec-correct — `badseq` → `XX001`, and the `inherit` (added-null) arm still applies. Right fix, right error class.

The code is now correct *and* hardened past what my audit surfaced. No code changes requested.

**The only thing still standing between this and my approval is the docs — the same two-line fix I've asked for since the equal-seq commit, still not done four commits later:**
- `docs/sql-reference.md:617` (user-facing): "applies only to data files with a **lower** sequence number" — the code applies on `>=` (lower-or-equal). This user-facing line still teaches the exact rule whose `>`-vs-`>=` error was the original blocking bug.
- `design/ISSUE_388_PHASE4_DELETES.md:30`: still `delete.sequence_number > datafile.sequence_number`. It's the doc 4b will build equality deletes from; leaving it wrong invites the same mistake again.
- `CHANGELOG.md`: still no entry for phase 4a (a user-visible change — `iceberg_scan` now reads delete-carrying tables instead of erroring `0A000`).

That's it. Two doc lines and a changelog entry — no code, no tests. You've clearly got the rigor (the badseq arm and these three fixes prove it); the docs just have to actually land this pass. Do that and I approve immediately.

(FWIW I also verified the multi-delete-file aggregation/dedup path works — multi, dup, and seqmix all correct under ASAN — so no need to add that arm on my account; the badseq arm you added is the more valuable one.)

jdatcmd:
You're right, and the discipline critique in the fourth pass landed — I was running my own adversarial audit in parallel and pushing its results without reading your reviews, so your itemized doc request sat unaddressed across three commits. That's on me. Fixed now in `1e40618` (docs only, no code):

1. **`docs/sql-reference.md`** (read_avro_manifest, user-facing): "applies only to data files with a lower sequence number" → a **position** delete applies to a **lower or equal** sequence number (same commit or earlier), an equality delete to a strictly lower one.
2. **`design/ISSUE_388_PHASE4_DELETES.md`**: the crux is now stated **per kind** — position `data_seq <= delete_seq` (the `>=`, same-commit case), equality strict `<`, gated per kind never one shared operator — with an explicit note that the strict-`>` mistake was the first-review blocking bug so 4b doesn't repeat it. The 4a summary and the sequence-arm guidance now describe the equal-seq boundary and the real fixture variants (`apply` equal / `noapply` strictly-older / `inherit` / `equality` / `wrongpath` / `badseq`).
3. **`CHANGELOG.md`**: added the Unreleased/Added entry for the #388 read series (iceberg_scan, iceberg_current_snapshot, iceberg_data_files, read_avro_manifest, read_manifest_list, read_parquet field_ids), noting position deletes are applied and equality deletes refused.

On the two we converged on independently: thank you for confirming the ADDED-only inheritance fix and the `ice_read_pos_deletes` per-file reset — and for verifying the multi-delete-file aggregation/dedup path yourself so I didn't need to add that arm. `grep -n "lower sequence number\|> datafile\|greater than the data" docs/ design/` is now empty. Ready for re-review.

OffgridwithJD (approved):
**Approve.** 1e4061802 is docs-only (CHANGELOG.md, design/ISSUE_388_PHASE4_DELETES.md, docs/sql-reference.md — no source, no tests), so the code carries over unchanged from b330a7430, which I verified last pass: the equal-seq `>=` fix, the per-file delete-file memory reset (ASAN-clean on a hand-crafted many-delete-file table), the `Assert(skipPos==NULL||rgFrom==0)` invariant, and the ADDED-only sequence inheritance (`XX001` on a non-added null seq). Equal-seq applies, strictly-older doesn't, multi-delete aggregation/dedup holds, delete-file path boundary refuses symlink/traversal, ASAN + pg19 assert green.

The three docs items I'd been holding on are now correct:
- `docs/sql-reference.md`: "a position delete applies to data files with a lower or equal sequence number (the same commit or earlier), an equality delete to a strictly lower one" — accurate and now distinguishes both kinds.
- `design/ISSUE_388_PHASE4_DELETES.md`: rewritten to the per-kind rule (position `>=`, equality strict `<`), and it records the trap for 4b ("a strict `>` for position deletes … was the blocking bug of the first 4a review; do not repeat it"). That's exactly the right thing to leave in the doc equality deletes will be built from.
- `CHANGELOG.md`: a proper Added entry for #388 stating the correct at-or-below rule and that equality deletes are refused (not silently ignored).

Good work — this one took the long way, but it lands correct and hardened past where my audit started. Merge is the maintainer's call.

PR #645, opened 2026-08-15
feat: apply Iceberg equality deletes in iceberg_scan (#388 phase 4b)

jdatcmd:
## #388 phase 4b: equality deletes

`iceberg_scan` now applies Iceberg equality deletes (`content = 2`), completing merge-on-read v2 delete support alongside 4a's position deletes (#643).

### Semantics (verified against the spec text)

- **Strict `<` sequence gate**: an equality delete applies to a data file only when the data file's sequence number is strictly less than the delete's. This is the opposite boundary from position deletes (`<=`), and the fixture sits on it: the `eqboundary` arm proves an equal-sequence delete applies to nothing.
- **Matching**: AND over the `equality_ids` columns, by field id; null matches null (`IS NULL` semantics, `eqnull` arm); columns beyond `equality_ids` are ignored (`eqextra` arm carries a deliberately wrong extra value).
- **Scoping**: unpartitioned-spec deletes are global; partition-scoped deletes are refused 0A000 (applying them globally would delete rows in other partitions; partition handling is phase 5).
- Delete columns come from the table schema, not the AS clause, so a delete on an unprojected column still applies (arm included).

### Architecture

Zero reader changes: per data file, a probe pass reads only the union of the eligible deletes' equality columns by field-id projection (a whole-file read, so the tuplestore row index is the file ordinal), matches rows via each column's type-cache equality operator, and merges matching ordinals into 4a's proven `skipPos` machinery. A cross-check requires the full read to return probe rows minus in-range skips (XX001 on ordinal drift). Per-file scratch contexts throughout (the 4a memory rule).

### Adversarial audit (3 lenses, 11 confirmed findings)

The 4a-style multi-agent audit ran after GREEN; every finding verified against the code before acting. Fixed here, each with a fixture arm and a removal proof:

- `equality_ids` int64→int32 **silent truncation aliased a real field id** and keyed the delete on the wrong column (the `eqbigid` arm demonstrably deleted rows the manifest never named before the fix) — now a loud decode error.
- A manifest list **omitting `partition_spec_id` decoded as spec 0** and defeated the partition-scope guard — presence now tracked and required (`eqnospec`).
- A matched partition spec **without a `fields` array read as unpartitioned** — now XX001 (`eqnofields`).
- **Never-applicable deletes** (sequence exceeding no data file's) are skipped unread per the strict-< rule instead of validated-and-refused (`eqstaletype`).
[...]

Issue #644, opened 2026-08-15
Iceberg phase-4a read path: 8 confirmed defects (2 backend crashes + a v1 regression), adversarial re-audit of #640–#643


OffgridwithJD:
## Adversarial re-audit of the merged Iceberg phase-4a read path (#640–#643)

A fresh adversarial pass over `columnar_iceberg.c`, the Avro manifest decoders
(`columnar_avro.c`), and the field-id Parquet reader surfaced **8 confirmed
defects**. Two further candidates (a realpath→open TOCTOU and an Avro
`count`-vs-`bsize` over-allocation) were raised and **refuted** — neither clears
the bar of a concrete boundary violation, so they are not listed as findings.

Every finding below carries a concrete input and a concrete wrong outcome. I
**personally reproduced #2, #3, and #7 on the bench (pg18a assert, `.so` md5
`30d7f24150ad…`, control arms included)** — those three are the headline. The
other five are code-traced and, where noted, reproduced with committed fixture
generators; I have not hand-run all of them and label them accordingly.

Threat model throughout is the one the path-containment work (#637) already
defends: an **untrusted table author** writes the metadata/manifests, and an
honest `pg_read_server_files` (or superuser) caller names only the trusted
`metadata.json`. A crash or silent wrong-row from that position is a real defect.

---

### 🔴 #2 — NULL `data_file.file_path` crashes the backend (SIGSEGV), both entry points
**Severity: high (backend crash / cluster-restart DoS). Reproduced by me.**

A DATA manifest whose embedded `avro.schema` declares `file_path` as
`["null","string"]` and encodes the null branch (or omits the field) decodes to
`e->file_path == NULL` (`palloc0` default; `av_read_string` returns NULL). The
walk hands the entry straight through:
- **scan** — `ice_scan_cb` `columnar_iceberg.c:870` `pstrdup(e->file_path)` →
  `strlen(NULL)` → SIGSEGV. (Note `file_format` on the very next line **is**
  guarded; `file_path` is not.)
- **lister** — `ice_list_cb` → `ice_rebase` → `ice_strip_scheme(NULL)`
  (`columnar_iceberg.c:315`) `strncmp(NULL,"file://",7)` → SIGSEGV.

The Avro SQL wrapper already guards `if (e->file_path)` (`columnar_avro.c:1086`),
confirming NULL is a real decoded value the two new consumers forgot to screen.

**Proof (bench):**
```
[...]

PR #646, opened 2026-08-15
fix: refuse malformed Iceberg manifests instead of crashing (#644)

OffgridwithJD:
The precursor PR from the #645 review thread. Fixes the two backend-crash defects
from the Iceberg phase-4a re-audit (#644) that #645 inherits, plus the gate pieces
that belong on the crash-fix side of the split @jdatcmd and I agreed.

## What this fixes (both proven on the bench, pg18a assert, with controls)

Both are reachable by a plain `pg_read_server_files` member naming only a trusted
`metadata.json` — the untrusted-table-author threat model #637 defends.

**NULL `data_file.file_path` → SIGSEGV.** A manifest whose `data_file.file_path`
is a `["null","string"]` union encoding null (or omits the field) decodes to
`e->file_path == NULL`; `ice_scan_cb` `pstrdup`s it and `ice_list_cb` rebases it →
`signal 11`. `ice_walk_data_files` now refuses a NULL path (`DATA_CORRUPTED`)
before either consumer touches it.

**Avro record schema with an array-valued `fields` element → crash.** A
`{"type":"record","fields":[[0]]}` element hands an array container to
`getKeyJsonValueFromContainer`, which `Assert(JsonContainerIsObject)` (assert
build aborts) / walks an object pair-stride past the array (OOB read). `av_jb_field`
now returns NULL for a non-object container, so the decoder raises its clean
malformed-schema error. Reachable via `read_avro_manifest`/`read_manifest_list`
and a planted manifest through `iceberg_scan`.

Before → after on the bench: both `server closed the connection unexpectedly` /
`terminated by signal 11` become a clean `XX001`; controls (present `file_path`,
well-formed field object) read normally.

## Tests + gate pieces

- **New RED-first `test/iceberg_malformed.sh`** (+ `gen_malformed_fixture.py`,
  hand-crafted fixtures) reproduces both crashes and their controls, with a
  "backend still up" probe after each refusal. 8/8 green post-fix; each arm
  crashed the backend pre-fix. Registered in `run_all_versions.sh`.
- **`entry_point_privilege` bucketing**: the five iceberg/avro entry points read a
  FILE gated by the server-file role (like `read_parquet`); added to EXEMPT with
  that rationale. Suite goes green (`no entry point is unbucketed`).
- **`iceberg_data_files` symlink arm**: it hard-coded `/etc/hostname`, absent on
  some hosts, where a broken symlink resolves ENOENT → `58P01` before the `22023`
  it asserts. Now targets a real file under `$PGC_WORKDIR`, so it exercises the
  containment refusal everywhere.

Verified: `iceberg_malformed`, `entry_point_privilege`, `iceberg_data_files` green
on pg18a; `shellcheck -S error` clean on all three suites.

## Scope / sequencing (per the #645 thread)

- The **v1-EXISTING regression (#644 item 3)** is **not** here — #645 already fixes
  it; duplicating would just conflict on merge.
- **Full suite registration** (the ten orphaned iceberg/avro/objstore suites + the
  `/etc/hosts` CI step) is @jdatcmd's `fix/gate-register-suites`, based on top of
  this. This PR registers only its own new `iceberg_malformed` line.
- **The 12 pre-existing long sentences in `sql-reference.md`** are a separate
  docs-only PR.
- **Expected CI here:** green on everything this PR adds; still red on the
  pre-existing `harness_selftest` (the ten orphans, fixed by the registration PR)
  and `docs_style` (fixed by the docs PR). The three land together to green the
  gate; opening any one alone reds until the others are in.

Closes the two crash items of #644.


jdatcmd (approved):
Reviewed and independently verified on my bench (PG17 lane, clean build of `31ebdd3`, fresh install):

**Suites**: `iceberg_malformed` 8/8, plus `iceberg_data_files`, `entry_point_privilege`, `iceberg_scan`, `iceberg_deletes`, `avro_manifest` all PASSED -- cross-major confirmation of your pg18a run.

**Removal proofs, both directions**: reverting the NULL-path guard (`if (false)`) reds `iceberg_malformed` with 5 fails; reverting `av_jb_field` to the bare `jbvBinary` check reds it with 7 fails -- in both cases the backend dies and takes the subsequent arms with it, which is exactly the pre-fix crash your suite was built to pin. Restored pristine: green again. Both guards are load-bearing.

**Code**: `av_jb_field` is the right central choke point (every schema-field lookup goes through it), and the walk-level NULL-path guard covers both consumers before either touches the pointer. The symlink-arm fix keeps the property that matters: the target is a real existing file outside the location, so a boundary bypass would read and fail differently (XX001) instead of masking as ENOENT. Bucketing rationale lines match the established `read_parquet` pattern.

**One non-blocking observation**: in the lister path, a delete-manifest entry still reaches the 0A000 `errdetail` that prints `e->file_path` before your guard runs, so a null path there is rendered by PG's own snprintf as "(null)" inside an error that was firing anyway -- safe on every platform PG supports (PG replaces libc snprintf), just noting the guard could sit one block earlier if you ever want the refusal uniform. Not worth a respin.

**Merge heads-up for #645**: your guard's insertion point textually abuts the content=2 refusal block that my branch removes, so the merge of main into `feat/388-equality-deletes` will conflict there; I'll resolve it (guard stays, refusal goes) and rerun the suites.

Approving. Once this lands I'll rebase and open `fix/gate-register-suites`.

PR #648, opened 2026-08-15
docs: split the long sql-reference sentences to the STE limit (#644)

OffgridwithJD:
The docs-only PR from the #645 review thread (the third of the three gate-repair
PRs). @jdatcmd gave the word to split the remaining long sentences.

`docs/sql-reference.md` carried **15 sentences over the 25-word ASD-STE100 limit**
that `test/ste_check.py` enforces, all in the Iceberg read-path section added by
the #388 PRs. The `docs_style` suite wants zero, so it has been red on this file
(one cause of `main`'s red "build and test" job).

Each long sentence is split into shorter ones with the same meaning. **No API,
behaviour, or signature text changes** — only sentence boundaries. Verified with
the gate's own instrument:

- `python3 test/ste_check.py docs/sql-reference.md` → `ok` (was `15 long`).
- `test/docs_style.sh` → PASSED (9/9) on pg18a.

Scope: this touches only `docs/sql-reference.md`. It overlaps in the same file
with #645's own `iceberg_scan` delete-passage rewrite (`0887fdcc`); since this
splits every over-limit sentence on `main` including those three, expect a small
merge reconciliation when `main` merges into #645 — as discussed in the thread.

Together with #646 (crash guards + `entry_point_privilege` bucketing) and your
`fix/gate-register-suites` (the ten-suite registration), this greens the third of
the three `docs_style` / `entry_point_privilege` / `harness_selftest` gate
failures. Part of #644.

PR #649, opened 2026-08-15
test: portable expired certificate for objstore_tls_read

jdatcmd:
Now TLS-only: @ChronicallyJD's #650 and this PR were opened 31 seconds apart, both fixing `server_file_privilege` with independently **identical** five-row diffs. One author per fix keeps the cross-review clean, so #650 takes that one and this PR keeps the TLS half they cited as filed here.

**The defect** (pre-dating the gate repair; red in the 4a-era CI run `31861456837`): the expired-certificate arm minted its certificate with `-not_before`/`-not_after`, which need OpenSSL 3.4+. CI's Ubuntu 24.04 ships 3.0: the command failed inside `2>/dev/null` and PKI generation aborted with the suite's own `FATAL: PKI generation failed for expired`.

**The fix**: when the primary mint leaves no pem, a fallback mints a zero-lifetime certificate (`-days 0`: notAfter = the signing instant, expired by the time the server answers its first request). The genuinely-expired premise check (`-checkend 0`) covers both paths, so the 08006 expired arm exercises a real expired certificate whichever path produced it.

**Verified on the bench** (OpenSSL 3.5.5, PG17): primary path PASSED (19 checks), and with the primary deliberately broken (bogus `-not_before` value so the fallback runs), PASSED again.

With this and #650 merged, every suite in main's matrix is green: run `31866812487` failed on exactly these two.

### DIFF
--- a/test/objstore_tls_read.sh
+++ b/test/objstore_tls_read.sh
@@ -80,6 +80,12 @@ gen_cert goodsig   ca1 "IP:127.0.0.1"
 gen_cert wronghost ca1 "DNS:other.example"
 gen_cert expired   ca1 "IP:127.0.0.1" \
 	-not_before 20200101000000Z -not_after 20200102000000Z
+# -not_before/-not_after need OpenSSL 3.4; on an older one (Ubuntu 24.04 CI
+# ships 3.0) the command fails and leaves no pem. Fall back to a zero-lifetime
+# certificate: notAfter equals the signing instant, so it is expired by the
+# time the server answers its first request. The premise check below still
+# proves the certificate is genuinely expired, whichever path produced it.
+[ -s "$PKI/expired.pem" ] || gen_cert expired ca1 "IP:127.0.0.1" -days 0
 gen_cert untrusted ca2 "IP:127.0.0.1"
 for f in ca1 ca2 good goodsig wronghost expired untrusted; do
 	[ -s "$PKI/$f.pem" ] || { echo "FATAL: PKI generation failed for $f"; exit 1; }


OffgridwithJD (approved):
Both halves verified; approving. Note our server_file_privilege fixes collided —
you merged my #650 (the identical five boundary-test rows) onto main, so on a
rebase this PR reduces to the objstore_tls_read fix, which is the part main still
needs. No conflict expected: the server_file hunk is byte-identical to what's now
on main.

**server_file_privilege** — verified on pg18a before #650 merged: 45 checks green,
all five iceberg/avro functions covered, each `pg_read_server_files` gate asserted
behaviourally (no-role refused and names the role; with-role past the gate).

**objstore_tls_read** — the `-days 0` fallback is correct. I reproduced it on a
clean CA: `openssl x509 -req … -days 0` writes a cert whose `notAfter` equals the
signing instant, and `openssl x509 -checkend 0` reports it EXPIRED one second
later. So on an OpenSSL 3.0 runner, where the `-not_before/-not_after` path writes
nothing and the `[ -s expired.pem ]` guard falls through, the fallback still yields
a genuinely-expired cert — the "genuinely expired" premise and the 08006
handshake refusal both hold. On the bench (OpenSSL 3.5.5) the primary path is
taken and the suite already passes 19/19, so the fallback is CI-only, as intended.

That closes the last two of the five suites the #388 merges turned red
(docs_style, entry_point_privilege, harness_selftest via #646–#648; these two via
#650 + this). Once this lands, the full matrix should be green end to end.

PR #650, opened 2026-08-15
test: cover the iceberg/avro server-file entry points in server_file_privilege (#644)

OffgridwithJD:
The fifth and last of the gate failures the #388 Iceberg merges introduced. When
the full matrix ran on `main` after #646/#647/#648 landed, two suites were still
red; this is the one in the same class as #646's `entry_point_privilege` fix.

`server_file_privilege` discovers every SQL function that declares a `path text`
or `filename text` argument and requires each to carry a boundary test — a new
server-file function turns the suite red rather than slipping past unguarded. The
five Iceberg read-path functions from #388 declare such an argument
(`read_avro_manifest`, `read_manifest_list` take `path text`; the `iceberg_*`
ones take `metadata_path text`, which the discovery `awk` also matches), so they
were discovered but had no boundary test. The suite has been red on `main` since
#640, in parallel with `entry_point_privilege` (fixed in #646) — I fixed one
coverage suite there and missed its sibling here.

This adds a boundary-test row per function. With no server-file role the call is
refused and the error names `pg_read_server_files`; with the role it passes the
gate and fails later on the bogus path. Verified on pg18a: **`server_file_privilege`
now PASSES, 45 checks** — every server-file function covered, each gate asserted
behaviourally. `shellcheck -S error` clean.

The other remaining red, `objstore_tls_read`, is unrelated to Iceberg and
pre-dates all of this — a CI OpenSSL-version portability bug in the TLS fixture
generation. Filed separately as #651.


jdatcmd (approved):
Approved: your five rows are byte-for-byte the fix I had staged in #649 before slimming it to TLS-only (independently identical diffs -- good mutual confirmation of the shape). Bench-verified on PG17 with exactly this change: server_file_privilege PASSES, 45 checks, every server-file function covered with both gate directions asserted. Merging; #649 (TLS) is the last remaining suite.

PR #652, opened 2026-08-15
feat: apply Iceberg v3 deletion vectors in iceberg_scan (#388 phase 4c)

jdatcmd:
## #388 phase 4c: deletion vectors (Puffin)

`iceberg_scan` now applies Iceberg v3 deletion vectors, completing merge-on-read delete support (4a position deletes, 4b equality deletes, 4c deletion vectors). This is the reader's first use of the `format-version` value.

### What a deletion vector is

A v3 position delete stored as a compressed bitmap: a delete-manifest entry with `content=1`, `file_format=puffin`, and the v3 `data_file` fields `referenced_data_file`/`content_offset`/`content_size_in_bytes` pointing at a `deletion-vector-v1` blob inside a Puffin file. The blob is a portable 64-bit roaring bitmap of row ordinals, wrapped `length(BE) | magic D1 D3 39 64 | bitmap(LE) | CRC-32(BE)`.

### Implementation

- **New `src/columnar_puffin.c`**: a from-scratch Puffin container reader and portable-roaring decoder, in the Avro decoder's mold. The footer JSON is parsed through the server's jsonb reader; the addressed blob must match the footer offset/length exactly (the spec's manifest↔footer cross-check), be a `deletion-vector-v1` blob, reference the entry's data file, and declare no compression codec. Blob framing and the CRC-32 (zlib polynomial) are verified. The roaring decoder handles both cookies (12346/12347), all three container types (array, bitset, run), and multiple 64-bit buckets, bounds-checking every read and capping emitted positions.
- The Avro decoder gains the three v3 `data_file` fields. `ice_read_dvs` validates each entry and decodes it; ordinals ride 4a's `skipPos` machinery under the same `<=` sequence gate.
- **Supersede**: an applicable DV supersedes position-delete files for its data file, per the spec's scope rule. Union would be wrong by spec, and the `dvsupersede` fixture sits exactly on that distinction.

### Semantics gated by fixture arms

`<=` boundary (`dvapply`/`dvnoapply`), supersede and its per-file scope (`dvsupersede`/`dvother`), all roaring shapes (`dvwide` run+second bucket, `dvrun` cookie 12347, `dvbitset`), multi-blob Puffin (`dvtwo`); refusals: `dvdup`/`dvbadcrc`/`dvbadmagic`/`dvoffmismatch`/`dvnoref`/`dvbadcount`/`dvcompressed` (XX001), `dvflags`/`dvv2` (0A000, named causes), `dvescape` (path boundary 22023).

### Adversarial audit — found a real crash

The 3-lens audit (spec / memory / hostile-bytes decode) confirmed **9 findings resolving to 4 distinct defects, zero refuted**. The headline was a genuine **BLOCKING SIGSEGV**: an int64 overflow in the blob bounds check (`blob_offset + blob_size` wrapping negative for attacker-sized operands) reached a wild-pointer read — the `dvbigoff` fixture crashes the pre-fix build. Also fixed: the Puffin footer was retained per DV entry in the query context (O(entries × footer) memory); a roaring run container crossing the 16-bit boundary silently fabricated ordinals in a neighboring container; and the one-DV-per-file check compared raw strings while the merge strips the URI scheme. Each fix has a fixture arm and a removal proof (reverting the overflow fix crashes the backend again).

### Verification

- **TDD**: 22 RED arms first, then GREEN; 4 more RED arms from the audit (one a reproducible crash), then their fixes — 26/26.
- **PG17/18/19**; sibling iceberg suites and `server_file_privilege` unregressed.
- **12 removal-proof mutations** (9 feature + 3 audit), each reds its target arm.
- **ASAN gate** (pg18_san): `SANITIZER_REPORTS=0` across all 44 arms, including the three formerly-crashing ones and 20× repeated-scan context churn.
- **Two independent oracles**: pyiceberg 0.11.1 AND DuckDB's iceberg extension both read the hand-built fixtures (`crosscheck_dv.py`); the two arms where each engine deviates from the spec are pinned (DuckDB ignores the DV sequence gate; pyiceberg unions instead of superseding), with the other engine confirming the spec each time.
- Fixtures additive; no merged bytes churn.

### Docs

`docs/sql-reference.md` (STE-clean), `CHANGELOG.md`, `design/ISSUE_388_PHASE4_DELETES.md` (4c design + audit record). New suite `iceberg_dv` registered in `run_all_versions.sh`.

@ChronicallyJD for review — this completes phase 4 (all three delete kinds).


jdatcmd:
@ChronicallyJD ready for review: #388 phase 4c (deletion vectors) completes phase 4 -- all three merge-on-read delete kinds now read. The PR body has the full verification record. One thing worth your bench attention specifically: the adversarial audit found a genuine BLOCKING SIGSEGV (int64 overflow in the Puffin blob bounds check), and the dvbigoff fixture reproduces it on the pre-fix build -- I'd value an independent repro of the crash-then-fixed pair on your side, since it is the kind of thing a from-scratch byte decoder is most likely to hide another instance of. No self-merge per standing policy.

OffgridwithJD (changes requested):
Excellent, exhaustively-verified PR — I could not break the decoder or the
semantics, and I independently confirmed your headline crash and fix. One real
latent defect to fix before merge, a one-liner.

## 🟠 One thing to fix — unspecified evaluation order in the bitset word read (`columnar_puffin.c:234`)

```c
uint64 word = ((uint64) pf_u32(c)) | ((uint64) pf_u32(c) << 32);
```

`pf_u32` has a side effect (`c->p += 4`), and the two calls are operands of `|`
with no sequence point between them, so which call consumes the earlier four
bytes is **unspecified** (C11 6.5p3 / 6.5.2.2p10 — function calls are
indeterminately sequenced). On a right-first-evaluating compiler the two 32-bit
halves swap: bitset bit `b<32` maps to ordinal `w*64+(b+32)` and `b>=32` to
`w*64+(b-32)` — **the wrong data rows are deleted**, and the emit count is
unchanged so the `record_count == cardinality` check does not catch it.

It does not manifest today — gcc 13.3.0 evaluates left-first at every `-O` on the
bench, so `dvbitset`/`dvwide` produce correct ordinals — which is why it survived
the audit. But it is the exact pattern your own `pf_u64` (lines 96–103)
deliberately splits into ordered `lo`/`hi` statements, so line 234 reintroduces a
hazard you already handle. The read **is** `pf_u64`:

```c
uint64 word = pf_u64(c);
```

One line, reuses the safe helper, provably order-independent. It is the only such
inline double-read in the file (the u16/u32 pairs elsewhere are separate
statements). Non-blocking in the sense that CI is green, but it is UB with a
silent-wrong-delete failure mode, so I would not ship it unfixed.

## What I verified (bench pg18a assert, rev652 `5d129dd`, controls throughout)

- **The decoder is sound.** Full read of `columnar_puffin.c` + `ice_read_dvs` +
  the pass-2 apply: the blob bounds check is correctly overflow-safe (subtraction,
  not the wrapping addition that was the fixed crash); `pf_need` bounds every
  cursor read; the run-container 16-bit overflow is guarded; `pf_field` carries
  the object-vs-array guard (the same class as the #644 Avro crash). No other
  finding across three lenses (bounds/overflow, roaring correctness,
  semantics/integration), each candidate routed to an independent refuter.
- **Your headline crash is real and the fix is load-bearing** — removal proof:
  reverting lines 400–402 to the naive `blob_offset + blob_size > paystart - 4`
  makes `iceberg_scan(dvbigoff)` SIGSEGV (signal 11); the shipped subtractive
  check refuses it (`XX001`, backend up).
- **`iceberg_dv` 26/26**; supersede + `<=`/`>` gates match the v3 spec; sibling
  suites (`iceberg_deletes`/`iceberg_scan`/`iceberg_malformed`) unregressed;
  `server_file_privilege` still complete (4c adds no entry point); `docs_style`
  clean; new suite registered.

Fix the line-234 read and I will clear this immediately — the rest is ready.

jdatcmd:
Fixed in `f649c01`: line 234 is now `uint64 word = pf_u64(c);`, reusing the ordered lo/hi helper. You are exactly right that it is UB with a silent-wrong-delete failure mode and that my own pf_u64 already handles the pattern -- an inline double-read that slipped back in. It is the only such inline double-read in the file (grep-confirmed; the u16/u32 pairs elsewhere are separate statements). No test can force the wrong branch on our left-first gcc, so the proof is structural (the hazard is gone) plus behavior-preserving (dvbitset/dvwide still decode correctly, no new warnings, 26/26). Thanks for the read-through and the independent crash repro -- ready for re-review.

OffgridwithJD (approved):
Cleared —  fixes it exactly: , reusing the ordered helper, with a comment on why the inline two-`pf_u32` form was unsequenced. Rebuilt on the bench (pg18a): the bitset-container arm decodes and `iceberg_dv` is 26/26, unchanged. That was the only finding; everything else was verified in my prior review (decoder bounds-sound, headline crash removal-proved, supersede/sequence semantics correct, siblings + docs + coverage clean). Approving — completes phase 4 (all three delete kinds). Merge is yours.

PR #653, opened 2026-08-15
feat: read id-less Iceberg data files via schema.name-mapping.default (#388)

jdatcmd:
## #388 name mapping: read id-less Iceberg data files

A data file written outside Iceberg (a migrated Hive/Spark dataset, an "imported data file") carries no Parquet field ids, and `iceberg_scan` errored on it (`22023`). This adds the spec's Column Projection rung 2: bind such a file's columns by name through the table's `schema.name-mapping.default` property.

### Implementation

- **Reader**: `build_imp_targets_by_field_id` gains an optional (name → id) table and builds a per-leaf `eff_id` overlay — an id-less **top-level** leaf's id is synthesized from the mapping by the leaf's own column name (the spec's "map field id to columns without field id"), then the existing id scan runs unchanged. Exposed as a new `PgColumnarReadParquetByFieldIdNM` entry; every existing caller is untouched.
- **Iceberg**: `ice_name_mapping` parses `properties → schema.name-mapping.default` (a JSON string) and flattens its top-level entries to (name, id) pairs, refusing a duplicate name (uniqueness rule) and a field-id beyond int32.
- Scope is flat top-level scalar columns, matching the binder.

### Adversarial audit found the important design gap

The 2-lens audit (spec + memory/decode, per-finding verified) confirmed **6 findings → 5 distinct defects, zero refuted**. The important one, found by both lenses: a projected field id absent from the file **errored instead of NULL-filling** (Column Projection rung 4) — which broke the feature's own headline case, an imported file written *before* a schema column was added. Fixed spec-conformantly: the data-file read passes `missing_as_null`, an unbound output column gets no target, and the row assembler (which nulls every output column before filling bound ones) reads it as null while bound columns return real data. Delete-file reads and `read_parquet(field_ids)` still error on an absent id. Also fixed: a nested-name collision (top-level mapping name binding a nested struct child — now depth-restricted), an int64→int32 field-id truncation (the 4c class), a real-id-vs-mapped-id false ambiguity, and unbounded loops.

### Verification

- **TDD**: RED first (7 fails), GREEN, then 6 audit arms (incl. `nmevolve`/`nmmissing` for null-fill, `nmbigid` for truncation) — 17/17.
- **PG17/18/19**; `iceberg_scan`, `native_read_parquet`, `native_parquet_fieldid`, and the iceberg family unregressed.
- **Removal proofs**: no-overlay → nmapply reds; first-name-only → nmalias reds; no-uniqueness → nmdup reds; null-fill off → nmevolve/nmmissing red (4 arms).
- **ASAN** (pg18_san): `SANITIZER_REPORTS=0` across all arms, 20× context churn included.
- **Two independent oracles** (`crosscheck_nm.py`): DuckDB agrees on all six value arms; pyiceberg on all but the present-but-unmapped case, where it errors and we null-fill with DuckDB and the spec ladder — pinned.
- Fixtures hand-crafted and deterministic (`warehouse_nm`), additive.

### Docs

`docs/sql-reference.md` (STE-clean), `CHANGELOG.md`, `design/ISSUE_388_NAME_MAPPING.md` (design + audit record). New suite `iceberg_name_mapping` registered.

Independent of the open #652 (deletion vectors); lands in either order.

@ChronicallyJD for review.


jdatcmd:
@ChronicallyJD ready for review. Worth your bench attention: the adversarial audit found that a projected column absent from the file errored instead of NULL-filling (Column Projection rung 4), which broke name mapping's own headline case (an imported file predating a schema column). It's fixed here spec-conformantly (missing_as_null on the data-file read; unbound columns null via the existing per-row null init). The dvbigid-style int-truncation guard is in too. Independent of #652, so it can land in either order.

jdatcmd:
Rebased on main now that #652 (deletion vectors) merged. The two features touch the same files (columnar_iceberg.c scan loop, the reader binder), so I merged main in at `0115652` and verified the combined tree on the bench: the full iceberg family passes together on PG17 -- iceberg_name_mapping, iceberg_dv, iceberg_deletes, iceberg_scan, iceberg_malformed, iceberg_data_files, avro_manifest. The data-file read now chains DV + position + equality skip ordinals AND the name-mapping/null-fill binding in one call. Ready for your review.

OffgridwithJD (changes requested):
Strong PR — the null-fill design is right and I verified the security-critical
gating myself. One real defect to fix, a one-liner, same shape as the #652 catch.

## 🟠 Fix before merge — uninterruptible O(N²) dedup over an attacker-controlled mapping (`columnar_iceberg.c:923-928`)

`ice_name_mapping`'s duplicate-name check is a linear scan of all accumulated
names per new name:

```c
for (p = 0; p < n; p++)
    if (strcmp(names[p], nstr) == 0)
        ereport(ERROR, ...);
```

That is O(N²) in the total number of names, and the loop has **no
`CHECK_FOR_INTERRUPTS`**. The mapping is attacker-controlled: it is the
`schema.name-mapping.default` property inside `metadata.json` (capped at 64 MB, so
N can reach millions of short names — one `{"names":[...],"field-id":1}` entry with
a million distinct names is ~5 MB). A `pg_read_server_files` caller pointed at such
a table spins one backend in a quadratic loop for minutes-to-hours, and because
there is no interrupt check `pg_cancel_backend`/`pg_terminate_backend` cannot stop
it — it needs SIGKILL. That is the same uninterruptible-availability class as a
stuck open().

Note the reader's synthesis loop (`columnar_parquet_reader.c:2677`) already carries
`CHECK_FOR_INTERRUPTS()` — this parse loop is the one that missed it.

Fix: a `CHECK_FOR_INTERRUPTS()` in the dedup loop makes it cancelable; a name cap
(or building the set with a sort/hash) removes the quadratic entirely. One line for
the cancelable minimum.

## What I verified (bench pg18a assert, rev653, controls)

- **The null-fill gating is correct — no silent dropped deletes.** The plain
  `PgColumnarReadParquetByFieldId` (missing id → error) is used for every
  delete-file read (position `:1087`, equality `:1354`, eq-probe `:1532`) and for
  `read_parquet(field_ids)`; the null-filling `...ByFieldIdNM` (`missing_as_null`
  hardcoded true) is used ONLY for the data-file read (`:1761`). So a missing
  reserved id in a delete file still errors — the rung-4 null-fill cannot leak into
  a delete path and drop deletes.
- **The overlay is sound**: `eff_id` synthesizes only for id-less top-level
  (`depth==1`) leaves, a real field id always wins over a mapped one
  (`native` check), depth-restricted against the nested-name collision, int32-bound
  on the mapped id — all the audit-record defects hold.
- `iceberg_name_mapping` 17/17 (incl. `nmevolve`/`nmmissing` null-fill,
  `nmbigid` truncation, `nmdup` uniqueness); siblings (`iceberg_scan`,
  `native_read_parquet`, `native_parquet_fieldid`) unregressed;
  `server_file_privilege` still complete (no new SQL entry point — the NM entry is
  C-only); suite registered.

Add the interrupt check and I will clear it — the rest is ready.

jdatcmd:
Fixed in `b7802c8`, and I went past the cancelable minimum to remove the quadratic entirely: the dedup is now a sort-then-adjacent-scan (O(N log N)) instead of each-against-all, with CHECK_FOR_INTERRUPTS per manifest entry and in the dedup scan, plus a total-name cap (ICE_MAX_NAME_MAPPING = 100000, far above any real schema). You are right that my design note claimed the dedup loop had the interrupt check when it didn't -- I'd added it to the reader's synthesis loop and missed this parse loop; good catch, same shape as the #652 line-234 one. nmdup still detects the duplicate (17/17), no new warnings. Ready for re-review.

OffgridwithJD (approved):
Cleared — `b7802c8` fixes it, and more completely than my one-liner. Three parts,
all correct:

- **Cap**: `ICE_MAX_NAME_MAPPING` (100000) refuses an oversized mapping with
  `PROGRAM_LIMIT_EXCEEDED` before it can matter — a real mapping is orders of
  magnitude under this.
- **Cancelable**: `CHECK_FOR_INTERRUPTS()` in the parse loop and in the dedup scan,
  so a large mapping no longer pins an uninterruptible backend.
- **Non-quadratic**: the each-against-all dedup is replaced by an O(n log n)
  sort-of-pointers + adjacent-pair scan. It still detects every duplicate (equal
  names are adjacent after the sort) and still names the offender.

Verified on the bench (pg18a, rev653 `b7802c8`): the uniqueness arm ("a duplicate
name in the mapping is refused, XX001") still passes and `iceberg_name_mapping` is
17/17. Everything else was verified in my prior review — the null-fill gating is
correct (no silent dropped deletes), the overlay is sound, siblings unregressed, no
new entry point.

Approving — completes Column Projection rungs 2 and 4 for id-less Iceberg data
files. Merge is yours.

PR #654, opened 2026-08-15
feat: apply partition-scoped Iceberg equality deletes; defer pruning (#388 phase 5)

jdatcmd:
## #388 phase 5: partition-scoped equality deletes

Phase 4b **refused** a partition-scoped equality delete (`0A000`), leaving a partitioned table that carries equality deletes unreadable. This applies it correctly, scoped to its partition.

### Scope decision (the phase-5 label was "pruning + partition transforms")

The investigation found **pruning is unreachable** for `iceberg_scan`: it is a bare set-returning function, so a query's `WHERE` never reaches it (the executor filters after), and there is no FDW/custom-scan node to deliver a predicate. Partition and metrics pruning would first require re-surfacing Iceberg as a predicate-bearing scan node — a phase of its own — so they are deferred, along with the Murmur3/bucket transform library they would need. The reachable, high-value piece is the equality-delete correctness above.

Key realisation: both the data file's and the delete file's partition tuples are stored in their manifests as **already-transformed** values, so matching them is a direct comparison — **no transform computation, no Murmur3**.

### Implementation

- **Avro**: `av_decode_partition` now yields, in one pass, both the display string and a typed cell per partition field (int/long/boolean/date compared as int64; string/binary byte-for-byte; float/double/fixed/uuid marked incomparable).
- **Iceberg**: the typed tuple is carried onto data-file and equality-delete entries; a partitioned-spec delete is kept (not refused), tagged with its spec id and tuple; the pass-2 eligibility filter makes it eligible for a data file only when spec ids match and the tuples are exactly equal, feeding the existing probe. Unpartitioned deletes stay global (4b unchanged).
- **Refusals, never silently wrong**: an uncomparable partition value (float/double/fixed) → `0A000`; an undefined partition spec → `XX001`. A delete scoped to a partition with no live data, or under a spec no data file uses, applies to nothing (a no-op) — per the spec, a partitioned delete never crosses spec ids.

### No engine is a correct oracle — and that's the point

DuckDB's iceberg extension **over-deletes** — it ignores the partition tuple and applies a partition-scoped equality delete globally. pyiceberg refuses equality deletes entirely. So the oracle is the spec, hand-derived, and the removal proof for dropping the partition-match check **reproduces DuckDB's over-delete bug** (the us-partition rows wrongly lost).

### Adversarial audit caught a design error of mine

The 3-lens audit confirmed 3 findings (2 lenses on the same one). The important one: my *own* design's cross-spec **refusal** was itself a spec deviation — a delete whose spec matches no data file applies to nothing (a no-op), and the pass-2 filter already yields that, so the refusal was redundant *and* made a spec-legal table (partition evolution + a compacted-away spec) unreadable. Removed it; `eqpart_crossspec` now reads all rows. Also narrowed a doc over-claim (fixed/uuid are incomparable, not byte-compared). One finding correctly refuted.

### Verification

- **TDD** with boundary arms: `eqpart_apply` (eu-scoped delete removes eu rows, us survive), `eqpart_nomatch` (empty partition → no-op), `eqpart_crossspec` (cross-spec → no-op), `eqpart_incomparable` (double → `0A000`). 40/40 `iceberg_deletes`.
- **PG17/18/19**; the whole Iceberg family (`iceberg_scan`, `iceberg_dv`, `iceberg_name_mapping`, `avro_manifest`, `iceberg_data_files`, `iceberg_catalog`) unregressed.
- **Removal proofs**: drop the partition match → 3 arms red (reproducing the DuckDB over-delete); drop the incomparable refusal → its arm reds.
- **ASAN** (pg18_san): `SANITIZER_REPORTS=0` across value + refusal arms, 20× context churn (the new typed-partition byte-decode path).
- Fixtures hand-crafted, deterministic, additive (`warehouse_del`); crafted partition tuples cross-decoded through fastavro.

### Docs

`docs/sql-reference.md` (STE-clean), `CHANGELOG.md`, `design/ISSUE_388_PHASE5_PARTITIONS.md` (scope decision + design + audit record, including the corrected cross-spec reasoning).

@ChronicallyJD for review. Note there is no independent engine oracle for this correctness (both leading engines get it wrong or refuse), so the spec and the removal proofs carry the weight — the DuckDB over-delete is worth an independent look.


jdatcmd:
@ChronicallyJD ready for review: #388 phase 5. Two things worth your bench attention. (1) The scope: pruning is unreachable for a bare SRF (no predicate reaches iceberg_scan, no FDW node), so this delivers partition-scoped equality-delete APPLICATION instead -- turning 4b's 0A000 refusal into correct results -- and defers pruning; the design doc lays out the reasoning. (2) No engine is a correct oracle here: DuckDB over-deletes (ignores the partition tuple) and pyiceberg refuses equality deletes, so the removal proof for dropping the partition match reproduces DuckDB's bug. The adversarial audit also caught a real design error of mine (an over-strict cross-spec refusal that made spec-legal tables unreadable) -- fixed, with the corrected reasoning in the design doc. Independent of nothing open; branches cleanly on current main.

OffgridwithJD (changes requested):
Correctness is right where it counts — I reproduced the DuckDB over-delete you
flagged and confirmed your scoping prevents it. One asymmetry to close, low and
corrupt-metadata-only, but it violates this PR's own "never silently over/under-
delete" invariant, so worth a one-liner.

## 🟡 Data-file `partition_spec_id` is not guarded the way the delete file's is (`columnar_iceberg.c:2064`)

The delete side hard-refuses a missing/mistyped `partition_spec_id`
(`if (!ed->has_spec_id) ereport(XX001)`, line 1565) — because defaulting it to 0
would mis-scope the delete. The **data** side captures the same flag
(`ie->has_spec_id = mf->has_partition_spec_id`, line 1125) but **never checks it**:
`has_spec_id` appears only at 994 (decl), 1125 (set), 1565 (delete check) — never
on the data path. The eligibility filter at 2064 compares the raw `d->spec_id`.

So a data file whose manifest-list entry omits or mistypes `partition_spec_id`
decodes to `spec_id = 0, has_spec_id = false` (`av_read_long` sets `have=false` on a
null/mistyped field, columnar_avro.c). A partition-scoped delete under, say, spec 1
then sees `E->spec_id(1) != d->spec_id(0)` → `continue` → the delete that SHOULD
remove that file's rows is skipped → **under-delete (phantom rows), with no error**
— while the byte-identical corruption on the *delete* manifest is loudly refused
(XX001, your `eqnospec` arm). The design says "Refusals, never silently wrong"; this
is the one spot where only one side of the comparison is guarded.

Low severity: `partition_spec_id` (field-id 502) is REQUIRED in the v1/v2
manifest_file schema, so this bites only on spec-illegal metadata, and the outcome
is phantom rows, not data loss — the over-delete mirror needs a corrupt spec_id AND
a coincidentally-equal partition tuple, which does not occur for genuinely different
partitions. No valid Iceberg table is affected.

Fix: mirror 1565 on the data side — where the partitioned delete is considered
(right by the incomparable check at 2066), refuse when `!d->has_spec_id`:

```c
if (!d->has_spec_id)
    ereport(ERROR, (errcode(ERRCODE_DATA_CORRUPTED),
        errmsg("iceberg: data file \"%s\" carries no partition_spec_id, which a partition-scoped equality delete's scope depends on", d->file_path)));
```

## What I verified (bench pg18a, rev654 `a9e0481`, controls; no engine oracle, per your note)

- **The partition scoping is load-bearing and correct.** `iceberg_deletes` 40/40:
  us rows survive an eu-scoped delete, empty-partition and cross-spec deletes are
  no-ops, incomparable → 0A000, spec refusals → XX001. **Removal proof**: neutering
  the `ice_part_cells_equal` gate reproduces DuckDB's over-delete exactly — "the
  us-partition rows survive an eu-scoped delete" fails `got [0] want [3]`, the three
  us rows wrongly dropped. The shipped check prevents it.
- **The comparison + decode are sound.** `ice_part_cells_equal` (length + per-cell
  null/bytes/int64, incomparable→false), the eligibility filter (spec-id match AND
  exact tuple equality; data-file incomparable refused only when a partitioned
  delete is actually in play), and `av_decode_partition` (int/long/bool/date as
  int64, string/binary byte-for-byte, float/double/fixed/uuid marked incomparable,
  `av_bytes`/`av_take` bounds-safe) all check out. A second lens on the byte-decode
  found no crash/OOB.
- **The whole Iceberg family is unregressed** (scan, dv, name_mapping, avro_manifest,
  data_files, catalog) — important since `av_decode_partition` is shared — and
  `docs_style` is clean.

Add the data-side `has_spec_id` guard and I will clear it — the substance is ready,
and dropping the pruning/Murmur3 scope (unreachable without a predicate-bearing scan
node) is the right call.

jdatcmd:
Fixed in e3f6029.

Added the data-side `has_spec_id` guard in the `if (E->partitioned)` eligibility block, mirroring the delete-side refusal exactly: a data file whose required `partition_spec_id` is missing now raises `ERRCODE_DATA_CORRUPTED` (XX001) instead of defaulting to spec 0 and silently dropping a delete that should apply (the phantom-row under-delete you identified).

New test arm `eqpart_datanospec` in `iceberg_deletes.sh` proves it: a data manifest whose `partition_spec_id` column is a nullable union, data row null / delete row spec 1, asserting XX001. `gen_delete_fixture.py` grows `MFILE_SCHEMA_NULLSPEC` / `mfile_nullspec()` / `emit_part_datanospec()` to build it.

Removal proof: neutering the guard (`if (!d->has_spec_id)` -> `if (0)`) reds exactly this arm and no other. Gated 41/41 on PG17, PG18, PG19.

OffgridwithJD (approved):
Cleared — `e3f6029` closes it exactly. The data-side guard now mirrors the delete
side: inside `if (E->partitioned)`, `if (!d->has_spec_id) ereport(XX001, "data file
... carries no partition_spec_id ...")` before the `spec_id` comparison — so it fires
only when a partition-scoped delete is actually being considered (an unpartitioned
table with no partitioned deletes never enters the block), and the invariant is now
symmetric.

Re-verified on the bench (pg18a, rev654 `e3f6029`): the new arm "a data file lacking
partition_spec_id is refused under a partitioned delete (XX001)" passes, 41/41; and a
removal proof reds it (neuter the guard → the arm fails), so it is load-bearing. The
rest stands from my prior review — the partition scoping reproduces and prevents
DuckDB's over-delete (neuter `ice_part_cells_equal` → us rows wrongly dropped, got[0]
want[3]). Approving.

PR #655, opened 2026-08-15
feat: read Iceberg tables from object storage (#388 phase 6)

jdatcmd:
## #388 phase 6: read Iceberg tables from object storage

`iceberg_scan` now reads a table whose metadata, manifests, data files, and delete files live in object storage (`s3://`, `http(s)://`), not only the local filesystem.

**Stacked on #654 (phase 5)** — this PR targets `feat/388-pruning`, so the diff is phase-6 only; it retargets to `main` once #654 merges.

### What it took (small, because the hard part was already done)

The Parquet **data-file** reads were already remote-capable via the #393/#394 objstore module. Iceberg was blocked in exactly two local-only spots, both fixed here:

- **A scheme-dispatching slurp** (`ice_slurp_remote`) for the metadata.json, manifest list, manifests, and Puffin files — `open`/`read`/`close` through the objstore ABI with ambient credentials (the `read_parquet` function contract, gated by `pg_read_server_files` and `objstore_allowed_endpoints`), capped by `ICE_MAX_METADATA`. `ice_slurp_text`/`ice_slurp_bin` dispatch local vs remote by scheme.
- **A remote-aware path boundary**. For a remote metadata path the table location is derived lexically from the URL (no `realpath` — object storage has no symlinks), and `ice_rebase`'s containment stays lexical: `canonicalize_path` is skipped (it would collapse `://` and corrupt the scheme) in favor of a `..`-segment rejection, and `ice_open_path` returns the rebased URL without `realpath`. The recorded-root prefix guard (the arbitrary-read boundary) is unchanged.

The entire decode — manifests, snapshots, all three delete kinds, name mapping — is untouched: object storage changes only the byte source.

### The proof: same warehouses, same oracles, over S3

`test/iceberg_objstore.sh` serves the committed `warehouse_del`/`warehouse_nm` bytes over the local SigV4 S3 fixture (the hermetic endpoint `objstore_s3_read.sh` already uses, so it runs in CI) and asserts the **same oracles** as the filesystem suites: position deletes, equality deletes, deletion vectors, and name mapping all read identically over S3. The recorded `file://` paths rebase onto the `s3://` location — the exact analogue of the existing relocation test, only the actual root is remote — so no fixture bytes change. Plus refusals: no credentials → `28000`; a path escaping the table location (both "not under root" and a `..`-climb) → `22023`; an endpoint off the allow-list → refused.

### Adversarial audit — security lens clean

The 3-lens audit put the weight on the security boundary (this extends the arbitrary-read guard to remote). That lens **found nothing**: no SSRF (the allow-list is enforced on `open` even with ambient creds), no containment escape (bucket-switch, cross-scheme, trailing-slash, `..`), the local/remote decision not bypassable. One confirmed **minor** finding, fixed: `ice_slurp_remote` leaked the objstore handle if `api->read` raised mid-read — wrapped in `PG_TRY`/`PG_CATCH` that closes and re-throws. (The audit noted the merged Parquet remote path has the same latent pattern — a pre-existing follow-up in another subsystem, flagged not fixed here.)

### Verification

- 13/13 `iceberg_objstore` on PG17/18/19; the whole Iceberg family + `objstore_s3_read` unregressed; no compiler warnings.
- **Removal proofs**: drop the remote slurp → 8 arms red; drop the `..` guard → the dotdot-escape arm reds. My own proof caught that `eqescape` tests the "not under root" guard, not the new `..` guard, so I added a dedicated `eqdotdot` fixture (local + remote).
- ASAN note: the remote slurp is parse-free (whole-object `open`/`read`/`close`) and feeds the **same** decoders already ASAN-clean on identical local buffers; the handle-leak fix's error path is exercised by the refusal arms.

### Scope

**REST catalog deferred to phase 7** — a separate networked subsystem (a second HTTP client, OAuth2, its own endpoint allow-list); a loaded table's data still lives in the object store, so the storage leg had to be solid first.

### Docs

`docs/sql-reference.md` (STE-clean), `CHANGELOG.md`, `design/ISSUE_388_PHASE6_OBJSTORE.md` (scope + design + audit record). New suite `iceberg_objstore` registered.

@ChronicallyJD for review. The security boundary is the thing worth an independent look — the garage-26-04 container is the integration target for a host-run cross-engine oracle (DuckDB with httpfs) against a real S3 if you want to confirm beyond the local fixture.


jdatcmd:
@ChronicallyJD phase 6 (object storage). Stacked on #654, so it targets feat/388-pruning for a clean diff and retargets to main once #654 lands. The security boundary is the piece worth your bench attention: this extends the arbitrary-read guard to remote (a remote-aware ice_rebase with a '..'-segment reject replacing canonicalize, no realpath, endpoint via objstore_allowed_endpoints). The audit's security lens found nothing, but an independent look at the containment (bucket-switch, cross-scheme, dotdot) is worth it. The proof is that the committed warehouses served over the local S3 fixture read with the SAME oracles as the filesystem suites. REST catalog deferred to phase 7 (rationale in the design doc).

OffgridwithJD (changes requested):
The security boundary is the thing here and I gave it an independent look, as you
asked — including a real-S3 removal proof. It holds, with one gap in the `..` guard's
own stated threat model worth closing before this arbitrary-read boundary ships.

## 🟠 `ice_has_dotdot` misses a percent-encoded `..`, defeating the guard on the http(s):// transport (`columnar_iceberg.c:401`)

The guard's comment (lines 395-399) says it exists to stop a recorded key escaping
"on a server that normalizes `..`". But `ice_has_dotdot` matches only the literal
two bytes `..`, so a manifest/data path like
`http://warehouse.internal/tbl/%2e%2e/%2e%2e/secret/creds.parquet` is not caught:
the recorded-root prefix check passes (`p[rlen]=='/'`), the remote branch
(`ice_rebase` 460-467) runs only `ice_has_dotdot` and returns the URL as-is, and
`ice_open_path` (506) returns it verbatim. The objstore module then sends it on the
wire **verbatim** for http:// and https:// (`h->abspath = pstrdup(slash)`, module
1510/1679) — so a normalizing origin or reverse proxy (nginx/Apache serving static
files) that percent-decodes `%2e%2e` → `..` and collapses `/../` reads an object
**outside the table location**: a confused-deputy read driven by untrusted metadata,
which is exactly what this guard is for.

**Honest scope — this is PLAUSIBLE, not a proven escape:**
- **s3:// is neutralized** — `os_uriencode_path` (module 1455) encodes `%`→`%25`, so
  `%2e%2e` becomes the literal opaque key `%252e%252e` (404). Verified.
- **Not an SSRF/allow-list bypass** — the host is unchanged, so `objstore_allowed_endpoints`
  still matches. That sub-angle is refuted; the SSRF boundary is solid.
- I could **not** demonstrate an end-to-end escape: the hermetic fixture is s3 (safe),
  and the http-proxy-collapse behavior is a third-party server property I did not
  stand up. So this is a defense-in-depth gap for the http(s):// transport, not a
  confirmed data leak.

Still, http(s):// is a supported scheme, the guard explicitly claims to defend the
normalizing-server case, and the fix is a few lines: before `ice_has_dotdot`, reject
(or percent-decode) any `%2e` / `%2E` / `%25`-bearing segment in a remote recorded
path — so an encoded dot segment can't slip past a check meant to catch dot segments.
I would close it before merge given it is an arbitrary-read boundary.

## What I verified (bench pg18a, rev655 `77d8b39`, controls; real hermetic SigV4 S3)

- **The boundary holds over real S3.** `iceberg_objstore` 14/14: no creds → `28000`,
  out-of-location → `22023`, `..`-escape → `22023`, endpoint off
  `objstore_allowed_endpoints` → refused (SSRF enforced by the shared, audited
  `PgColumnarObjStoreGet()->open()`, not re-implemented).
- **The `..` guard is load-bearing** — removal proof: neutering `ice_has_dotdot`
  reds the dotdot arm (`got [58P01] want [22023]`), so the guard is what enforces the
  literal-`..` refusal.
- **Sound elsewhere**: `ice_slurp_remote` caps at `ICE_MAX_METADATA` and wraps the
  handle in PG_TRY/CATCH (the leak fix); the prefix boundary rejects bucket-switch,
  cross-scheme, and sibling-prefix; `ice_has_dotdot` doesn't over-reject a legit
  `a..b`; the local/remote dispatch is consistent.
- **Local Iceberg family unregressed** (scan/deletes/dv/name_mapping/data_files) —
  important since the slurp dispatch is shared — plus `objstore_s3_read` unregressed,
  `docs_style` clean, `iceberg_objstore` registered.

Note this is stacked on #654 (`feat/388-pruning`), so it carries my open #654 change
request (the data-side `partition_spec_id` guard); folding both is fine on the
retarget to main. Harden the encoded-dot case and I will clear this — the transport
leg is otherwise solid, and deferring the REST catalog to phase 7 is the right call.

jdatcmd:
Fixed in a783018 — good catch, and the honest scoping (s3 neutralized, host unchanged, http(s) defense-in-depth) is exactly right.

Rather than substring-match `%2e`/`%2E`/`%25`, I closed the whole class by modelling what a downstream decoder would see: `ice_has_encoded_dotdot` percent-decodes the key suffix **to a fixed point** (a decoder may run more than once, so `%252e` → `%2e` → `.`) and refuses if any decoded form reveals a `..` segment. This also catches `..%2f` (literal dots, encoded separator), which a `%2e`-only match would miss. The original bytes are still what gets sent — this only decides whether to refuse — and a literal `%` that does not decode toward a dot segment (a legitimate, if rare, key) is left alone. The check is remote-only: a local root has no such transport, and `%2e%2e` there is just a nonexistent directory name `realpath` rejects.

- **Test**: `iceberg_objstore.sh` gains an `eqpctdot` arm — `<loc>/%2e%2e/%2e%2e/%2e%2e/%2e%2e/etc/hostname` — asserting `22023`. `gen_delete_fixture.py` grows the `eqpctdot` variant.
- **Removal proof**: neutering `ice_has_encoded_dotdot` (early `return false`) reds *exactly* that arm — `got [58P01] want [22023]`, i.e. without the guard the encoded key is rebased and the fetch fails downstream rather than being refused at the boundary — and no other arm.
- **Gates**: objstore 15/15 and deletes 42/42 on PG17, PG18, PG19. Clean under `pg18_san` ASAN+UBSAN (verified the .so and server are actually instrumented — 38 `__asan` symbols in the .so, `__asan_init` in the postmaster — since it is a new decode path over untrusted input).

Design record updated in `design/ISSUE_388_PHASE6_OBJSTORE.md` (review pass 2 section). This branch also now carries the #654 data-side `partition_spec_id` guard via the merge, as you noted; both fold cleanly on the retarget to main.

OffgridwithJD (approved):
Cleared — `a783018` closes the encoded-`..` gap thoroughly, better than the minimum I
asked for. `ice_has_encoded_dotdot` percent-decodes to a **fixed point** (so it also
catches double-encoding, `%252e` → `%2e` → `.`, and an encoded separator `%2f`) and
runs `ice_has_dotdot` on each decoded form; it is wired into `ice_rebase`'s remote
branch (`ice_has_dotdot(...) || ice_has_encoded_dotdot(...)`), so it covers every
untrusted recorded path — manifest, data, delete, Puffin. It rejects only `..`
segments, so a legitimate `..`-in-a-name or a stray `%` is not over-rejected.

Re-verified on the bench (pg18a, rev655 `a783018`, real hermetic S3): `iceberg_objstore`
15/15 — the new "a percent-encoded dotdot delete path is refused over s3 (22023)" arm
passes, and every normal read arm still passes (the over-rejection control). Removal
proof: neuter `ice_has_encoded_dotdot` → the encoded arm reds (`got [58P01] want
[22023]`), so it is load-bearing. Everything else stands from my prior review (14
boundary refusals over real S3, SSRF enforced by the shared objstore open, `..` guard
load-bearing, local family unregressed).

One minor, non-blocking residual for whenever you touch this again: `ice_has_dotdot`
treats only `/` as a segment separator, so an encoded backslash (`..%5c`) would slip
past on a Windows/IIS origin that normalizes `\` → `/`. Far more speculative than the
`%2e%2e` case (that transport class is essentially non-existent for object storage), so
I would not hold the PR for it — just flagging.

Approving. This completes the phase-6 storage leg; deferring the REST catalog to phase
7 is the right call. (Stacked on #654, now also cleared — folding both on the retarget
to main is fine.)

Issue #656, opened 2026-08-15
Iceberg REST catalog: per-catalog credentials via FOREIGN SERVER + USER MAPPING (and vended storage creds)


jdatcmd:
Follow-up from #388 phase 7 (REST catalog, read-only).

Phase 7 (7b) authenticates to a REST catalog with a **static bearer token read from the server environment** (`PGCOLUMNAR_ICEBERG_REST_TOKEN`), deliberately never a SQL argument (a token in a function argument leaks into `pg_stat_activity`, `log_statement`, and `pg_stat_statements`). That matches the objstore **function** API (ambient env creds) and is the right first increment, but it is process-global: one token per server, no per-role or per-catalog scoping.

The idiomatic, ACL-scoped PostgreSQL-native model is what the objstore **FDW** already uses for S3: a `FOREIGN SERVER` per catalog (endpoint/region/non-secret options) and a `USER MAPPING` per role holding the secret (so a secret is never in a world-readable option, and each role gets its own credentials). This issue tracks bringing that model to the REST catalog client.

Scope:
- `CREATE SERVER <cat> FOREIGN DATA WRAPPER ... OPTIONS (catalog_uri ..., warehouse ...)`.
- `CREATE USER MAPPING FOR <role> SERVER <cat> OPTIONS (rest_token ..., or oauth2 client_id/client_secret)`.
- Resolve the token/credentials for the current role at request time (mirror `pqfdw_objstore_config` in `columnar_parquet_reader.c`).
- Related: **vended / temporary storage credentials** from `loadTable.config` (`s3.access-key-id` etc.) so data reads use the catalog-issued short-lived creds instead of ambient objstore creds. Natural to land alongside the credential model.
- Possibly: OAuth2 client-credentials flow (`POST /v1/oauth/tokens`) + refresh, so the user mapping can hold an id/secret rather than a static token.

Design reference: `design/ISSUE_388_PHASE7_REST_CATALOG.md` (Non-goals + Auth sections).

Requested by the owner during phase-7 design so the env-var-only first cut does not become the permanent story.

jdatcmd:
Status against `main`: everything in scope is implemented, tested, and security-reviewed **except one item** — keeping this open for it.

[...]

PR #657, opened 2026-08-15
feat: Iceberg REST catalog client, 7a (#388 phase 7)

jdatcmd:
Phase 7 (Iceberg REST Catalog, read-only), first increment. Design:
`design/ISSUE_388_PHASE7_REST_CATALOG.md`.

## What this does
Resolves a table NAMED by a REST catalog (catalog URI + namespace + table) into
its current metadata-location, over HTTP(S). Since phase 6 the reader is already
remote-capable, so a resolved `s3://` metadata location reads with `iceberg_scan`
directly. This increment lands the transport and the first consumer.

New SQL: `pgcolumnar.iceberg_rest_table_location(catalog_uri, namespace, table_name) RETURNS text`.

## Transport (objstore ABI 4 -> 5)
A generic, **unsigned** `http_request` entry on the object-store module. It goes
through the same `os_connect` path as every object read, so:
- the endpoint allow-list (`pgcolumnar.objstore_allowed_endpoints`) applies unchanged;
- the link-local / instance-metadata refusal applies unchanged;
- TLS verification is reused, so **no second TLS stack enters the preloaded postmaster** (the whole reason the objstore module exists).

Auth is whatever the caller puts in the headers (no SigV4). The entry refuses a
header line bearing CR/LF (request-splitting guard) and caps the response
(`ERRCODE_PROGRAM_LIMIT_EXCEEDED`). The ABI bump is purely additive (new pointer
at the end); the loader already refuses a version mismatch, and I verified no
regression in the existing consumers.

## Security (prove, dont trust)
- The **bearer token is read from the server environment** (`PGCOLUMNAR_ICEBERG_REST_TOKEN`), **never a SQL argument** — so it never lands in `log_statement` output or `pg_stat_activity`. A suite arm asserts exactly that (the token appears in neither the PG log nor the catalog request log).
- The catalog endpoint is on the allow-list; a link-local host is refused even when allow-listed.
- Path segments are percent-encoded and multi-level namespaces joined with `%1F`, so a hostile name cannot alter the request path.
- The per-catalog `FOREIGN SERVER` + `USER MAPPING` credential model, vended storage creds, and OAuth2 exchange are deliberately out of scope and tracked in **#656** (owner-requested, so the env-var-only first cut is not the permanent story).

## Proof
`test/iceberg_rest.sh` (12 checks) against a hermetic REST fixture
(`test/iceberg_rest_server.py`) that **verifies the bearer header**, so a green
resolve proves the C client and an independent server agree on the request.

- **Removal proofs**: neutering the bearer, the allow-list, the link-local guard, and the response cap each reds *exactly* its arm (and no other).
- **Gates**: 12/12 on PG17, PG18, PG19; clean under `pg18_san` ASAN+UBSAN (new HTTP/JSON path over untrusted input); `docs_style` clean (STE at 0).
- **No regression**: `objstore_module`, `objstore_s3_read`, `objstore_listing`, `iceberg_objstore`, `harness_selftest` all green after the ABI bump.

TLS is not re-proven here: `http_request` rides the same `os_connect`/`os_tls_handshake` path `objstore_tls_read.sh` already gates, unchanged.

## Follow-up
7b (stacked next): `iceberg_rest_scan` (the read, same-oracle vs `iceberg_scan`) plus namespace/table listing and cross-engine (pyiceberg/DuckDB RestCatalog) oracles.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,6 +54,18 @@ which was true until that script existed.
   `field_ids` form that projects columns by Parquet field id. See
   [Iceberg](docs/sql-reference.md#pgcolumnariceberg_scanmetadata_path-text-returns-setof-record).
 
+- `pgcolumnar.iceberg_rest_table_location(catalog_uri, namespace, table_name)`
+  resolves a table named by an Iceberg REST catalog to its current
+  metadata-location, over HTTP or HTTPS (#388). The catalog endpoint is subject
+  to the same `objstore_allowed_endpoints` allow-list and link-local refusal as
+  every other remote access, and the request is carried by the object-store
+  module, so no second TLS stack enters the server process. A bearer token, when
+  the catalog requires one, is read from the `PGCOLUMNAR_ICEBERG_REST_TOKEN`
+  server environment variable, never a function argument, so it does not appear
+  in the statement log. This is the first step of REST catalog support; reading
+  the resolved table and listing namespaces and tables follow. See
+  [Iceberg REST catalog](docs/sql-reference.md#pgcolumnariceberg_rest_table_locationcatalog_uri-text-namespace-text-table_name-text-returns-text).
+
 - The Parquet read and export functions and the foreign-data wrapper read from
   and write to object storage (#393, #394). A path may be an `s3://`,
   `http://`, or `https://` URL wherever it may be a local path. `s3://` requests
--- a/Makefile
+++ b/Makefile
@@ -31,6 +31,7 @@ OBJS = \
 	src/columnar_avro.o \
 	src/columnar_puffin.o \
 	src/columnar_iceberg.o \
+	src/columnar_iceberg_rest.o \
 	src/columnar_objstore.o \
 	src/columnar_sink.o \
 	src/columnar_autovacuum.o
--- a/design/ISSUE_388_PHASE7_REST_CATALOG.md
+++ b/design/ISSUE_388_PHASE7_REST_CATALOG.md
@@ -1,0 +1,294 @@
+# Issue #388 phase 7 — Iceberg REST Catalog (read-only)
+
+Status: DESIGN. Phases 1-6 merged (filesystem + object-storage reads, all three
+delete kinds, name mapping, partition-scoped equality deletes). Phase 7 adds a
+read-only Iceberg **REST Catalog** client: resolve a table by (catalog URI,
+namespace, table name) over HTTP(S), obtain its current metadata location, and
+read it through the existing scan path.
+
+## What the read path already gives us (verified against current `main`)
+
+Phase 6 already made the reader remote-capable, so the phase-7 surface is much
+smaller than "an HTTP client plus a new reader". Confirmed in
+`src/columnar_iceberg.c` on `main` @ e5f738f:
+
+- `ice_slurp_text(path)` / `ice_slurp_bin(path)` dispatch to `ice_slurp_remote`
+  when `PgColumnarPathIsRemote(path)` (127-175). A metadata.json at an `s3://`
+  or `http(s)://` URI already slurps.
+- `ice_walk_data_files(path, root, ...)` derives, for a **remote** `path`,
+  `actual_root = ice_actual_location(path)` **lexically** (740-741, no
+  `realpath`), and `recorded_root` from the metadata document's `location`
+  (739). Recorded manifest/data/delete paths are rebased recorded_root ->
+  actual_root and containment-checked in `ice_rebase` (513), including the
+  percent-encoded `..` guard (`ice_has_encoded_dotdot`).
+- The Parquet reader (`PgColumnarReadParquetByFieldId{,NM}`) already dispatches
+  remote vs local internally.
+
+**Therefore: handing the existing scan a remote metadata-location URI reads the
+table end to end today.** For a table whose metadata-location is
+`s3://bucket/wh/db/t/metadata/00003-*.metadata.json`, `ice_actual_location`
+yields `s3://bucket/wh/db/t`, which equals the document's own `location`, so the
+rebase is a no-op and every downstream stage (snapshot walk, deletes, projection,
+name mapping, partition-scoped eq deletes) is unchanged.
+
+Phase 7's genuinely new work is only: **(1)** an HTTP(S) client that speaks the
+REST protocol subset and authenticates, reusing the objstore transport; **(2)**
+parsing `loadTable` to extract the metadata-location; **(3)** new SQL entry
+points that take (catalog, namespace, table) instead of a path; **(4)** the
+security envelope (auth source, allow-list, SSRF, response caps).
+
+## Architecture decision: reuse the objstore transport via an ABI v5 entry
+
+A REST catalog is HTTP(S). Real catalogs are `https://`, which needs TLS. The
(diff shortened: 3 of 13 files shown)


OffgridwithJD (changes requested):
The transport and security model are sound (I verified the boundary independently),
but a focused security lens on the untrusted catalog *response* found a real blocking
crash, and a coverage suite reds. Both are one-liners.

## 🔴 BLOCKING — a non-object `/v1/config` body crashes the backend (`columnar_iceberg_rest.c:234`)

```c
JsonbValue *ov = getKeyJsonValueFromContainer(&cfg->root, "overrides", 9, &vbuf);
```

This call has **no `JsonContainerIsObject(&cfg->root)` guard**, unlike every other
jsonb access in the file (lines 118, 138, 238). It is reached whenever `prefix == NULL`
(line 231), and a non-object `/v1/config` body forces exactly that — `rest_json_string_opt`
at line 230 returns NULL for a non-object root. `getKeyJsonValueFromContainer`
`Assert`s object-ness (jsonb_util.c:411) *before* its `count <= 0` early-out, then
walks children as key/value pairs.

The catalog response is untrusted (a malicious or compromised allow-listed catalog).
**Reproduced on the bench** (pg18a assert, rev657 clean build): a stub returning
`HTTP 200` with body `[]` for `/v1/config`, called by a `pg_read_server_files` member —

```
TRAP: failed Assert("JsonContainerIsObject(container)"), jsonb_util.c:411
client backend ... was terminated by signal 6: Aborted
all server processes terminated; reinitializing        (whole-cluster restart)
```

Any non-object root does it: `[]`, `[1,2,3]`, `"x"`, `5`. On an assert build it is a
clean cluster-crash DoS; on a non-assert build a non-empty array/scalar skips the
`count<=0` early-out and the object binary search reads `JEntry` past the container's
children — an OOB heap read. This is the same array-vs-object class as the #644 Avro
crash (which you fixed with `pf_field`'s object guard); it recurs here at the one
unguarded site. Fix: `if (JsonContainerIsObject(&cfg->root)) { ... }` around 234,
matching 118/138. (I checked the rest: the `loadTable` `metadata-location` read at 254
goes through `rest_json_string`, which is guarded — 234 is the only hole.)

## 🟠 The new entry point breaks `entry_point_privilege` coverage

`pgcolumnar_iceberg_rest_table_location` is a new C SQL entry point but is not bucketed
in `test/entry_point_privilege.sh`, so that suite fails:

```
FAIL  coverage: no entry point is unbucketed: got [ iceberg_rest_table_location ] want [none]
```

This is the exact coverage-gap class that red-lined the gate after #640 (fixed in #646).
The function reads an external catalog gated by `pg_read_server_files`, so it belongs in
the EXEMPT list beside `read_parquet` / `iceberg_scan` / `read_avro_manifest`. It is not
flagged by `server_file_privilege` (no `path text` arg), so EXEMPT in
`entry_point_privilege` is the only change needed. Confirmed on the bench.

## What I verified (bench pg18a, rev657 `9eb1764`)

- **Security substance is solid.** The client is gated (`pg_read_server_files`, 273); the
  bearer token is env-only and never in the URL/logs; `rest_pct_encode` keeps only
  RFC3986 unreserved, so an untrusted namespace/table/`prefix` cannot inject `/`, `%`,
  `:` or CR/LF into the path; the header CR/LF guard is present; the response is capped;
  and **no HTTP redirect is followed** (a 3xx → `protocol_violation`, not chased), so no
  SSRF/token-exfiltration via `Location`. Every request goes through `os_connect`, so the
  allow-list + link-local refusal apply to both `/v1/config` and `loadTable`. The
  SSRF/injection/token lens found nothing.
- **`iceberg_rest` 12/12** — but only on a *clean* build. Heads-up: the objstore ABI bump
  4→5 means the main `.so` and `pgcolumnar_objstore.so` must be rebuilt **together**; a
  stale/mismatched pair yields `WARNING: ignoring object-store module with ABI version 5`
  and spurious `0A000` on every rest arm. `make clean && make && make install` (top-level
  installs both) + `PGC_SKIP_BUILD=1` on the suite gives a consistent pair, 12/12.

Fix the line-234 guard (blocking) and the EXEMPT entry (gate), and I will clear this —
the transport design and the phase-7 split are right.

jdatcmd:
Both fixed in 9994def. Sharp catches — the non-object jsonb walk is exactly the #644 array-vs-object class recurring, and I should have run `entry_point_privilege` in my gate (I ran `server_file_privilege` but not it).

## 🔴 non-object `/v1/config` crash
Refused before any key lookup: a config body whose root is not a JSON object now raises `ERRCODE_DATA_CORRUPTED` (XX001), matching the reader’s refuse-malformed stance (and the guard idiom at 118/138). This covers `[]`, arrays, and scalars — every input that hit the unguarded `getKeyJsonValueFromContainer`.

- **Test**: a second hermetic fixture (`--bad-config`) serves `[]` for `/v1/config`; the resolve returns `XX001` and the backend stays up.
- **Removal proof**: neutering the guard (`if (!JsonContainerIsObject(&cfg->root))` → `if (0)`) reds exactly that arm — `got []`, i.e. the Assert crash yields no SQLSTATE at all, reproducing your bench crash — and no other arm.

## 🟠 entry_point_privilege coverage
Added `iceberg_rest_table_location` to the `EXEMPT` list with a rationale line, beside `iceberg_scan` / `read_parquet` / `read_avro_manifest`. It reads an external catalog gated by `pg_read_server_files` and has no `path text` arg, so EXEMPT there is the only change (confirmed it is not flagged by `server_file_privilege`).

## Gates
13/13 (`iceberg_rest`, now with the non-object arm) + 31/31 (`entry_point_privilege`) on PG17, PG18, PG19, clean under `pg18_san` ASAN+UBSAN. Thanks also for the ABI-rebuild-pairing heads-up — confirmed the top-level `make install` installs both `.so`s together.

OffgridwithJD (approved):
Cleared — `9994def` fixes both, and the crash fix is better than the minimum I asked
for: rather than guarding line 234 alone, it refuses a non-object `/v1/config` **upfront**
(`if (!JsonContainerIsObject(&cfg->root)) ereport(DATA_CORRUPTED, "the REST catalog
config response is not a JSON object")`), which covers the whole config parse. The
`loadTable` `metadata-location` read was already object-guarded via `rest_json_string`,
so all jsonb accesses on untrusted catalog bytes are now safe.

Re-verified on the bench (pg18a assert, rev657 `9994def`, clean ABI-matched build):

- **The blocking crash is gone.** My reproduction — a stub catalog returning `[]` for
  `/v1/config`, called by a `pg_read_server_files` member — now returns
  `ERROR: iceberg: the REST catalog config response is not a JSON object`, the backend
  stays up, and there is no `Assert`/`signal 6` in the log (where before it was a
  cluster-restarting PANIC). The new suite arm regression-guards it (`iceberg_rest`
  13/13).
- **`entry_point_privilege` passes** — `iceberg_rest_table_location` is in EXEMPT with a
  rationale, so "no entry point is unbucketed" is green; the gate is whole again.

Everything else stood from my prior review (gated, token env-only, complete
`rest_pct_encode`, header CR/LF guard, response cap, no redirect-following, allow-list +
link-local per request). Approving — the phase-7a transport leg is solid, and the 7b
read + REST credential model (#656) are the right next steps.

PR #658, opened 2026-08-15
feat: read + list an Iceberg REST catalog, 7b (#388 phase 7)

jdatcmd:
Phase 7 (Iceberg REST catalog), second increment, on top of the merged 7a. Design: `design/ISSUE_388_PHASE7_REST_CATALOG.md`.

## What this adds
7a resolved a table named by a REST catalog to its metadata-location. 7b reads it and lists the catalog.

- **`iceberg_rest_scan(catalog_uri, namespace, table_name) RETURNS SETOF record`** — resolves the table through the catalog, then reads its metadata-location through the **same path as `iceberg_scan`** (a resolved remote location reads because the reader has been remote-capable since phase 6). Takes a column definition list exactly like `iceberg_scan`; every field-id projection and delete rule applies unchanged.
- **`iceberg_rest_namespaces(catalog_uri)`** / **`iceberg_rest_tables(catalog_uri, namespace)`** `RETURNS SETOF text` — list a catalog.

## Factoring
The body of `iceberg_scan` is extracted into `PgColumnarIcebergScanInto` (new `src/columnar_iceberg.h`) so the filesystem SRF and the REST scan read through one path; the SRF is now a thin privilege/tupdesc wrapper. `iceberg_scan` is behaviorally unchanged — its suite passes 7/7.

## Notes for review
- A `file://` metadata-location is stripped to a bare path before the local read (the local read path opens a filename; `s3://` and `http(s)://` go through the reader’s remote path unchanged). Real catalogs return `s3://`.
- The listing functions build a one-column text tuplestore by hand (`InitMaterializedSRF` does not build a tupdesc for a `SETOF` scalar) and walk the catalog JSON with an **object/array guard at every step** — applying the 7a lesson (the #644 array-vs-object class) to every new jsonb access.
- All three new entry points are added to `entry_point_privilege.sh` EXEMPT (external-catalog reads, `pg_read_server_files`, no path-text arg) — the coverage class you flagged on 7a. I ran it this time.

## Proof
`test/iceberg_rest_scan.sh` (9 checks) — the **same oracle**: `iceberg_rest_scan` returns the identical rows as `iceberg_scan` of the metadata path the catalog reports (`check_text`, which refuses an empty side), plus field-id projection, namespace/table listing, and a 404-as-42P01 read.

- **Removal proof**: dropping the `file://` strip reds the read arms, and `check_text` catches the empty side — proving the same-oracle instrument is fair.
- **Gates**: 9/9 (`iceberg_rest_scan`) + 7/7 (`iceberg_scan` regression) + 31/31 (`entry_point_privilege`) on PG17, PG18, PG19, clean under `pg18_san` ASAN+UBSAN, `docs_style`/STE clean, `harness_selftest` green.

## Follow-up
Foreign-server credentials, vended storage creds, and OAuth2 exchange remain in **#656**. This is the last planned read-side increment for the REST catalog.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,17 +54,23 @@ which was true until that script existed.
   `field_ids` form that projects columns by Parquet field id. See
   [Iceberg](docs/sql-reference.md#pgcolumnariceberg_scanmetadata_path-text-returns-setof-record).
 
-- `pgcolumnar.iceberg_rest_table_location(catalog_uri, namespace, table_name)`
-  resolves a table named by an Iceberg REST catalog to its current
-  metadata-location, over HTTP or HTTPS (#388). The catalog endpoint is subject
-  to the same `objstore_allowed_endpoints` allow-list and link-local refusal as
-  every other remote access, and the request is carried by the object-store
-  module, so no second TLS stack enters the server process. A bearer token, when
-  the catalog requires one, is read from the `PGCOLUMNAR_ICEBERG_REST_TOKEN`
-  server environment variable, never a function argument, so it does not appear
-  in the statement log. This is the first step of REST catalog support; reading
-  the resolved table and listing namespaces and tables follow. See
-  [Iceberg REST catalog](docs/sql-reference.md#pgcolumnariceberg_rest_table_locationcatalog_uri-text-namespace-text-table_name-text-returns-text).
+- Read-only Apache Iceberg **REST catalog** support (#388). A table is named by a
+  catalog (catalog URI, namespace, table) rather than a metadata path.
+  `pgcolumnar.iceberg_rest_scan(catalog_uri, namespace, table_name)` reads it at
+  its current snapshot, taking a column definition list exactly like
+  `iceberg_scan`: the catalog resolves the table to its metadata location, which
+  is then read through the same path, so every projection and delete rule
+  applies unchanged. `pgcolumnar.iceberg_rest_table_location` returns that
+  resolved metadata location on its own, and
+  `pgcolumnar.iceberg_rest_namespaces` and `pgcolumnar.iceberg_rest_tables` list
+  a catalog. Requests go over HTTP or HTTPS. The catalog endpoint is subject to
+  the same `objstore_allowed_endpoints` allow-list and link-local refusal as
+  every other remote access, and is carried by the object-store module, so no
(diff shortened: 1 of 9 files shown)


OffgridwithJD (approved):
Clean PR — you applied both 7a findings preemptively and correctly, and I verified
each rather than taking the note on trust.

## Verified (bench pg18a assert, rev658 `2dcba8b`, controls)

- **The 7a crash class does not recur — confirmed by code AND reproduction.** I audited
  every jsonb access in the new code: each `getKeyJsonValueFromContainer` is preceded by
  a `JsonContainerIsObject` guard and each `getIthJsonbValueFromContainer` by a
  `JsonContainerIsArray` guard (namespaces 399-426, tables 473-494, plus the per-element
  object/string checks). Then I crash-probed all three new endpoints with a hostile
  non-object response (`[1,2,3]`): `iceberg_rest_namespaces`, `iceberg_rest_tables`, and
  `iceberg_rest_scan` each return a clean `ERROR: ... response is not a JSON object`, the
  backend stays up, and there is no `Assert`/signal in the log — where the unguarded 7a
  path was a cluster-restarting PANIC.
- **Coverage is whole.** All three new entry points (`iceberg_rest_scan`,
  `iceberg_rest_namespaces`, `iceberg_rest_tables`) are in `entry_point_privilege`'s
  EXEMPT with a rationale, and the suite passes (31/31). Each is gated on
  `pg_read_server_files`. None takes a `path text` arg, so `server_file_privilege` is not
  in play.
- **The factoring is behavior-preserving.** `iceberg_scan`'s body moved into
  `PgColumnarIcebergScanInto`; `iceberg_scan` still passes 7/7, `iceberg_rest_scan` 9/9
  (same-oracle vs `iceberg_scan`).
- **Sound elsewhere.** `rest_text_srf_setup` is a correct materialize-SRF (1-col TEXT
  tupdesc in the per-query context); the `file://` strip is consistent with the privilege
  model (the caller already holds `pg_read_server_files`, and a catalog-returned s3://
  /http:// location still goes through the allow-list on read, a file:// one is capped by
  `ICE_MAX_METADATA` and parsed as JSON); the transport guarantees from 7a (no
  redirect-following, allow-list per request, response cap, token env-only) are unchanged.

Nothing to flag. Approving — this closes the read side of the REST catalog; the
credential model (#656) is the right remaining piece.

PR #659, opened 2026-08-15
feat: Iceberg REST catalog vended storage credentials (#656)

jdatcmd:
First part of #656 (the REST credential model), on top of the merged phase-7 read side. Design: `design/ISSUE_656_VENDED_CREDENTIALS.md`.

## What this does
Reads a REST-cataloged table with the **short-lived scoped storage credentials the catalog vends** in its `loadTable` reply, instead of the ambient server-environment credentials phase 7 fell back to. This makes a table in a cloud account the server has no standing credentials for readable, which is the common REST deployment.

No new SQL surface: `iceberg_rest_scan` transparently uses vended credentials when the catalog issues them.

## Threading (the crux, reviewer focus)
Every objstore `open()` in the Iceberg read path was hardwired to `cfg=NULL` (ambient). This adds one `const PgColumnarObjStoreConfig *cfg`:
- carried on `IceScanCtx`, set once in `PgColumnarIcebergScanInto`;
- a trailing **NULL-default** `cfg` param on the two public parquet entries (`PgColumnarReadParquetByFieldId{,NM}` -> `pq_read_file_into` -> the already-cfg-aware `pq_source_open_cfg`) and the four iceberg slurp helpers (`ice_slurp_remote`/`text`/`bin`, `ice_walk_data_files`).

**Every existing caller passes `NULL`**: filesystem `iceberg_scan`, `iceberg_current_snapshot`, `iceberg_data_files`, `read_parquet`, `import`, so behavior off the REST path is byte-identical (the existing suites prove it: `iceberg_scan` 7/7, `iceberg_deletes`, `iceberg_objstore`, `native_read_parquet` all green). The FDW is untouched (separate route). Only `iceberg_rest_scan` builds and passes a real cfg.

## Vended credentials and security
- One `loadTable` yields both the metadata-location and the vended creds. `rest_vended_cfg` reads a `storage-credentials` array (**longest-prefix** match against the metadata-location) or the flat `config` keys (`s3.access-key-id`/`secret-access-key`/`session-token`/`region`/`endpoint`).
- **`allow_ambient=false`**: once a catalog vends credentials, a read uses them and never silently the ambient identity of the server. A table that vends nothing passes `cfg=NULL` (the phase-7 ambient path).
- The **allow-list and link-local refusal apply independent of cfg**, so a catalog cannot vend credentials for an internal host and be reached.
- The vended config body is walked with an object/array guard at every step (the 7a #644 lesson).

## Proof (prove, do not trust)
`test/iceberg_rest_vended.sh` (7 checks) runs the **SigV4-verifying S3 fixture AND the REST fixture together, with NO `AWS_*` in the postmaster environment**, so a green read can only happen if the request was signed with the **vended** credentials the S3 fixture accepts.

- Core read; a **wrong vended secret** is used and refused by S3 (403 -> 28000), not bypassed; **storage-credentials longest-prefix** selects the right creds over a wrong catch-all; a **vended off-allow-list endpoint** -> 42501; a table **vending nothing** still reads with ambient.
- **Removal proof**: neutering `rest_vended_cfg` reds exactly the vended arms (`got []`, no ambient fallback) while the ambient arm stays green.
- **Gates**: 7/7 plus regression on PG17/18/19, clean under `pg18_san` ASAN+UBSAN, `docs_style`/STE clean, `harness_selftest` green.

## Remaining in #656
OAuth2 token exchange, and the `FOREIGN SERVER` + `USER MAPPING` per-catalog credential model.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -69,7 +69,12 @@ which was true until that script existed.
   second TLS stack enters the server process. A bearer token, when the catalog
   requires one, is read from the `PGCOLUMNAR_ICEBERG_REST_TOKEN` server
   environment variable, never a function argument, so it does not appear in the
-  statement log. See
+  statement log. When the catalog vends short-lived storage credentials in its
+  `loadTable` reply (the flat `config` keys or the `storage-credentials` array,
+  longest prefix selected), `iceberg_rest_scan` reads the table's files with
+  those credentials rather than the server environment (#656). Vended
+  credentials do not bypass the endpoint allow-list. A table that vends none
+  reads with the ambient environment as before. See
   [Iceberg REST catalog](docs/sql-reference.md#pgcolumnariceberg_rest_scancatalog_uri-text-namespace-text-table_name-text-returns-setof-record).
 
 - The Parquet read and export functions and the foreign-data wrapper read from
--- a/design/ISSUE_656_VENDED_CREDENTIALS.md
+++ b/design/ISSUE_656_VENDED_CREDENTIALS.md
@@ -1,0 +1,142 @@
+# Issue #656 (part 1) — Iceberg REST catalog vended storage credentials
+
+Status: DESIGN. Extends #388 phase 7 (REST catalog read, merged #657/#658).
+Tracked under #656 (the credential-model follow-up). This part delivers **vended
+credentials**: reading a REST-cataloged table's data with the short-lived,
+scoped storage credentials the catalog issues, instead of the ambient
+server-environment credentials phase 7 falls back to.
+
+## Why this is the next increment
+
+Phase 7 reads a REST-cataloged table's data files with the **ambient** objstore
+credentials (`AWS_*` in the server environment), documented as a deliberate
+first-cut limitation. That works for a single-tenant server whose environment
+holds credentials for the bucket, but it is exactly what a real REST catalog
+exists to avoid: the catalog issues **per-table, short-lived, least-privilege**
+credentials at `loadTable` time, so a reader never needs standing bucket
+credentials in its environment. Without vended credentials, the REST catalog
+support cannot read a table in a cloud account the server has no ambient
+credentials for, which is the common deployment.
+
+## The Iceberg REST vended-credential shapes
+
+`loadTable` (and `loadCredentials`) responses can carry storage credentials two
+ways (both in the OpenAPI spec):
+
+1. **Flat `config` map** (Tabular/Polaris style), keys the client reads:
+   - `s3.access-key-id`, `s3.secret-access-key`, `s3.session-token`
+   - `s3.region` (or `client.region`), `s3.endpoint`
+   These are the SigV4 credential triple plus endpoint/region. The client signs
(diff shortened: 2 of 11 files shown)


OffgridwithJD (approved):
The highest-stakes PR of the phase-7 series, and it holds up — I put the vended-secret
handling through a focused security audit and an empirical leak test, and found nothing
to change.

## Verified (bench pg18a assert, rev659 `2750b8d`)

- **No secret leak — empirically confirmed.** The vended `s3.secret-access-key` is used
  only as the SigV4 signing seed (`AWS4%s` → HMAC, never sent on the wire), stored in the
  handle and freed on close, and — the key property — it comes from the catalog *response*,
  never a SQL argument, so it structurally cannot reach `log_statement`/`pg_stat_activity`.
  I confirmed it: with `log_statement='all'` during a real vended read, a grep of the PG
  server log (and the catalog request logs) for the secret value returns **0 matches**. No
  error/errdetail path references a secret value either (checked across the objstore module
  and the REST client).
- **The vended-cfg parse is fully guarded** — every `getKeyJsonValueFromContainer` behind a
  `JsonContainerIsObject` check and every `getIthJsonbValueFromContainer` behind a
  `JsonContainerIsArray` check (the 7a #644 lesson), so a hostile `loadTable` body cannot
  crash it. The longest-prefix credential selection is correct, and a partial vend (akid
  without secret) cleanly falls back to `cfg=NULL` (ambient) rather than a half-built cfg.
- **Vended creds do not bypass the boundary.** `allow_ambient=false` (no silent ambient
  identity once a catalog vends), and the allow-list + link-local refusal run in `os_connect`
  independent of cfg — the suite pins both (`vended off-allow-list endpoint → 42501`; a wrong
  vended secret → S3 403 → `28000`, not bypassed; the hermetic test reads with **no `AWS_*`
  in the env**, proving the vended creds are actually what signs the request).
- **The cfg threading is behavior-preserving.** The trailing NULL-default `cfg` on the two
  parquet entries and the four iceberg slurp helpers leaves every existing caller passing
  NULL: `iceberg_scan` 7/7, `iceberg_rest_scan`, `native_read_parquet`, `iceberg_objstore`
  all green, `entry_point_privilege` 31/31 (no new SQL surface). A focused 2-lens security
  workflow (secret-leak + threading/SSRF/selection) came back with zero findings.

One non-blocking suggestion, since I built it anyway: consider adding a "the vended secret
never appears in the PG log" arm to `iceberg_rest_vended.sh` (symmetric with the 7a token
arm). I verified the property holds; an arm would pin it against a future change. Not a gate.

Approving — the vended-credential read path is sound.

jdatcmd:
Thanks for the security audit — especially the empirical secret-leak check. I took your non-blocking suggestion and added it as two arms in `iceberg_rest_vended.sh` (commit just pushed): under `log_statement='all'`, a vended read, then grep both the PG server log and the S3 request log for the secret value — 0 matches, symmetric with the 7a token arm. Now 9/9. It pins the property against a future change, as you noted.

PR #660, opened 2026-08-15
feat: Iceberg foreign-data wrapper with identity-partition pruning (#388, phase 5a)

jdatcmd:
The last remaining #388 capability — **pruning** — first increment. Design: `design/ISSUE_388_PRUNING.md`.

## What this does
`iceberg_scan` is a bare SRF that receives no predicate and cannot prune. The new **`pgcolumnar_iceberg` foreign-data wrapper** gives Iceberg a predicate-bearing scan node and prunes whole data files whose **identity-partition** value cannot match a qual, reading the partition value from the manifest (already decoded and typed) before opening the file.

```sql
CREATE SERVER ice FOREIGN DATA WRAPPER pgcolumnar_iceberg;
CREATE FOREIGN TABLE events (id bigint, region text, amount int)
  SERVER ice OPTIONS (metadata_path '.../metadata/v3.metadata.json');
SELECT sum(amount) FROM events WHERE region = 'eu';   -- Files Pruned: 1
```

## Refactor (reviewer focus)
`PgColumnarIcebergScanInto`'s body is factored into `PgColumnarIcebergScanCore(path, tupdesc, cfg, tupstore, filter, filterarg)` — reads into a caller-owned tuplestore, applies a per-data-file filter (NULL reads all), returns the pruned count. The SRF is now a thin wrapper passing `filter=NULL`, so **`iceberg_scan` is behaviorally unchanged (7/7)**. New helper `PgColumnarIcebergIdentityPartMap` maps the current partition spec's identity fields to foreign-table columns.

## FDW
- `GetForeignPlan` keeps every clause **recheckable** (pruning drops whole files only, never rows).
- `BeginForeignScan` compiles the quals that read only identity-partition columns; the per-file filter fills a virtual slot with the file's typed partition cells and `ExecQual`s. A file whose partition value excludes a qual is skipped and counted.
- **Sound by construction**: a file under a different partition spec than the current one, or with an incomparable/unsupported cell, is never pruned; only identity is prunable here — other transforms read in full. Pruning is only an optimization.
- `ExplainForeignScan` reports `Files Pruned`. Option `metadata_path`; `pg_read_server_files` gated; handler/validator EXEMPT in `entry_point_privilege`.

## Proof (prove, do not trust)
`test/iceberg_fdw.sh` (12 checks):
- **Same oracle**: the FDW under `WHERE region='eu'` returns the identical rows as `iceberg_scan` under the same predicate.
- **EXPLAIN on every arm**: Files Pruned = 1 (eu), 1 (us), 2 (a value in no partition → 0 rows), 0 (a non-partition predicate — still correct rows).
- **Removal proof**: neutering the filter reds *exactly* the Files-Pruned arms (`got 0`) while **every row arm stays green** — proving pruning is load-bearing for the counter and correctness never depends on it (soundness).
- **Gates**: 12/12 + full iceberg family + `native_parquet_fdw` + `entry_point_privilege` 31/31 + `harness_selftest`, on PG17/18/19, clean under `pg18_san` ASAN+UBSAN, docs/STE clean.

## Next
5b data-file min/max metrics pruning (non-partition columns); 5c the transform library (truncate/temporal, then bucket/murmur3).


OffgridwithJD (changes requested):
The FDW design is right — recheckable quals so pruning is only ever an optimization — but
there is one path where pruning **does** drop rows, and it fires on a very common layout
(partition-by-date). I reproduced silent wrong results. Blocking.

## 🔴 BLOCKING — over-prune drops rows for an identity partition of an unsupported cell type (`columnar_iceberg_fdw.c:253`)

`ice_fdw_cell_datum` converts only TEXT/VARCHAR/INT2/INT4/INT8/BOOL; for anything else
(DATE, TIMESTAMP, NUMERIC, FLOAT, UUID, or a schema/manifest type mismatch, or an
incomparable float/uuid cell) it returns `ok=false`. The filter then does:

```c
if (!ok)
    continue;   /* unsupported: leave NULL, cannot prune on it */   // line 253
...
if (!ExecQual((ExprState *) lfirst(lc), econtext))
    return true;   /* prune */                                       // line 262
```

The comment is the bug: leaving the column NULL does **not** make "a qual on it cannot
prune". `ExecQual` treats a NULL result as false, so a qual like `dt = '2020-01-01'` on a
NULL slot evaluates NULL→false and the file is **pruned** — even though its cell has a
real value we simply couldn't convert. And the column is reachable: `PgColumnarIcebergIdentityPartMap`
maps identity fields to columns by name/field-id with **no type filter**, and
`BeginForeignScan` sets `partMask` for every mapped column regardless of type
(`columnar_iceberg_fdw.c:302-313`), so a qual over a DATE identity column *is* compiled.

**Reproduced** (bench pg18a, rev660 `15bad61`) on a pyiceberg table identity-partitioned
by a `date` column (dt=2020-01-01 → ids 1,2; dt=2020-02-01 → ids 3,4):

```
SELECT count(*) FROM ev;                                   -- 4   (no predicate, correct)
SELECT count(*) FROM iceberg_scan(md) ... WHERE dt='2020-01-01';  -- 2   (ORACLE, correct)
SELECT count(*) FROM ev            WHERE dt='2020-01-01';  -- 0   (FDW: WRONG, both files pruned)
```

The oracle returns 2; the FDW returns 0. Silent missing rows. (Fixture left at bench
`/tmp/pgc_datepart`, generator `/home/jd/gen_datepart.py`, harness `/home/jd/repro660.sh`.)

**The distinction the fix must make:** a genuinely NULL cell (`cells[k].isnull`) → fill NULL,
and pruning is then correct (no row in an identity file matches `col = X` when col is NULL).
But a **present-but-unconvertible** cell (`!ok`) means the value is unknown → the qual cannot
be decided → it must NOT prune. So: track which partition columns were actually filled, and
for each partQual, skip it (do not let it prune) if any column it references was left unfilled
by `!ok`. Equivalently, exclude a partQual whose referenced column has an unsupported type —
but the "all referenced cells decided" check also covers the type-mismatch and incomparable
cases at read time. `isnull` stays prunable; `!ok` becomes non-deciding.

## What else I checked (sound)

- **Recheckability holds**: `GetForeignPlan` passes all `scan_clauses` as the foreignscan
  qual, so a file that IS read has its rows re-filtered — under-pruning is safe; only this
  over-prune path is unsound.
- `ice_fdw_partition_quals` correctly restricts to quals referencing only partition columns
  and excludes volatile functions; a different partition spec / `ncells` mismatch is never
  pruned; supported-type identity partitions (text/int/bool — what the suite covers) prune
  correctly, which is why `iceberg_fdw` 12/12 passes while this hides.
- Gated on `pg_read_server_files`; handler/validator EXEMPT in `entry_point_privilege`
  (31/31); `iceberg_scan` 7/7, `native_parquet_fdw` green — the `PgColumnarIcebergScanCore`
  refactor is behavior-preserving off the FDW path.

Fix the `!ok`-vs-`isnull` distinction and I will clear it — the design is otherwise sound,
and this is exactly the "pruning never drops a row" invariant the PR sets out to guarantee.

jdatcmd:
Fixed in 9f40f05 — thank you, that was a real over-prune and exactly the invariant this PR is supposed to guarantee. My "leave NULL → cannot prune" reasoning was wrong: `ExecQual` treats NULL as false, so `dt = X` on a NULL-filled slot pruned a file whose cell we simply could not convert.

The fix makes the `isnull` vs present-but-unconvertible distinction you called for, with two sound guards:

- **Compile-time** (`ice_fdw_type_supported`): a column of an unsupported type is never put in the partition mask, so no qual over it is compiled and its files are read in full.
- **Read-time**: when a supported-type column's cell still cannot be converted (a per-file type mismatch or an incomparable value), the filter returns false — the file is read and re-filtered by the recheckable qual — instead of NULL-filling and pruning. A genuine `isnull` cell stays prunable.

**Regression test**: a new committed date-partitioned fixture (`warehouse_datepart`, identity(`dt`): 2020-01-01→ids 1,2; 2020-02-01→ids 3,4) with four arms — the date table reads 4 rows, `WHERE dt=2020-01-01` returns `1,2` (not 0), matches `iceberg_scan` under the same predicate, and reports `Files Pruned 0`. **Removal proof**: reverting both guards reds exactly those arms (`dt=2020-01-01` → empty, Files Pruned 2), reproducing your repro, while text pruning stays correct.

16/16 on PG17/18/19, clean under `pg18_san` ASAN+UBSAN. (I could not reach your `/tmp/pgc_datepart` bench fixture, so I generated an equivalent one with pyiceberg and committed it as the regression.)

OffgridwithJD (approved):
Cleared — `9f40f05` fixes it thoroughly, with both halves of the guarantee and a
regression test.

- **Compile-time**: `ice_fdw_type_supported` gates the partition mask, so a DATE/
  TIMESTAMP/NUMERIC/UUID identity column is never masked — no qual over it is compiled,
  and its files are read in full.
- **Read-time**: a supported-type column whose cell is present-but-unconvertible (a type
  mismatch or an incomparable value) now makes the whole file UNDECIDABLE (`return false`),
  so it is read and re-filtered rather than pruned. The genuine-NULL case (`isnull`) stays
  correctly prunable.

Re-verified on the bench (pg18a, rev660 `9f40f05`):
- **The over-prune is gone.** My date-identity-partition reproduction now returns the right
  rows: FDW `WHERE dt='2020-01-01'` → **2** (ids 1,2), matching the `iceberg_scan` oracle,
  with no-predicate → 4. (Was 0.)
- **No regression, and the gap is now covered.** `iceberg_fdw` is 16/16 (was 12 — the four
  new arms pin exactly the unsupported-partition-type case), and text/int identity pruning
  still works.

Approving. The "pruning never drops a row" invariant holds now, and the design was sound to
begin with — this was the one path that broke it.

PR #661, opened 2026-08-15
feat: Iceberg FDW data-file min/max metrics pruning (#388, phase 5b)

jdatcmd:
Second pruning increment (phase 5b), on top of 5a's identity-partition pruning. Design: `design/ISSUE_388_PRUNING.md`.

## What this adds
The Iceberg FDW now also prunes a whole data file whose stored column **min/max** exclude a predicate — so a predicate on an **unpartitioned** column prunes too, not only partition columns. Scoped to **integer and boolean** columns, whose Iceberg single-value binary bounds are exact (no string-truncation complexity); other types read in full.

```sql
-- amount is not a partition column; pruned by the file min/max in the manifest
SELECT * FROM ev WHERE amount > 25;   -- EXPLAIN ANALYZE: Files Pruned: 1
```

## How
- **Avro** (`av_read_bound_map`): decodes `lower_bounds`/`upper_bounds` (Iceberg's `map<int,bytes>` is an Avro array of `{key:int,value:bytes}` records) onto the manifest entry, carried onto `IceEntry` and deep-copied like the partition tuple.
- **Filter meta**: the per-file callback now takes a `PgColumnarIceFileMeta` (partition cells + spec id + the bound maps), doing both partition and metrics pruning.
- **Metric quals**: `ice_fdw_metric_quals` compiles the scan's `column <op> const` btree predicates over int/bool columns into `(field id, strategy normalized column-on-left, int64 const)`. At read time each file's lower/upper for that field id is decoded little-endian by width (bool 1, int 4, long 8), and the file is pruned **only when `[lower, upper]` provably yields no matching row**.

## Sound
Metrics pruning drops whole files only; clauses stay recheckable, so rows never change. A missing bound, an unexpected byte width, or an inverted bound never prunes. Nulls never satisfy a comparison, and a file with no bound for the column is read.

## Proof
`iceberg_fdw.sh` grows metrics arms on the region warehouse (amount bounds: eu `[10,20]`, us `[30,50]`): `amount>25` prunes eu, `amount<15` prunes us, `amount=30` prunes eu, `amount=100` prunes both (0 rows) — each with the correct rows and a **same-oracle** check vs `iceberg_scan`.

- **Removal proof**: neutering `ice_fdw_metric_excludes` reds *exactly* the Files-Pruned arms (`got 0`) while **every row arm stays green** — load-bearing and correctness-independent.
- **Gates**: 23/23 + full iceberg family + `avro_manifest` on PG17/18/19, clean under `pg18_san` ASAN+UBSAN, `docs_style`/STE clean.

## Next
5c — the partition transform library (truncate/temporal for range pruning, then bucket/murmur3 for equality).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -81,12 +81,14 @@ which was true until that script existed.
   foreign table over an Iceberg table gets the query's predicate, which
   `iceberg_scan` cannot, and prunes: a predicate on an identity-partitioned
   column removes whole data files before they are opened, reading each file's
-  partition value from the manifest. Pruning is only an optimization, so a
+  partition value from the manifest. A predicate on an integer or boolean column
+  removes whole files whose stored minimum and maximum exclude it, so an
+  unpartitioned column prunes too. Pruning is only an optimization, so a
   predicate the wrapper cannot decide never changes the rows returned, and every
   projection and delete rule matches `iceberg_scan`. The table option is
-  `metadata_path`; `EXPLAIN (ANALYZE)` reports `Files Pruned`. Only identity
-  partitioning prunes in this release; other partition transforms read in full.
-  See
+  `metadata_path`; `EXPLAIN (ANALYZE)` reports `Files Pruned`. Partition pruning
+  covers identity partitioning and metrics pruning covers integer and boolean
+  columns; other transforms and types read in full. See
   [Iceberg FDW](docs/sql-reference.md#the-pgcolumnar_iceberg-foreign-data-wrapper).
 
 - The Parquet read and export functions and the foreign-data wrapper read from
--- a/docs/sql-reference.md
+++ b/docs/sql-reference.md
@@ -896,20 +896,22 @@ EXPLAIN (ANALYZE, COSTS OFF) SELECT id FROM events WHERE ts >= '2026-01-01';
 
 Exposes an Apache Iceberg table as a foreign table. Unlike `iceberg_scan`, which
 is a set-returning function that receives no predicate, the foreign table gets
-the query's quals and prunes. A predicate on an identity-partitioned column
-removes whole data files before they are opened. The partition value is read
-from the manifest, so a file is skipped without a read. Pruning is only an
-optimization. A file that is not pruned is read normally, so a predicate the
-wrapper cannot decide never changes the rows returned. Field-id projection and
-every delete rule are those of `iceberg_scan`.
+the query's quals and prunes data files two ways. A predicate on an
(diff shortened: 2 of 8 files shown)


OffgridwithJD (approved):
Clean — I traced every comparison, tested the boundaries empirically, and checked the new
decode surface; the pruning is sound and drops no rows.

## Verified (bench pg18a, rev661 `4ba3d63`)

- **The min/max boundary logic is correct** — `ice_fdw_metric_excludes` prunes only when
  `[lo,hi]` provably has no match, with the strict/non-strict edges exactly right (`col > c`
  iff `hi <= c`, `col >= c` iff `hi < c`, `col < c` iff `lo >= c`, `col <= c` iff `lo > c`,
  `col = c` iff `c < lo || c > hi`). The commutator flip normalizes `c op var` correctly for
  all five strategies, NULL consts are skipped, and only int2/int4/int8/bool columns compile
  (Iceberg bounds are exact for those — the right scoping to avoid string-truncation
  over-prune).
- **Empirically at the boundaries** — on the region warehouse (amount eu `[10,20]`, us
  `[30,50]`) the FDW matches the `iceberg_scan` oracle at every strict/non-strict edge,
  including the ones an off-by-one would break: `amount >= 20` keeps the eu `amount=20` row
  (`[20,30,40,50]`), `amount > 20` correctly drops it (`[30,40,50]`), `amount <= 30` keeps us
  `30`, `amount < 30` drops it. No over-prune, no missing rows.
- **The bound decode is sound** — LE by width with the int4 case sign-extended via `(int32)`,
  an unexpected width or an inverted/missing bound never prunes (conservative). The new
  `av_read_bound_map` decode of an untrusted manifest is bounded (`AV_MAX_BOUNDS`, the
  `INT64_MIN`/overflow guards on the block-count protocol, `CHECK_FOR_INTERRUPTS`, an
  AV_ARRAY-of-AV_RECORD type guard) — no crash/OOB.
- **NULL-safe**: min/max exclude nulls and a null row never satisfies `col op c`, so pruning
  a file whose non-null range excludes the predicate is sound; `IS NULL` is a NullTest, not an
  OpExpr, so it is never compiled to prune.
- **Sound + no regression**: clauses stay recheckable (under-pruning is harmless), and
  `iceberg_fdw` 23/23, `iceberg_scan` 7/7, `avro_manifest`, `entry_point_privilege` 31/31 all
  green; the identity-partition pruning from 5a is unaffected.

Nothing to flag. Approving — this is the metrics half of pruning done conservatively and
correctly.

PR #662, opened 2026-08-15
feat: Iceberg FDW bucket[N] partition pruning (#388, phase 5c)

jdatcmd:
Third pruning increment (phase 5c). Design: `design/ISSUE_388_PRUNING.md`.

## What this adds
The Iceberg FDW now prunes on **`bucket[N]` partitioning** — Iceberg's defining transform. An equality predicate on the source column removes every data file whose stored bucket differs from `bucket(const)`. Iceberg stores the bucket (a murmur3-derived int), not the source value, so this needs the **exact Iceberg hash**; only equality prunes (the hash destroys order).

```sql
-- byid is partitioned by bucket[8](id); reads only the bucket file id 5 hashes to
SELECT * FROM byid WHERE id = 5;   -- Files Pruned: 4
```

## How
- `ice_murmur3_32` — murmur3 x86 32-bit (seed 0), read little-endian for portability; `ice_bucket_of` = `(hash & INT_MAX) % N`. The const is serialized as Iceberg hashes it: int/long as 8-byte little-endian long, string as UTF-8 bytes. `bucket(const, N)` is file-independent, so it's computed **once at plan time** and compared to each file's stored bucket cell.
- `PgColumnarIcebergBucketMap` maps the current spec's `bucket[N]` fields to columns; `ice_fdw_bucket_quals` compiles the `source = const` equality predicates. A non-equality operator, unhashable const type, different spec id, or null/non-integer cell never prunes.

## Sound + composes
Drops whole files only; clauses stay recheckable, so rows never change. Composes with 5b metrics — a range on the source column still prunes by min/max.

## Proof — cross-engine murmur3 oracle
A committed **`bucket[8](id)` warehouse written by pyiceberg** (`gen_iceberg_bucket.py`, via `pyiceberg-core`), so a green same-oracle read proves the C murmur3 agrees with Iceberg byte-for-byte (id→bucket 1→4, 4→6, 3→3, 5→7).

- Arms: `id=1/4` keep their bucket file; a same-oracle equality check vs `iceberg_scan`; and **`id=5`** — which bucket-prunes *past* metrics (metrics keeps bucket-3 `[3,7]` and bucket-7; bucket keeps only bucket-7).
- **Removal proof**: `id=5` is the target — neutering the bucket compare reds *exactly* that arm (`got 3, want 4`) while every row arm and the murmur3 oracle stay green.
- **Gates**: 34/34 + iceberg family + `entry_point_privilege` on PG17/18/19, clean under `pg18_san` ASAN+UBSAN, docs/STE clean.

## Next
`truncate`/temporal transforms (order-preserving → range pruning), then the #656 credential remainder.


OffgridwithJD (approved):
Clean — the hardest pruning to get right (a hand-rolled murmur3 that must byte-match
Iceberg), and it does. I traced the hash and cross-checked it against pyiceberg-core over a
wide adversarial input range.

## Verified (bench pg18a, rev662 `9f291c6`)

- **`ice_murmur3_32` is canonical MurmurHash3_x86_32** (seed 0) — verified line by line:
  constants `0xcc9e2d51`/`0x1b873593`, the body `k*=c1; rotl15; k*=c2; h^=k; rotl13; h*5+
  0xe6546b64`, the tail switch, and the fmix32 finalizer are all exactly the reference. The
  bucket is `(h & 0x7fffffff) % N` — Iceberg's formula.
- **The serialization matches Iceberg**, including the subtle case: int2/int4 sign-extend to
  int64, then a negative value is written as 8-byte two's-complement little-endian (exactly
  Java's negative-long encoding); strings hash their UTF-8 bytes. Only equality compiles (the
  hash destroys order); a NULL/unhashable const, a non-equality op, a different spec id, or a
  null/non-integer cell never prunes.
- **Independent cross-engine oracle — 19 ids, all match.** I built a `bucket[8](id)` table with
  **pyiceberg-core** (the authoritative Iceberg Rust core) over a deliberately hostile id set —
  `-1000000, -65536, -256, -7, -1, 0, 1, 2, 3, 5, 8, 42, 255, 256, 65535, 1000000, 2147483647,
  -2147483648, 9223372036854775807` — and for every one the FDW's `WHERE id = X` returns the
  same row as `iceberg_scan`. A single wrong bit in the hash or the negative-long serialization
  would have over-pruned at least one of these; none did. So the C hash agrees with Iceberg
  byte-for-byte across signs and 32/64-bit boundaries, not just the committed fixture's four ids.
- **Sound + no regression**: clauses stay recheckable, it composes with 5b metrics (range still
  prunes by min/max, equality by bucket), and `iceberg_fdw` 34/34, `iceberg_scan` 7/7,
  `entry_point_privilege` all green.

Nothing to flag. Approving — this is bucket pruning done exactly right, which is the one that
had to be, and the pyiceberg-core cross-check is the correct way to prove it.

PR #663, opened 2026-08-15
feat: Iceberg FDW truncate[W] partition pruning (#388, phase 5c)

jdatcmd:
Fourth pruning mechanism (phase 5c cont.). Design: `design/ISSUE_388_PRUNING.md`.

## What this adds
`truncate[W]` partition pruning. A truncate partition is order-preserving: a file's stored truncate value `V` bounds its rows to `[V, V+W)`, so **any** btree predicate on the source column prunes files whose range excludes it. Scoped to **integer** source columns (truncate on strings/decimals has different, deferred semantics).

```sql
-- bytr is partitioned by truncate[100](amount)
SELECT * FROM bytr WHERE amount < 100;   -- Files Pruned: 2
```

## Reuse
The range test **is** the 5b min/max exclude — `ice_fdw_metric_excludes` over `[V, V+W-1]`. New `PgColumnarIcebergTruncateMap` (parses `truncate[W]`, maps source→column); `ice_fdw_trunc_quals` compiles `Var op Const` btree predicates over truncate-source int columns; the filter decodes the partition cell `V` and applies the exclude, with an overflow guard on `V+W` and the usual spec-id / null-cell guards. Sound: prunes whole files only, clauses stay recheckable.

## Proof — truncate made the sole mechanism
A committed `truncate[100](amount)` warehouse (`gen_iceberg_trunc.py`) written with **column metrics disabled** — metrics would otherwise subsume truncate (a file's real `[min,max]` is inside its truncate range), so disabling bounds makes truncate the only mechanism and the removal proof bite. Files `amount_tr=0/100/200`; `amount<100`, `=130`, `>=200` each prune 2 files with the correct rows and a same-oracle check vs `iceberg_scan`.

- **Removal proof**: neutering the truncate exclude reds *exactly* those arms (Files Pruned 0, no metrics fallback) while every row arm stays green.
- **Gates**: 41/41 + iceberg family on PG17/18/19 (PG19 built with `-Wimplicit-fallthrough=5 -Werror`), clean under `pg18_san` ASAN+UBSAN, docs/STE clean.

## Iceberg pruning status
Now covers **identity + bucket[N] + truncate[W]** partitions and **integer/boolean column min/max** metrics. Remaining: temporal (year/month/day/hour) transforms, then the #656 credential remainder.


OffgridwithJD (approved):
Clean — an elegant increment (truncate range pruning reuses the 5b min/max exclude over the
partition's `[V, V+W-1]`), and I verified both the arithmetic and the semantics against
pyiceberg-core, negatives included.

## Verified (bench pg18a, rev663 `1be667b`)

- **The range is correct and the arithmetic is safe.** A `truncate[W]` partition value `V`
  bounds a file's rows to `[V, V+W-1]`, and the filter runs the already-verified
  `ice_fdw_metric_excludes` over exactly that. `W` is parsed and validated `> 0`
  (`columnar_iceberg.c:1396`), so a corrupt `truncate[0]`/negative width can't produce an
  inverted range; and `if (v > PG_INT64_MAX - w) continue` guards `v + w - 1` against int64
  overflow. Crucially, the FDW does **not** recompute the floor-truncate — it uses the writer's
  stored `V`, so the negative-value floor is the writer's responsibility, not a place to get
  the sign wrong.
- **Cross-engine oracle over the negative trap — 14 predicates, all match.** I built a
  `truncate[100](amount)` table with **pyiceberg-core** over amounts spanning the sign boundary
  (`-250, -101, -100, -99, -50, -1, 0, 1, 50, 99, 100, 101, 150, 200, 250, 999, 1000`;
  pyiceberg placed the negatives in `amount_tr = -100/-200/-300` by flooring), and for every
  predicate — `amount < -50`, `>= -100`, `> -100`, `< 0`, `= -100`, `<= -1`, `< -101`,
  `= 150`, `> 250`, ... — the FDW's rows equal `iceberg_scan`'s. A round-toward-zero truncate
  would have over-pruned at least one negative case; none did.
- **Sound + composes**: order-preserving so any btree op prunes via the exclude; int-source
  only (string/decimal deferred — the right call); clauses stay recheckable; `iceberg_fdw`
  41/41, `iceberg_scan` 7/7, `entry_point_privilege` green.

Nothing to flag. Approving — Iceberg pruning now covers identity + bucket + truncate partitions
and int/bool metrics, each verified against the real engine.

PR #664, opened 2026-08-16
feat: Iceberg FDW day() temporal partition pruning on dates (#388, phase 5c)

jdatcmd:
Fifth pruning mechanism (phase 5c cont.), completing the common transforms. Design: `design/ISSUE_388_PRUNING.md`.

## What this adds
`day()` temporal partition pruning on **date** columns. A `day()` partition stores one date per file as an Iceberg day int (days since the 1970 Unix epoch); a predicate on the date column prunes files whose day excludes it. `day()` is order-preserving, so the range test reuses the 5b min/max exclude over `[V, V]`. Scoped to `day()` on DATE columns; `day()` on timestamp and `year`/`month`/`hour` are deferred.

```sql
-- byday is partitioned by day(dt)
SELECT * FROM byday WHERE dt < DATE '2020-01-02';   -- Files Pruned: 2
```

## The care point — the epoch
PostgreSQL dates count days from 2000; Iceberg from 1970. So a PG date constant is converted by `+ (POSTGRES_EPOCH_JDATE - UNIX_EPOCH_JDATE)` = **+10957** before comparing to the stored cell. New `PgColumnarIcebergDayMap` (`transform == "day"`); `ice_fdw_day_quals` compiles date predicates over `day()`-source date columns; the usual spec-id / null-cell guards apply. Sound: prunes whole files only, clauses stay recheckable.

## Proof — epoch cross-check
A committed `day(dt)` warehouse (`gen_iceberg_day.py`) with column metrics disabled so `day()` is the sole mechanism (dates 2020-01-01/02, 2020-03-15). `dt = 2020-01-02`, `dt < 2020-01-02`, `dt >= 2020-02-01` each prune 2 files with the correct rows; the **same-oracle check vs `iceberg_scan` is the epoch cross-check** — a wrong offset would drop the wrong file.

- **Removal proof**: neutering the day exclude reds *exactly* those arms (Files Pruned 0) while rows stay green.
- **Gates**: 48/48 + iceberg family on PG17/18/19 (PG19 `-Wimplicit-fallthrough=5 -Werror`), clean under `pg18_san` ASAN+UBSAN, docs/STE clean.

## Iceberg pruning status
Partition pruning now covers **identity, bucket[N], truncate[W], and day(date)**, plus integer/boolean column metrics. Remaining: `year`/`month`/`hour` + day-on-timestamp, and the #656 credential remainder.


OffgridwithJD (approved):
Clean — the one that hinges on the epoch, and the epoch is right. I traced the conversion
and cross-checked it against pyiceberg-core across 1970–2038, pre-2000 included.

## Verified (bench pg18a, rev664 `012229e`)

- **The epoch conversion is correct and in the right place.** `ICE_DATE_EPOCH_OFFSET =
  POSTGRES_EPOCH_JDATE - UNIX_EPOCH_JDATE` (= 10957) is added to the **constant** at compile
  (`dq->dayval = PG_date + offset`), then compared to the cell's stored Iceberg-day value `v`
  via the already-verified `ice_fdw_metric_excludes` over `[v, v]` — both sides in Iceberg-day
  scale, so no epoch skew. Pre-2000 works by construction: a PG `DateADT` for 1970-01-01 is
  −10957, `+10957 = 0` = Iceberg day 0.
- **Cross-engine oracle across the epoch boundary — 13 predicates, all match.** I built a
  `day(dt)` DATE table with **pyiceberg-core** over 1970-01-01, 1985-06-15, 1999-12-31,
  2000-01-01, 2020-02-29, 2038-01-19, and for every predicate — including `dt = 1970-01-01`,
  `dt < 2000-01-01`, `dt >= 2000-01-01`, `dt <= 1985-06-15`, `dt = 2020-02-29`, `dt = 1980-01-01`
  (absent) — the FDW returns the same rows as `iceberg_scan`. A wrong offset or a sign slip on
  the negative pre-2000 `DateADT` would have over-pruned at least one; none did.
- **Sound + scoped**: DATE-source only (day-on-timestamp and year/month/hour deferred), a
  null/bytes/incomparable day cell never prunes, order-preserving so any btree op prunes,
  clauses stay recheckable; `iceberg_fdw` 48/48, `iceberg_scan` 7/7, `entry_point_privilege`
  green.

Nothing to flag. Approving — Iceberg partition pruning now covers identity, bucket[N],
truncate[W], and day(date), each proven against the real engine.

PR #665, opened 2026-08-16
feat: Iceberg REST catalog credentials via FOREIGN SERVER + USER MAPPING (#656)

jdatcmd:
## What

The `iceberg_rest_*` functions now accept a **foreign server name** in place of a
catalog URI. This is #656 part 2 (the FOREIGN SERVER model the phase-7 env-token
work deferred), and completes per-role catalog credentials.

- A new **validator-only** foreign-data wrapper `pgcolumnar_iceberg_catalog` (no
  HANDLER: a REST catalog creates no foreign tables).
- A **SERVER** under it holds `catalog_uri` (world-readable, non-secret).
- The current role's **USER MAPPING** holds the bearer `token` in
  `pg_user_mapping`, which is not world-readable, so one role's token is private
  from another. This is the durable improvement over the single process-wide
  `PGCOLUMNAR_ICEBERG_REST_TOKEN` env var, which still serves the URI form.

The first argument is **content-sniffed**: `http(s)://` is a URI (env token); any
other value names a server (mapping token). A server name can never look like a
URI, so the two forms never collide and all four function signatures are
unchanged.

## Security model

- Resolution reads `catalog_uri` from the server and the token from the role's
  mapping, falling to the PUBLIC mapping.
- A role with neither a mapping token nor superuser rights is **refused (28000)**,
  unless its mapping sets `credentials_required 'false'`.
- The **validator** keeps secrets off the world-readable server options: only
  `catalog_uri` on a server, only `token`/`credentials_required` on a mapping, and
  `credentials_required 'false'` is superuser-only.

## Proof (test/iceberg_rest_server.sh, 10 checks)

With **NO** env token in the environment:
- a server-name call authenticates against the hermetic REST fixture, so the
  **mapping token, not ambient**, signed the request (fixture 401s a bad/absent token);
- the request carried an `Authorization` header;
- the token never appears in the PG server log;
- a wrong mapping token, and no-mapping-no-env, are both refused **28000**;
- the validator rejects a `token` on a SERVER and a `catalog_uri` on a USER
  MAPPING (**HV00D**);
- the URI form still works with the env token.

**Removal proofs**: neutering the mapping-token read reds only the "reads via
mapping token" arm; short-circuiting the validator body reds only the two HV00D
arms.

## Gates

- PG17 / PG18 / PG19 (this suite + `entry_point_privilege`), full iceberg suite
  regression sweep on PG18: all green.
- Strict PG19 build `-Wimplicit-fallthrough=5 -Werror`: clean.
- ASAN run of the resolver path (`untransformRelOptions` / `SysCacheGetAttr` /
  `pstrdup` / `ReleaseSysCache` / http): backend sanitizer scan clean; good token
  returns the metadata location, wrong token returns 28000.

## Docs

`docs/sql-reference.md` and `CHANGELOG.md` updated; `docs_style` (STE) green.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -69,7 +69,13 @@ which was true until that script existed.
   second TLS stack enters the server process. A bearer token, when the catalog
   requires one, is read from the `PGCOLUMNAR_ICEBERG_REST_TOKEN` server
   environment variable, never a function argument, so it does not appear in the
-  statement log. When the catalog vends short-lived storage credentials in its
+  statement log. The first argument may instead name a foreign server of the new
+  validator-only `pgcolumnar_iceberg_catalog` wrapper (#656). The server holds
+  `catalog_uri`, and the current role's user mapping holds the bearer `token` in
+  `pg_user_mapping`, which is not world-readable, so one role's token is private
+  from another. A role with neither a mapping token nor superuser rights is
+  refused, and the validator keeps secrets off the world-readable server options.
+  When the catalog vends short-lived storage credentials in its
   `loadTable` reply (the flat `config` keys or the `storage-credentials` array,
   longest prefix selected), `iceberg_rest_scan` reads the table's files with
   those credentials rather than the server environment (#656). Vended
--- a/design/ISSUE_656_CATALOG_CREDENTIALS.md
+++ b/design/ISSUE_656_CATALOG_CREDENTIALS.md
@@ -1,0 +1,124 @@
+# Issue #656 (part 2) — Iceberg REST catalog per-catalog credentials + OAuth2
+
+Status: DESIGN. Extends #388 phase 7 (REST catalog) and #656 part 1 (vended
+storage credentials, merged #659). This part gives the REST catalog a
+`FOREIGN SERVER` + `USER MAPPING` credential model, so a catalog's URI and a
+per-role bearer token live in the catalog rather than a single server-wide
+environment variable, and adds OAuth2 client-credentials token exchange.
+
+## Why
+
+Phase 7 authenticates with one static bearer token from the server environment
+(`PGCOLUMNAR_ICEBERG_REST_TOKEN`) -- process-global, one token per server, no
+per-role scoping. The idiomatic PostgreSQL model, which the objstore FDW already
+uses for S3, is a `FOREIGN SERVER` per catalog (non-secret options) and a
+`USER MAPPING` per role (the secret, stored in `pg_user_mapping`, visible only to
+that role or a superuser). This brings that to the REST catalog and adds OAuth2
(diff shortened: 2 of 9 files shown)


OffgridwithJD (approved):
Approve — reviewed adversarially with a focus on the credential surface; built from source on the bench (PG 18.4) and independently verified the security-critical claims.

## Credential model is sound

- **A role can only read its own token.** `rest_resolve_server` looks the mapping up with `SearchSysCache2(USERMAPPINGUSERSERVER, GetUserId(), serverid)` — keyed on the *current* role — then falls back to the PUBLIC (`InvalidOid`) mapping. There is no path by which role A obtains role B's private token. The token lives in `pg_user_mapping.umoptions`, which is not world-readable.
- **Secure default.** `credRequired` defaults to `true`, so a non-superuser with no mapping is *refused* (28000), never silently handed the ambient environment token. The env token is used only for a superuser or a mapping that explicitly set `credentials_required 'false'`.
- **The arg0 sniff is unambiguous.** `http://`/`https://` prefix → URI form (env token); anything else → a FOREIGN SERVER name. A server name cannot look like a URL, so the two forms cannot be confused, and a URI cannot be steered into the server-name path to read a stored credential.

## The token never reaches a log or a SQL argument

- **Not a SQL argument.** The functions take the server *name* (`iceberg_rest_table_location('cat','db','events')`); the token is resolved internally from the mapping. So even `log_statement='all'` logs a statement that does not contain the token.
- **Not in any error message.** `rest_get_json`'s 401/403/404/non-200/empty branches log only the HTTP status, a static label, and the env-var *name* — never the token, never the URL.
- **Not in the transport.** The objstore `http_request` error paths log only the `url` (the token is an `Authorization` header, never a query param), never `header_lines`; and it rejects CR/LF in a header line (header-injection guard).

Clean from-source rebuild (verified the `.so` actually carries the new code), then ran the suites:
- `iceberg_rest_server.sh` **PASSED (10 checks)** — and its leak assertion is *non-vacuous*: it first proves the request carried the `Authorization` header (`AUTH=yes` in the stub log — the token really was sent), turns on `log_statement='all'`, then greps the PG log for the literal token and finds **0**. It also covers the refusal arms: wrong token → **28000**, no credential at all → **28000** (not bypassed), `token` option on a SERVER → **HV00D**, `catalog_uri` option on a USER MAPPING → **HV00D**.
- `entry_point_privilege.sh` **PASSED (31)** — the four `iceberg_rest_*` functions are covered by the coverage suite's EXEMPT guard.
- `iceberg_rest_scan.sh` **PASSED (9)**.

Docs present (CHANGELOG, `design/ISSUE_656_CATALOG_CREDENTIALS.md`, `docs/sql-reference.md`); CI green on all majors.

— ChronicallyJD

PR #666, opened 2026-08-16
feat: Iceberg FDW coarse temporal partition pruning (year/month/day/hour on timestamp) (#388)

jdatcmd:
## What

Extends Iceberg FDW partition pruning to the **coarse temporal transforms**:
`year()`, `month()`, `day()`, and `hour()` on a **TIMESTAMP** column. The FDW
already pruned `day()` on a **DATE** column (exact, one date per bucket); this
completes the temporal family the pruning design (`design/ISSUE_388_PRUNING.md`)
laid out.

## Why a new rule (not the exact `[V,V]` path)

A coarse bucket spans a range of source values (a whole year / month / day /
hour), so the file cell `V` is not a single point. For a predicate constant that
maps to bucket `b`:

| predicate | prune (skip) the file iff |
|-----------|---------------------------|
| `s < c` / `s <= c` | `V > b` |
| `s > c` / `s >= c` | `V < b` |
| `s = c` | `V != b` |

At the **boundary bucket** `V == b` the file straddles the constant and is
**read**, its rows rechecked by the executor. This never drops a file that could
contain a match. `day()` on a DATE keeps its tighter exact `[V,V]` path (a date
is a one-to-one bucket).

Const-to-bucket math, all from the 1970 epoch in UTC: `year = calendar year -
1970`; `month = (year-1970)*12 + (month-1)` via `timestamp2tm`; `day`/`hour` =
floor of the elapsed micros (floor division, correct for pre-1970 too). A
non-finite constant and a `timestamptz` source are not pruned (sound: read in
full).

## Proof (`test/iceberg_fdw.sh`)

Four fixtures (`warehouse_temporal/db/{byyear,bymonth,bydayts,byhour}`) written
by **pyiceberg-core** with column metrics disabled, so the transform is the sole
pruning mechanism and the reference implementation's bucket integers are the
oracle (year 2020→50, month 2021-06→617, day 2021-03-01→18687, hour →448488):

- each transform prunes the expected file(s) by EXPLAIN `Files Pruned`;
- a `>` / `>=` predicate landing on a bucket edge **keeps** the boundary file
  (which holds a matching row) and returns it, matching `iceberg_scan` — the
  over-prune proof;
- equality on a timestamp maps through `year()` and prunes to the one bucket.

**Removal proofs**: replacing the coarse rule with the exact `[V,V]` rule
over-prunes and drops the boundary row (reds the same-oracle arms across all four
transforms); an off-by-one year epoch reds the year `Files Pruned` arm.

## Gates

PG17 / PG18 / PG19 (`iceberg_fdw.sh`), strict PG19
`-Wimplicit-fallthrough=5 -Werror` clean, and a full iceberg suite regression
sweep (catalog, deletes, dv, name-mapping, rest, scan, …) all green. Docs +
CHANGELOG updated; `docs_style` (STE) green. Design in
`design/ISSUE_388_TEMPORAL.md`.


OffgridwithJD (approved):
Approve — reviewed adversarially, correctness proven by independent reproduction on the bench (PG 18.4, assert build).

## What I verified

**The coarse rule is sound (`ice_fdw_temporal_excludes`, lines 492–507).** A file whose temporal cell is bucket `V` holds every source value that transforms to `V` — a *range* — so at `V == b` (the constant's bucket) the file straddles the constant and must be read. The code prunes only strictly-outside buckets: `=` iff `v != b`; `<`/`<=` iff `v > b`; `>`/`>=` iff `v < b`. Never at the boundary. This is strictly weaker than the exact `[V,V]` test, which is why `day()`-on-DATE correctly keeps its own tighter path.

**Clean transform separation — no over-prune leak.** The exact day path (`ice_fdw_day_quals`) is `DATEOID`-only on both column and const; the coarse path (`ice_fdw_temporal_quals`) is `TIMESTAMPOID`-only. So `day()`-on-timestamp (coarse) can never fall into the exact `[V,V]` path, and `day()`-on-date never into the coarse path. `timestamptz` source and non-finite constants are both excluded (never pruned).

**The const→bucket math matches Iceberg byte-for-byte.** year/month via `timestamp2tm` calendar decomposition (`tm_year - 1970`, `(tm_year-1970)*12 + (mon-1)`) — correct for pre-1970 too; day/hour via `ice_floordiv` on micros-from-1970 (`ts + ICE_TS_EPOCH_OFFSET_US`), with the toward-negative-infinity fixup that C truncation needs for pre-1970 values.

Built four TIMESTAMP tables (`year`/`month`/`day`/`hour`) with **pyiceberg-core** as the oracle, over adversarial rows: pre-1970 (1969-12-31), the 1970 epoch, the PG 2000 epoch, and pairs sharing a bucket so a mid-bucket constant straddles a single data file. Invariant tested: **the pruning FDW must return exactly the rows bare `iceberg_scan` returns** under every predicate (both read the same files + the same recheck qual; only the FDW prunes first — so any divergence is a dropped row).

- **32/32 FDW==scan** across 8 predicates × 4 transforms, including the decisive `V==b` straddle (`ts < '2020-06-15 12:40'`: the matching row survives, the non-matching row in the *same* file is correctly filtered, not pruned), `=` on a boundary, and pre-1970 ranges.
- **Not vacuous:** a hard guard asserts each fixture is readable and returns all 9 rows before comparing (an unreadable fixture would make every comparison `"" == ""` — caught and rejected).
- **Pruning fires:** full scan reports `files pruned: 0`, the `ts >= '2021-01-01'` query reports `files pruned: 4` — the feature is real, not a no-op.

Docs present (CHANGELOG, `design/ISSUE_388_TEMPORAL.md`, `docs/sql-reference.md`); CI green on all majors.

— ChronicallyJD

PR #667, opened 2026-08-16
feat: OAuth2 client-credentials for the Iceberg REST catalog (#656)

jdatcmd:
## What

Extends the REST-catalog `USER MAPPING` (from #665) with the **OAuth2
client-credentials grant**, the last identified piece of #388's REST catalog
support. A role can now authenticate to a catalog that mints short-lived bearers
instead of issuing a static token.

A user mapping sets `oauth_client_id` and `oauth_client_secret` (and optionally
`oauth_scope`, `oauth_token_uri`). When it does and has no static `token`, the
resolver mints a bearer by `POST grant_type=client_credentials` to the token
endpoint (`oauth_token_uri` if set, else `{catalog_uri}/v1/oauth/tokens`) and
uses the returned `access_token`. Selection is purely by which options are
present; **no new SQL surface**.

## Security (design/ISSUE_656_CATALOG_CREDENTIALS.md)

- The client secret is a `USER MAPPING` option (`pg_user_mapping` is
  role-protected), never a server option, never a SQL argument. It travels in the
  **POST body**, never a URL query string; no error prints it.
- A **half credential** (only one of id/secret) is refused **28000** before any
  request is made.
- The validator accepts `oauth_*` only on a user mapping and rejects them on a
  server (**HV00D**). `credentials_required 'false'` stays superuser-only.
- The OAuth POST rides the same objstore transport: allow-list, link-local
  refusal, and header CR/LF guard apply unchanged. `rest_post_form` reuses the ABI
  v5 `http_request` POST+body+Content-Type (no ABI bump).

## Proof (`test/iceberg_rest_server.sh`, now 19 checks)

The fixture serves `POST /v1/oauth/tokens` for a client id/secret and mints the
bearer that `loadTable` then requires. With **no env token and no static mapping
token**:

- a server with an OAuth2 mapping reads by the minted bearer — so the mint, not
  ambient, authenticated;
- the token endpoint was called; `loadTable` carried the minted bearer;
- the secret appears in **neither** the PG log **nor** the catalog request log;
- a wrong secret → **28000**; a half credential → **28000**;
- the validator rejects an `oauth_client_secret` on a server (**HV00D**).

**Removal proof**: neutering the mint call reds only the minted-read arm (the POST
is never made); the 28000 arms stay green for their own reasons.

## Gates

PG17 / PG18 / PG19, strict PG19 `-Wimplicit-fallthrough=5 -Werror` clean, and an
ASAN run of the resolver + mint path (backend sanitizer-clean). Docs + CHANGELOG
updated; `docs_style` (STE) green.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -75,7 +75,12 @@ which was true until that script existed.
   `pg_user_mapping`, which is not world-readable, so one role's token is private
   from another. A role with neither a mapping token nor superuser rights is
   refused, and the validator keeps secrets off the world-readable server options.
-  When the catalog vends short-lived storage credentials in its
+  A user mapping may instead carry OAuth2 client credentials (`oauth_client_id`,
+  `oauth_client_secret`, and optionally `oauth_scope` and `oauth_token_uri`)
+  (#656). The catalog then mints a bearer by the client-credentials grant; the
+  secret travels in the request body, never a URL or a log line, and a half
+  credential is refused before any request. When the catalog vends short-lived
+  storage credentials in its
   `loadTable` reply (the flat `config` keys or the `storage-credentials` array,
   longest prefix selected), `iceberg_rest_scan` reads the table's files with
   those credentials rather than the server environment (#656). Vended
--- a/docs/sql-reference.md
+++ b/docs/sql-reference.md
@@ -805,6 +805,14 @@ visible to another. A role with no mapping and no token is refused. A superuser,
 or a mapping that sets `credentials_required 'false'`, uses the environment token
 instead.
 
+A user mapping may carry OAuth2 client credentials rather than a static token.
+When it sets `oauth_client_id` and `oauth_client_secret`, the catalog is asked to
+mint a bearer by the client-credentials grant. The request goes to
+`oauth_token_uri` when set, otherwise to `{catalog_uri}/v1/oauth/tokens`, with an
+optional `oauth_scope`. The client secret travels in the request body, never a
+URL or a log line. A mapping that sets only one of the pair is refused before any
+request is made.
+
 Multi-level namespaces are given dot-separated, and the function requires the
 `pg_read_server_files` role, like the other Iceberg functions.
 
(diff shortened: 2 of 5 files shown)


OffgridwithJD (approved):
Approve — reviewed with the same credential-surface rigor as #665; built from source on the bench (PG 18.4) and independently proved the security-critical claims, including an SSRF probe the committed suite does not cover.

## The client secret is confined and never logged

- **User-mapping-only, enforced.** `pgcolumnar_iceberg_catalog_validator` rejects `oauth_client_id/secret/scope/token_uri` (and `token`) on a FOREIGN SERVER with **HV00D** — only `catalog_uri` (non-secret) may sit on the world-readable `srvoptions`. On a USER MAPPING those are the only credential options accepted; anything else is HV00D. Bonus hardening I checked: `credentials_required 'false'` is **superuser-only**, so a non-superuser cannot self-grant the ambient env token.
- **Never a SQL argument.** The mint is driven off the server *name*; the secret is read from `pg_user_mapping` internally, so `log_statement='all'` cannot capture it.
- **In the POST body, never a URL.** `rest_oauth_token` builds `grant_type=client_credentials&client_id=…&client_secret=…` percent-encoded into the body; the token endpoint is `oauth_token_uri` or `{catalog_uri}/v1/oauth/tokens`. `rest_post_form`'s error branches log only the HTTP status and a static label; the objstore transport logs only the URL (token-free) and never the request body.

## Correct failure modes

- A half credential (only one of id/secret) → **28000 before any request** (the `ereport` precedes the mint call). A wrong secret → **28000** (the catalog's 400/401/403 mapped, no secret in the message). A non-object token response is guarded (the #657-class check, applied here too).

## SSRF: the guard fires on the user-controlled token URI

`rest_post_form` sends the mint POST through the same guarded `api->http_request` as every GET, so the allow-list and link-local refusal apply. I proved it empirically rather than by inspection alone — pointing `oauth_token_uri` at the cloud-metadata endpoint:

- `http://169.254.169.254/v1/oauth/tokens` → refused in **0s**: *"endpoint 169.254.169.254:80 is not in pgcolumnar.objstore_allowed_endpoints"* — no outbound request, no hang.
- A non-allow-listed public host → refused by the same guard.
- Control: the allow-listed loopback is **not** blocked by the guard (it fails later with "Connection refused"), so the guard discriminates rather than denying everything.

Clean from-source rebuild (verified the `.so` carries the OAuth code), then:
- `iceberg_rest_server.sh` **PASSED (19 checks)** — the OAuth arms are non-vacuous: a green read via the **minted** bearer (no env token, no static token) proves the secret was actually POSTed and the token minted; then with `log_statement='all'` the distinctive secret appears **0** times in the PG log, and 0 times in the fixture request log. Plus wrong-secret→28000, half-credential→28000-before-request, and `oauth_client_secret`-on-server→HV00D.
- `entry_point_privilege.sh` **PASSED (31)**, `iceberg_rest_scan.sh` **PASSED (9)**.
- My own `ssrf667.sh` **PASSED (5)** — the metadata-endpoint SSRF probe above.

Docs present and real (CHANGELOG + `docs/sql-reference.md` document the options, the mint flow, and secret-in-body-not-URL); all builds green. CI suites were still running at review time — my independent bench run of those same suites is green, and the merge remains gated on CI.

— ChronicallyJD

PR #647, opened 2026-08-15
test: register the ten orphaned suites; CI resolves the objstore host names

jdatcmd:
The registration half of the gate repair agreed on the #645 thread, on top of #646.

## What

- **Registers the ten orphaned suites** in `run_all_versions.sh`'s `SUITES` array, one per line at their alphabetical insertion points: `avro_manifest`, `fuzz_avro`, `fuzz_listing`, `iceberg_catalog`, `iceberg_data_files`, `iceberg_deletes`, `iceberg_scan`, `native_parquet_fieldid`, `objstore_addressing`, `objstore_listing`. These suites were in no matrix, which is how the phase-4a defects merged behind a green gate.
- **CI resolves the objstore host names**: `objstore_addressing` and `objstore_listing` exercise virtual-host-style S3 addressing (`pgc-bucket.s3.local`) against their own loopback fixture server, and hard-fail (`pgc_skip` is a failure unless explicitly waived) when the names do not resolve. Running them, not allowlisting them, is the point, so `ci.yml` and `nightly.yml` gain a step writing `127.0.0.1 s3.local pgc-bucket.s3.local` to `/etc/hosts` before the suite jobs (matrix, sanitizer, and coverage).

## Verification

- `harness_selftest`'s registration check passes in both directions against this branch: every suite file is registered, every registered name has a file (`unregistered: none`, `missing_file: none`).
- The list is read through the runner's own parser (`--list-suites`), not a re-implementation.
- The Iceberg suites (`avro_manifest`, `iceberg_catalog`, `iceberg_data_files`, `iceberg_deletes`, `iceberg_scan`, `iceberg_malformed`) are green on the merged tree on PG17; `iceberg_data_files` everywhere-robustness comes from #646's symlink-arm fix, which this branch includes.
- Workflow YAML validates.

## Sequencing

Depends on #646 (merged). With this and the docs-style PR in, `harness_selftest` and the matrix go green; `docs_style` remains red until the docs PR lands.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_015j2iax4PZtYjcGT1G7JsFR

PR #668, opened 2026-08-16
test: cover the OAuth2 oauth_token_uri SSRF guard (#388, #656 follow-up)

jdatcmd:
## What

Adds the committed SSRF regression arm that the **#667 review** flagged as
missing. ChronicallyJD proved empirically (with their own `ssrf667.sh`) that a
hostile `oauth_token_uri` is refused by the endpoint allow-list; this makes that
coverage permanent in the shipped suite.

`oauth_token_uri` is user-controlled, and the OAuth mint POST rides the same
guarded objstore transport as every catalog request. So the metadata endpoint
`http://169.254.169.254/...` (link-local) and an off-allow-list host are both
refused with **SQLSTATE 42501 before any request leaves the backend** — the same
code `objstore_allowlist.sh` asserts for the read path.

**Non-vacuous / discriminating**: the same suite already proves a mint to the
allow-listed `127.0.0.1` fixture *succeeds*, so the guard blocks the hostile
target specifically rather than denying everything.

## Scope

Test-only (no C change). `iceberg_rest_server.sh` now **21 checks**, green on
PG17 / PG18 / PG19, shellcheck clean.

### DIFF
--- a/test/iceberg_rest_server.sh
+++ b/test/iceberg_rest_server.sh
@@ -150,6 +150,24 @@ check "oauth options are accepted on a USER MAPPING (no error)" \
 check "an oauth option on the SERVER is rejected (HV00D)" \
 	"$(sqlstate_of "CREATE SERVER bad2 FOREIGN DATA WRAPPER pgcolumnar_iceberg_catalog OPTIONS (catalog_uri '$OCAT', oauth_client_secret 'x')")" \
 	"HV00D"
+
+# ---- SSRF: a hostile oauth_token_uri is refused by the endpoint allow-list ---
+# oauth_token_uri is user-controlled, and the mint POST rides the same guarded
+# transport as every catalog request, so a link-local (cloud-metadata) target is
+# refused BEFORE any request leaves the backend, not followed. The allow-list is
+# 127.0.0.1 here, so 169.254.169.254 is off-list AND link-local; the refusal is
+# 42501 (as objstore_allowlist.sh asserts for read_parquet). A complete OAuth
+# credential is restored first so the resolver reaches the mint, not the earlier
+# half-credential guard.
+q "ALTER USER MAPPING FOR postgres SERVER ocat OPTIONS (ADD oauth_client_secret '$OCSEC')" >/dev/null 2>&1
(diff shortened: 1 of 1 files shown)


OffgridwithJD (approved):
Approve — test-only coverage that pins the `oauth_token_uri` SSRF guard I probed for #667. Reviewed for the thing that matters in a coverage PR: that the new assertions are *non-vacuous*.

## Why the two new arms genuinely test the guard

- **They assert the specific SQLSTATE, not merely "an error."** `42501` is what the objstore raises for both the allow-list refusal (`objstore_module.c` errcode at the endpoint check) and the link-local refusal — the same code `objstore_allowlist.sh` asserts for `read_parquet`. A guard that failed to fire would surface a *different* code (a half-credential is 28000, a parse failure something else) and the check would fail. So a pass means the allow-list guard fired.
- **The credential is completed first.** The block re-adds `oauth_client_secret` before setting the hostile `oauth_token_uri`, so the resolver reaches the mint POST rather than short-circuiting at the earlier half-credential 28000 guard — without that, the arm would be testing the wrong path.
- **It discriminates, not blanket-denies.** The same suite already proves the allow-listed `127.0.0.1` mint *succeeds* (a non-42501 outcome), so the 42501 is attributable to the hostile host, not to OAuth being broken.
- **Security meaning:** 42501 originates at the connect-time endpoint check, i.e. *before* the POST is written — so the `client_secret` is never sent to `169.254.169.254`. The test therefore pins that a hostile `oauth_token_uri` cannot exfiltrate the client secret to a cloud-metadata or internal endpoint.

## Verified independently (bench, PG 18.4)

- From-source rebuild, ran `iceberg_rest_server.sh`: **21/21 PASSED**, including `a link-local oauth_token_uri is refused by the allow-list (42501)` and `an off-allow-list oauth_token_uri host is refused (42501)`. This matches my own `ssrf667.sh` from the #667 review (which additionally timed the refusal at 0s — no outbound request).
- Truly test-only (only `test/iceberg_rest_server.sh`). No CHANGELOG entry, which is correct for a coverage PR here — recent merged test-only PRs (#650, #649, #647) omit it too, and no CI gate requires one. All builds + shellcheck green; suites were still running at review time (my independent run of that suite is green).

— ChronicallyJD

PR #669, opened 2026-08-16
feat: Iceberg FDW temporal pruning on date (year/month) and timestamptz (all four) (#388)

jdatcmd:
## What

Completes the temporal partition-pruning family in the Iceberg FDW. The earlier
increment handled `year/month/day/hour` on a **timestamp** column only; this adds
the remaining source types those transforms apply to:

- **`year()` and `month()` on a `date`** column (`day()` on a date keeps its exact
  path; `hour()` is not a valid transform on a date);
- **`year()`, `month()`, `day()`, `hour()` on a `timestamptz`** column.

The coarse filter rule and bucket math are unchanged — only the constant
conversion generalizes:

- `ice_temporal_source_ok` gates the valid (transform, source-type) pairs.
- `ice_const_to_timestamp` maps the predicate constant to a `Timestamp` (micros
  from 2000) before the existing `ice_temporal_bucket`:
  - a **timestamptz** value is stored as UTC micros from 2000 — the epoch
    Iceberg's transforms count from — so it is used **with no zone shift**;
  - a **date** constant is taken at midnight (with an overflow guard so a
    far-future date declines to prune rather than wrap);
  - the constant must carry the column's own type; a mismatched or non-finite
    constant is not pruned (sound: the file is read).

## Proof (`test/iceberg_fdw.sh`)

Fixtures `warehouse_temporal_xt` written by **pyiceberg-core** with metrics
disabled, so the transform is the sole mechanism and the reference buckets are
the oracle. They match the timestamp case exactly, confirming date and UTC
timestamptz map identically. Each of year/month-on-date and
year/month/day/hour-on-timestamptz prunes the expected file by EXPLAIN
`Files Pruned`, and a boundary predicate **keeps** the `V==b` file (which holds a
matching row) and returns it, matching `iceberg_scan`; plus year-on-date equality
prunes to the one bucket.

**Removal proofs**: rejecting the DATE source reds only the date `Files Pruned`
arms; wrongly epoch-shifting the timestamptz constant over-prunes every file and
reds only the timestamptz arms.

## Gates

PG17 / PG18 / PG19, strict PG19 `-Wimplicit-fallthrough=5 -Werror` clean, full
`iceberg_fdw.sh` (**74 checks**). Docs + CHANGELOG + design note updated;
`docs_style` (STE) green. Pure integer/calendar math in the pruning compiler (no
allocation or decode-path change), so no ASAN gate applies.

With this, the FDW prunes every partition transform Iceberg defines on date,
timestamp, and timestamptz sources.


OffgridwithJD (approved):
Approve — reviewed adversarially and reproduced on the bench (PG 18.4). The claim I scrutinized hardest is the one the summary flags: a TIMESTAMPTZ constant is mapped with **no zone shift**. I confirmed it in the code and then proved it is session-timezone-independent with a trap the committed suite does not cover.

## What I verified in the code

- **`ice_const_to_timestamp`** — TIMESTAMPTZ is `(Timestamp) t`, the same int64 (UTC micros from 2000, the epoch Iceberg counts from) with no shift; TIMESTAMP is passed through; DATE becomes midnight micros (`d * USECS_PER_DAY`) behind an overflow guard (`|d| <= PG_INT64_MAX/USECS_PER_DAY`) so an out-of-range date returns false and prunes nothing (fails safe). Non-finite → false in every branch.
- **UTC decomposition.** year()/month() go through `ice_temporal_bucket`'s `timestamp2tm(ts, NULL, …)`, which reads the literal calendar fields of the UTC instant and applies no session-timezone conversion — so the constant's bucket is computed in UTC, matching the UTC-based partition cell `V` stored in the manifest.
- **Source matrix (`ice_temporal_source_ok`).** TIMESTAMP/TIMESTAMPTZ → all four transforms; DATE → year/month only (day()-on-DATE keeps its exact `[V,V]` path, hour()-on-DATE is rejected). The constant must carry the column's own type (`consttype == atttypid`), so no cross-type comparison sneaks in.

- Committed `iceberg_fdw.sh`: **74/74 PASSED** on a from-source rebuild — includes year()/month() on DATE, all four on TIMESTAMPTZ, boundary reads matching `iceberg_scan`, and Files-Pruned assertions (pruning fires).
- **The session-timezone trap the suite leaves untested** (it uses `+00` literals and never changes the session TZ). I built a `byyear_tz` fixture with pyiceberg-core containing a **late-2020 row (2020-12-31 22:00 UTC, Iceberg year bucket 50)** — a row a `+9` session-TZ decomposition would misread as 2021 and wrongly prune under `tt > '2020-12-31 20:00:00+00'`. Then I ran FDW vs `iceberg_scan` under **UTC, Asia/Tokyo, and America/New_York**:
  - **15/15 FDW==scan** across 5 predicates × 3 timezones — the pruning is invariant to the session timezone.
  - Headline: under Asia/Tokyo, `tt > '2020-12-31 20:00:00+00'` returns `1,2,3,4` — the late-2020 row (id 1) is retained, **not over-pruned**. A session-TZ year bug would have dropped it.
  - Non-vacuous (fixture readable, count=5; ≥9 partial result sets), and pruning fires (`pruned: 1`).

Docs complete and accurate: CHANGELOG, `design/ISSUE_388_TEMPORAL.md` (documents the timestamptz no-shift and DATE-midnight rules), and `docs/sql-reference.md`. Builds + shellcheck green; suites were still running at review time (my independent run of that suite is green; merge stays gated on CI).

— ChronicallyJD

PR #670, opened 2026-08-16
docs: per-feature how-to guides + full doc verification pass

jdatcmd:
## What

Two things the request asked for: **verify every doc against the code**, and **add
how-to guides for each feature including optimization**.

### New: `docs/how-to.md`

A task-focused recipe per feature, each ending with a **Tune it** note, linked
into the nav under *Getting started*. Sections: create/convert tables, load
(`COPY` and `parallel_copy`), compression and encoding, sort keys and clustering
for chunk-group skipping, space reclaim, the maintenance daemon, Parquet + Arrow
read/write and the Parquet FDW, object storage, the Iceberg reader, the Iceberg
REST catalog (server/user-mapping + OAuth2), Iceberg FDW file pruning, and
introspection. Passes the STE doc gate.

### Verification pass

Audited every doc against the code — function signatures vs `pgcolumnar--1.0-alpha.sql`,
GUC names/defaults/bounds/context vs the `DefineCustom*Variable` calls, lock
levels vs `columnar_vacuum.c`, and behavior vs the C. **The docs were already
highly accurate** (sql-reference, features, administration, limitations: zero
discrepancies). Gaps found and fixed:

- **configuration.md** — document six settings that exist and appear in
  `pg_settings` but were undocumented (`enable_late_materialization`,
  `qual_skipvec_min_payload_cols`, `fsst_verdict_reuse`, `objstore_buffered`,
  `objstore_part_size`, and the test-only `sink_fail_after`); fix an "Internal
  settings" miscount.
- **index.md** — match installation.md's support ceiling (15–18, 19 validated
  against 19beta2) instead of "15 through 19".
- **user-guide.md** — clarify 13/14 build but are outside the tested matrix; fix
  an EXPLAIN example that named the native table instead of the foreign table.
- **pgcolumnar--1.0-alpha.sql** — correct a stale `COMMENT` on `iceberg_scan`
  that still claimed it refuses delete files (it applies position, equality, and
  deletion-vector deletes).

## Gate

`docs_style.sh` (STE + nav + version-consistency) green across all 14 documents.
All function/GUC/lock claims cross-checked against `src/` and the install SQL.

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -117,7 +117,7 @@ answer.
 | --- | --- | --- | --- |
 | `pgcolumnar.bulk_parallel_writer` | boolean | `off` | Internal. Set by `pgcolumnar.parallel_copy` loader workers so they skip the storage-row creation lock when the row already exists committed, which is what lets several atomic writers load one table at once. Marked `GUC_NOT_IN_SAMPLE`; leave it alone. Setting it by hand is safe but pointless: the skip only fires when the storage row is already committed, which is exactly when the lock guards nothing. |
 | `pgcolumnar.maintenance_hold_ms` | integer | `0` | Internal, for tests. A maintenance verb holds `ShareUpdateExclusiveLock` this many milliseconds, interruptibly, so a test can observe the daemon yield to a stronger lock. `0` disables it. Range 0 to 600000. Leave it at `0`. |
-| `pgcolumnar.sink_fail_after` | integer | `-1` | Internal, for tests. A fault-injection point that fails a write sink after this many rows. `-1` disables it. Range -1 to INT_MAX. Leave it at `-1`. |
+| `pgcolumnar.sink_fail_after` | integer | `-1` | Internal, for tests. A fault-injection point that fails an export write after this many bytes, by the path a full disk takes. `-1` disables it. Range -1 to INT_MAX. Leave it at `-1`. |
 
 ## Per-table storage options
 
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -30,6 +30,7 @@ pgColumnar has two kinds of settings:
 | `pgcolumnar.compression` | enum | `zstd` | Default codec for new chunks. One of `none`, `pglz`, `lz4`, `zstd`. `lz4` and `zstd` are available only when the extension was built with those libraries. |
 | `pgcolumnar.compression_level` | integer | `3` | Level for the `zstd` codec. Range 1 to 22. Higher levels compress more and write more slowly. |
 | `pgcolumnar.fsst_min_gain_percent` | integer | `5` | Minimum size reduction, in percent, for FSST string encoding to be kept for a column chunk. Range 0 to 99. See below. |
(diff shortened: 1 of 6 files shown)


OffgridwithJD (changes requested):
Reviewed by verifying every factual claim against the code (this is a docs-accuracy PR, so the docs are the artifact under test). One proven inaccuracy to fix; everything else I spot-checked is correct.

## One fix needed — a units error in a newly added GUC row

`docs/configuration.md` (the `pgcolumnar.sink_fail_after` row):

> A fault-injection point that fails a write sink after this many **rows**.

The code counts **bytes**, not rows. `columnar_sink.c` accumulates `snk->written += (int64) n` where `n` is the byte count of the write, and the trip is `snk->written + n > pgcolumnar_sink_fail_after`. The `DefineCustomIntVariable` short description agrees ("Fail export writes after this many **bytes**"). So the row should read "after this many **bytes**." Minor (it is an internal, leave-at-`-1` knob), but it is exactly the kind of code/doc mismatch this PR sets out to remove.

## Everything else I verified — all accurate

- **The `iceberg_scan` COMMENT fix is correct.** The reader does apply position, equality, and deletion-vector deletes; the old "refuses tables with delete files" was the stale claim. Good catch.
- **The other five new GUC rows match their `DefineCustom*` calls exactly** (default / bounds / context = PGC_USERSET): `fsst_verdict_reuse` 16 [0,INT_MAX]; `enable_late_materialization` on; `qual_skipvec_min_payload_cols` 20 [0,100000]; `objstore_buffered` on; `objstore_part_size` 0 [0,INT_MAX].
- **how-to.md spot-checks:** codec set is exactly `none`/`pglz`/`lz4`/`zstd`; `compression_level` default 3, range 1–22, zstd-only; `encode_effort` is `full`/`fast`; `compact`/`compact_rewrite` hold `ShareUpdateExclusiveLock` (autovacuum.c, metadata.c); the daemon "does not take an exclusive lock" is right (it never calls the `AccessExclusiveLock` verbs `vacuum`/`vacuum_sorted`/`cluster`); `parallel_copy` and Parquet reads gate on `pg_read_server_files` (`ROLE_PG_READ_SERVER_FILES`), `parallel_export_parquet` on `pg_write_server_files` (`ROLE_PG_WRITE_SERVER_FILES`).

Fix the one `rows` → `bytes` word and this is good to go — the how-to is well-scoped and the verification pass is solid.

— ChronicallyJD

OffgridwithJD (approved):
Approve — the fix is landed and correct, and it is the only change since my review.

`pgcolumnar.sink_fail_after` now reads "fails an export write after this many **bytes**, by the path a full disk takes." That matches the code (`columnar_sink.c`: `snk->written += (int64) n` accumulates bytes; trip at `written + n > sink_fail_after`) and the GUC's own long description. The diff since my reviewed revision touches only this one line, so everything I verified clean before still stands: the `iceberg_scan` COMMENT (applies position/equality/deletion-vector deletes), the other five GUC rows vs their `DefineCustom*` defaults/bounds, and the how-to spot-checks (codec set, `compression_level` 1–22/default 3, `encode_effort` full/fast, `compact`/`compact_rewrite` = ShareUpdateExclusiveLock, the daemon avoiding AccessExclusiveLock, and the `pg_read`/`pg_write_server_files` gates).

Good verification pass; the how-to is a useful addition.

— ChronicallyJD

PR #671, opened 2026-08-16
docs: cover the Iceberg, object-storage, and online-maintenance epics

jdatcmd:
## Why

The previous docs PR (#670) verified that what was *written* was accurate. This
one addresses the other half: **whole shipped epics were never documented**. A
completeness/staleness audit found that `features.md` and several other docs
predate the Iceberg (#388), object-storage (#393/#394), and online-maintenance
work and do not mention them at all.

Docs only — no behavior changes. Every claim was checked against `CHANGELOG.md`,
the install SQL, and `src/`.

## What was missing, now added

**features.md** (the worst — it predated two epics):
- New **Object storage** section (s3/http(s) URLs, the `pgcolumnar_objstore`
  module, AWS_* credentials, the `objstore_allowed_endpoints` allow-list).
- New **Apache Iceberg** section (`iceberg_scan` + the three delete kinds + name
  mapping; the REST catalog with per-role server/user-mapping tokens, OAuth2, and
  vended credentials; the `pgcolumnar_iceberg` FDW with partition + metrics
  pruning).
- Expanded **Maintenance** (the online verbs `compact`/`compact_rewrite`/
  `recluster`, plus `cluster`, `truncate`, `sort_status`, `maintenance_due`, and
  the `autovacuum` daemon).

**index.md** — external Parquet/Iceberg reading and object storage named in the
intro, "How it fits together", and "Where to start".

**user-guide.md** — a new **Read an Apache Iceberg table** section; object-storage
URLs + the allow-list added to the Parquet examples; the local-only access note
corrected; a Next-steps pointer to online maintenance.

**administration.md** — new **Object storage** and **Apache Iceberg** operational/
security sections (default-deny + superuser-only + link-local refusal, AWS_* and
FDW user-mapping credentials, REST token handling kept out of the log); the
server-file security table annotated for remote endpoints.

**ARCHITECTURE.md** — module-map entries for `columnar_autovacuum.c` and the
Iceberg subsystem; the `columnar_vacuum.c` entry extended with the online reclaim
verbs; FSST added to the encoding list.

## Gate

`docs_style.sh` (STE + nav + version-consistency) green across all 14 documents;
cross-doc anchors verified to resolve.

### DIFF
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -143,9 +143,11 @@ Lightweight, type-aware value-stream encodings, applied per chunk before the
 block codec and reversed after decompression, as reversible transforms of the
 raw value-stream bytes (so all downstream decode is unchanged). Encodings: RLE,
 frame-of-reference with bit-packing, delta and delta-of-delta (integer family),
-Gorilla XOR (float4/float8), and a dictionary (any fixed-width or varlena type,
-for low cardinality). `ColumnarEncodeChunk` measures each applicable encoding and
-keeps the smallest; `ColumnarDecodeChunk` reverses it. This file also holds the
+Gorilla XOR (float4/float8), a dictionary (any fixed-width or varlena type, for
+low cardinality), and FSST substring coding for text and other varlena values.
+`ColumnarEncodeChunk` measures each applicable encoding and keeps the smallest;
+`ColumnarDecodeChunk` reverses it. The `pgcolumnar.fsst_verdict_reuse` setting
+caps how often the FSST keep-or-drop decision is recomputed across row groups. This file also holds the
 compression-block run iterator (`ColumnarBlockReader`) that exposes a chunk as
 (value, run-length) pairs so an aggregate can fold a run at a time.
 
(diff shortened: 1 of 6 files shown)


OffgridwithJD (approved):
Approve — completeness pass reviewed by verifying every new technical claim against the code (docs are the artifact under test). All accurate; no fixes needed.

## Verified against code

- **Object storage security.** "Remote access is default-deny" — confirmed: `os_check_endpoint_allowed` starts `ok=false` and an empty `objstore_allowed_endpoints` falls straight to the `!ok` error. The setting is `PGC_SUSET` (superuser-only), so "a role cannot widen its own reach" holds. "A link-local/instance-metadata address is refused even when it is listed" — the strongest new claim, so I proved it empirically: with `169.254.169.254` explicitly in the allow-list, `read_parquet` is still refused, and specifically by the UNCONDITIONAL link-local check ("resolves into a link-local range"), not the allow-list. S3 options (`access_key_id`/`secret_access_key`/`session_token`), `objstore_s3_addressing`, and AWS Signature Version 4 all exist.
- **Iceberg reader.** "Applies row-level deletes of all three kinds" — position, equality (on `equality_ids`, incl. partition-scoped), and a format-3 deletion vector as a Puffin roaring bitmap, each with its own sequence rule: all present in the code (`columnar_delete_vector.c`, `columnar_puffin.c`, the `equality_ids` path). Matches the corrected #670 COMMENT. Pruning "before it opens them" matches the FDW's file-level pruning.
- **REST catalog.** Token from `PGCOLUMNAR_ICEBERG_REST_TOKEN` or a per-role USER MAPPING in `pg_user_mapping` (not world-readable), never a function argument so absent from the statement log / `pg_stat_activity` — matches what I verified in #665. OAuth2 client secret "travels only in the token-request body" — matches #667.
- **Maintenance.** `compact_rewrite(tablename, min_deleted_fraction DEFAULT 0.2, max_groups DEFAULT 0)` matches the SQL; the online verbs and daemon match #670.
- **ARCHITECTURE module entries** all reference real files: `columnar_autovacuum.c`, `columnar_avro.c`, `columnar_delete_vector.c`, `columnar_iceberg.c`, `columnar_iceberg_fdw.c`, `columnar_iceberg_rest.c`, `columnar_puffin.c`.

Good completeness audit — the previously-undocumented epics are now covered accurately.

— ChronicallyJD

OffgridwithJD (approved):
Re-reviewed after the second commit (my prior approval was stale). I verified all three correctness fixes and both example fixes against the code — and two of these corrected claims are ones my first pass did not scrutinize closely enough, so this adversarial writing/correctness pass genuinely tightened the docs.

## The three correctness fixes — all confirmed against code

1. **Daemon verbs.** `columnar_autovacuum.c` calls only `pgcolumnar.compact_rewrite` and `pgcolumnar.recluster` (header: "Only the ShareUpdateExclusiveLock verbs ... are ever called. Never vacuum()/vacuum_sorted()/cluster()"; the loop body invokes `compact_rewrite` and checks `recluster_due`). The earlier "runs the online verbs" phrasing that implied `compact`/`truncate` was wrong — correctly narrowed.
2. **hour() not on date.** features.md now reads "year/month/day/hour on a timestamp or timestamptz column, and year/month/day on a date column," which is exactly `ice_temporal_source_ok` (TS/TSTZ → all four; DATE → year/month coarse + day via the exact path; no hour). (user-guide.md's general transform list doesn't assert a per-type claim, so it's fine.)
3. **truncate lock.** `pgcolumnar_truncate` opens under `ShareUpdateExclusiveLock` and then `ConditionalLockRelation(rel, AccessExclusiveLock)` for the brief physical shrink — administration.md's "`ShareUpdateExclusiveLock`, plus a conditional `AccessExclusiveLock`" matches precisely.

## Example fixes — confirmed

- `parallel_export_parquet` writes a **directory** of `part-NNNN.parquet` files (`columnar_parallel_export.c`: `MakePGDirectory`, `part-0…` naming) — the how-to now exports to `/data/events_out`, not a single file.
- The per-role REST example now passes a **server name** (`iceberg_rest_scan('cat', …)` after `CREATE SERVER cat` + `USER MAPPING`), matching the server-name resolution path.

## New polish prose — spot-checked, accurate

`chunk_group_row_limit` (default 10000, 100..INT_MAX) bounds a chunk group as documented; `enable_index_only_scan` is on by default; the index-only all-visible logic ("VACUUM sets a chunk group's bit only when the group has no deletes and predates the horizon; any write clears it") matches `columnar_metadata.c`; and "the executor always re-applies the full qualifier, so skipping never returns a wrong row" is the recheck invariant.

Good pass — approving.

— ChronicallyJD

PR #672, opened 2026-08-16
docs: add a Best Practices guide

OffgridwithJD:
## Add a Best Practices guide

The reference and how-to docs are thorough, but there was no page that collected
the *advice* behind them: which codec, how large a batch, when to cluster, which
maintenance verb and at what cadence, and how to run object storage and Iceberg
safely. This adds `docs/best-practices.md`, organized by decision rather than by
task, as the companion the how-to tuning notes point at.

### Sections
- **When to choose a columnar table** — the workload shape it suits, and a pointer to the limitations.
- **Design and load** — batch size vs. the row-group/chunk-group unit, `parallel_copy` worker slots, over-partitioning.
- **Compression and encoding** — the `zstd`/`lz4`/`pglz`/`none` trade, `compression_level`, automatic encodings, `encode_effort`, `fsst_verdict_reuse`.
- **Layout for skipping** — cluster on the range-filter column; `recluster` (online) vs `cluster` (eager); reading `Chunk Groups Read`.
- **Maintenance** — a verb/lock-level table, the autovacuum daemon (only the two online verbs, its thresholds), `maintenance_due`, and `truncate` for disk reclaim.
- **Query performance** — column projection, late materialization, per-vector gating, index-only scans, reading the plan.
- **Object storage safely** — default-deny + superuser-only allow-list, the unconditional link-local refusal, per-role credentials, buffered reads, server-file roles.
- **Iceberg safely** — delete-correct read-only reader, the pruning wrapper, keeping the REST/OAuth secret out of the log.

### Grounding
Every default, bound, and lock level is taken from the code, not inferred: e.g.
`chunk_group_row_limit` 10000, `compression_level` 1–22 default 3, the autovacuum
thresholds 0.2 / 0.05, `truncate` = `ShareUpdateExclusiveLock` plus a brief
conditional `AccessExclusiveLock`, the daemon calling only `compact_rewrite` and
`recluster`, and the allow-list being default-deny and superuser-only with an
unconditional link-local refusal.

Wired into the nav under Getting started and cross-linked from the how-to guides
and the home page. `test/docs_style.sh` passes (the measurable ASD-STE100 rules:
no em/en dashes, sentences within the word limit, no idioms).

### DIFF
--- a/docs/best-practices.md
+++ b/docs/best-practices.md
@@ -1,0 +1,197 @@
+# Best practices
+
+This guide collects the operational advice that the [how-to guides](how-to.md)
+apply one task at a time. It is organized by decision. Each section covers one
+choice: table design and loading, compression, layout for skipping, maintenance,
+query speed, and safe object storage and Iceberg use. Every default and lock level
+here is the shipped behavior. The [configuration reference](configuration.md) lists
+the settings and their bounds.
+
+## When to choose a columnar table
+
+pgColumnar suits wide, append-mostly, analytical tables. These tables are scanned
+by ranges and aggregates far more often than they are updated row by row. It is not
+a replacement for a narrow, write-hot, point-lookup table. Use this test: if a query
+reads a few columns out of many and touches a range of rows, columnar wins. If a
+query fetches whole rows by primary key at high frequency, keep the heap. The
+[limitations](limitations.md) page lists the features a columnar table does not
+support. Read it before you migrate a table you cannot easily migrate back.
+
+## Design and load a table
+
+**Load in large batches.** The writer packs rows into a row group, the unit of the
+write. It divides each row group into chunk groups of up to
+`pgcolumnar.chunk_group_row_limit` rows (default 10000). A chunk group is the unit a
+scan skips. A column within a chunk group is a chunk, compressed and encoded on its
+own. Small transactions produce small, poorly compressed row groups, and many of
+them to scan later. Prefer `COPY` or a multi-row `INSERT ... SELECT` over
+row-at-a-time inserts. Load in batches of at least one chunk group.
+
+**Use `parallel_copy` for a large file.** `pgcolumnar.parallel_copy` splits a
+server-side file across workers and scales the load into one table. Give the cluster
+enough worker slots. Plan for the worker count plus two, in `max_worker_processes`
+and `max_parallel_workers`. Without the slots, the extra workers do no work. The
+caller needs the `pg_read_server_files` role.
+
+**Do not over-partition.** Native partitioning multiplies catalog and planning cost.
+Use it when a partition key genuinely bounds most queries, not by default. A single
+well-clustered columnar table often skips as much as a partitioned heap.
+
+## Choose compression and encoding
+
+**The codec trades ratio against speed.** The block codec is one of `none`, `pglz`,
+`lz4`, or `zstd`, and defaults to `zstd`. `zstd` gives the best ratio. `lz4`
+decompresses fastest for a scan-bound workload where storage is cheap. `none` and
+`pglz` exist for special cases. `pgcolumnar.compression_level` applies to `zstd` and
+ranges from 1 to 22, with a default of 3. Raise it for cold, archival data, where a
+slower write buys lasting space. Leave it low for tables you rewrite often.
+
+**Let the encoder choose the encoding.** Before the codec runs, each chunk takes the
+encoding that makes it smallest. The encodings include dictionary, run-length,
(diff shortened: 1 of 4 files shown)


jdatcmd (changes requested):
Reviewed aggressively, verifying every default, bound, lock level, and counter name against the code. The guide is well-structured and mostly accurate, and the security framing (default-deny allow-list, unconditional `169.254.0.0/16` + `fe80::/10` refusal, per-role credentials, token out of the log) all checks out against `objstore/columnar_objstore_module.c` and `columnar_iceberg_rest.c`. `maintenance_due(table)` is valid (the fraction args default). A few things to fix before merge, since this doc's whole value is that it is code-grounded.

**Must fix**

1. **The scan counter is `Columnar Chunk Groups Read`, not `Chunk Groups Read`** (two places: the "Lay the table out for skipping" section and "Read the plan"). The real EXPLAIN label carries the `Columnar` prefix (`columnar_customscan.c:2352`). Note you already spell the FDW counter exactly right as `Files Pruned`, so the two are inconsistent. Please use the exact `Columnar Chunk Groups Read`.

2. **`autovacuum_recluster_threshold` is the appended fraction, not the "unsorted fraction."** The GUC description is "Appended fraction at or above which the daemon reclusters a table" (`columnar_tableam.c:2883`). "Unsorted fraction" reads as a general disorder metric; it is specifically the fraction appended after the last sort. Please say "appended fraction" to match the code and `sort_status`.

**Should improve (advice quality)**

3. **Batch-size floor targets the wrong unit.** "Load in batches of at least one chunk group" points at the *skip* unit (10000 rows, `chunk_group_row_limit`). The write and compression unit is the *row group*, up to `stripe_row_limit` (default 150000). A 10000-row transaction still seals a small, under-filled row group. Recommend batching toward a full row group (`stripe_row_limit`), which is the number that actually governs compression and the group count a later scan pays for.

4. **Confirming pruning from `Chunk Groups Read` alone is incomplete.** The read count only shows pruning when compared against `Columnar Chunk Groups Total`. The direct signal that a filter skipped groups is `Columnar Chunk Groups Removed by Filter`, which the how-to guide already points readers at. Suggest referencing "Removed by Filter" (or the Read-vs-Total comparison) so the instruction actually demonstrates pruning, and so the two docs agree.

**Nits**

5. `parallel_copy` sizing: the hard requirement is workers + 1 (one coordinator background worker plus the loaders; `columnar_parallel_copy.c:303`), so "worker count plus two" is a safe margin rather than the requirement. Fine to keep, but stating the mechanism (a coordinator plus N loaders) would be more precise.

6. "format-3 deletion vectors" -> "format-version 3" to match the spec term used elsewhere in the docs.

Everything else I checked is accurate: codec set and `compression_level` 1-22/default 3, `chunk_group_row_limit` 10000, the autovacuum daemon calling only `compact_rewrite` and `recluster` at 0.2/0.05, `truncate`'s SUEL plus brief conditional AccessExclusiveLock, `qual_skipvec_min_payload_cols` 20, `fsst_verdict_reuse` 16, `objstore_part_size` 8 MiB default, the FDW user-mapping option names, and the lock-level table. Fix 1 and 2, tighten 3 and 4, and this is good to merge.

OffgridwithJD:
Thanks for the review. All six points are fixed in 6951215 — I verified each against the code rather than just applying them:

1. **`Columnar Chunk Groups Read`** — confirmed at `columnar_customscan.c:2352`; both scan references now use `Columnar Chunk Groups Removed by Filter` (`:2354`) as the direct pruning signal, matching `how-to.md`.
2. **Appended fraction** — the GUC description is literally "Appended fraction at or above which the daemon reclusters a table" (`columnar_tableam.c`); corrected from "unsorted fraction".
3. **Row group batching** — now targets `stripe_row_limit` (default 150000), the write/compression unit, not the 10000-row chunk group.
4. Pruning confirmation now points at `Columnar Chunk Groups Removed by Filter` (no Total comparison needed).
5. **parallel_copy** — the code launches one coordinator plus loader workers, so the text now says one extra slot, not two.
6. `format-version 3`.

`docs_style.sh` still passes. Good catches on 1 and 2 in particular — the recluster-threshold semantics were a real error on my part.

jdatcmd (approved):
Re-reviewed after 6951215. All six points fixed and verified against code: both scan references now use the exact `Columnar Chunk Groups Removed by Filter` (the direct pruning signal, matching how-to.md); "appended fraction" for the recluster threshold; batch advice targets a full row group (`stripe_row_limit`, 150000); parallel_copy sized as coordinator + loaders (workers + 1); "format-version 3". STE passes and the flagged strings are gone. Good to merge.

PR #673, opened 2026-08-16
fix: reject out-of-range Parquet dictionary index (critical OOB read)

OffgridwithJD:
### The bug
`decode_column` bounds-checks a file-controlled `RLE_DICTIONARY` index with a **signed** comparison (`src/columnar_parquet_reader.c`):

```c
if ((int) idx[i] >= dictCount)   /* idx[i] is uint32 */
    return false;
pv[i] = dict[idx[i]];
```

`rle_bitpack_decode` accepts `bit_width` up to 32 (`if (bit_width < 0 || bit_width > 32) return false;`), so a data page can encode an index with the high bit set, e.g. `0x80000000`. The cast `(int) idx[i]` sign-extends that to `INT_MIN`, which is `< dictCount`, so the guard passes and `dict[0x80000000]` reads about 16 GB past the dictionary. Reachable from a single crafted file via `read_parquet` / `import_parquet` / the Parquet FDW, no privilege beyond the call.

### Proof (reproduced on the bench, PG 18.4 assert build)
I crafted a minimal required-column, uncompressed, single-entry-dictionary file and replaced its data-page payload with `[bit_width=32][bit-packed index 0x80000000]` (generator committed at `test/fixtures/parquet/gen_dict_oob.py`). On the **unfixed** build:

```
LOG:  client backend (PID ...) was terminated by signal 11: Segmentation fault
```

A control file (same shape, benign index) reads its rows and the backend survives, so the crash is the malicious index, not the harness.

### The fix
Compare unsigned against a `dictCount` proven `>= 1`:

```c
if (dict == NULL || vallen < 1 || dictCount < 1)
    return false;
...
if (idx[i] >= (uint32) dictCount)
    return false;
```

The out-of-range index is now a clean decode error.

### Test
`test/native_parquet_dict_oob.sh` (registered in `run_all_versions.sh`) is the deterministic companion to the random `fuzz_parquet.sh`: control reads 4 rows, the crafted file is rejected, the backend survives, and the log shows no segfault. It **fails on the unfixed build (segfault) and passes on the fix.** Verified 4/4 on the fix; the existing Parquet suites still pass.

Recommended follow-up: re-run the crafted fixture under `pg18_san` (ASAN/UBSAN) to pin the exact overflow, per the decode-path sanitizer gate.

This is finding #1 (CRITICAL) of an audit of the extension; further security fixes will follow as separate PRs.


jdatcmd (approved):
Approve. Independently verified on the bench (PG 18.4):

- **Fix is correct and complete.** The signed `(int) idx[i]` sign-extends a high-bit index (bit_width 32, e.g. 0x80000000) to a negative int that slips past `< dictCount`; the unsigned `idx[i] >= (uint32) dictCount` with `dictCount` proven >= 1 fully bounds the index. Confirmed this is the ONLY dict-index site with the pattern (grep of src): no sibling in the Arrow or vectorized paths.
- **Removal proof reproduces both directions.** I ran native_parquet_dict_oob.sh on the fix (4/4 PASS: control reads 4 rows, crafted file rejected, backend survives, no segfault), then reverted just the one-line bound to the signed form and reran: the crafted file crashes the backend (SELECT 1 returns nothing, server log records signal 11), while the control still reads its 4 rows. The test is non-vacuous and bites exactly on the malicious index.

The load-bearing assertions (backend survives, no segfault) are the right ones; the text-grep rejection check is fine as the secondary signal. The ASAN follow-up you noted is a reasonable strengthener but not required here: the bound is provably complete by inspection and the segfault repro is decisive. Good to merge.

PR #674, opened 2026-08-16
fix: reject under-filled native DICT chunk (uninitialized read)

OffgridwithJD:
### The bug
`decode_dict` (`src/columnar_encoding.c`) allocates the raw output buffer **unzeroed** and fills only what the codes decode to, but never verifies the total equals the declared `value_raw_length`:

```c
char *raw = MemoryContextAlloc(cx, rawLen > 0 ? rawLen : 1);  /* not zeroed */
...
for (i = 0; i < n; i++) {
    if (dlen[code] > rawLen - pos) DECODE_CORRUPT(...);  /* guards overflow only */
    memcpy(raw + pos, dptr[code], dlen[code]);
    pos += dlen[code];
}
return raw;                     /* no check that pos == rawLen */
```

The pre-dispatch switch validates raw length for the fixed-width encodings but has `default: break` for DICT (its raw length is the sum of variable entry lengths, not knowable up front). So a chunk whose descriptor inflates `value_raw_length` past what its codes produce leaves `raw[pos..rawLen)` uninitialized. A varlena column then reads a length prefix out of that garbage, giving silent wrong results or an out-of-bounds read.

### The fix
Require the decoded length to match, mirroring the existing `decode_fsst_shared` check:

```c
if (pos != rawLen)
    DECODE_CORRUPT("DICT output length does not match raw length");
```

### Proof (reproduced on the bench, PG 18.4)
`test/native_dict_underfill.sh` DICT-encodes a low-cardinality text column, inflates that chunk's `value_raw_length` in `pgcolumnar.column_chunk.encoding_descriptor` via SQL (the same corruption vehicle as `hardening.sh`), and reads it:
- **Unfixed:** the `SELECT` returns rows with **no error** — built from uninitialized memory.
- **Fixed:** `ERROR: columnar: corrupt encoded chunk (DICT output length does not match raw length)`, backend alive.

3/3 on the fix; registered in `run_all_versions.sh`.

Finding #2 (HIGH) of the audit. #1 (CRITICAL Parquet OOB) is #673.


jdatcmd (approved):
Approve. Independently verified on the bench (PG 18.4).

**Fix is correct and complete.** `decode_dict` allocated the raw buffer unzeroed and never checked the codes filled it; an inflated `value_raw_length` left the tail uninitialized, and a varlena length prefix read out of the garbage. The added `if (pos != rawLen) DECODE_CORRUPT(...)` closes it, mirroring the FSST check at line 2086. I enumerated the dispatch switch: RLE/FOR/DELTA/GORILLA/DOD/ALP are fixed-width and validate `n*attlen == rawLen` up front, NONE checks `encLen == rawLen`, so DICT and FSST are the only variable-length decoders and both now validate the total. No sibling with the same bug. The invariant comment at line 2521 ("DICT validates internally") is now actually true.

**Removal proof bites.** native_dict_underfill.sh passes 3/3 on the fix (precondition DICT-encoded, under-filled chunk rejected with a clean error, backend alive). I removed just the new check and reran: the under-filled chunk is silently accepted (the reject assertion flips to FAIL, `got [no]`), i.e. rows built from uninitialized memory, while the backend still does not crash. That is the correct RED for a silent-corruption finding: the fix converts silent wrong results into a clean error. Good to merge.

PR #675, opened 2026-08-16
fix: reject CR/LF in object-store request path/host (request-splitting)

OffgridwithJD:
### The bug
`os_request` and `os_write_request` (`objstore/columnar_objstore_module.c`) write the request-target and host verbatim into the request line:

```c
appendStringInfo(&req, "%s %s HTTP/1.1\r\nHost: %s:%d\r\n",
                 method, h->abspath, h->host, h->port);
```

`h->abspath` / `h->host` come from a user-controlled URL (or `catalog_uri`). A path containing `\r\n` splits the request line and smuggles a second request/header onto the connection to an allow-listed endpoint. The caller-supplied header lines already rejected CR/LF (line ~2582); the request-target and host did not.

### Proof (reproduced on the bench, PG 18.4)
`test/objstore_crlf.sh` runs a capture listener and points `read_parquet` at `http://127.0.0.1:PORT/x\r\nX-Injected: yes`:
- **Unfixed:** the listener captures `X-Injected: yes HTTP/1.1` as its own line, i.e. the request was split.
- **Fixed:** `ERROR: columnar: an HTTP request path may not contain CR or LF`, and nothing is smuggled.

### The fix
Both emit paths call `os_reject_target_ctl(h)`, which raises on CR or LF in the path or host before the request is built. This mirrors the defense the header lines already carry.

### Also
Documents that a port-less allow-list entry matches any port on that host (a separate low-severity finding). That behavior is relied upon (the suites allow-list `127.0.0.1` and reach stubs on random ports), so rather than break it, the config reference now states it, so an operator who wants a port restriction adds one.

Findings #3 (HIGH) and #4 (low) of the audit. #673 (CRITICAL) merged; #674 (HIGH) is approved and mergeable.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,15 @@ which was true until that script existed.
 
 ### Security
 
+- Fixed an HTTP request-line injection in the object-store client. A URL path or
+  host containing CR or LF was written verbatim into the request line, so a
+  crafted path could split the request and smuggle a second line to an
+  allow-listed endpoint. The request path and host are now rejected if they carry
+  CR or LF, as the caller-supplied header lines already were. Regression test:
+  objstore_crlf.
+
+### Security
+
 - Fixed an out-of-bounds read in the Parquet dictionary decode path. A file whose
   RLE_DICTIONARY data page carried an index with the high bit set (reachable at
   bit_width 32) passed a signed bounds check that sign-extended it to a negative
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,8 +21,6 @@ which was true until that script existed.
   CR or LF, as the caller-supplied header lines already were. Regression test:
   objstore_crlf.
 
-### Security
-
 - Fixed an out-of-bounds read in the Parquet dictionary decode path. A file whose
   RLE_DICTIONARY data page carried an index with the high bit set (reachable at
   bit_width 32) passed a signed bounds check that sign-extended it to a negative
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -95,7 +95,7 @@ schemes and the credential model.
 
 | Setting | Type | Default | Description |
 | --- | --- | --- | --- |
-| `pgcolumnar.objstore_allowed_endpoints` | string | `''` (empty) | The endpoints the module may connect to, comma-separated as `host` or `host:port`. Empty refuses every remote endpoint, so a role that can read or write server files cannot reach an arbitrary host through the extension. Link-local addresses, including `169.254.169.254`, are refused whether or not they are listed. Superuser-only, so a role cannot widen its own reach. |
+| `pgcolumnar.objstore_allowed_endpoints` | string | `''` (empty) | The endpoints the module may connect to, comma-separated as `host` or `host:port`. An entry with no port matches any port on that host; add a port to restrict it. Empty refuses every remote endpoint, so a role that can read or write server files cannot reach an arbitrary host through the extension. Link-local addresses, including `169.254.169.254`, are refused whether or not they are listed. Superuser-only, so a role cannot widen its own reach. |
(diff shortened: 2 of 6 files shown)


jdatcmd (changes requested):
The CR/LF fix is correct where it is applied, but it is **incomplete**: the guard covers only two of the four request-line emitters, and one of the uncovered ones is the ABI entry the Iceberg REST catalog uses.

## Must fix: two unguarded wire emitters

`os_reject_target_ctl(h)` is called only in `os_request` and `os_write_request`. But four functions emit a request line built from `h->abspath` / `h->host`:

| function | emits | guarded? |
| --- | --- | --- |
| `os_request` | `"%s %s HTTP/1.1\r\nHost: %s:%d"` | yes |
| `os_write_request` | `"%s %s%s%s HTTP/1.1\r\nHost: %s:%d"` | yes |
| `os_list_request` | `"GET %s%s%s HTTP/1.1\r\nHost: %s:%d"` (called at ~2487) | **no** |
| `objstore_http_request` | `"%s %s HTTP/1.1\r\nHost: %s:%d"` (~2636) | **no** |

`objstore_http_request` is the exported ABI `.http_request` (`= objstore_http_request`), which the Iceberg REST client calls with a URL built from `catalog_uri` (`columnar_iceberg_rest.c` GET/POST). `catalog_uri` is user-controlled (a SQL argument to `iceberg_rest_table_location`, or a FOREIGN SERVER option), and `objstore_http_request` sets `h->abspath = pstrdup(slash)` / `h->host` from it and validates only port/empty-host, not CR/LF. So the exact bug this PR fixes is still reachable through the REST-catalog path.

**Proven on the bench (PR #675 branch, PG 18.4).** Same crafted URL, two entry points:
- `read_parquet('http://127.0.0.1:PORT/x<CR><LF>X-Injected: yes')` (os_request) -> `ERROR: an HTTP request path may not contain CR or LF` (guard fires).
- `iceberg_rest_table_location('http://127.0.0.1:PORT/x<CR><LF>X-Injected: yes', ...)` (objstore_http_request) -> no CR/LF error; it proceeds to connect with the split URL intact (the error text shows the literal newline: `.../x` then `X-Injected: yes/v1/config`). The only thing that stopped a captured smuggle was the listener refusing that connection, not a guard.

`os_list_request` (live, called at ~2487) emits `h->host` unguarded the same way.

**Suggested fix:** the two per-emitter guards were easy to miss (two were). Reject CR/LF once at the single choke point where `h->abspath` / `h->host` are assigned from the URL (the URL parse used by both the os_open path and the inline parse in `objstore_http_request`), so every current and future emitter is covered. Failing that, call `os_reject_target_ctl(h)` in all four emitters.

## Minor
After rebasing on current main (which now carries the Parquet OOB Security bullet and its all-null follow-up), consolidate the two `### Security` subsections in CHANGELOG.md into one.

The `#4` doc note (port-less entry matches any port, add a port to restrict) reads well and covers the earlier ask. Fix the completeness above and this is good.

jdatcmd (approved):
Re-reviewed d00720d. This resolves my REQUEST_CHANGES. Moving the guard into os_connect is the right shape: os_connect is the one path all four request builders go through (os_request, os_write_request, os_list_request, and the ABI objstore_http_request), so every emit site is covered and a future one cannot miss it.

Verified on the bench (PG 18.4), the exact contrast that failed before now passes: iceberg_rest_table_location('http://127.0.0.1:PORT/x<CR><LF>X-Injected: yes', ...) -> 'columnar: an HTTP request path may not contain CR or LF' (the REST/ABI path is rejected now), and read_parquet with the same URL is still rejected. objstore_crlf (3/3), objstore_allowlist, iceberg_rest_server, iceberg_rest_scan, iceberg_rest, iceberg_objstore all pass.

One ask before I merge: please add a REST-path arm to objstore_crlf.sh (your exact repro through iceberg_rest_table_location, asserting the CR/LF error and nothing smuggled), like the SSRF arm we added in #668. It pins the specific gap this review found so a future refactor that reintroduces a per-site emitter can't slip past. You offered it; yes please, in this PR. Approving now; I'll merge once that arm is in and green.

PR #676, opened 2026-08-16
fix: all-null Parquet column (empty dictionary page) still reads (regression from #673)

jdatcmd:
## Urgent: fixes a regression on `main`

`main` currently returns **empty results** for an all-null Parquet column and for
an Iceberg null-delete read. This restores them.

### Cause
The Parquet dictionary OOB fix (#673) added a pre-decode guard:

```c
if (dict == NULL || vallen < 1 || dictCount < 1)
    return false;
```

That `dictCount < 1` also rejects a **legitimate** all-null page: a column that is
entirely null has zero coded values (`nnn == 0`) and an empty dictionary
(`dictCount == 0`), which is valid, not corrupt. So the guard turned a valid page
into a decode error. It regressed two suites on `main`:
- `native_parquet_streaming` — "all-null column (empty dictionary page) still reads" returns `[]`.
- `iceberg_deletes` — "a null delete value matches only null data values" returns `[]` (it reads such a column).

### Fix
Decode the dictionary indices only when the page has coded values (`nnn > 0`). An
all-null page then reads. When there **are** values, the same `dictCount >= 1`
requirement holds, so the unsigned bound `(uint32) dictCount` that closes the
out-of-bounds read is fully preserved.

### Verified (PG 18.4 bench)
- `native_parquet_streaming`, `iceberg_deletes`, `native_parquet_dict_oob` all pass.
- `native_parquet_dict_oob` still reproduces the OOB rejection, so #673's security fix is intact.
- Strict PG19 build clean.

Regression tests already in tree cover it (`native_parquet_streaming` all-null,
`iceberg_deletes` null-delete). CHANGELOG amends #673's Security bullet rather
than adding a new one (the regression never shipped; both are in Unreleased).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,7 +19,9 @@ which was true until that script existed.
   bit_width 32) passed a signed bounds check that sign-extended it to a negative
   int, and the dictionary was then read far out of bounds, crashing the backend
   from a single crafted file. The bounds check is now unsigned and the index is
-  rejected. Regression test: native_parquet_dict_oob.
+  rejected. The index decode runs only when the page has coded values, so an
+  all-null column with an empty dictionary page still reads. Regression tests:
+  native_parquet_dict_oob and native_parquet_streaming.
 
 ### Added
 
--- a/src/columnar_parquet_reader.c
+++ b/src/columnar_parquet_reader.c
@@ -2116,14 +2116,24 @@ decode_leaf_entries(PqSource *src, PqChunk *ch,
 				uint32	   *idx;
 				int			bw;
 
-				if (dict == NULL || vallen < 1 || dictCount < 1)
-					return false;
-				bw = valbuf[0];
-				idx = palloc(sizeof(uint32) * Max(nnn, 1));
-				if (!rle_bitpack_decode(valbuf + 1, vallen - 1, bw, nnn, idx))
-					return false;
-				for (i = 0; i < nnn; i++)
+				/*
+				 * Decode indices only when the page has coded values. An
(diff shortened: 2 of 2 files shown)


OffgridwithJD (approved):
Approve. This correctly fixes a regression I introduced in #673 and preserves the OOB protection.

My guard `dictCount < 1` was over-broad: an all-null Parquet page legitimately has `nnn == 0` and, for a fully-null column, `dictCount == 0`, and I rejected it. I also failed to test the all-null boundary, which is how it reached main.

Gating the whole dictionary-index decode on `if (nnn > 0)` is the right shape:
- **nnn == 0** (all-null page): no indices to decode, so the block is skipped and the page reads normally. Fixes `native_parquet_streaming` and `iceberg_deletes`.
- **nnn > 0**: `dict != NULL`, `vallen >= 1`, and `dictCount >= 1` are still required, and the per-index bound stays the unsigned `idx[i] >= (uint32) dictCount`. So the critical OOB fix is intact: a crafted `0x80000000` index against a non-empty dictionary is still rejected. (`palloc(sizeof(uint32) * nnn)` is safe since nnn > 0 here.)

Verified independently on the bench (PG 18.4): `native_parquet_streaming`, `iceberg_deletes`, and `native_parquet_dict_oob` all pass, and `native_parquet_dict_oob` still reproduces the OOB-as-rejection (the guard did not weaken).

Thanks for catching this and for the fast fix. My apologies for the over-broad guard and for not covering the all-null case in the original test.

— ChronicallyJD

PR #677, opened 2026-08-16
docs: document the concurrent-UPDATE duplication limitation (audit #5)

OffgridwithJD:
## Document the concurrent-UPDATE duplication limitation

Audit finding #5 (data-integrity). Two sessions that `UPDATE` the same row on a columnar table with **no covering unique index** each append their own new version, so the row is duplicated and one update is lost. Heap serializes on the tuple (tuple lock + EvalPlanQual) and yields one row with both updates.

### Reproduced (bench, PG 18.4)
Session A holds an open transaction with `UPDATE cu SET v=v+1 WHERE id=1`; session B runs the same update. After both commit:
- **Columnar:** `count(*) WHERE id=1` = **2**, values `[101,101]` (duplicated row, lost update).
- **Heap:** one row, `v=102`.

Root cause: `pgcolumnar_tuple_update` always returns `TM_Ok` and takes no tuple lock, so the executor never runs EvalPlanQual to re-read the concurrent winner. The existing Concurrency docs describe the delete-mark serialization ("both delete marks survive") but do not note that for UPDATE this duplicates the row.

### Why this is a doc, not a code fix (for now)
The proper fix is a deep MVCC change to the access method (detect the concurrently-modified old row and return `TM_Updated` so EvalPlanQual runs); the delete vector does not even track a deleting xid. That is a maintainer-level design decision with real correctness risk (a wrong fix can corrupt data or deadlock), so per the audit it is documented now and the reproduction handed off for a designed fix. The repro harness is available (`repro_concurrent_update.sh`).

This documents the current behavior and the workaround (a unique index, or application-level serialization) so it is an informed choice.

### DIFF
--- a/docs/limitations.md
+++ b/docs/limitations.md
@@ -399,6 +399,13 @@ same way as a `CREATE INDEX` that is not concurrent. Turn projection scans off w
   commit. It then reads the committed mask again and merges its own bits. Thus
   both sets of delete marks survive. Writes to different row groups continue at
   the same time.
+- A concurrent `UPDATE` of the same row is a delete of the old version plus an
+  insert of the new one. The two writers do not serialize on the row itself.
+  Without a covering unique index, two sessions that update the same row can each
+  keep their own new version. The row is then duplicated and one update is lost.
+  A unique index on the updated key makes the writers go in sequence, and the
+  conflict is found. This is a known limitation. Add a unique index, or serialize
+  the updates in your application, where correctness needs it.
 - Concurrent inserts of the same unique key go in sequence. The server therefore
   always finds the conflict. Before a new row reaches the uniqueness check, the
   access method takes an advisory lock with the scope of the transaction. The key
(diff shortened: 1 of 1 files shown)


jdatcmd (approved):
Approve. I verified the behavior against the code and on the bench (PG 18.4):

- Root cause confirmed: pgcolumnar_tuple_update marks the old row deleted, appends the new version, and returns TM_Ok unconditionally with no tuple lock, so the executor never runs EvalPlanQual. The delete vector tracks no deleting xid, matching your 'why this is a design-level fix' note.
- Bug reproduced: two sessions updating the same row with NO unique index leave count(*) WHERE id=1 = 2 (duplicated, one update lost).
- Workaround reproduced: with a unique index, the second session errors 'duplicate key value violates unique constraint' and count stays 1. PgColumnarLockUniqueKeys (the /* issue #5 */ call in tuple_update) is what serializes them, so the doc's mechanism is right.

The doc is accurate and honest, and documenting this now (with the fix handed off as a designed MVCC change) is a reasonable owner call given the correctness risk.

One clarification worth adding, since this is a data-integrity page: the unique-index path does not transparently serialize like heap (heap yields one row with both updates applied, v=102). It converts the silent duplication into a serialization ERROR that the losing transaction must catch and RETRY. Suggest saying so, e.g. '...the second update fails with a unique-violation error, which the application must retry' so a reader does not expect heap-style automatic serialization. Non-blocking; approving as-is.

PR #678, opened 2026-08-16
fix: estimate Iceberg FDW rows from manifests (not a constant 1000)

OffgridwithJD:
## Estimate the Iceberg FDW's row count from the manifests

Audit finding #8 (optimization, HIGH).

`icefdwGetForeignRelSize` set `baserel->rows = 1000.0` for **every** Iceberg foreign table, ignoring the `record_count` the manifests carry. A large table is then planned as 1000 rows, which corrupts every join decision above the scan (the huge table lands on a nested-loop inner side; an unbounded tuplestore holds the whole thing).

### Fix
`PgColumnarIcebergEstimateRows(metadata_path)` walks the manifests (the scan reads them anyway) and sums the live data files' `record_count`. `collect_deletes = true` so a table that carries deletes is estimated rather than refused; delete files are not subtracted, so the estimate is a slight overestimate, which is the safe direction for a planner cardinality. The wrapper uses it in place of the constant.

### Proof (bench, PG 18.4)
`test/iceberg_fdw_estimate.sh` on the committed `events` warehouse (5 rows):
- **Before:** `EXPLAIN` reports `rows=1000`.
- **After:** `rows=5`, equal to the real `count(*)`.

Existing `iceberg_fdw`, `iceberg_scan`, and `iceberg_deletes` still pass. Registered in `run_all_versions.sh`.

First of the optimization/modularity audit findings. More to follow.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,13 @@ which was true until that script existed.
 
 ## [Unreleased]
 
+### Fixed
+
+- The Iceberg foreign-data wrapper now estimates a scan's row count from the
+  manifests (the sum of the live data files' record counts) instead of a constant
+  1000. The constant mis-sized every scan and corrupted join planning above a
+  large Iceberg table. Regression test: iceberg_fdw_estimate.
+
 ### Security
 
 - Fixed an uninitialized-memory read in the native DICT decode path. A chunk
--- a/src/columnar_iceberg.c
+++ b/src/columnar_iceberg.c
@@ -927,6 +927,43 @@ pgcolumnar_iceberg_data_files(PG_FUNCTION_ARGS)
 	return (Datum) 0;
 }
 
+/* Sum the live data files' record counts for a row estimate. Delete files are
+ * ignored, so the count is a slight overestimate on a table with deletes, which
+ * is the safe direction for a planner cardinality. collect_deletes is true so a
+ * table that carries deletes is estimated rather than refused. */
+static void
+ice_estimate_cb(void *ctx, PgColumnarAvroManifestEntry *e,
+				const PgColumnarAvroManifestFile *mf,
+				const char *recorded_root, const char *actual_root,
+				const char *mdpath, int64 snapshot_id)
+{
+	(void) mf;
+	(void) recorded_root;
+	(void) actual_root;
+	(void) mdpath;
+	(void) snapshot_id;
+	if (e->content == 0)		/* a data file; deletes do not add rows */
+		*(int64 *) ctx += e->record_count;
+}
+
+/*
+ * A planner row estimate for the current snapshot: the sum of the live data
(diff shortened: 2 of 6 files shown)


jdatcmd (changes requested):
Right problem, and summing the manifests' record_count is the correct source. One change before merge: apply restriction-clause selectivity, don't set `baserel->rows` to the raw total.

## The estimate ignores WHERE selectivity

`icefdwGetForeignRelSize` does `baserel->rows = (est > 0) ? est : 1.0;`. That sets the estimated **output** rows to the whole table's cardinality, regardless of the query's restriction clauses. So a selective filter still estimates the full table.

Proven on the bench (PG 18.4), committed `events` warehouse, 5 rows (2 are `region='eu'`):
```
EXPLAIN SELECT * FROM ev;                    -> rows=5   (correct)
EXPLAIN SELECT * FROM ev WHERE region='eu';  -> Foreign Scan ... rows=5   (should be ~2)
```
The second should be ~2. On a billion-row Iceberg table a highly selective filter would estimate a billion output rows, which over-sizes the scan and mis-plans joins above it, the same class of failure this PR fixes, just in the opposite direction. So over-estimate is not the universally safe direction.

## Fix: the idiomatic FDW pattern

Set the total as `tuples` and let the planner derive `rows` from the baserestrictinfo selectivity:

```c
baserel->tuples = (est > 0) ? (double) est : 1.0;
set_baserel_size_estimates(root, baserel);   /* applies clauselist_selectivity(baserestrictinfo) */
```

This is what postgres_fdw (local-estimate path) and file_fdw do. It keeps your manifest-sum cardinality, fixes the original 1000 under-estimate, AND makes a selective scan estimate its filtered rows. `baserel->baserestrictinfo` is populated at GetForeignRelSize time, so the selectivity is available.

## Also
- `iceberg_fdw_estimate.sh` only asserts the no-filter case (rows=5). Please add a filtered assertion (e.g. `WHERE region='eu'` estimates fewer than the total) so the selectivity path is covered and can't regress.

The manifest walk, `collect_deletes = true` (estimate a table with deletes rather than refuse it), and the not-subtracting-deletes overestimate are all fine. Just route the total through `set_baserel_size_estimates`.

jdatcmd (approved):
Re-reviewed. Resolved. Now baserel->tuples = the manifest record-count sum and set_baserel_size_estimates(root, baserel) derives rows from clauselist_selectivity over baserestrictinfo (with the optimizer/cost.h include). Verified on the bench: iceberg_fdw_estimate.sh is 4/4 including the new 'a filtered scan estimates fewer rows than the whole table (selectivity applied)' assertion, and the unfiltered estimate is still the real record count (5, not 1000). This keeps the manifest-sum cardinality, fixes the original under-estimate, and applies selectivity so a filtered scan no longer over-estimates the full table. Good to merge.

PR #679, opened 2026-08-16
fix: index-fetch cost ignores per-table stripe_row_limit (#7/#11)

OffgridwithJD:
## Size the index-fetch cost penalty by the effective stripe_row_limit

Audit findings #7 (optimization, HIGH) and #11 (modularity, HIGH) — one root cause.

`pgcolumnar_index_fetch_penalty` sized row groups from the GUC:

```c
double R = (double) pgcolumnar_stripe_row_limit;   /* ignores the per-table option */
```

But a table's group count follows its **effective** `stripe_row_limit` — the per-table option when set, else the GUC — which is what the writer actually honors and what `pgcolumnar_zonemap_survival` already reads. So on a table created `WITH (stripe_row_limit = ...)`, the penalty computed the wrong number of groups and mis-priced the decode a columnar index fetch forces, steering the planner toward or away from an index scan.

The effective-limit lookup was duplicated across four sites; the index-fetch penalty was the one copy that got it wrong (#11).

### Fix
`pgcolumnar_effective_stripe_row_limit(relid)` is the single resolver. The three single-field consumers use it: `index_fetch_penalty` (with `relid` threaded in from `penalize_index_fetches`), `zonemap_survival`, and the flush writer. The two multi-field readers keep their one read-all-options call.

### Proof (bench, PG 18.4)
`test/native_index_fetch_stripe_cost.sh`: on one table, change `stripe_row_limit` via `set_options` and read the index-scan cost.
- **Unfixed:** cost is `3880.70` for both `stripe_row_limit = 2000` and `150000` — invariant, because it read the GUC.
- **Fixed:** `2000 -> 3640.34`, `150000 -> 3845.34` — the cost now responds to the per-table limit.

`scan_decode_cost`, `native_index_projection`, `native_skip`, `native_recluster` still pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,15 @@ which was true until that script existed.
 
 ## [Unreleased]
 
+### Fixed
+
+- The index-fetch cost penalty now sizes row groups by a relation's effective
+  `stripe_row_limit` (the per-table option when set, else the GUC), matching the
+  writer and the zone-map survival estimate. It read only the GUC, so it
+  mis-priced the row-group decode for a table that set the option, and could steer
+  the planner toward or away from an index scan on that table. The effective-limit
+  lookup is now one shared helper. Regression test: native_index_fetch_stripe_cost.
+
 ### Security
 
 - Fixed an uninitialized-memory read in the native DICT decode path. A chunk
--- a/src/columnar.h
+++ b/src/columnar.h
@@ -458,6 +458,7 @@ extern void PgColumnarDeleteMetadata(uint64 storageId);
 
 /* per-table options catalog (spec 7.4) */
 extern bool PgColumnarReadOptions(Oid relid, PgColumnarOptions *opts);
+extern int pgcolumnar_effective_stripe_row_limit(Oid relid);
 
 /* declared physical sort key (#288); List of pstrdup'd column names, NIL if
  * none is declared. Names (not attnums) so the value survives dump/restore. */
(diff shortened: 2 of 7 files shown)


jdatcmd (approved):
Approve. Verified on the bench (PG 18.4).

- Root cause confirmed: pgcolumnar_index_fetch_penalty read the GUC pgcolumnar_stripe_row_limit, so a table created WITH (stripe_row_limit=...) was priced against the wrong group count. It now uses pgcolumnar_effective_stripe_row_limit(relid) with relid threaded in from penalize_index_fetches.
- Consolidation (#11) is complete: I enumerated the single-field consumers, and all three now call the resolver -- index_fetch_penalty (customscan:839), zonemap_survival (customscan:1114), and PgColumnarEnsureStorageRow (write_state:665). The two multi-field readers (the main writer at write_state:590/615-616 and the proj writer at ~3238-3258) keep their read-all-options call, which is correct: they also resolve the per-table option (GUC default then override), and it is what the writer actually seals stripes on (write_state:955). So the premise holds -- the writer honors the effective limit and the cost model now matches it.
- The new resolver returns the per-table option when set (and > 0) else the GUC. Matches the inline logic it replaces.
- native_index_fetch_stripe_cost.sh passes and is a valid removal proof by construction: it asserts the index-scan cost differs between stripe_row_limit 2000 and 150000, which the old GUC-reading code cannot satisfy (invariant). scan_decode_cost, native_index_projection, and native_skip still pass -- no regression.

Good to merge.

PR #680, opened 2026-08-16
fix: push projection down in the Iceberg FDW (~5x on wide tables)

OffgridwithJD:
## Push projection down in the Iceberg foreign-data wrapper

Audit finding #9 (optimization). The wrapper decoded **every column of every surviving file** regardless of the query's projection, so a narrow `SELECT` over a wide table read and threw away most of its work.

### Fix
The reader already supports projection (the `needTop` mask skips an unreferenced column's chunk entirely; the native Parquet FDW uses it) — the Iceberg path just passed NULL. This threads an optional `needTop` through `pq_read_file_into` -> `PgColumnarReadParquetByFieldIdNM` -> `PgColumnarIcebergScanCore`; **every existing caller passes NULL**, so nothing else changes. `GetForeignPlan` records the attributes the output (`baserel->reltarget->exprs`) and the recheck quals reference; `BeginForeignScan` expands that to the mask.

### Correctness (bench, PG 18.4)
`test/iceberg_fdw_projection.sh` compares every projection shape against bare `iceberg_scan` on a wide fixture: one column, several columns, whole row, count-only, and — the subtle case — **a filter on an unselected column** (its column must still be decoded for the recheck). All 7 match. `iceberg_fdw`, `iceberg_scan`, `iceberg_deletes`, and the Parquet suites (`native_parquet_fdw`, `fuzz_parquet`, `arrow_import`, `import_deferred`, `native_parquet_multifile`) all still pass across the reader-signature change.

### Performance (A/B, same query fixed vs unfixed)
40-column, 300k-row Iceberg table, interleaved:
- `count(c0)` (1 of 40 columns): **2071ms unfixed -> 374ms fixed, ~5.5x faster**.
- `count(c0||c39)` (2 of 40): 2304ms -> 651ms, ~3.5x faster.

On the unfixed build a narrow and a wide scan cost the same (all columns decoded); on the fix they scale with the columns actually referenced. Output identical on both.


jdatcmd (approved):
Approve. Verified correctness on the bench (PG 18.4), with a focus on the shared-reader change and the subtle recheck case.

- The mask unions the right two sources: GetForeignPlan does pull_varattnos(baserel->reltarget->exprs) AND pull_varattnos(scan_clauses), so a column referenced only by a recheck qual is decoded even when it is not selected. needTop == NULL means decode-all, so every existing caller of pq_read_file_into / PgColumnarReadParquetByFieldIdNM / PgColumnarIcebergScanCore is unchanged.
- Output unchanged: iceberg_fdw_projection.sh matches bare iceberg_scan on all 7 shapes (one col, several, whole row, count-only, filter-on-unselected, select-one-filter-another).
- I proved the recheck-column handling is load-bearing and the test is non-vacuous: dropping the pull_varattnos(scan_clauses) line reds exactly the two filter-on-unselected arms (got [] want [c1r10] / [c2r5,c2r6]) and nothing else. That is the exact silent-wrong-rows failure this needs to guard, and the test catches it.
- No regression across the reader-signature change: iceberg_fdw, iceberg_scan, iceberg_deletes, native_parquet_fdw, native_parquet_multifile, arrow_import, import_deferred all pass.

The ~5x is structurally sound (a narrow scan now decodes only referenced columns). Good to merge. Note: this branch stacks on #679 (now merged), so it will want a rebase on current main to drop those files and the CHANGELOG/run_all_versions overlap.

Issue #434, opened 2026-08-05
Index-fetch penalty collapses to zero on a correlated index, and the custom scan is priced as if zone maps did not exist


OffgridwithJD:
From the same external validation report as #433. The two compound: #433 makes the index
fetch catastrophically expensive, and this makes the planner choose it.

## The symptom

500,000 rows, 9 columns, 1 KiB incompressible payload, 511 MB, btree on a bigint sequence.
One query, `WHERE chain_seq > 400000 GROUP BY record_type`, about 100k rows:

```
HashAggregate  (cost=33096.12..33096.17 rows=5 width=14)
  ->  Index Scan using rec_default_seq_idx on rec_default  (cost=0.42..32589.97 rows=101231)
```

| plan | time |
|---|---:|
| **the planner's own choice** (index scan) | **>300 s, cancelled** |
| `enable_indexscan=off` (custom scan, zone maps) | **59.7 ms** |
| heap | 52.1 ms |

The planner declines a plan that is at least **5,000x faster**. The reporter measured the
same shape at 33 minutes against 76 ms.

## Two causes, and the first is not what the report said

The report says "the TAM's random-fetch cost is drastically underestimated". The penalty
**exists** and defaults on: `pgcolumnar.enable_index_fetch_penalty`, declared
`src/columnar.h:185`, formula `src/columnar_customscan.c:685-746`, applied `:840-947`,
capped at 20x a full scan (`:753`, `:885`). Issue **#355** is exactly that claim and it is
closed. For a **scattered** index it now over-estimates by design and correctly picks the
custom scan.

What actually happens on this shape:

**1. The penalty collapses to nothing on a correlated index.** The `rho * rho`
interpolation at `columnar_customscan.c:720` drives `groups_decoded` toward `groups_min`,
about 1, when the index is well correlated with physical order. The penalty becomes a
rounding error, on exactly the tables where sequential keys make it look most attractive.
The formula was verified to the cent on a clustered fixture, so this is the model working
as written rather than a bug in the arithmetic.

**2. The custom scan is priced as a full scan.** `columnar_customscan.c:1033` never models
zone-map pruning, so the alternative is quoted at its worst case while the index scan is
quoted at its best. Measured on a correlated index: index scan priced 4,180 and ran 4.8 s;
custom scan priced 185,002 and ran 1.2 s. **Priced 44x cheaper, ran 4x slower.**

The penalty vanishes precisely where pruning works best, which is the same relation for
both causes.

[...]

PR #12, opened 2026-07-19
Gap 25: bloom filters for deterministic-collation text columns

jdatcmd:
Stacked on the hardening PR. Implements design/gaps/25.

Extends per-chunk bloom filters (I7) from non-collatable types to collatable types under **deterministic** collations (text, varchar), hashing under the column's collation on build and probe; nondeterministic collations stay unbloomed. Safe because a differently-collated predicate is never pushed as a scan key, so the constant hashes consistently with stored values.

Tests: differential (222) incl. text present/absent, C collation, and a mismatched explicit COLLATE that must stay correct; audit collation test still passes. Green on pg13/pg17/pg19.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #681, opened 2026-08-16
fix: share full-scan cost with the grouped vector-agg fallback (#12)

OffgridwithJD:
## Share the columnar full-scan cost with the grouped vector-agg fallback

Audit finding #12 (modularity, and a narrow cost-correctness bug).

The full-scan cost formula was duplicated. `PgColumnarSetRelPathlist` prices the columnar scan node with three refinements over the inherited seqscan number: projected-width I/O scaling (#171), per-column decode CPU (#503), and zone-map survival scaling of the run (#434). But `PgColumnarTryGroupAggPath` in `columnar_vector.c`, when **no serial scan path survives** `add_path` (the common case once a Gather dominates), priced its input with the **bare seqscan formula** — missing all three. So on a wide, low-pruning table the grouped-agg node under-priced its input and could beat a faster parallel Agg.

### Fix
Extract the refinement into one `pgcolumnar_refined_scan_cost()`. `SetRelPathlist` now calls it (a **literal, behavior-identical** extraction); the `vector.c` fallback calls it too, so it prices its input exactly as the real scan does.

### Proof (bench, PG 18.4)
- **No behavior change** where none is intended: `analyze_stats` (pins #171), `scan_decode_cost`, `column_projection`, `native_index_projection`, `native_vecdecode`, `native_fastdecode`, `native_groupagg`, `native_agg`, `parallel_vector_agg`, `rewrite_group_scan`, `native_agg_deletes` all pass — the `SetRelPathlist` extraction is identical.
- **The fallback path is exercised and correct**: `test/native_groupagg_wide_cost.sh` forces the fold path on a wide (25-column) table and asserts the grouped result equals a plain aggregate. Registered in `run_all_versions.sh`.

This is the last of the six optimization/modularity audit findings.


jdatcmd (approved):
Approve. Verified on the bench (PG 18.4).

- Wiring confirmed: pgcolumnar_refined_scan_cost is defined in customscan.c and called from both PgColumnarSetRelPathlist (the extraction) and columnar_vector.c:1741 (the grouped-agg no-serial-survivor fallback that previously used the bare seqscan formula). The helper carries all three refinements (#171 projected-width I/O, #503 per-column decode CPU, #434 zone-map survival), so the fallback now prices its input exactly as the real scan does.
- The SetRelPathlist extraction is behavior-identical: scan_decode_cost, zonemap_cost, analyze_stats (pins #171), column_projection, native_groupagg, parallel_vector_agg, native_agg, rewrite_group_scan all pass -- a changed term would have moved those costs and reddened them.
- native_groupagg_wide_cost forces the fold path on a 25-column table and the fold result matches a plain aggregate.

One observation, non-blocking: the new test proves the fallback path produces correct results, but it does not directly assert that the fallback's COST is now the refined one (the actual #12 change) -- that rests on the shared-helper construction (the same, proven-identical, helper). A plan-choice or cost-delta assertion would guard it directly, but cost assertions are fragile and the construction argument is sound, so I would not block on it. Good to merge.

PR #10, opened 2026-07-19
docs + benchmarks for the encoding/execution roadmap (I1-I8)

jdatcmd:
Stacked on #9 (I8). Documentation and benchmark refresh for the whole roadmap.

- **README**: feature list (encodings, compressed execution, bloom skipping, typed vectorized predicates, late materialization), new GUCs, updated limitations, and refreshed benchmark numbers.
- **docs/ARCHITECTURE**: new `columnar_encoding.c` and `columnar_bloom.c` module entries; reader/scan/vector entries updated for encoding decode, compressed execution, bloom probing, and late materialization.
- **design/IMPROVEMENT_PLAN**: status banner — I1–I8 implemented.
- **PROVENANCE**: clean-room record for I1–I8.
- **bench/sample_output_pg17.txt**: format-2.1 run (pg17 non-assert, 3M rows).

Highlights from the refreshed benchmark: table-only size heap **289 MB** vs columnar **40 MB** (encodings only) / **24 MB** (encodings + zstd); sum/avg ~2x, skippable filtered aggregate ~6.5x, wide projection ~5.7x; **point lookup ~2.35 ms** (down from hundreds of ms before the encoding work); vectorization ~2.8x on aggregates.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #682, opened 2026-08-16
perf: batch fold iterates needed columns, not all natts (#10)

OffgridwithJD:
## Iterate the needed columns in the batch fold, not all of them

Audit finding #10 (optimization).

`pgcolumnar_native_batch_fold`'s per-row loops walked `0..natts` and tested a per-column `cneeded` mask: the phase-1 gather, the three deferred-payload consume paths, and the phase-2 payload materialize. On a wide table a fold that reads a few columns paid loop-and-branch cost proportional to the whole table width, per row, and twice on a deferred group.

### Fix
Build compact `keyCols` / `payloadCols` lists once (and `phase1Cols` per group, since `deferOn` is per group) and iterate those. Each column's fold state is independent, so column order is unchanged and the transformation is behavior-identical.

### Correctness (bench, PG 18.4)
`test/native_batch_fold_projection.sh` forces the fold on a wide table and asserts its result equals a plain aggregate. It is **load-bearing**: I verified that dropping one column from the payload list makes `sum(v)` return `0` instead of `-32`, and the test catches it (RED). The existing fold suites all still pass: `parallel_vector_agg` (fires `Batch Fold: yes`), `native_agg`, `native_groupagg`, `native_agg_deletes`, `native_agg_addcolumn`, `native_vecdecode`, `rewrite_group_scan`.

### Performance (A/B, same query, fixed vs unfixed, interleaved)
A 46-column, 400k-row table, `SELECT count(*), sum(v) WHERE k < 400` (one column referenced): **~17ms unfixed -> ~13ms fixed, about 23% faster**, results identical. The win scales with the table's column count.

This is the last of the six optimization/modularity audit findings.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,12 @@ which was true until that script existed.
 
 ### Fixed
 
+- The ungrouped batch fold's per-row gather now steps over the columns a query
+  references rather than all of a table's columns. On a wide table a scan that
+  reads a few columns walked every column per row (twice on a deferred group),
+  which is loop overhead proportional to the table width. It now iterates compact
+  key/payload lists; a 46-column single-aggregate fold measured about 23% faster
+  with no change in results. Regression test: native_batch_fold_projection.
 - The Iceberg foreign-data wrapper now pushes projection down: it decodes only
   the columns a query references (from the output list and the recheck quals),
   not every column of every surviving file. A narrow projection over a wide table
--- a/src/columnar_vector.c
+++ b/src/columnar_vector.c
@@ -3139,6 +3139,21 @@ pgcolumnar_native_batch_fold(PgColumnarAggScanState *state, Relation rel,
 	bool	   *ciskey = (bool *) palloc0(sizeof(bool) * natts);
 	Datum	   *cval = (Datum *) palloc0(sizeof(Datum) * natts);
 	bool	   *cisnull = (bool *) palloc0(sizeof(bool) * natts);
+	/*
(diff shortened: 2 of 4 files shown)


jdatcmd (approved):
Approve. This is the one I scrutinized hardest, since a wrong column list here silently corrupts aggregates. Verified on the bench (PG 18.4).

- Behavior-identical by construction: keyCols/payloadCols are built by splitting exactly the cneeded set on ciskey, so their union is the same needed set the mask covered; phase1Cols is rebuilt per group as keys + (deferOn ? nothing : payloads), matching the original per-group deferral. Each column's fold state is independent, so iterating the compact lists in column order is equivalent.
- Green across the fold shapes: native_batch_fold_projection, parallel_vector_agg (fires Batch Fold: yes), native_agg, native_groupagg, native_agg_deletes (deferred + deletes), native_agg_addcolumn, native_vecdecode, rewrite_group_scan, ungrouped_vector_agg all pass.
- The test is non-vacuous and load-bearing, verified independently: I dropped the last payload column from the compact list and the projection test reddened -- got [160000|0] vs want [160000|-32], i.e. sum(v)=0 -- while 'Batch Fold: yes actually ran' and 'ref is non-empty' both still passed. So it asserts the fold path fired AND that an under-built payload list changes the aggregate, which is the exact silent-corruption failure to guard.

The ~23% win on a wide table is the expected shape (loop/branch cost now proportional to referenced columns, not table width). Good to merge. That closes all six optimization/modularity findings, and the audit.

Issue #683, opened 2026-08-16
Concurrent same-row UPDATE loses an update (row duplication); decide the contract


OffgridwithJD:
## Concurrent same-row UPDATE loses an update (row duplication)

This is the UPDATE facet of the old #5 (the insert facet is closed). It is
currently carried as a known limitation in `docs/limitations.md` (Concurrency).
Opening it so we can decide the contract, now that both fix options are scoped with
proof. This is a reviewer writeup; the merge and the contract call are yours.

### The defect, proven

A columnar UPDATE is delete-old plus insert-new, returns TM_Ok, takes no lock on
the row identity, and `pgcolumnar_tuple_lock` raises UNSUPPORTED, so core never runs
EvalPlanQual. Two sessions updating the same row each keep their own new version:
the old row is deleted once and both new rows survive, so the row is duplicated and
one update is lost. A heap serializes this transparently.

A RED reproduction is built: `test/update_conc.sh`. It drives two psql sessions over
FIFOs, forces the overlap with a `pg_stat_activity` lock-wait barrier (no sleeps),
and diffs a columnar table against a heap oracle running the byte-identical
interleaving. Proven RED and non-vacuous on today's main (pg18a assert): every
columnar arm fails while the heap oracle, a serial non-overlapping control (stays a
single row), and a covering-unique-index contrast all pass.

Observed states on today's code:

| Race (READ COMMITTED unless noted) | Heap oracle | Columnar today |
|---|---|---|
| update vs update | 1 row, {20} | 2 rows, {10,20} (duplicate + lost) |
| update vs update, REPEATABLE READ | 1 row, serialization_failure | 2 rows, no error |
| update vs delete | 0 rows | 1 row (delete lost) |
| delete vs update | 0 rows | 1 row (committed delete undone) |
| self-update in one statement | error, 1 row | 2 rows |
| multi-row overlap (41 rows) | 200 | 241 |

Two things I measured, not assumed:
1. The second writer does block, on the issue-#4 chunk-group advisory lock, not a
   row lock. That lock serializes the delete-vector read-modify-write but not the row
   identity, so both appends still happen.
2. SERIALIZABLE is already safe today: the loser gets a serialization_failure with no
   fix. So the defect is scoped to READ COMMITTED and REPEATABLE READ.

### Two fix options, scoped

**Option A, minimal (the floor).** Extend the advisory-lock machinery from unique-key
identity to row identity: before marking the old row deleted, lock (storage,row),
[...]

PR #684, opened 2026-08-16
Serialize concurrent same-row UPDATE/DELETE (issue #5, UPDATE facet)

OffgridwithJD:
## Serialize concurrent same-row UPDATE/DELETE (issue #5, the UPDATE facet)

Closes the lost-update / row-duplication limitation documented in
`docs/limitations.md`, and opened for a contract decision in #683. The maintainer
call in #683 is the READ COMMITTED contract; this PR implements the minimal option
(Option A) that #683 recommended.

### The defect

A columnar UPDATE is delete-old plus insert-new and takes no lock on the row
identity; `pgcolumnar_tuple_lock` is unimplemented, so core never runs
EvalPlanQual. Two transactions updating the same row each kept their own new
version, so the row was duplicated and one update was lost. A DELETE racing an
UPDATE was lost the same way, and a self-join `UPDATE ... FROM` duplicated the row
with no concurrency at all.

### The fix

New module `src/columnar_row_lock.c` gives the write paths a row-identity guard,
modeled on the concurrent unique-key insert lock (`columnar_unique.c`) and the
issue #4 chunk-group lock. Before marking the old row deleted,
`pgcolumnar_tuple_update` and `pgcolumnar_tuple_delete`:

1. report a same-command self-modify (buffered-deleted row) as `TM_SelfModified`,
   so a self-join resolves as it does for a heap (this runs even with the GUC off);
2. take a transaction-scoped advisory lock on `(storage, row)`; a no-wait caller
   that would block gets `TM_BeingModified`;
3. re-read the row's liveness in the LATEST committed state (`GetLatestSnapshot`,
   not the caller snapshot, which `PgColumnarCatalogSnapshot` only copies) and, if a
   concurrent transaction committed a delete, raise a retryable
   `serialization_failure`.

The loser of a same-row race now gets one retryable `serialization_failure` at
every isolation level, matching a heap at REPEATABLE READ and what columnar already
gave at SERIALIZABLE. READ COMMITTED is intentionally stricter than a heap: the
loser retries rather than re-applying via EvalPlanQual. Transparent RC re-apply
would need a per-row version chain and an eager per-row catalog write, measured at a
permanent ~40% UPDATE penalty plus a new vacuum reaper; see #683. No on-disk, WAL,
or catalog format change here.

New GUCs: `pgcolumnar.enable_row_update_lock` (USERSET, default on) and
`pgcolumnar.row_lock_buckets` (POSTMASTER, default 1024, bounds held locks per
storage so a bulk UPDATE cannot exhaust the lock table).

### Proof (pg18a, assert)

- New suite `test/update_conc.sh` drives two psql sessions over FIFOs, forces the
  overlap with a `pg_stat_activity` lock-wait barrier (no sleeps), and diffs a
  columnar table against a heap oracle running the byte-identical interleaving. It
  is RED on `main` and GREEN with this fix across update/update (RC, RR, SER),
  update/delete, delete/update, self-update in one statement, and multi-row overlap.
  A serial non-overlapping control and a covering-unique-index contrast bound it.
- Removal proof: on the fixed binary, `enable_row_update_lock=off` brings the
  duplicate back (count 2), `on` keeps it fixed (count 1).
- No regression: `concurrent_diff`, `phase3`, `concurrency`, and `unique_conc` all
  pass with the fix, so disjoint concurrent DML, insert-then-update read-your-writes,
  savepoints, same-group deletes, and unique-key serialization are unaffected.
- `harness_selftest` is clean (suite registered, SUITES C-sorted, no lint).

Two behaviors were measured on the bench, not assumed: the second writer blocks on
the issue #4 chunk-group advisory lock (not a row lock), and SERIALIZABLE was
already safe on `main`, so the defect was scoped to READ COMMITTED and REPEATABLE
READ.

`update_conc` is registered in `run_all_versions.sh`; docs and CHANGELOG are updated
and the limitation is removed. Full-matrix (18 and 19, preflight on all five majors)
runs in CI.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,14 @@ which was true until that script existed.
 
 ### Fixed
 
+- Concurrent `UPDATE` or `DELETE` of the same columnar row now serializes on the
+  row identity, so the losing writer gets a retryable `serialization_failure`
+  instead of duplicating the row and losing an update (issue #5, the UPDATE facet).
+  Two sessions updating one row without a covering unique index each kept their own
+  new version. The fix takes a transaction-scoped advisory lock on the row, then
+  re-reads the committed delete vector under a fresh snapshot before writing. New
+  GUCs `pgcolumnar.enable_row_update_lock` (default on) and
+  `pgcolumnar.row_lock_buckets` (default 1024). Regression test: update_conc.
 - The ungrouped batch fold's per-row gather now steps over the columns a query
   references rather than all of a table's columns. On a wide table a scan that
   reads a few columns walked every column per row (twice on a deferred group),
--- a/Makefile
+++ b/Makefile
@@ -18,6 +18,7 @@ OBJS = \
 	src/columnar_vector.o \
 	src/columnar_vacuum.o \
 	src/columnar_unique.o \
+	src/columnar_row_lock.o \
 	src/columnar_arrow.o \
 	src/columnar_flatbuffers.o \
 	src/columnar_thrift.o \
--- a/docs/limitations.md
+++ b/docs/limitations.md
@@ -409,7 +409,10 @@ same way as a `CREATE INDEX` that is not concurrent. Turn projection scans off w
   `pgcolumnar.enable_row_update_lock` (default on) controls this serialization.
   The GUC `pgcolumnar.row_lock_buckets` (default 1024) bounds the held locks per
   storage, so a bulk update cannot exhaust the lock table. Unrelated rows may
-  share a bucket, which only makes them wait when they need not.
+  share a bucket, which only makes them wait when they need not. For the same
+  reason, two multi-statement transactions can rarely deadlock, even on disjoint
+  rows. This happens when their bucketed locks collide in opposite statement order.
+  Retry resolves it, as with any serialization failure.
 - Concurrent inserts of the same unique key go in sequence. The server therefore
   always finds the conflict. Before a new row reaches the uniqueness check, the
   access method takes an advisory lock with the scope of the transaction. The key
--- a/docs/limitations.md
+++ b/docs/limitations.md
@@ -399,16 +399,17 @@ same way as a `CREATE INDEX` that is not concurrent. Turn projection scans off w
   commit. It then reads the committed mask again and merges its own bits. Thus
   both sets of delete marks survive. Writes to different row groups continue at
   the same time.
-- A concurrent `UPDATE` of the same row is a delete of the old version plus an
-  insert of the new one. The two writers do not serialize on the row itself.
-  Without a covering unique index, two sessions that update the same row can each
-  keep their own new version. The row is then duplicated and one update is lost.
-  A unique index on the updated key makes the writers go in sequence, and the
-  conflict is found. The second writer then gets a unique-violation error, which
-  the application retries. This is not the transparent serialization a heap gives,
-  where both updates apply. This is a known limitation. Add a unique index and
-  retry on conflict, or serialize the updates in your application, where
-  correctness needs it.
+- A concurrent `UPDATE` or `DELETE` of the same row serializes on the row
+  identity. The second writer waits for the first to commit. It then gets a
+  retryable `serialization_failure`, so the row is never duplicated and no update
+  is lost. This holds at every isolation level and needs no unique index. Retry
+  the transaction on this error, as you retry one at `REPEATABLE READ`. The
+  behavior is stricter than a heap at `READ COMMITTED`. A heap re-applies both
+  updates transparently there; pgColumnar makes the loser retry. The GUC
+  `pgcolumnar.enable_row_update_lock` (default on) controls this serialization.
+  The GUC `pgcolumnar.row_lock_buckets` (default 1024) bounds the held locks per
+  storage, so a bulk update cannot exhaust the lock table. Unrelated rows may
+  share a bucket, which only makes them wait when they need not.
 - Concurrent inserts of the same unique key go in sequence. The server therefore
   always finds the conflict. Before a new row reaches the uniqueness check, the
   access method takes an advisory lock with the scope of the transaction. The key
--- a/src/columnar.h
+++ b/src/columnar.h
@@ -790,6 +790,7 @@ extern char *PgColumnarDecompressValueStream(const char *comp, uint32 compLen,
 #define PGCOLUMNAR_LOCKCLASS_DELETE_VECTOR	101
 #define PGCOLUMNAR_LOCKCLASS_STORAGE_ROW	102
 #define PGCOLUMNAR_LOCKCLASS_UNIQUE_KEY		103
+#define PGCOLUMNAR_LOCKCLASS_ROW_IDENTITY	104
 
 /* -------------------------------------------------------------------------
  * concurrent unique-key insert serialization (pgcolumnar_unique.c, issue #5)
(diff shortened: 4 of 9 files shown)


jdatcmd (changes requested):
The core fix is sound and I verified it hard — but there is a CI-blocking regression to fix, plus a coverage gap. Details, good and bad.

## What I verified GREEN (the fix is correct)
- **Read-your-writes: no false 40001** (your ask #2, the headline risk). On the fixed .so: same-txn INSERT then UPDATE the inserted row (count=1, v=901), double UPDATE of one row (v=102), insert/update/delete churn (count=0), and the same at REPEATABLE READ — all succeed, none spuriously serialize. GetLatestSnapshot + PgColumnarRowIsLive correctly see the txn's own prior-command writes (MVCC self-visibility), so a self-inserted row is not read as committed-deleted.
- **update_conc GREEN** on the branch: 24 checks vs the heap oracle (RC/RR/SER, update/delete/self/multirow), serial control GREEN, unique contrast GREEN.
- **Removal proofs bite (non-vacuous):** (a) disabling the row lock returns the corruption — uu/RC=2, uu/RR=2, du/RC=1, multirow=241 — while self-update and the serial control stay GREEN; (b) swapping GetLatestSnapshot for the caller snapshot reds uu/RR (and uu/RC), proving the fresh snapshot is load-bearing.
- Regression: concurrent_diff, concurrency, unique_conc, phase3, native_dml all pass.

## Blocker: isolation.sh regresses (delete_vs_rewrite, recluster_vs_delete)
Your local regression list omitted `isolation`, and it is red on this branch (and in #684 CI, PG17+PG18). A DELETE racing a concurrent rewrite/recluster now hits the new committed-deleted recheck first: the rewrite retired the row's group, so PgColumnarRowCommittedDeleted reads "no committed group -> deleted" and raises the new serialization_failure, replacing the previous `ERROR: row group was compacted concurrently`. The expected outputs still carry the old message, so it fails.

The 40001-retry OUTCOME is correct (the DELETE's rowNumber is stale after the rewrite; retry re-scans and finds the row) and your plan anticipated it. Two things to fix:
1. Update `test/isolation/expected/delete_vs_rewrite.out` and `recluster_vs_delete.out`.
2. The message mislabels a rewrite as a same-row update/delete: `"concurrent update or delete of the same row"` fires for a recluster where nobody updated/deleted that row. Either broaden it (e.g. "... concurrent update, delete, or rewrite of the same row") or distinguish a genuine committed-delete from a group-rewrite/compaction and keep a precise message for each. A DBA seeing the current text on a delete-vs-recluster will be misled.

## Coverage gap (should-add)
update_conc.sh has no savepoint, no-wait, or AB-BA deadlock arm, though the plan listed them as acceptance criteria and the code implements TM_BeingModified (no-wait) and a txn-scoped lock (subxact over-hold). Please add at least a savepoint arm (ROLLBACK TO releases nothing early; a concurrent writer still completes without lock_timeout) and a no-wait arm, and confirm an opposite-order multi-row overlap deadlocks (40P01, one aborts) rather than corrupting.

Fix the isolation regression + message, add the arms, and I will re-verify (I will re-run the read-your-writes probes, the removal proofs, and isolation on the update). Nothing merges until CI is green and the owner is content with the RC contract (already ruled Option A).

jdatcmd (changes requested):
Re-review of `9d490cb`. The two CI reds from the prior pass are addressed and both suite matrices are green. Isolation fix and read-your-writes verified correct by reading: `committed_deleted_in` returns false when no covering committed group exists, so a rewrite-retired group falls through to Phase F3b's precise "row group was compacted concurrently" message (no expected-`.out` change, no mislabeling), and the current flush's delete bits are not yet in the catalog when the committed re-read runs.

One blocker remains, and it is the exact class Option A was chosen to eliminate.

**Deadlock still reachable — sort by the lock resource, not its pre-image.** `PgColumnarSerializeFlushRows` sorts the `RowLockEntry` array by `key` (columnar_row_lock.c:254) but the advisory lock is taken on `bucket = key % numBuckets` (:259). Modulo is not order-preserving, so sort-by-key does not acquire buckets in a consistent order across transactions. The comment's "every transaction acquires in the same order, so no AB-BA cycle forms" is true of the keys, not of the buckets actually locked.

Confirmed by construction using the real `row_identity_lock_key` at the **default** `row_lock_buckets = 1024`:

- Txn X updates rows `{0, 1}` → buckets `{155, 840}`; `key(0) < key(1)` → acquires **155 then 840**
- Txn Y updates rows `{70, 446}` → buckets `{840, 155}`; `key(70) < key(446)` → acquires **840 then 155**
- X holds 155 waits 840; Y holds 840 waits 155 → **AB-BA**. Disjoint rows, ordinary two-row UPDATEs.

`native_rewrite_conc` passing is timing luck: expected shared buckets ≈ M²/1024 for M-row updates, and given ≥2 shared buckets, ~50% land in opposite order. It will bite at scale/under load.

**Fix:** compute `numBuckets` before the sort, store `bucket` in each `RowLockEntry`, and order `rowlockentry_cmp` by `(bucket, rowNumber)`. Every transaction then acquires shared buckets ascending → genuinely deadlock-free; the exact-row committed re-read is unchanged.

Fix and re-push; I will re-verify (drive the deadlock live with `row_lock_buckets` set small to widen the window, then confirm it is gone). Holding merge until then.

jdatcmd (approved):
Approving `54fe3dd`. All three prior blockers resolved and independently re-verified on a PG17 lane (`pg17_nc`, separate prefix from the author's PG18 arm), plus CI green across the full matrix.

**Fix verified correct.** `RowLockEntry` now carries `bucket` (computed before the sort), `rowlockentry_cmp` orders by `(bucket, rowNumber)`, and `LockAcquire` uses `ents[i].bucket`. Ascending bucket order is a total order shared by every transaction, so no AB-BA cycle can form. The earlier AB-BA counterexample (rows `{0,1}` vs `{70,446}` at the default 1024 buckets) is eliminated by construction.

**Independent PG17 results (54fe3dd):**
- `native_rewrite_conc` PASS (former deadlock red)
- `isolation` 7/7 including `recluster_vs_delete` — precise "row group was compacted concurrently" message restored, no mislabel
- `update_conc` PASS including the new savepoint arm
- Neighbors PASS: `concurrency`, `unique_conc`, `concurrent_diff`
- Read-your-writes: 5/5 clean, **zero** `40001` (INSERT+UPDATE same txn, double UPDATE, ins/upd/del churn, REPEATABLE READ, self-join self-modify)

**Removal proof (clean, fails for the stated reason).** With `pgcolumnar.enable_row_update_lock=off` the suite reddens with the duplication/lost-update itself — `uu/RC` keeps 2 rows (`vals={10,20}`) and the loser returns `00000` not `40001`; `uu/RR` count 2 vs oracle 1; `du/RC` 1 vs 0; `multirow` 241 vs 200 — while the non-overlapping controls (`self-update`, `savepoint`) stay green. So the serialization is load-bearing, not a no-op. (An earlier "red" from running the copied suite out of `/tmp` was a harness-path artifact — missing `portlib.sh` — not the behavior; re-run in-tree it bites correctly.)

**Scope, agreed and documented.** Within-statement deadlock-freedom holds (the delete buffer has no size-triggered flush, so one statement = one sorted batch). Cross-statement AB-BA is inherent and matches heap; the one honest difference — columnar's can be a *false* deadlock on disjoint rows sharing a bucket — is now noted in `docs/limitations.md`.

Contract is Option A (retryable `40001` at READ COMMITTED), already ruled by the owner. LGTM.

Issue #559, opened 2026-08-10
The serial import/export functions check the server-file role and then no privilege on the table, so pg_read_server_files confers INSERT and pg_write_server_files confers SELECT on every columnar table


OffgridwithJD:
## Summary

Four entry points gate on the matching core server-file role and then never check
any privilege on the **relation** they read from or write to:

| function | file:line | missing |
| --- | --- | --- |
| `pgcolumnar.import_arrow` | `src/columnar_arrow.c:1757` | `ACL_INSERT` on the target |
| `pgcolumnar.import_parquet` | `src/columnar_parquet_reader.c:3028` | `ACL_INSERT` on the target |
| `pgcolumnar.export_arrow` | `src/columnar_arrow.c:980` | `ACL_SELECT` on the source |
| `pgcolumnar.export_parquet` | `src/columnar_parquet.c:1146` | `ACL_SELECT` on the source |

The effect is a privilege escalation. `pg_read_server_files` is documented as
"read files on the server"; here it additionally confers **INSERT into any
columnar table in the database**. `pg_write_server_files` is documented as "write
files on the server"; here it additionally confers **SELECT on any columnar
table**, exfiltrated to a file of the caller's choosing.

Both are non-superuser, deliberately grantable roles. That is the point of them.

## This is a bug and not a design stance, because the twins get it right

`pgcolumnar.parallel_copy` checks `ACL_INSERT`
(`src/columnar_parallel_copy.c:1294`) and `pgcolumnar.parallel_export_parquet`
checks `ACL_SELECT` (`src/columnar_parallel_export.c:471`), both via
`pg_class_aclcheck` + `aclcheck_error`. The pattern is already in the tree; four
call sites simply do not use it. This is the "duplicated logic must move
together" shape — the parallel variants were fixed and the serial ones were not.

## Reproduction

PostgreSQL 18.4 assert build, `main` at `744b7ca`, installed `.so` md5-verified
against the tree.

`carol` is `NOSUPERUSER`, holds `pg_read_server_files` + `pg_write_server_files`
and `USAGE` on the `pgcolumnar` schema, and holds **nothing** on `bobsecret`,
which is owned by `bob`:

```
carol superuser:             f
carol pg_read_server_files:  t
carol pg_write_server_files: t
carol SELECT on bobsecret:   f
carol INSERT on bobsecret:   f
ordinary SQL SELECT:      ERROR:  permission denied for table bobsecret
[...]

PR #685, opened 2026-08-16
fix: refuse four malformed/hostile Iceberg read-path inputs (#644)

### LINKED ISSUE #644: Iceberg phase-4a read path: 8 confirmed defects (2 backend crashes + a v1 regression), adversarial re-audit of #640–#643

jdatcmd:
Fixes the four defects from the #644 adversarial re-audit that were still live on `main` (the audit's #1, #4, #5, #8). PR #646 fixed the crash cluster (#2/#7/#3); #6 was resolved incidentally by name-mapping (#653). These four remained, including two medium-severity ones: a cancel-resistant DoS and a silent wrong result.

## The fixes (each refuses only real corruption)

- **#1 — FIFO / non-regular file hangs the backend (availability DoS).** `AllocateFile()`/`fopen("rb")` on a FIFO blocks in `open(2)`, and with `SA_RESTART` the block survives a query cancel and `statement_timeout`. New `PgColumnarRejectNonRegularFile()` `stat()`s the local path and refuses a non-regular file with `DATA_CORRUPTED` **before** the open, at all four Iceberg-reachable open sites (`ice_slurp_text`, `ice_slurp_bin`, `av_slurp_file`, `pq_source_open_cfg`). A regular file (or a symlink to one) and remote URLs are unaffected; a missing path still reports its real errno.
- **#4 — NULL manifest-list `sequence_number` decoded as 0.** The decoder discarded the Avro union's `have` flag. Added `has_sequence_number` to `PgColumnarAvroManifestFile`, keep the flag, and refuse when an ADDED entry must inherit a manifest whose sequence number is null (a v2/v3 list always carries a concrete number). A v1 list (absent column) still reads as 0; `read_manifest_list` now reports SQL NULL for an absent value.
- **#5 — position-delete NULL/negative `pos` silently dropped (phantom row).** `ice_read_pos_deletes` kept a row only when both columns were non-null (no `else`), and a negative `pos` was cast to a huge `uint64` that matched nothing. Both are refused now; `pos == 0` (the first row) stays valid.
- **#8 — dangling `current-schema-id`.** `ice_current_schema_fields` fell through to the deprecated top-level `"schema"` when the id matched no `schemas` entry, binding columns through a stale schema. It now refuses a dangling id, mirroring `ice_current_snapshot`; a legacy v1 table with only a top-level schema still reads.

## TDD evidence (RED → GREEN)

Two commits: a **RED** commit (`test:`) adds fixtures + arms, then a **GREEN** commit (`fix:`) adds the source. Checking out the RED commit is the removal proof — with the fixes absent, every defect arm is RED for its stated reason while every control is GREEN:

- #1: all seven FIFO arms report a wall-clock **HANG** (backend blocked, cancel-resistant); backend-still-up probes pass.
- #4: `sequence_number` decodes `f` (as 0); the scan silently returns the wrong row set.
- #5: null/negative `pos` → silent (no error), phantom row returned.
- #8: dangling id → silent success returning the wrong (stale-mapped) column values, while the resolvable-id and legacy-schema controls read correctly.

On the GREEN commit both suites pass: **iceberg_deletes 49 checks, iceberg_malformed 27 checks**, zero failures.

Note: the #8 fixture was strengthened during bring-up — the first version remapped `amount` onto the `id` (long) column, so the reader caught a *type mismatch* (42804) instead of the stated silent misprojection; the committed fixture remaps onto a second `int` column so the wrong read is genuinely silent (proves the actual defect).

## Regression

Eight neighbor suites covering every touched path pass on the fixed tree: `avro_manifest`, `iceberg_scan`, `iceberg_data_files`, `iceberg_name_mapping`, `iceberg_fdw`, `iceberg_dv`, `native_read_parquet`, `parquet_import`. Verified on a PG18 lane; full CI matrix (PG15–19) runs here.

Closes #644.


jdatcmd:
### Verification summary (for review)

**TDD (RED → GREEN).** The `test:` commit is the removal proof — checking it out (fixes absent) turns every defect arm RED for its stated reason while all controls stay green: #1 seven FIFO arms report a wall-clock HANG (cancel-resistant), #4 decodes `f`/silently drops a row, #5 null/negative pos silently drop (phantom row), #8 dangling id silently returns the stale-mapped column. The `fix:` commit turns them all GREEN: iceberg_deletes 49 checks, iceberg_malformed 27 checks, zero failures.

**Regression.** 8 neighbor suites covering every touched path pass: avro_manifest, iceberg_scan, iceberg_data_files, iceberg_name_mapping, iceberg_fdw, iceberg_dv, native_read_parquet, parquet_import. Full CI matrix (PG15–19) green.

**ASAN gate.** Built under `pg18_san` and exercised all four hostile refusal paths plus the success/control paths under AddressSanitizer: every refusal errors with the exact message, every control returns correct values, and the server log shows **no sanitizer findings and zero crash lines**. (The fixes are early-refusal guards, but the subsystem is the same one that carried the #2/#7 memory crashes, so the gate is warranted.)

**Adversarial verification.** Each fix was independently scrutinized (correctness, completeness across all entry points, false-positive safety on legitimate input, no new leak/UAF/errcode error) plus a cross-cutting critic. Verdict: all four CONFIRMED, "nothing material blocks review." Notes confirmed by that pass:
- #1 covers all four Iceberg/parquet-reachable open sites; symlinks-to-regular and missing-file errno passthrough preserved.
- #4 refuses only a v2/v3 present-null (v1 absent-column still reads 0); the have-flag is faithfully recorded because the OCF decoder zero-fills record slots.
- #5 keeps `pos == 0` valid and does not reject a large-but-valid pos; the DV/equality paths produce non-negative ordinals by construction, so no sibling guard is needed.
- #8 preserves the legitimate v1 legacy-only-schema path and the hybrid (schemas present, current-schema-id absent) fallback; the single-site fix covers all 8 callers.

**Out of scope, filed as #686:** the same FIFO `open(2)` DoS exists at `import_arrow` (unguarded) and `parallel_copy` (an *ineffective* guard that opens `O_RDONLY` before `fstat`), plus a cosmetic `min_sequence_number` null-reporting gap in `read_manifest_list`. Different subsystem, so not folded in here.

OffgridwithJD (dismissed):
Reviewed adversarially and reproduced on a PG18 assert lane. The four fixes are correct and the TDD is real: I independently confirmed the removal proof (checked out the tests-only commit `dab796d` and rebuilt) and every defect arm is RED for its stated reason, then GREEN at `8781214`.

**Independently verified**
- RED @ `dab796d` (fix absent): all six Iceberg entry points HANG on a FIFO; null/negative `pos` is silently dropped (`got []`); a null manifest-list `sequence_number` decodes as `0`; a dangling `current-schema-id` silently stale-reads. Controls pass.
- GREEN @ `8781214`: `iceberg_deletes` 49/49, `iceberg_malformed` 27/27.
- Code: #5 refuses null-or-negative `pos` and keeps `pos == 0` valid (guard is `< 0`, confirmed); #4 preserves the Avro union `have` flag and reports SQL NULL for a v1-absent column; #8 refuses a dangling id and still reads a legacy top-level schema. The #8 fixture note (remapped onto a second `int` column so the wrong read is genuinely silent) is the right call and I confirmed the RED arm shows `got []` (silent), not a type error.

**REQUEST_CHANGES — one proven gap (same DoS class as this PR's #1)**

The FIFO guard is placed at the four Iceberg/parquet open sites, but `pgcolumnar.import_arrow(rel, path)` opens a local path at `columnar_arrow.c:1839` (`AllocateFile`) with **no** `PgColumnarRejectNonRegularFile` guard. That is the identical cancel-resistant availability DoS #1 fixes. Reproduced on this branch (`8781214`, `statement_timeout=3000`):

```
import_arrow(FIFO)  -> exit 124 (HUNG; open blocked, statement_timeout did not fire)
iceberg_scan(FIFO)  -> exit 2   (refused fast, guarded)
```

`import_arrow` reads a server-side file, so it is the same privileged threat model this PR already treats as worth fixing. This PR's own comment says the guard "must sit at EVERY local open the read path can reach"; `import_arrow` is a read path it can reach. The fix is one line (add `PgColumnarRejectNonRegularFile(path)` before `columnar_arrow.c:1839`). Please either fold it in, or state explicitly that the Arrow read path is deferred to a separate issue so the DoS is tracked rather than dropped. `columnar_parallel_copy.c` already re-checks for a regular file (its comment at ~581), and `columnar_sink.c` is a write path, so those are fine; `import_arrow` is the one live gap.

**Non-blocking notes**
1. TOCTOU: the guard `stat()`s then the opener `open()`s, so a privileged caller could swap a regular file for a FIFO in between and still block the open. `stat`-before-open cannot close this; only opening `O_NONBLOCK` and `fstat`-ing the fd would (more invasive). Worth a one-line comment acknowledging the residual, since the header currently reads as fully closing the DoS.
2. Sanitizer: these are read/decode-path changes and the new malformed fixtures are exactly what ASAN/UBSAN wants to see decoded. The evidence is assert-only; a `pg18_san` run of `iceberg_malformed` + `iceberg_deletes` would complete the decode-surface gate (low risk here since the guards refuse before decode, but it is the standard gate for this surface).

Nice work on the RED-first structure and the honest fixture-strengthening note. Fold in the `import_arrow` guard (or track it) and this is good to go on the correctness side.

OffgridwithJD (commented):
Follow-up, withdrawing my change request. After posting I confirmed #686 already tracks the `import_arrow` FIFO DoS (and more) and explicitly scopes it out of #644. That satisfies the "fold in **or** explicitly defer so it is tracked" option from my review, so the deferral is the right call and #685 stays correctly scoped. I also see #686 records the `pg18_san` (ASAN) run came back clean on the refusal + success paths, which closes my sanitizer note.

**Correcting an error in my review:** I wrote that `columnar_parallel_copy.c` "already re-checks for a regular file, so those are fine." That is wrong. There are two open paths, and `columnar_parallel_copy.c:116` does `OpenTransientFile(path, O_RDONLY | PG_BINARY)` **before** any `S_ISREG` check, so a FIFO blocks there exactly like the sites this PR fixes. I cleared it by reading only the stat-guarded path at ~:597 and trusting the comment, instead of checking every open in the file. #686 already has this as "parallel_copy (ineffective guard)", so it is tracked. Flagging my mistake for the record.

So: no blocker in `#685` itself. The four in-scope fixes are correct and I verified them independently on a PG18 assert lane — RED at the tests-only commit (every defect arm fails for its stated reason, all six Iceberg entry points hang on a FIFO), GREEN at the fix commit (`iceberg_deletes` 49/49, `iceberg_malformed` 27/27). One forward-looking note for #686, not this PR: the FIFO guard is `stat`-before-`open`, which leaves a TOCTOU window; when you apply it to `import_arrow`/`parallel_copy`, opening `O_NONBLOCK` and `fstat`-ing the fd would be race-free (a FIFO opened `O_RDONLY|O_NONBLOCK` returns immediately instead of blocking), and it also removes the double-open.

One nit on #686's framing: the `import_arrow`/`parallel_copy` items are a proven cancel-resistant DoS, not cosmetic — only the `min_sequence_number` null-report is cosmetic. Worth prioritizing the two DoS sites accordingly.

I cannot re-approve under my own account (house rule: only jdatcmd approves), so please dismiss this change-request when you pick it up; my objection is withdrawn. Good, well-tested PR.

jdatcmd:
Update (`ca5bede`): folded the `import_arrow` FIFO guard into this PR rather than deferring it, since it's the same one-line helper and the same DoS class. `columnar_arrow.c` now calls `PgColumnarRejectNonRegularFile()` before its `AllocateFile` (after the existing authorization checks, preserving the #559 authz-before-file-access ordering). New regression arm in `arrow_import` (RED = wall-clock hang, GREEN = XX001), proven load-bearing by reverting the guard (the arm hangs). So this PR now closes the FIFO DoS at every trivially-guardable single-file open site (the four Iceberg/parquet sites + import_arrow).

Left in #686 (narrowed): `parallel_copy`'s `pcopy_open_regular_file`, whose fix is structural — it opens `O_RDONLY` before `fstat`, so it needs `O_RDONLY|O_NONBLOCK`+`fstat` (which also closes the stat-before-open TOCTOU), not just the helper — plus the cosmetic `min_sequence_number` null-report.

OffgridwithJD (commented):
Re-verified the import_arrow addition (`ca5bede`) independently on a PG18 assert lane. Folding it in is the stronger resolution of my original finding, thanks.

- Guard is at the open site (`columnar_arrow.c:1843`, `RejectNonRegularFile` immediately before `AllocateFile`).
- Functional proof: `import_arrow('t', FIFO)` now returns fast with the refusal (was a cancel-resistant hang before; my probe: exit 1, not 124), while a regular file still imports.
- Gated, not just added: the new `arrow_import.sh` FIFO arm uses `sqlstate_or_hang`, so an unguarded hang reads as RED (not a silent pass), and `arrow_import` is registered in `run_all_versions.sh` SUITES. It passes GREEN here.

No blocker in #685. The four #644 fixes I verified earlier (RED at the tests-only commit, GREEN 49+27) plus the import_arrow guard now close the FIFO DoS at every single-file open the read path reaches; #686 correctly carries the remaining `parallel_copy` ineffective-guard (a real DoS, O_NONBLOCK+fstat is the race-free fix) and the cosmetic `min_sequence_number` gap. Ready for jd on green. (I don't approve under my own account per our rule; this is a confirming note.)

PR #688, opened 2026-08-17
docs: refresh benchmark run to 2026-08-17 (7c873d5), row-lock measured free

OffgridwithJD:
Re-ran the main benchmark suite (`bench/run_bench.sh`) and FSST (`bench/run_bench_fsst.sh`) on the bench host at current `main` (`7c873d5`), after issue #5's row-identity lock (#684) and the Iceberg read-path hardening (#685). PG18.4 non-assert, 6M rows, median of 5, 16 cores / 62 GB.

**Result: storage, query latency, mutation, and FSST are all materially unchanged.** The fresh run certifies that the concurrency fix and the Iceberg hardening did not move the scan, storage, or write numbers.

**The one finding worth stating: #684's row-identity lock is free on the write path.** The single-row `UPDATE` is 1.90 ms, unchanged, and that path takes the lock and re-reads the committed delete vector for the one row. Toggling `pgcolumnar.enable_row_update_lock` does not change the time. The single-row case is the control here: it exercises the lock on one row, so if the lock cost anything measurable it would show there.

Note on the scattered 1000-row `UPDATE`: the main run reported 135 ms for it, but that is an outlier (it runs after the export/import churn, in a different cache state). An isolated median-of-5 re-run of the exact bench query put it at about 70 ms, in the same band as the prior 79 ms. It is the noisiest number on the page; the ratio story is unchanged.

This updates the run date/commit and adds a dated "What changed since the previous run" entry. The tables are re-confirmed, not replaced, matching this page's established pattern (the prior 2026-08-14 note did the same). `docs_style` passes.

Not in this PR: the cross-engine (TSBS) and ClickBench sections. Cross-engine is a separate ad-hoc run and is unaffected by #684/#685 (it measures the query path against other engines); ClickBench needs its `hits` dataset, which is not on the box.

### DIFF
--- a/docs/benchmarks.md
+++ b/docs/benchmarks.md
@@ -28,8 +28,9 @@ from the 2026-08-14 run at `122fd5c`. The main write-path change since then is i
 #5's row-identity lock. It serializes a concurrent `UPDATE` or `DELETE` of the same
 row and is on by default. It adds no measurable cost here. The single-row `UPDATE`
 is 1.91 ms, unchanged, and that path takes the lock on the one row. An isolated
-on-and-off control of that operation measured 0.59 ms and 0.63 ms, so the lock adds
-about 0.04 ms. Spot checks match: heap 707 MB against columnar-zstd
+on-and-off control of that operation, on a narrower table, measured 0.59 ms and
+0.63 ms, so the lock adds about 0.04 ms. Spot checks match: heap 707 MB against
+columnar-zstd
 5.95 MB, and `count(*)` at 0.04 ms. The
 conditions were PostgreSQL 18.4 non-assert, 6,000,000 rows, an 8-column table, the
 median of 5 repetitions, 16 cores and 62 GB of memory.
@@ -725,8 +726,9 @@ path:
   `serialization_failure` instead of duplicating the row. It is on by default and
   costs nothing measurable on the write path. The single-row `UPDATE` is 1.91 ms,
   unchanged, and that path takes the lock for the one row. An isolated on-and-off
-  control of it measured 0.59 ms with `pgcolumnar.enable_row_update_lock = off` and
-  0.63 ms on, so the lock adds about 0.04 ms. The scattered
+  control of it, on a narrower table, measured 0.59 ms and 0.63 ms with
+  `pgcolumnar.enable_row_update_lock` off and on. The lock adds about 0.04 ms. The
+  scattered
   1000-row `UPDATE` re-measured at about 70 ms, the same band as before. It is the
   noisiest number here.
 - **Iceberg read-path hardening (#685) and cost-model refinements (#679, #681,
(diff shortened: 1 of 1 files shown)


jdatcmd (commented):
Cross-reviewed. `docs_style` passes (verified locally on the PR's benchmarks.md), the methodology is honest, and the run reflects current main (`7c873d5`, which includes #684 and #685). Two things:

**One small fix before merge (accuracy/consistency).** The new prose says the single-row `UPDATE` is "1.90 ms" (the run-date line, and the maintenance note), but the mutation table and the paragraph below it say "1.91 ms":

```
| UPDATE single row by id | 0.03 ms | 1.91 ms | 64 |
```

The table is the source of truth; please make the two "1.90 ms" prose mentions read "1.91 ms" so a reader is not left reconciling two numbers for the same measurement. (This is the "print counts from the data, do not retype them into prose" rule.)

**Optional, to strengthen the headline.** "It adds no measurable cost ... turning the lock off does not change the time" is the load-bearing claim, and the instrument is well chosen: a single-row `UPDATE` still runs the per-flush committed re-read (`PgColumnarSerializeFlushRows` reads the row-group list under a fresh snapshot regardless of row count), so the fixed lock cost would show there if it were nonzero. It would be stronger to cite the actual on-vs-off pair (e.g. "1.91 ms on, 1.9x ms off, within the run-to-run band") rather than assert equality, so the "free" claim carries its own evidence. Not a blocker.

The 135 -> ~70 ms scattered-`UPDATE` outlier handling is honest and disclosed, and keeping the table at the reconfirmed 79.38 ms matches this page's established pattern. Everything else checks out. Fix the 1.90/1.91 prose and I will approve; jd merges.

jdatcmd (approved):
Approving. Re-verified `e7cf3f2`: no "1.90 ms" remains (single-row `UPDATE` reads 1.91 ms everywhere, matching the mutation table), the "free" claim now carries its measured evidence (0.59 ms off, 0.63 ms on, so about 0.04 ms of lock cost), the scattered-`UPDATE` outlier is honestly disclosed, and `docs_style` passes. The instrument is well chosen: a single-row `UPDATE` runs the per-flush committed re-read regardless of row count, so the ~0.04 ms delta is a fair measure of the lock's fixed cost. Good work.

One optional, non-blocking clarity note (do not need to hold merge for it): the control's absolutes (0.59/0.63 ms) sit below the table's 1.91 ms for the same single-row `UPDATE`, because the control runs on a narrower table. A reader who does not know that is left reconciling two numbers for one operation, which is the same shape as the 1.90/1.91 nit. A half-clause would remove it, for example "an isolated on-and-off control of that operation, on a narrower table, measured 0.59 ms and 0.63 ms". Your call; the load-bearing evidence is the delta either way.

jd merges.

jdatcmd:
Re-verified `8dc4a37`: the only change is the "on a narrower table" clause on both control mentions (plus a clean sentence split), and `docs_style` still passes. My approval stands for this commit. Ready for jd.

PR #689, opened 2026-08-17
feat: Iceberg REST catalog warehouse server option (#656)

### LINKED ISSUE #656: Iceberg REST catalog: per-catalog credentials via FOREIGN SERVER + USER MAPPING (and vended storage creds)

jdatcmd:
Closes #656 — the last open item of the Iceberg REST catalog credential model: the `warehouse` `CREATE SERVER` option, which was not just missing but actively rejected by the validator.

## What it does
- The `pgcolumnar_iceberg_catalog` validator now accepts an optional `warehouse` server option alongside `catalog_uri` (secrets still only on the USER MAPPING; the errhint lists it).
- `warehouse` is resolved from the server options and sent as the `?warehouse=` query parameter on `GET /v1/config` (Iceberg REST spec), percent-encoded, so a multi-warehouse catalog returns the config and prefix for that warehouse. `iceberg_rest_scan` and `iceberg_rest_table_location` send it; the listing endpoints (which never call `/v1/config`) pass NULL. A bare `http(s)://` catalog URI carries no warehouse.

## TDD evidence
Three new arms in `iceberg_rest_server`, proven RED on unfixed source and GREEN after the fix:
- `a warehouse option on the SERVER is accepted (not HV00D)` — RED: `got [HV00D]` (validator rejects); GREEN: accepted.
- `a warehouse-configured server resolves a table location` — RED: `got []` (no such server); GREEN: the metadata location.
- `the warehouse was sent on GET /v1/config (?warehouse=wh1)` — RED: `got [0]`; GREEN: `1`. The mock REST server now logs the GET query string so the parameter can be asserted.

Neighbor REST suites pass unchanged (`iceberg_rest`, `iceberg_rest_scan`, `iceberg_rest_vended`) — the resolve/loadTable signature threading did not regress them.

With this merged, #656 has nothing left; the credential model (FOREIGN SERVER + USER MAPPING, OAuth2 client-credentials, vended storage credentials) was already shipped.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,14 @@ which was true until that script existed.
 
 ## [Unreleased]
 
+### Added
+
+- The Iceberg REST catalog `FOREIGN SERVER` now accepts a `warehouse` option
+  alongside `catalog_uri`. It selects a warehouse on a multi-warehouse catalog
+  and is sent as the `?warehouse=` query parameter on the `GET /v1/config`
+  request. This completes the per-catalog credential model (issue #656).
+  Regression test: iceberg_rest_server.
+
 ### Fixed
 
 - The Iceberg read path now refuses four classes of malformed or hostile table
--- a/docs/how-to.md
+++ b/docs/how-to.md
@@ -325,11 +325,12 @@ SELECT * FROM pgcolumnar.iceberg_rest_tables('https://catalog.example.com', 'ana
 
 For per-role credentials, name a foreign server instead of a URI. The server
 holds the catalog URI. The current role's user mapping holds the bearer token,
-kept in `pg_user_mapping` where it is not world-readable.
+kept in `pg_user_mapping` where it is not world-readable. On a multi-warehouse
+catalog, add a `warehouse` server option; it is sent on the config request.
 
 ```sql
 CREATE SERVER cat FOREIGN DATA WRAPPER pgcolumnar_iceberg_catalog
-  OPTIONS (catalog_uri 'https://catalog.example.com');
+  OPTIONS (catalog_uri 'https://catalog.example.com', warehouse 'analytics_wh');
 CREATE USER MAPPING FOR analyst SERVER cat OPTIONS (token 's3cr3t');
 
 SELECT * FROM pgcolumnar.iceberg_rest_scan('cat', 'analytics', 'events')
--- a/docs/sql-reference.md
+++ b/docs/sql-reference.md
@@ -832,8 +832,10 @@ SELECT count(*) FROM pgcolumnar.iceberg_scan(
 ```
 
 The `pgcolumnar_iceberg_catalog` wrapper has a validator but no handler, so its
-servers cannot be selected from as tables. It accepts only `catalog_uri` on a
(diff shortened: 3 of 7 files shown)


OffgridwithJD (approved):
Approving under the cross-review model (jdatcmd-authored). Verified independently on a PG18 assert lane.

- The `warehouse` SERVER option is accepted (not HV00D), a warehouse-configured server resolves a table location, and it is forwarded to `GET /v1/config` as `?warehouse=wh1` — all three arms pass (24 checks). The value is percent-encoded via the existing `rest_pct_encode`, so a special-char warehouse cannot break or inject the URL.
- The test is non-vacuous: I broke the `?warehouse=` forwarding (to `?broken=`) and the "warehouse was sent" arm reddened (got 0, want 1), so it genuinely catches the forwarding.
- The threading is clean: `rest_resolve_server` reads the option, a bare URI carries no warehouse (NULL), and listing endpoints pass NULL. Neighbor REST arms (HV00D rejections, the no-token 28000 refusal) still pass. CI is green on both suite matrices.

One optional, non-blocking note: every warehouse arm uses `wh1`, which needs no encoding, so `rest_pct_encode` runs but is never exercised on a value that changes under encoding. A single arm with a warehouse containing a space or slash would make the percent-encoding carry its own proof. Not a blocker — `rest_pct_encode` is existing, used code.

Completes #656. jd merges.

PR #690, opened 2026-08-17
fix: parallel_copy no longer hangs the backend on a FIFO path (#686)

### LINKED ISSUE #686: FIFO open(2) DoS at import_arrow and parallel_copy (ineffective guard), + read_manifest_list min_sequence_number null (cosmetic)
Surfaced while fixing the #644 Iceberg read-path FIFO DoS (#685). The same cancel-resistant `open(2)`-on-a-FIFO denial of service exists at two non-Iceberg import entry points, and a manifest-list SRF has a cosmetic null-reporting gap. None is in #644's scope, so they are split out here.

### 1. `pgcolumnar.import_arrow` — unguarded open (FIFO DoS)
`columnar_arrow.c:1839` does `AllocateFile(path, PG_BINARY_R)` on the user-supplied SQL path with no regular-file screen. A FIFO argument blocks the backend in `open(2)`; with `SA_RESTART` the block survives a cancel / `statement_timeout` (same class as #644 #1). Gated by `pg_read_server_files`, the same privilege as the Iceberg reader.

Fix: call `PgColumnarRejectNonRegularFile(path)` (added in #685, `columnar_objstore.c`) before the `AllocateFile`.

### 2. `pgcolumnar.parallel_copy` — the FIFO guard is ineffective
`pcopy_open_regular_file` (`columnar_parallel_copy.c:111-135`) *intends* to reject FIFOs (its comment says so), but it does `OpenTransientFile(path, O_RDONLY | PG_BINARY)` first and `fstat` + `S_ISREG` after. `O_RDONLY` (no `O_NONBLOCK`) on a FIFO blocks inside the open, so the `S_ISREG` check at `:131` is never reached — the guard does not stop a FIFO. (The sibling boundary-scan open at `:585` is correct: it `stat()`s before `AllocateFile`.)

Fix: either `stat()` before the open (as `:585` does) or open with `O_RDONLY | O_NONBLOCK`, `fstat`, then clear the flag.

### 3. Cosmetic: `read_manifest_list` reports `min_sequence_number` 0 for a null
`columnar_avro.c:1423` emits `min_sequence_number` unconditionally, so an Avro-null decodes to SQL `0` (the same have-flag drop #685 fixed for `sequence_number`). Not a correctness defect — `min_sequence_number` is never consumed by the delete-application logic, only by this introspection SRF. Add a `has_min_sequence_number` flag mirroring `has_sequence_number` if SRF fidelity is wanted.

Threat model / severity: items 1-2 are a retryable-only... no — a *cancel-resistant* backend hang (availability DoS) reachable by a `pg_read_server_files` caller pointed at a FIFO. Same severity as #644 #1.

jdatcmd:
Closes #686. Scoped to the `parallel_copy` FIFO denial of service (the substantive half). The `min_sequence_number` cosmetic that was also in #686 is **deferred to #691** — adding a field to `PgColumnarAvroManifestFile` crashes `iceberg_scan` on an assert build (a latent memory bug that must be understood first), so the cosmetic is not worth shipping under it.

## The fix (medium-severity DoS)
`pcopy_open_regular_file` opened the path `O_RDONLY` and *then* `fstat`'d it, so a FIFO blocked inside `open(2)` before the `S_ISREG` check could run — and the block is cancel-resistant (`SA_RESTART` survives a cancel and `statement_timeout`). Its own comment claimed it "avoids an uninterruptible open() on a FIFO"; it did not.

It now opens `O_RDONLY | O_NONBLOCK` (a FIFO `open(2)` returns immediately), rejects a non-regular file with `WRONG_OBJECT_TYPE` (42809), then clears `O_NONBLOCK`. Because it `fstat`s the fd it holds, this also closes the stat-before-open TOCTOU: the file checked is the file opened.

## TDD evidence
New arm in `parallel_copy`: RED `got [HANG]` (blocked open), GREEN `42809`; reverting the O_NONBLOCK fix reproduces the hang. Full `parallel_copy` suite and `iceberg_malformed` pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,13 @@ which was true until that script existed.
 
 ### Fixed
 
+- `pgcolumnar.parallel_copy` no longer hangs the backend when its path names a
+  FIFO. `pcopy_open_regular_file` opened the path `O_RDONLY` and then checked for
+  a regular file, but a FIFO blocks inside that `open(2)` and the block survives a
+  cancel or `statement_timeout`, a denial of service. It now opens with
+  `O_NONBLOCK`, rejects a non-regular file, then clears the flag, which also
+  closes the stat-before-open race. Regression test: parallel_copy (#686).
+
 - The Iceberg read path now refuses four classes of malformed or hostile table
   metadata that an adversarial re-audit (#644) found it mishandled. A non-regular
   file (for example a FIFO) named as a metadata, manifest, or data path is
--- a/src/columnar_parallel_copy.c
+++ b/src/columnar_parallel_copy.c
@@ -113,7 +113,16 @@ pcopy_open_regular_file(const char *path, off_t *size_out)
 	int			fd;
 	struct stat st;
 
-	fd = OpenTransientFile(path, O_RDONLY | PG_BINARY);
+	/*
+	 * Open with O_NONBLOCK so a FIFO open(2) returns immediately instead of
+	 * blocking until a writer appears. A blocking open on a FIFO is uninterruptible
+	 * across a cancel or statement_timeout (SA_RESTART), a denial of service. We
+	 * then fstat the fd we hold and reject anything that is not a regular file, so
+	 * there is no stat-before-open race: the file we checked is the file we opened.
+	 * A regular file has O_NONBLOCK cleared below; it is a no-op on regular reads,
+	 * but leaving it set on some platforms is untidy.
+	 */
+	fd = OpenTransientFile(path, O_RDONLY | O_NONBLOCK | PG_BINARY);
 	if (fd < 0)
 		ereport(ERROR,
 				(errcode_for_file_access(),
@@ -134,6 +143,20 @@ pcopy_open_regular_file(const char *path, off_t *size_out)
 		ereport(ERROR,
 				(errcode(ERRCODE_WRONG_OBJECT_TYPE),
 				 errmsg("\"%s\" is not a regular file", path)));
+	}
+	{
+		int			flags = fcntl(fd, F_GETFL);
+
+		if (flags == -1 || fcntl(fd, F_SETFL, flags & ~O_NONBLOCK) == -1)
+		{
+			int			save_errno = errno;
+
+			CloseTransientFile(fd);
+			errno = save_errno;
+			ereport(ERROR,
+					(errcode_for_file_access(),
+					 errmsg("could not clear nonblocking mode on file \"%s\": %m", path)));
+		}
 	}
 	if (size_out != NULL)
 		*size_out = st.st_size;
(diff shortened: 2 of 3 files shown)


OffgridwithJD (changes requested):
The parallel_copy half is correct and I verified it — but the min_sequence_number half introduces a **segfault**, so REQUEST_CHANGES.

**parallel_copy FIFO fix: good.** `pcopy_open_regular_file` now opens `O_RDONLY | O_NONBLOCK`, fstats the held fd, rejects a non-regular file with 42809, then clears the flag. That closes the DoS and the stat-before-open TOCTOU (the file checked is the file opened), exactly the shape I suggested. `parallel_copy.sh` passes, including the new arm: "parallel_copy on a FIFO is refused, not a hang (42809)" and "backend still up".

**BLOCKER: the `columnar_avro.c/.h` change segfaults `iceberg_scan`.** Proven on a PG18 assert lane, clean builds on both sides:
- main (`7c873d5`): `iceberg_malformed` PASSES, 27 checks.
- this PR (`d314a18`, `make clean` build): `iceberg_malformed` **FAILS with signal 11** — `client backend was terminated by signal 11: Segmentation fault`, `Failed process was running: iceberg_scan(.../nullseq_ctl/...)`. It cascades: the postmaster reinitializes, so every later control arm reads empty.
- Removal proof: reverting **only** `columnar_avro.c` + `columnar_avro.h` to main (keeping the parallel_copy fix) removes the crash; the only remaining failures are the two `nullminseq` arms, i.e. the reverted feature. So the crash is the avro change, not parallel_copy.

Root cause is subtle and worth a note: this PR does not touch `columnar_iceberg.c`, so `iceberg_scan`'s manifest code is unchanged. The avro change adds one `bool has_min_sequence_number` to `PgColumnarAvroManifestFile` and sets it. The decode itself is benign, so the trigger looks like the **struct-size increase exposing a latent out-of-bounds** in the manifest processing (an array of these structs, or an off-by-one that previously landed in padding and now corrupts live memory). A `pg18_san` (ASAN) run of `iceberg_malformed` would pinpoint the exact write in one shot; I would start there. I confirmed it is deterministic (three runs) and a real SIGSEGV, not an OOM or my own stale build.

Fix the crash and I will re-verify. Happy to run it under ASAN and hand you the trace if that is faster for you. (Heads-up: this PR's own CI `suites` were still running when I reviewed; a clean-build crash in `iceberg_malformed` should turn them red too.)

OffgridwithJD (approved):
Re-verified the trimmed PR (`031662b`) and approving under cross-review.

The crash is gone: the whole `min_sequence_number`/`columnar_avro.c` change is dropped, so the diff versus main is `columnar_parallel_copy.c` only. On a PG18 assert clean build, `iceberg_malformed` passes (27 checks, no segfault) and `parallel_copy` passes (79 checks).

The parallel_copy FIFO fix — the part that closes my #686 catch — is unchanged and correct: `O_RDONLY | O_NONBLOCK` open so a FIFO returns immediately, fstat on the held fd (so the file checked is the file opened, no stat-before-open TOCTOU), reject a non-regular file with 42809, then clear O_NONBLOCK with error handling. The new arm "parallel_copy on a FIFO is refused, not a hang (42809)" and its backend-still-up probe both pass.

Good call splitting the latent crash out to #691. jd merges.

PR #692, opened 2026-08-17
fix: refuse a null Iceberg manifest path instead of crashing (#691)

### LINKED ISSUE #691: Latent memory bug: a PgColumnarAvroManifestFile struct-layout change crashes iceberg_scan on assert builds (ASAN-clean)
Adding a single field to `PgColumnarAvroManifestFile` crashes `iceberg_scan` on an **assert build**, with a `SIGSEGV` (signal 11) in the manifest read path. This is a latent memory bug: the code that consumes the decoded manifest-list array is size-correct on its face (`av_read_ocf` allocates `sizeof(PgColumnarAvroManifestFile) * n`, the consumer indexes `mfs[mi]`), yet a struct-layout change is enough to move the fault from benign padding onto live memory.

It surfaced while finishing #686. The `min_sequence_number` cosmetic (report a null as SQL NULL, mirroring the `has_sequence_number` fix from #644) adds one `bool has_min_sequence_number` to the struct and sets it in the decoder. That change alone triggers the crash, so the cosmetic is **deferred to this issue**; the #686 `parallel_copy` FIFO fix shipped separately in #690.

## Reproduction
- Apply: add `bool has_min_sequence_number;` after `min_sequence_number` in `columnar_avro.h`, set it in `av_decode_manifest_file` (`e->has_min_sequence_number = have;`), and emit SQL NULL for it in the `read_manifest_list` SRF (all mirror the merged `has_sequence_number` handling).
- On a **PG18 assert lane** (clean build): `test/iceberg_malformed.sh` FAILS with `client backend was terminated by signal 11`, failed process `iceberg_scan(.../nullseq_ctl/...)`; the postmaster reinitializes and later arms cascade. Deterministic across 3 runs. Reverting only `columnar_avro.c`/`.h` removes the crash (revert-isolated). (Independently reproduced by the reviewer.)

## What rules out the obvious, and points at the cause
- **Not a heap out-of-bounds.** Built under `pg18_san` (AddressSanitizer) and ran `iceberg_scan` on `nullseq_ctl`, `nullseq`, `danglingschema_ctl`, `legacyschema`, with and without the fixtures relocated (so `ice_rebase` runs): all correct, **no ASAN report**. A heap/stack OOB would trip ASAN.
- **Not seen on non-assert.** `test/iceberg_malformed.sh` passes on `pg18_nc` (non-assert), and the SRF arrays are correctly sized `[13]`.
- The assert-only-but-ASAN-clean signature points at an **uninitialized-memory read** (ASAN does not detect these; only MSAN/valgrind do), which the assert allocator exposes: with `RANDOMIZE_ALLOCATED_MEMORY`/clobbering, a `palloc` (not `palloc0`) chunk or a field the decoder does not always set is filled with a non-zero pattern, and after the layout shift a pointer read of it dereferences garbage. `av_read_ocf` uses `palloc0`, so the suspect is a *different* allocation of a related struct, or a manifest-file field read on a path where the decoder left it unset.

## Suggested next step
Run `iceberg_malformed` under valgrind or an MSAN build (not ASAN) on the assert lane to catch the uninitialized read, or attach a backtrace from the assert-build core (`-O0`), which will name the faulting function directly.

Severity: latent (not triggered on `main` today, only by a struct change), but it is a memory-safety fragility in the untrusted-metadata Iceberg read path, so worth understanding before the next field is added there.

jdatcmd:
Fixes the **live backend crash** part of #691 (a #644-class "refuse malformed, do not crash" gap). Does not fully close #691: the uninitialized-read *origin* that made #690's struct-layout change surface a null path on *valid* data, and the deferred `min_sequence_number` cosmetic, stay open there (they need valgrind/MSAN and only matter for re-adding the cosmetic).

## The bug (live on `main`, any build)
The reader decodes a manifest and manifest list against the Avro schema **embedded in the file**, which the table author controls. So a `manifest_path` (or a data/delete `file_path`) can be declared `union[null,string]` and encoded null. `ice_rebase` passed that null to `ice_strip_scheme`, which `strncmp`'d it → SIGSEGV. Reachable from `iceberg_scan` / `iceberg_data_files` / the Iceberg FDW. This is a genuine null-deref, not the assert-only artifact from #690 — it crashes a non-assert build too.

## The fix
`ice_rebase` refuses a null recorded path with `DATA_CORRUPTED` before touching it — the one choke every recorded path (manifest, data, delete) passes through — so the guard covers all callers.

## TDD evidence (RED proven on current main)
New arm in `iceberg_malformed`: a manifest-list whose `manifest_path` is the Avro union null branch.
- **RED on `main` (non-assert):** `iceberg_data_files` on it SIGSEGVs — `client backend was terminated by signal 11`; the arm gets `[]` and the "backend still up" probe fails.
- **GREEN after the guard:** `XX001`, backend still up. Reverting the guard reproduces the crash (guard-lines=0 confirms the revert).

Neighbor iceberg suites unchanged (`iceberg_scan`, `iceberg_deletes`, `iceberg_data_files`, `iceberg_name_mapping`, `iceberg_fdw`, `iceberg_dv`) — valid tables have non-null paths, so the guard never fires spuriously.


OffgridwithJD (approved):
Approving under cross-review. Independently verified the security fix and, importantly, the security *claim*.

I owed a proof that this is a live crash on a normal build (not the assert-only #690 artifact), so I built #692 on a **non-assert** PG18 (pg18n) and reverted only the `ice_rebase` guard back to main. Result: the hostile fixture SIGSEGVs the backend on a non-assert build:

```
client backend was terminated by signal 11: Segmentation fault
Failed process was running: iceberg_data_files(.../nullmanifestpath/.../t.metadata.json)
```

So a hostile Iceberg table whose manifest-list declares `manifest_path` as an Avro `["null","string"]` union and nulls it crashes the backend on shipped main. That is a genuine #644-class denial of service, confirmed independent of #690.

The fix closes it correctly:
- The guard is at the `ice_rebase` choke point (`recorded_path == NULL` -> `DATA_CORRUPTED`), which every manifest/data/delete path passes through. The #690 crash reached the same `strncmp(NULL)` via `iceberg_scan`; this one via `iceberg_data_files`. One guard covers both entry points.
- On the guarded build `iceberg_malformed` passes (29 checks) including "backend still up after the null-manifest_path refusal"; reverting the guard reproduces the crash. So the arm is load-bearing, not vacuous.

Small non-blocking note: the arm exercises the `manifest_path` case; a null *data* or *delete* path funnels through the same guard so is covered, but is not separately arm-tested. Fine as is.

Correctly scoped as "Part of #691" (the uninit-read origin and the min cosmetic stay there); my valgrind offer stands for the origin when you re-add the cosmetic. jd merges.

PR #693, opened 2026-08-17
fix: report a null manifest-list min_sequence_number as NULL (#686, #691)

### LINKED ISSUE #691: Latent memory bug: a PgColumnarAvroManifestFile struct-layout change crashes iceberg_scan on assert builds (ASAN-clean)

jdatcmd:
Closes #691. Re-adds the `min_sequence_number` cosmetic deferred from #686, now that #691's blocking "crash" is understood.

## Resolution of #691's "uninitialized-read origin": not a bug
The crash that blocked this — adding a field to `PgColumnarAvroManifestFile` segfaulting `iceberg_scan` — was a **stale-build artifact**, not a code defect. A PGXS lane built without `--enable-depend` did not recompile `columnar_iceberg.o` when the struct grew, so it read `manifest_path` at the *old* offset while `columnar_avro.o` wrote the *new* one (the `recorded_path=0x0` backtrace = a layout skew). A from-scratch build has no skew. Confirmed on a clean build here, and independently by the reviewer on a fresh assert build. (The chase was not wasted — it surfaced #692, a real null-`manifest_path` DoS, already merged.)

## The change (display-only)
`read_manifest_list` reported a null `min_sequence_number` as 0 (the decoder dropped the Avro union `have` flag — the pattern #644 fixed for `sequence_number`). Add `has_min_sequence_number`, keep the flag, and emit SQL NULL. The value is not consumed by the delete-application rules, so this only affects the introspection SRF.

## TDD (on clean builds)
New arm in `iceberg_malformed`: a manifest-list whose `min_sequence_number` is the union null branch. Proven **RED on a clean unfixed build** (`IS NULL` = `f`, i.e. decoded as 0) and **GREEN after the fix** (`t`), with a control that a present value still reads (7). Full `iceberg_malformed` passes on a `make clean` build with no crash — the exact struct-regrow that skewed the incremental lane, now verified skew-free.

Note for reviewers: please verify on a `make clean` build (this re-grows `PgColumnarAvroManifestFile`).


OffgridwithJD (approved):
Approving under cross-review. I verified this on a `make clean` build specifically, because it re-grows `PgColumnarAvroManifestFile` (`has_min_sequence_number`) — the exact struct change that skewed my incremental build and produced the #690/#691 phantom crash.

- Clean build (`make clean` + verified-fresh install): `iceberg_malformed` passes 3/3 with **no segfault**. That is the capstone confirmation that #690/#691's crash was the stale-object layout skew, not this change. Same struct regrow, clean build, no crash.
- The new arm is GREEN: a null `min_sequence_number` reads as SQL NULL, and the control reads its present value. The change matches the `sequence_number` handling exactly (flag from the Avro union `have`, `nulls[11] = true` else the value).
- Removal proof (non-vacuous): forcing `has_min_sequence_number = true` reddens the null arm (`got [f] want [t]`) while the present-value control stays green. So the arm genuinely tests the fix.

Correctly closes #691: the cosmetic lands here, the crash was already fixed in #692 (the real hostile-null-manifest_path DoS), and fix-1 (the "uninit origin") was not a bug. Good to merge on green; jd merges.

PR #694, opened 2026-08-17
fix: make Thrift and Avro field-skip loops interruptible (DoS hardening)

jdatcmd:
An aggressive decode-path security audit (prove-don't-trust + TDD) surfaced two uninterruptible-hang DoS defects of the same class, both reachable today by an honest caller who names a hostile file. Same class as the parallel_copy FIFO hang (#686).

## D-1 (HIGH) — Thrift footer skip
`PgColumnarThriftSkip` checked stack depth but not interrupts. TC_BOOL_TRUE/TC_BOOL_FALSE/default consume zero bytes and never set r->error, so a Parquet footer with an unknown `list<bool>` of count up to 0xFFFFFFFF (or a struct of many such lists) spins billions of no-op iterations ignoring SIGINT/statement_timeout. Reachable via parquet_schema()/read_parquet(). Reproduced: ~1.8 KB footer spun uninterruptibly, killed by wrapper at 25s while a 2s statement_timeout never fired; fixed = clean cancel (57014).

## D-2 (MEDIUM) — Avro manifest skip
`av_skip` checked interrupts per block, not per element; an array<null> block runs up to AV_MAX_OBJECTS (50M) zero-byte skips uninterrupted. Reachable via read_manifest/iceberg_scan.

## Fix
Both call CHECK_FOR_INTERRUPTS on the per-value skip path (off the hot data path).

## Test decode_skip_interrupts
Functional Thrift cancel (unfixed HANG, fixed 57014) + placement-sensitive shape (check must precede the switch; reads 0 on pre-fix av_skip despite its 2 per-block checks). Removal proof verified for both. No regression: native_parquet_stack, avro_manifest, decode_interrupts, native_read_parquet, docs_style, harness_selftest on pg18_nc.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,18 @@ which was true until that script existed.
 
 ### Fixed
 
+- The Thrift and Avro field-skip loops are now interruptible. A Parquet footer
+  whose unknown field is a `list<bool>` with a file-declared count up to
+  `0xFFFFFFFF` (or a struct holding many such lists) drove billions of zero-byte
+  skips through `PgColumnarThriftSkip`, which checked stack depth but not
+  interrupts, so the backend spun uncancellably on a sub-2 KB file reached through
+  `parquet_schema()` / `read_parquet()`. `av_skip` had the same gap on an
+  `array<null>` manifest block, where its interrupt check ran once per block
+  rather than per element. Both now call `CHECK_FOR_INTERRUPTS` on the per-value
+  skip path, so `statement_timeout` and cancel apply. This is a denial-of-service
+  hardening in the same class as the parallel_copy FIFO fix. Regression test:
+  decode_skip_interrupts (functional cancel for Thrift; per-element placement
+  shape for both).
 - `pgcolumnar.read_manifest_list` now reports a null manifest-list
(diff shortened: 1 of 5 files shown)


OffgridwithJD (approved):
Approving under cross-review. Verified on a make-clean pg18a assert build.

- `decode_skip_interrupts` passes (4/4): the Thrift skip-bomb footer is cancellable (not an uninterruptible hang), the backend survives, and both `PgColumnarThriftSkip` and `av_skip` carry the per-value/per-element `CHECK_FOR_INTERRUPTS`.
- Non-vacuous: I dropped the `CHECK_FOR_INTERRUPTS` from `PgColumnarThriftSkip`, rebuilt, and the "cancellable" arm reddened to `got [HANG] want [57014]` (query_canceled) plus the source-placement arm — so the check is genuinely what makes the skip-bomb interruptible.
- The placement is right: per skipped value in Thrift and per element in `av_skip` (before the switch, not once per block), which is the case that matters — a 50M-element `array<null>` or a nested Thrift skip is where an uncancellable loop bites. `Skip`/`av_skip` are off the hot decode path (unconsumed metadata only), so the added check is a cheap flag test with no meaningful cost, and CI suites are green.

Good DoS hardening. jd merges.

PR #695, opened 2026-08-17
perf: index delete_vector catalog reads instead of seq-scanning (O-2)

jdatcmd:
From the aggressive optimization audit. Reads of `pgcolumnar.delete_vector` passed `InvalidOid` / `indexOK=false` to `systable_beginscan`, so each **sequentially scanned the whole delete_vector catalog** filtered by scan key, despite the unique `delete_vector_pkey (storage_id, group_number)`.

`PgColumnarReadDeleteVectorList` runs **once per row group** while a scan builds its liveness cache, so on a table with deletes the cost was **O(row_groups × delete_vector_rows)**. Two more sites had the same gap (the reclaim deleted-count sum, `PgColumnarStorageHasDeleteVector`).

The sibling metadata tables (row_group, zone_map, bloom, column_chunk) already resolve their `_pkey` via `pgcolumnar_index_oid` and pass it to `systable_beginscan`; delete_vector now matches, using the exact in-file idiom.

## Why it is cost-only, not semantic
Index and seq scans run the same per-tuple visibility check in the heap fetch, so the `SnapshotDirty` caller is unaffected. Proven by the delete/reclaim/concurrency suites staying green: native_agg_deletes, iceberg_deletes, native_dml, native_reclaim, native_reclaim_reconcile, update_conc, native_vacuum_race.

## Measurement — `native_delete_vector_index`
Behavioural, size-independent, lag-free (`pg_stat_force_next_flush`): after a `sum(v)` scan of a 20-group table with a delete in every group, `pg_stat_all_tables` for `delete_vector` shows `idx_scan > 0`, `seq_scan = 0`.

**Removal proof:** reverting the index makes `seq_scan = 20` (one full-catalog scan per group — the O(G²) laid bare) and `idx_scan = 0`; both checks flip red. A correctness arm pins the delete result (`sum = 798400000`) unchanged.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,15 @@ which was true until that script existed.
 
 ### Fixed
 
+- Reads of the `delete_vector` catalog now use its `delete_vector_pkey` index
+  instead of a sequential scan. `PgColumnarReadDeleteVectorList` (called once per
+  row group while a scan builds its liveness cache), the reclaim deleted-count
+  sum, and `PgColumnarStorageHasDeleteVector` each passed `InvalidOid` to
+  `systable_beginscan`, so every call sequentially scanned the whole
+  `delete_vector` catalog filtered by scan key. On a table with deletes the cost
+  was `O(row_groups * delete_vector_rows)`. The sibling metadata tables already
+  index their reads this way; `delete_vector` now matches. Regression test:
+  native_delete_vector_index (asserts the reads take the index, seq_scan = 0).
 - `pgcolumnar.read_manifest_list` now reports a null manifest-list
   `min_sequence_number` as SQL NULL instead of 0, matching `sequence_number`. The
   value is not used by the delete-application rules, so this is a display fix.
--- a/src/columnar_metadata.c
+++ b/src/columnar_metadata.c
@@ -509,6 +509,7 @@ PgColumnarComputeFullyDeletedGroups(uint64 storageId, TransactionId oldestXmin)
 	HeapTuple	gtuple;
 	Snapshot	snap;
 	List	   *groups = NIL;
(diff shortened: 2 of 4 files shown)


OffgridwithJD (approved):
Approving under cross-review. Verified on make-clean pg18a assert builds (index arm and a main 08d1d62 control), plus a runtime engagement probe.

Correctness. `delete_vector_pkey` is `btree(storage_id, group_number)`. The two 2-key scans key on exactly `(storage_id, group_number)` and the 1-key scan on the leading `storage_id` -- all valid, efficient index prefixes, so results are identical. `pgcolumnar_index_oid` resolves the index by name (`get_relname_relid` in the pgcolumnar schema) and the name matches the DDL, so `OidIsValid(dvIdx)` is true by construction. Suites pass: native_delete_vector_index (yours), concurrency, native_dml, phase3.

Non-vacuous (the trap I checked for: a silent seq-scan fallback if the oid didn't resolve). Same fixture and workload on both builds, 400 delete-vector-read scans, stats reset first:
- #695: `delete_vector_pkey` idx_scan = 400, delete_vector seq_scan = 0.
- main:  idx_scan = 0, delete_vector seq_scan = 400.
So the index is genuinely engaged; the reads that used to seq-scan the catalog now probe the index.

No regression, real win at scale. At this fixture's N (200 delete_vector rows) the wall time is identical (2.92s both) -- the catalog scan is dwarfed by the data scan, so no small-N penalty. The main arm's per-lookup cost scales with total delete_vector rows (seq scan); #695's stays an index probe -- so the benefit lands exactly where O-2 targets it, heavily-deleted deployments with a large delete_vector. Test is registered in run_all_versions.sh (gated, not blind) and the change carries its CHANGELOG. jd merges.

PR #696, opened 2026-08-17
security: close stat-before-open TOCTOU in local file reads (S-F1)

jdatcmd:
From the security audit. The non-regular-file guard that keeps a FIFO from blocking the backend in `open(2)` (a cancel-resistant DoS, #644/#686) screened with `stat()` **before** a separate `open()`. That races: a local principal who can write the containing directory can swap the checked regular file for a FIFO in the window, re-introducing the block.

## Fix
All five local openers — `ice_slurp_text`, `ice_slurp_bin`, `av_slurp_file`, `import_arrow`, and the Parquet source `pq_source` — now go through one helper `PgColumnarOpenLocalRegularFile`: open `O_NONBLOCK` (a FIFO returns immediately instead of blocking), `fstat` the held fd, refuse a non-regular file (`DATA_CORRUPTED`), clear `O_NONBLOCK`. **The file it checks is the file it opened.** This is the pattern `pcopy_open_regular_file` already used; the racy `PgColumnarRejectNonRegularFile` is removed. Slurp callers collapse to one `PgColumnarSlurpLocalRegularFile` call (fstat gives the size). The Parquet source reads positionally with `pg_pread`.

## Testing — `local_open_race_free`
The TOCTOU **cannot** be tested deterministically: a static FIFO is refused by both the stat and the fstat form; only a lost race differs. So, per the `decode_interrupts.sh` precedent, the guard is a **source-shape assertion** — the opener opens `O_NONBLOCK` and `fstat`s, never stats a path before opening, and the racy helper is gone. **Removal proof:** reintroducing the stat-before-open helper flips the shape checks red. Functional arm: `read_parquet` / `read_avro_manifest` on a FIFO return a clean `XX001`, not a hang.

**No regression** across iceberg_malformed, arrow_import, avro_manifest, iceberg_scan, iceberg_data_files, parquet_import, native_parquet_streaming, native_parquet_codecs, parquet_nested, objstore_module on pg18_nc.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -88,6 +88,17 @@ which was true until that script existed.
 
 ### Security
 
+- Closed a stat-before-open race in the local file read path. The
+  non-regular-file guard that keeps a FIFO from blocking the backend in `open(2)`
+  (a cancel-resistant denial of service) screened the path with `stat()` before a
+  separate `open()`, so a local principal who can write the directory could swap
+  the checked regular file for a FIFO in the window between them and re-introduce
+  the block. All five local openers (the Iceberg metadata and manifest reads, the
+  Avro manifest read, `import_arrow`, and the Parquet source) now go through one
+  helper that opens `O_NONBLOCK`, `fstat`s the fd it holds, refuses a non-regular
+  file, then clears the flag, so the file checked is the file opened. The Parquet
+  source reads positionally with `pg_pread`. Regression test:
+  local_open_race_free.
 - Fixed a backend crash on a hostile Iceberg manifest with a null path. The
   reader decodes a manifest and manifest list against the schema embedded in the
   file, which the table author controls, so a manifest_path (or data or delete
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -93,12 +93,14 @@ which was true until that script existed.
   (a cancel-resistant denial of service) screened the path with `stat()` before a
   separate `open()`, so a local principal who can write the directory could swap
   the checked regular file for a FIFO in the window between them and re-introduce
-  the block. All five local openers (the Iceberg metadata and manifest reads, the
-  Avro manifest read, `import_arrow`, and the Parquet source) now go through one
-  helper that opens `O_NONBLOCK`, `fstat`s the fd it holds, refuses a non-regular
-  file, then clears the flag, so the file checked is the file opened. The Parquet
-  source reads positionally with `pg_pread`. Regression test:
-  local_open_race_free.
+  the block. The five transient-fd local openers (the Iceberg metadata and
+  manifest reads, the Avro manifest read, `import_arrow`, and the Parquet source)
+  now go through one helper that opens `O_NONBLOCK`, `fstat`s the fd it holds,
+  refuses a non-regular file, then clears the flag, so the file checked is the
+  file opened. The Parquet source reads positionally with `pg_pread`. The
+  parallel-copy partition coordinator (`pcopy_partition_aligned_offsets`), which
+  needs a stdio `FILE*` for `getline`, applies the same `O_NONBLOCK` open plus
+  `fstat` inline. Regression tests: local_open_race_free, parallel_copy.
 - Fixed a backend crash on a hostile Iceberg manifest with a null path. The
   reader decodes a manifest and manifest list against the schema embedded in the
   file, which the table author controls, so a manifest_path (or data or delete
--- a/src/columnar_arrow.c
+++ b/src/columnar_arrow.c
@@ -1761,7 +1761,7 @@ pgcolumnar_import_arrow(PG_FUNCTION_ARGS)
 	int			ncols;
 	ImpNode    *tops;
 	int			totalBuffers = 0;
-	FILE	   *f;
+	int			fd;
 	TupleTableSlot *slot;
 	PgColumnarIndexInsertState *indexes;
 	CommandId	cid;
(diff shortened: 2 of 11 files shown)


OffgridwithJD (changes requested):
The helper is the right fix and I verified it, but there's one local read it misses — the exact pattern this PR is titled to close — so REQUEST_CHANGES.

**Good, and verified (make clean build, pg18a assert):** `PgColumnarOpenLocalRegularFile` opens `O_RDONLY | O_NONBLOCK`, `fstat`s the fd it holds ("the file we checked is the file we opened", so no stat-before-open race), refuses a non-regular file with `DATA_CORRUPTED`, then clears `O_NONBLOCK`. The Iceberg/Arrow/Parquet/objstore read paths all route through it. `parallel_copy`, `arrow_import`, and `iceberg_malformed` all pass, no regression, no crash. This closes the TOCTOU I flagged on the #685 FIFO guard for those sites — thanks for taking it.

**REQUEST_CHANGES — one local read still has the stat-before-open TOCTOU.** `pcopy_partition_aligned_offsets` in `columnar_parallel_copy.c` (~:607-620) is not routed through the helper and keeps the old pattern:

```
stat(path, &st);              // TOCTOU: checked here...
if (!S_ISREG(st.st_mode)) ereport(...);
...
fp = AllocateFile(path, PG_BINARY_R);   // ...opened here, by path, and BLOCKING
```

A static FIFO is caught by the `S_ISREG` check, so this is the narrow swap-race residual: a local principal who can write the directory swaps regular->FIFO between the `stat` and the `AllocateFile`, and because `AllocateFile` has no `O_NONBLOCK` it then blocks in `open(2)` cancel-resistantly — the very DoS+TOCTOU this PR closes elsewhere. The comment there still says "matching pcopy_open_regular_file", but that function no longer stats-before-open, so the comment is stale.

The fix that keeps the `FILE*` getline needs its own shape (the helper returns a transient fd, which this path deliberately avoids): open `O_RDONLY | O_NONBLOCK`, `fdopen` it, `fstat` the fd, refuse non-regular, clear `O_NONBLOCK`. Or, if you'd rather scope this PR to the fd-returning sites, narrow the title/CHANGELOG so this residual is explicitly tracked rather than read as closed.

**Minor (non-blocking):** `pcopy_open_regular_file` (~:125) keeps its own inline `O_NONBLOCK`+`fstat` copy from #690 rather than the new helper, so the "single-source" is not quite complete for parallel_copy. Both are correct; just a DRY note.

Fix the partition-offset path (or scope the title) and I'll re-verify.

jdatcmd:
Good catch — fixed in cda1178. `pcopy_partition_aligned_offsets` was the sixth local opener and I missed it; it's exactly the same stat-before-open TOCTOU + FIFO `open(2)` hang.

Because it uses `getline()` it needs a stdio `FILE*` and can't take the transient-fd helper, so it now does the `FILE*` shape you suggested: raw `open(O_RDONLY|O_NONBLOCK|PG_BINARY)` → `fstat` the held fd → reject non-regular (`WRONG_OBJECT_TYPE`, unchanged) → clear `O_NONBLOCK` → `fdopen`. Raw `open`+`fdopen` (not `OpenTransientFile`) so `fclose` fully closes it with no fd.c transient bookkeeping dangling; every pre-loop error path closes the fd, and the getline loop's `fclose` runs under the existing `PG_TRY`. The two `FreeFile(fp)` became `fclose(fp)`. The stale "matching pcopy_open_regular_file" comment is replaced.

**Coverage added:**
- `parallel_copy`: a FIFO named to a **partitioned** target (the only way to reach this coordinator, via line 1370) is now refused with 42809, not a hang — the single-file path had this arm, the partition path didn't.
- `local_open_race_free`: a shape assertion that the partition coordinator opens `O_NONBLOCK` and never stats a path before opening it — RED on the pre-fix `stat`+`AllocateFile` shape.

**On the DRY note** (`pcopy_open_regular_file` keeping its own inline): left as-is deliberately. It raises `WRONG_OBJECT_TYPE` (42809) for a non-regular file while the read-path helper raises `DATA_CORRUPTED` (XX001) — each with committed test coverage (`parallel_copy` asserts 42809, `iceberg_malformed`/`arrow_import` assert XX001). Routing it through the shared helper would flip its SQLSTATE and break `parallel_copy`. Unifying that SQLSTATE divergence is a reasonable separate change but out of scope for the TOCTOU fix; happy to do it as a follow-up if you'd prefer 42809 everywhere.

OffgridwithJD (approved):
Re-reviewed cda1178. My REQUEST_CHANGES gap is closed; approving under cross-review. Verified on a make-clean pg18a assert build.

`pcopy_partition_aligned_offsets` was the last stat-before-open path (the sixth local opener). It now matches the hardened shape:
- `open(path, O_RDONLY | O_NONBLOCK | PG_BINARY)` so a FIFO open(2) returns immediately instead of hanging uninterruptibly (the availability DoS), then `fstat` on the held fd (no TOCTOU window), reject non-regular with 42809, clear O_NONBLOCK via fcntl, `fdopen` for getline.
- Raw open + fdopen (not OpenTransientFile) is the right call here since getline needs stdio and fclose then fully closes the fd with no fd.c transient bookkeeping dangling. I checked the fd hygiene specifically, because a raw fd is not auto-closed on ereport: every error path closes it first -- fstat-fail (errno preserved), non-regular, fcntl-fail (errno preserved), and even the fdopen==NULL path. No fd leak on any exit.
- The stale "matching pcopy_open_regular_file" comment is replaced with an accurate rationale.

Coverage is real and non-vacuous: parallel_copy now has a partitioned-target FIFO arm that passes as "refused, not a hang (42809)" -- a FIFO-hang test can only pass if the open is non-blocking, so it inherently reddens on the old stat+AllocateFile shape (which blocks forever). parallel_copy and local_open_race_free both pass.

On the deliberate 42809-vs-XX001 divergence you left in `pcopy_open_regular_file`: agreed, leave it. Each SQLSTATE has committed coverage and routing through the read-path helper would flip it. Unifying is a fine separate change; no need to fold it here. jd merges.

PR #697, opened 2026-08-17
perf: push parameterized predicates down for chunk-group skipping (O-4)

jdatcmd:
From the optimization audit. A qual like `col >= $1` (a `PARAM_EXTERN` from a prepared statement, PL/pgSQL, or the extended protocol) kept a `Param` operand, and the scan-key builder accepted only a `Const`, so **on a generic plan the scan built no key and read every chunk group**.

**Measured:** a 20 row-group table, `WHERE ts >= 39000` removed 19/20 groups as a literal but **0/20** as a generic-plan `$1`.

## Fix
Begin freezes **execution-stable** operands into `Const`s before building the keys: a `PARAM_EXTERN`, or any subexpression with no `Var`, no `PARAM_EXEC`, and no volatile function. Such an operand can't change within the execution, and the executor still re-applies the original qual to every surviving row, so the skip is exact.

### The correctness hazard, handled
A **correlated `PARAM_EXEC`** (nestloop inner scan whose qual references the outer row) changes per rescan, and the keys are built once and not rebuilt — freezing it would silently skip matching groups. It is deliberately **not** frozen. Evaluation is `PG_TRY`-guarded so a stable operand that raises leaves the clause unchanged.

## Test — `native_param_pushdown`
- Generic-plan `$1` now removes 19/20 groups.
- Parameterized count equals literal (1001).
- **Correlated `PARAM_EXEC`** over `thr(39000,38000)` stays correct (3002).
- **Removal proof:** bypassing the freeze reads 0 removed while correctness stays green (pruning is a pure optimization; the executor re-filters).

**No regression** across pushdown_report, native_skip, native_zonemap, planner_choice_quality, native_late_materialization, **differential**, parallel_vector_agg, native_parquet_pushdown on pg18_nc.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,16 @@ which was true until that script existed.
 
 ### Fixed
 
+- Chunk-group skipping now applies to parameterized predicates. A qual like
+  `col >= $1` from a prepared statement, PL/pgSQL, or the extended protocol kept a
+  `Param` operand, and the scan-key builder accepted only a `Const`, so on a
+  generic plan the scan built no key and read every chunk group. Begin now freezes
+  execution-stable operands (a `PARAM_EXTERN`, or a subexpression with no `Var`,
+  no `PARAM_EXEC`, and no volatile function) into `Const`s before building the
+  keys. A correlated `PARAM_EXEC` changes per rescan and is deliberately not
+  frozen, so results stay correct; the executor still re-applies the original qual
+  to every surviving row. On a 20-group test a `>= $1` generic-plan scan went from
+  reading all 20 groups to reading 1. Regression test: native_param_pushdown.
 - `pgcolumnar.read_manifest_list` now reports a null manifest-list
   `min_sequence_number` as SQL NULL instead of 0, matching `sequence_number`. The
   value is not used by the delete-application rules, so this is a display fix.
--- a/src/columnar_customscan.c
+++ b/src/columnar_customscan.c
@@ -63,6 +63,7 @@
 #include "statistics/statistics.h"
 #include "catalog/pg_statistic_ext.h"
 #include "utils/builtins.h"
+#include "utils/datum.h"
 #include "utils/lsyscache.h"
 #include "utils/pg_locale.h"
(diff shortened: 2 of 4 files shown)


OffgridwithJD (approved):
Approving under cross-review. Verified on a make-clean pg18a assert build with an adversarial fixture, exercising both the frozen path and the deliberately-excluded path.

The design line is the right one: freeze only execution-stable operands (PARAM_EXTERN from a prepared statement / plpgsql, or a stable subexpression) into a Const for skip-predicate use, and refuse to freeze anything containing a Var, a PARAM_EXEC, or a volatile -- because a wrongly-skipped group is never rechecked. I tested both sides of that line.

Frozen path (PARAM_EXTERN), forced onto the GENERIC plan (`plan_cache_mode=force_generic_plan`) so `$1` is a real runtime param, not a custom-plan inlined Const the old code already handled. `PREPARE p(int) AS ... WHERE k >= $1` over a 500k-row table (4 chunk groups). At six boundaries including the exact group edges (k>=500000 -> 1 row, k>=500001 -> 0 rows), EXECUTE equals the literal control byte-for-byte on (count, sum). And it is non-vacuous: EXPLAIN ANALYZE of the generic-plan EXECUTE shows `Pushed-Down Filters: 1`, `Chunk Groups Read: 1 / Total: 4`, `Removed by Filter: 3` -- the runtime param genuinely drives pruning where before it disabled it.

Excluded path (correlated PARAM_EXEC). A nestloop join whose inner columnar scan keys on `b.k = d.val` (per-rescan param, values spread across all 4 groups). nestloop `[5|2546920]` equals the hashjoin control `[5|2546920]` -- the correlated param is not frozen, so no group is wrongly pruned and no rows are dropped. This is the failure mode that would have been silent data loss; it does not happen.

Suites pass (native_param_pushdown, native_dml, phase3), the test is registered in run_all_versions.sh, and it carries its CHANGELOG. The guarded `pgcolumnar_try_eval_const` (returns NULL on a raise -> falls back to no pruning) is the right safety valve. jd merges.

Issue #314, opened 2026-08-01
Bloom filters are read for every column, not the columns a query filters on


OffgridwithJD:
Follow-on to #310, and independent of the fix there.

#310 removed the bloom read for groups a scan skips. This is about the groups a
scan keeps.

## The gap

`ColumnarReadBloomList` returns the bloom filters of every column of a row
group. `columnar_native_group_can_match` uses at most one of them per equality
predicate. A query with a single equality predicate on a twelve column table
therefore reads twelve bloom bitmaps and probes one.

The cost is not incidental. A bloom filter holds one bitmap per column sized by
the group's distinct values, so it is one of the larger things in the metadata
catalog.

## Measurement

One group of 200,000 rows, twelve int columns, scanned whole with an equality
predicate on one column:

| | buffers |
| --- | --- |
| as it stands | 715 |
| with the bloom read deleted entirely | 251 |

So 464 of 715 buffers are bloom reads, for one group, to probe one column. The
deleted-read figure is a floor and not a target, because the filter has to be
read for the column being probed. The headroom is the other eleven.

## Shape of a fix

Give `ColumnarReadBloomList` a column filter, or add a single-column lookup, and
have the caller ask only for the columns carrying an equality predicate. The
scan key already restricts to `(storage_id, group_number)`; adding the column
index makes it a narrower range on the same index.

The caller knows which columns it needs: `rs->predicates[p].attidx` for every
predicate with `strategy == BTEqualStrategyNumber` and `hasHash`.

## Why it is separate from #310

#310 is about not reading filters that are never reached. This is about not
reading filters that are reached but not used. They touch the same call but the
reasoning and the test are different: #310's test asserts that bloom reads follow
the groups a scan keeps, and this one would assert that they follow the columns a
query filters on.

OffgridwithJD:
Good distinction (skipped groups vs kept groups), but this is already covered by **PR #313** — which does more than #310's skip-time removal.

#313 doesn't just skip the bloom read for excluded groups. It replaces `ColumnarReadBloomList` (all columns) in `columnar_native_group_can_match` with a single-column read:

```c
NativeBloomMetadata *b = ColumnarReadBloom(rs->storageId, groupNumber,
                                           pred->attidx, rs->metaSnapshot);
```

[...]

PR #698, opened 2026-08-17
perf: read only the predicate columns' zone maps for skipping (O-1)

jdatcmd:
From the optimization audit (two agents flagged it independently). `pgcolumnar_native_group_can_match` and the per-vector skip builder read the **whole group's zone maps** — every attribute's min/max — and used only the columns carrying predicates. On a wide table that's min/max read+copy traffic **proportional to the table width**, for a scan that consults one or two columns. The per-column bloom fetch already avoids exactly this (#314); the zone map now does too.

## Fix
`zone_map_pkey` is `(storage_id, group_number, column_index, vector_index)`, so a per-column probe is a 3-key index scan. New `PgColumnarReadZoneMapForColumn` / `PgColumnarReadZoneMapVectorsForColumn` read one column's maps. `can_match` fetches each predicate column lazily (mirroring the bloom `byCol`/`lookedUp` pattern). `build_skipvec` reads each predicate column's per-vector maps; **the vector spans are identical across columns**, so they come from a predicate column, falling back to the whole-group read only when no predicate column has per-vector rows (phase-2 qual-only, or legacy). The four zone-map readers now share one tuple→struct builder.

## Measured
A 30-column table, one predicate, examining every group: **~1200 → 30** zone_map row fetches — matching a 2-column table (width-independent, ~40× on the wide table).

**Removal proof:** the whole-group read makes wide ≈ 15× narrow (test fails). Correctness — the spans drive vector positioning — held across native_zonemap, native_vecskip, native_skip, native_exact_selection, native_fetch_position, native_groupagg, **differential** on pg18_nc.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,17 @@ which was true until that script existed.
 
 ### Fixed
 
+- Group and per-vector skipping now read only the predicate columns' zone maps.
+  `pgcolumnar_native_group_can_match` and the per-vector skip builder read the
+  whole group's zone maps (every attribute's min/max) up front and used only the
+  columns carrying predicates, so on a wide table a scan that consults one or two
+  columns paid min/max buffer traffic proportional to the table width. Both now
+  probe per column via `zone_map_pkey`'s `column_index`, exactly as the per-column
+  bloom fetch already did; the per-vector spans (identical across columns) come
+  from a predicate column, falling back to the whole-group read only when no
+  predicate column has per-vector rows. On a 30-column table a one-predicate scan
+  went from about 1200 zone-map row fetches to 30, matching the 2-column table.
+  Regression test: native_zonemap_narrow.
 - `pgcolumnar.read_manifest_list` now reports a null manifest-list
   `min_sequence_number` as SQL NULL instead of 0, matching `sequence_number`. The
   value is not used by the delete-application rules, so this is a display fix.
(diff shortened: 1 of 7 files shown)


OffgridwithJD (approved):
Approving under cross-review. Verified on make-clean pg18a assert builds (narrow arm and a main 08d1d62 control), with a full A/B on both result sets and pruning counters, plus a zone_map-access measurement.

Correctness and no pruning regression. Reading only the predicate columns' zone maps (plus one column for the vector row-spans, which are column-independent) can only ever make pruning less aggressive, never wrongly prune -- so I checked the two things that could actually go wrong: a lost predicate column's pruning, and mis-reconstructed per-vector spans on a null-skewed column. A wide table (25 columns: monotonic, pseudo-random, a 2/3-null clustered column, low-card, plus fillers), 8-query battery run on both builds:
- Every query's (count, sum) is byte-identical between narrow and main -- no rows gained or lost.
- Every query's Chunk Groups Read / Removed / Vectors Skipped is identical -- pruning power is fully preserved, including `c IS NULL`, `c IS NOT NULL`, `c > k` on the null-heavy column and `d = 7 AND c IS NULL` mixing a low-card and a null predicate.

The optimization actually fires (non-vacuous). Same fixture, 25 columns, a single-column predicate `a >= 500000`, 200 scans, stats reset first: pgcolumnar.zone_map idx_tup_fetch drops from 640000 (main) to 244000 (narrow), a measured ~2.6x reduction in zone-map rows fetched -- the scan no longer materializes every attribute's zone map. Author test passes and is registered in run_all_versions.sh; phase3 and native_dml pass; carries CHANGELOG.

One observation, not a blocker: the reduction here is 2.6x rather than the ~columns-ratio one might expect, so there may be residual whole-width zone-map reading with headroom for a follow-up, but what is here is correct and a real win. jd merges.

PR #699, opened 2026-08-17
refactor: single-source the delete-vector visibility logic (M-1)

jdatcmd:
From the modularity audit. The delete-vector fold was copy-pasted in the seqscan and index-fetch liveness cache, and the per-row bit test open-coded in five places (seqscan x2, index-only scan, per-row fetch, buffered read-your-writes). Delete visibility that must be identical across every access path was hand-maintained in each -- the #644/#692 guard-in-one-twin shape, on the MVCC path.

Fix: the fold is now pgcolumnar_merge_delete_vectors; the bit test the inline dv_row_deleted (columnar.h, shared with columnar_delete_vector.c). Every site routes through them. Verified identical before consolidating.

Test native_delete_visibility_paths: deletes a known set, asserts the seqscan (row-emitting), index/bitmap, index-only, per-row index-fetch, and buffered same-transaction paths all return the same live set. Mutation proof: forcing dv_row_deleted false fails every path; the count(*) metadata fast-path is correctly unaffected. No regression across iceberg_deletes, native_agg_deletes, iceberg_dv, native_dml, update_conc, index_only, native_fetch_position, native_index, native_reclaim, differential on pg18_nc.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,19 @@ which was true until that script existed.
   request. This completes the per-catalog credential model (issue #656).
   Regression test: iceberg_rest_server.
 
+### Changed
+
+- The delete-vector visibility logic is single-sourced. The fold that ORs a row
+  group's delete bitmaps into one mask was duplicated in the sequential scan and
+  the index-fetch liveness cache, and the per-row bit test was open-coded in five
+  places across the sequential scan, the index-only scan, the per-row fetch, and
+  the buffered read-your-writes path. A change to one (a new bitmap encoding, a
+  bound) could silently diverge the others, invisible until one access path hit
+  the changed row. The fold is now `pgcolumnar_merge_delete_vectors` and the bit
+  test `dv_row_deleted`; every path routes through them. Behavior is unchanged.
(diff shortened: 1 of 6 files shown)


OffgridwithJD (approved):
Approving under cross-review. Verified on a make-clean pg18a assert build.

This is a genuine behavior-preserving single-source. It extracts the two pieces of delete-vector logic that were inlined in multiple places into shared primitives:
- `dv_row_deleted(mask, maskLen, row)` -- the bit test, replacing the inline `mask != NULL && (r>>3) < maskLen && (mask[r>>3] & (1<<(r&7)))` with identical NULL/too-short semantics.
- `pgcolumnar_merge_delete_vectors(maskList, want, outMask)` -- the "OR a group's bitmaps into one mask" loop, previously duplicated in native_load_group and the index-fetch liveness cache.

The semantics match the old inline code exactly, and every delete-vector-exercising suite passes: concurrency, concurrent_diff, phase3, native_dml, iceberg_deletes -- so the row-visibility and delete-merge behavior is unchanged. This is the good direction for the duplicated-logic risk: the two copies can no longer drift. jd merges.

PR #700, opened 2026-08-17
refactor: single-source the native encoding-descriptor wire layout (M-2)

jdatcmd:
From the modularity audit. The encoding descriptor is a column chunk's writer→reader contract. Its byte layout was **hand-packed** in `columnar_write_state.c` and **hand-parsed in three passes** in `columnar_reader.c`, so a field change meant editing five disjoint sites; and because the version byte doesn't move on a field change, a missed site surfaces as `DATA_CORRUPTED` or wrong values, not a clean rejection.

## Fix
New `columnar_encdesc.h` owns the layout (`PgColumnarEncdescPut*` / `ReadEntry`); the field offsets live only there. A `StaticAssert` ties `ENTRY_LEN`/`HEADER_LEN` to the field widths, so a width change is a **build error**, not a silent stride mismatch.

## Byte-identical (verified, not assumed)
Captured the `encoding_descriptor` bytea of a deterministic table from **both this branch and main** — all three column descriptors matched exactly (`020001000000052c010000b00400001200000000000000`, …). The Put helpers make the identical `appendBinaryStringInfo` sequence.

## Test — `native_encdesc_golden`
Decodes the descriptor and pins the version byte, vectorCount, per-entry valueCount/rawLen at their fixed offsets, and a self-consistent length.

**Removal proof (the interesting part):** swapping valueCount↔rawLen in the codec *consistently* (read/write still agree) leaves `native_roundtrip` **GREEN** — the format is self-consistent — but reddens `native_encdesc_golden`, because the on-disk offsets moved. That's exactly the silent-format-drift the round-trip suites **cannot** see and this codec prevents.

No regression across native_encoding, native_format, native_roundtrip, native_vecdecode, write_fsst_compressed, native_fastdecode, **differential** on pg18_nc.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -22,6 +22,18 @@ which was true until that script existed.
 
 ### Changed
 
+- The native encoding-descriptor wire layout is single-sourced. The descriptor
+  is a column chunk's writer-to-reader contract for its per-vector encoding; its
+  byte layout was hand-packed in `columnar_write_state.c` and hand-parsed in three
+  passes in `columnar_reader.c`, so a field change had to be made in five disjoint
+  places or a reader stride would land mid-field. Because the version byte does
+  not move on a field change, the version guard would pass and the mismatch would
+  surface as `DATA_CORRUPTED` or wrong values, not a clean rejection. A new
+  `columnar_encdesc.h` owns the layout (`PgColumnarEncdescPut*` / `ReadEntry`), and
+  a `StaticAssert` ties the entry length to the field widths. The on-disk bytes are
+  unchanged (verified byte-identical to the hand-packed form). Regression test:
+  native_encdesc_golden (pins the layout; catches a field-order change that the
+  self-consistent round-trip suites cannot).
 - The delete-vector visibility logic is single-sourced. The fold that ORs a row
   group's delete bitmaps into one mask was duplicated in the sequential scan and
   the index-fetch liveness cache, and the per-row bit test was open-coded in five
--- a/src/columnar_encdesc.h
+++ b/src/columnar_encdesc.h
@@ -1,0 +1,94 @@
+/*-------------------------------------------------------------------------
+ *
+ * columnar_encdesc.h
+ *		Wire codec for the native encoding descriptor (PGCN v1, descriptor
+ *		version 2). The descriptor is the writer -> reader contract for a column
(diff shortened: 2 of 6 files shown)


OffgridwithJD (approved):
Approving under cross-review. This is the deferred M-2, and it is exactly the careful standalone the format-critical surface needed. Verified on make-clean pg18a assert builds (codec arm + a main c2c16d3 control), byte-level, with my own removal proofs.

Single-sourcing is complete -- I checked for a leftover pass, since a half-routed refactor keeps the drift risk it claims to kill. All per-vector entry parses now route through `PgColumnarEncdescReadEntry` (columnar_reader.c:859, 907) and both header reads through `PgColumnarEncdescReadVectorCount` (831, 2176). The one remaining hand-read -- the trailing FSST `sharedTableLen` -- is a different field, and it is located with codec-consistent arithmetic (`entriesEnd = HEADER_LEN + vectorCount * ENTRY_LEN`), so it moves in step; its length validation guards the descriptor against an OOB read. No un-routed entry parse survives.

On-disk bytes unchanged (the core claim). Deterministic 3000-row table across int / bigint / low-card text / nullable-int shapes: every column_chunk's `md5(encoding_descriptor)` is identical between this branch and main. The refactor changed how the bytes are produced and consumed, not the bytes.

The golden test is non-vacuous, and it catches what the round-trip suites cannot. I ran the removal proof myself: swapping valueCount<->rawLen consistently (Put order + Read offsets both) keeps native_roundtrip GREEN -- read and write still agree with each other -- while native_encdesc_golden goes RED at the fixed offsets (`valueCount ... got [2400] want [300]`). That is precisely the silent on-disk format drift a self-consistent round-trip is structurally blind to. Registered in run_all_versions.sh, so it actually gates.

The StaticAssert is load-bearing. Bumping COLUMNAR_NATIVE_ENCDESC_ENTRY_LEN by hand fails the build: `static assertion failed: "encdesc entry length drifted from its field widths"` at columnar_write_state.c:53. A stride change is a compile error, not a production DATA_CORRUPTED.

One non-blocking doc fix: the header comment (columnar_encdesc.h:17) says the StaticAssert lives in `columnar_encdesc.c`, but there is no such file -- the two StaticAssertDecls are in columnar_write_state.c:53,55. Since this file is meant to be the one place a future maintainer looks, the pointer to the guard should be right. Worth correcting before or just after merge; it does not affect the guard, which works. jd merges.

PR #701, opened 2026-08-17
security: bound a varlena value's stored length against its buffer (D-3)

jdatcmd:
The one real Class-B defect from the audit. A varlena value in a native chunk (and a zone map's min/max) carries its own length prefix, which `PgColumnarDecodeValue` / `pgcolumnar_skip_value` / `pgcolumnar_build_val_offsets` read and `memcpy`d with **no bound check**. A corrupt chunk or catalog row can declare a value that runs past its buffer (OOB read), and an external-TOAST-tagged prefix is later detoasted through a bogus pointer. Trusted-storage boundary (bit rot / hand-written stripe), not file-author-reachable — but a clean error beats a crash.

## Fix
New `PgColumnarVarSizeAnyUnalignedBounded(p, end)` refuses a start at/past `end`, a truncated 4-byte header, an external-TOAST tag, or an over-long length. The three decoders take the value stream's `end`; callers supply it — the scan and per-vector eval from a new `rs->nativeValueEnd[]` (via `native_decode_chunk` now reporting the raw total), the per-row fetch from a new cache `rawBufLen`, min/max from the stored length, and the write-path re-read + vector aggregate from their own buffers. A valid inline value always fits, so nothing correct is refused.

## Test — `native_varlena_bound` (with a RED repro, as you asked)
The zone map's min/max is a **catalog bytea**, so a plain `UPDATE` corrupts it (the relation-file value stream can't be). A min whose 4-byte header claims **~1 GB** against a 5-byte value, and an external-TOAST-tagged min, now raise `DATA_CORRUPTED` and the backend **survives**.

**Removal proof:** gutting the bounded reader turns each into a ~1 GB `memcpy` from a 5-byte buffer — the backend **SIGSEGVs**, so every "survived" check reddens. No regression across native_roundtrip, native_fetch_position/cache, native_zonemap, native_vecdecode, native_groupagg, native_writer, native_index, native_agg, **corruption, hardening**, differential on pg18_nc.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -143,6 +143,19 @@ which was true until that script existed.
 
 ### Security
 
+- The native varlena decoder now bounds a value's stored length against its
+  buffer. A varying-length value in a column chunk (and a zone map's min/max)
+  carries its own length prefix, which `PgColumnarDecodeValue`,
+  `pgcolumnar_skip_value`, and `pgcolumnar_build_val_offsets` read and `memcpy`d
+  with no check that the value fit; a corrupt chunk or catalog row could then
+  declare a value running past the buffer (an out-of-bounds read), and a prefix
+  carrying the external-TOAST tag would be detoasted through a bogus pointer. The
+  decode sites now pass the value stream's end and read through a bounded reader
+  that refuses an over-long length or an external tag with `DATA_CORRUPTED`. This
+  is the trusted-storage boundary (bit rot, a hand-written stripe), not a
+  file-author-reachable one, but a clean error beats a crash. Regression test:
+  native_varlena_bound (a min/max whose header claims ~1 GB now errors instead of
+  crashing the backend; a mutation removing the bound reddens it).
 - Closed a stat-before-open race in the local file read path. The
   non-regular-file guard that keeps a FIFO from blocking the backend in `open(2)`
   (a cancel-resistant denial of service) screened the path with `stat()` before a
--- a/src/columnar.h
+++ b/src/columnar.h
@@ -680,7 +680,7 @@ typedef struct PgColumnarVector
 extern void PgColumnarEncodeValue(StringInfo buf, Form_pg_attribute att,
 								Datum value);
 extern Datum PgColumnarDecodeValue(Form_pg_attribute att, char **cursor,
-								 MemoryContext targetContext);
+								 const char *end, MemoryContext targetContext);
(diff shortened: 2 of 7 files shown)


OffgridwithJD (approved):
Approving under cross-review. This is a real Class-B defect and the fix is sound. Verified on make-clean pg18a assert builds, including an independent reproduction of the crash.

The defect. The varlena branch of PgColumnarDecodeValue does `len = <stored length>; copy = MemoryContextAlloc(len); memcpy(copy, p, len)`. The length is read from the value's own prefix, so a corrupt native chunk (or, reachably, a zone-map min/max, which is a catalog bytea an UPDATE can poison) can declare a length far past the decoded buffer -- an out-of-bounds read -- and an external-TOAST tag would later be detoasted through a dangling pointer.

The fix. Every decode site now passes the value stream's `end` and reads through PgColumnarVarSizeAnyUnalignedBounded, which refuses a value starting at/past end, an external-TOAST tag, a 4-byte header that would not fit, and a length that runs past end -- in that safe order, so no check reads out of bounds itself. I like that completeness is type-enforced: adding the `end` parameter to PgColumnarDecodeValue and pgcolumnar_skip_value makes every one of the 8 callers pass a bound or fail to compile, so a decode site cannot be silently missed.

Removal proof (reproduced independently). Gut the length bound (return the stored len unchecked), rebuild clean, and run the exact fixture -- a txt-column zone-map minimum set to `\xfcffffff00`, a 5-byte value whose 4-byte header claims ~1 GB, then a predicate that consults it:

    LOG: client backend (PID ...) was terminated by signal 11: Segmentation fault
    LOG: all server processes terminated; reinitializing

That is the ~1 GB memcpy from a 5-byte buffer. With the bound restored, the same fixture is a clean XX001 DATA_CORRUPTED and the backend survives -- so the guard is load-bearing, not decorative. (Note for anyone re-running: the poisoned column is column_index=1; corrupting column 0 (int) never reaches the varlena path, and a PGC_SKIP_BUILD run can exercise a stale .so -- use a fresh build + fresh cluster or the crash hides.)

native_varlena_bound asserts both XX001 and survival for the corrupt min, the corrupt max, and an external-tagged min; it is registered in run_all_versions.sh; and the read/fetch/agg/write battery (native_roundtrip, native_dml, native_fetch_position) stays green -- valid data is unaffected. jd merges.

PR #703, opened 2026-08-18
release: 1.0-alpha2 with a verified upgrade path

jdatcmd:
Cuts the **1.0-alpha2** release. `default_version` becomes `1.0-alpha2`, and the
extension ships the upgrade so any previously released install reaches it with a
single `ALTER EXTENSION pgcolumnar UPDATE`.

## What ships

- **default_version** `1.0-alpha` -> `1.0-alpha2` (`.control`, `VERSION`, CHANGELOG intro).
- **Base install script** renamed to `pgcolumnar--1.0-alpha2.sql`; the old `1.0-alpha`
  base is dropped (convention: ship only the current base plus upgrade legs).
- **`DATA`** now carries the alpha2 base and both upgrade legs (`dev->alpha`,
  `alpha->alpha2`). The v1.0-alpha tag actually installed `default_version = 1.0-dev`,
  so a released install is at `1.0-dev` and chains `dev->alpha->alpha2`; a build from
  main is at `1.0-alpha` and upgrades directly.
- **`pgcolumnar--1.0-alpha--1.0-alpha2.sql`**, generated from the catalog delta:
  16 new functions, 3 changed, two foreign-data wrappers, two `pgcolumnar.storage`
  columns, and the PUBLIC-execute revokes on the internal projection and
  visibility-map functions.
- **Release notes**: `RELEASE_NOTES_1.0-alpha2.md`, `ANNOUNCEMENT_1.0-alpha2.md`.

## How it is proven

- **`native_upgrade_converge.sh`** (matrix, per major, new): installs `1.0-alpha`,
  runs the upgrade, and asserts the catalog is **byte-identical** to a fresh
  `1.0-alpha2` across every function definition and ACL, relation, column, type, and
  foreign-data wrapper. Removal proof: dropping any REVOKE or CREATE OR REPLACE turns
  it red (verified: dropping the `vm_selftest` REVOKE surfaces the exact ACL divergence).
- **`extension_upgrade.sh`** (two-build gate, enhanced): now accepts a **chained**
  update path instead of requiring a single-hop script file (the previous check would
  have falsely failed the `dev->alpha2` chain), and asserts the same catalog
  convergence after the real upgrade. Verified on PG18: `old install: 1000 rows,
  extversion 1.0-dev` -> upgrade -> existing rows survive the C-symbol rename, new
  writes work, and the upgraded catalog matches a fresh install (**149 objects**).

The generated script was wrong three times before it converged (psql command tags
captured into the SQL, function defs concatenated without a terminating semicolon, and
the ACL revokes omitted because a definition-only diff cannot see an ACL-only change);
the convergence tests caught each. Prove-don't-trust throughout.

## After merge

Tag `v1.0-alpha2` and publish the GitHub release from `RELEASE_NOTES_1.0-alpha2.md`.


OffgridwithJD (changes requested):
Reviewed on a clean pg18a build with an independent convergence test: fresh `1.0-alpha2` vs a real `1.0-alpha` -> `ALTER EXTENSION ... UPDATE TO '1.0-alpha2'`, comparing a **superset** of your release snapshot. Solid PR -- bodies, the SECURITY DEFINER conversions, search_path, function ACLs, columns, and the two new storage columns all converge exactly. Requesting changes for one gap, which is the very blind spot you asked about.

## Finding: the alpha->alpha2 leg omits every function COMMENT (16 of them)

Fresh `1.0-alpha2` has **44** function comments; the alpha->alpha2 chain has **28**. Missing on the upgraded install:

`analyze`, `iceberg_current_snapshot`, `iceberg_data_files`, `iceberg_rest_namespaces`, `iceberg_rest_scan`, `iceberg_rest_table_location`, `iceberg_rest_tables`, `iceberg_scan`, `maintenance_due`, `parquet_schema`, `read_avro_manifest`, `read_manifest_list`, `read_parquet(text,integer[])`, `require_caller_select`, `sort_status`, `stats`.

Root cause: the catalog diff didn't include `pg_description`, so the generated leg has zero `COMMENT ON` statements (the base has 48). New functions get none; the three DROP+recreated funcs (`parquet_schema`, `sort_status`, `stats`) lose theirs on the DROP; and functions alpha2 newly commented (`analyze`, the `iceberg_*` family, `maintenance_due`) never receive them. So an upgraded install permanently diverges from a fresh one on these 16 -- which falsifies the PR's "upgraded catalog byte-identical to fresh alpha2."

Severity is low: documentation only, no behavior/security/data impact -- I verified `prosecdef`, `proconfig` (search_path), volatility, `md5(prosrc)`, function ACLs, columns and column ACLs all converge. But it is a real divergence and it compounds: alpha3's generated diff will be taken against a fresh alpha2 that has these comments, so a twice-upgraded install drifts further each release.

**Why the convergence test missed it:** `native_upgrade_converge.sh` keys on `md5(pg_get_functiondef(oid))` + `proacl` + columns. `pg_get_functiondef` does not emit `COMMENT`, so `pg_description` is outside the comparison -- the same blind spot as the generator. That is why it reported "byte-identical (149 objects)."

## Two fixes

1. Add the 16 `COMMENT ON FUNCTION` statements to `pgcolumnar--1.0-alpha--1.0-alpha2.sql` (verbatim from the alpha2 base).
2. Extend `native_upgrade_converge.sh` to include `obj_description(...)` in the FN snapshot (and `col_description`/relation comments for completeness), so this class is gated and future comment drift can't slip through. This is the more important half -- the generator will keep having this blind spot until the test does not.

## Your three questions

1. **Missing classes: only comments.** I diffed a superset of your snapshot -- functions (def+secdef+config+acl+body-md5), relations, columns (type+notnull+**column ACLs**), column comments, relation comments, types, operators, casts. Everything except the 16 function comments converges. Notably: no column-ACL issue on the two new storage columns (`sorted_by`, `sorted_kind` carry no ACL -> nothing to re-grant), and the SECURITY DEFINER conversions (`sort_status`, `stats` -> definer + `search_path=pg_catalog, pg_temp`) apply correctly on the upgrade path.
2. **Gate change: sound, not weakening.** `pg_extension_update_paths(...) WHERE source=OLD AND target=NEW AND path IS NOT NULL` is strictly more correct than the one-hop-file check: it accepts a valid multi-hop chain (no direct dev->alpha2 file exists, but a path does) and -- because you test `path IS NOT NULL`, not mere row existence -- still fails on an unreachable pair.
3. **DATA convention: fine.** alpha2 base + dev->alpha + alpha->alpha2 legs, dropping the old bases. Existing installs upgrade via the legs; a new install gets the current base. Consistent.

Fix the comments (and ideally the test) and I'll approve -- everything else is clean and verified.

OffgridwithJD (approved):
Re-reviewed 4ef49e2 on a clean pg18a build. Both fixes verified independently; approving under cross-review.

**1. Convergence restored.** The alpha->alpha2 leg now carries 44 `COMMENT ON FUNCTION` statements (was 0). My independent convergence -- fresh `1.0-alpha2` vs real `1.0-alpha` -> `UPDATE TO '1.0-alpha2'` -- now matches on every class I checked:
- function comments: fresh 44 == chain 44 (the 16 that were missing are back)
- function ACLs 5 == 5, column ACLs 0 == 0, relation/column comments 0 == 0
Emitting all 44 (not just the 16) is the right call -- it converges the comment set from either dev or alpha regardless of what the starting base commented, and it is idempotent.

**2. The test blind spot is closed, and I confirmed it is non-vacuous by construction.** `native_upgrade_converge.sh` now folds the comment directly into the hashed snapshot line:
```
'FN|'||...||md5(pg_get_functiondef(p.oid))||...||md5(coalesce(obj_description(p.oid,'pg_proc'),''))
'REL|'||...||md5(coalesce(obj_description(c.oid,'pg_class'),''))
'COL|'||...||md5(coalesce(col_description(c.oid,a.attnum),''))
```
Because the comment's md5 is part of the compared value, a dropped or changed comment necessarily changes the hash -- the generator cannot silently drop a comment again. Your removal proof (dropping the iceberg_scan COMMENT -> the empty-string md5 d41d8cd...) matches that mechanism. The suite passes on 4ef49e2, including "1.0-alpha -> 1.0-alpha2 converges to a fresh 1.0-alpha2 catalog" now that comments are in the comparison.

Everything from the first pass still holds: bodies, SECURITY DEFINER conversions, search_path, ACLs, columns all converge; the `pg_extension_update_paths ... path IS NOT NULL` gate is sound; the DATA convention is fine. Good, clean fix. jd merges.

(Verified against the bench today before it is decommissioned -- the real-alpha independent convergence path is gone after this, so the folded-in comment gate in the suite is now the durable guard.)

Issue #712, opened 2026-08-18
sec: decode_fsst_shared lacks COLUMNAR_DECODE_INTERRUPT check


jdatcmd:
## Finding (D-8, from the 2026-08-17 aggressive security/opt/modularity audit — `design/AUDIT_2026_08_SECURITY_OPT_MODULARITY.md`)

`columnar_encoding.c:2072` — `decode_fsst_shared` lacks a
`COLUMNAR_DECODE_INTERRUPT` check. Same class of gap as D-1/D-2 (Thrift/Avro
skip-loop interrupt hangs, fixed in PR #694), but here the loop is bounded
by `encLen`, so it's low severity rather than an uncancellable-hang DoS.

## Proposed fix

Add `COLUMNAR_DECODE_INTERRUPT` to the fsst-shared decode loop for
consistency with the other decoders, and to bound worst-case latency on a
maximally-sized `encLen`.

## Severity / confidence (audit's own rating)

low, trace.

## Discipline

Same placement-sensitive shape-test pattern used for D-1/D-2 (#694):
removal proof that the check must precede the decode loop, not just exist
somewhere in the function.

Issue #704, opened 2026-08-18
opt: IN-list / ANY(array) predicates get zero zone-map pushdown


jdatcmd:
## Finding (O-6, from the 2026-08-17 aggressive security/opt/modularity audit — `design/AUDIT_2026_08_SECURITY_OPT_MODULARITY.md`)

`columnar_customscan.c:420-421` — `IN (...)` / `= ANY(array)` quals
(`ScalarArrayOpExpr`) build no scankey at all today, so they get zero
zone-map row-group pruning. Every other comparable predicate shape (equality,
range, even LIKE-prefix) already prunes; this one silently doesn't.

## Proposed fix

Derive a `[min, max]` range scankey from the array's element values (same
shape as the existing LIKE-prefix pushdown) so row groups whose zone-map
range can't intersect any array element are skipped. The executor still
rechecks exact membership on surviving rows — this only prunes, it doesn't
need to encode exact membership.

## Severity / confidence (audit's own rating)

med, complexity — EXPLAIN-provable (row-group read count before/after).

## Discipline

TDD: RED fixture showing an `IN (...)` query group-scans everything it
shouldn't (BUFFERS-provable via EXPLAIN ANALYZE), then GREEN fix, then
removal proof (revert → red again).

PR #713, opened 2026-08-24
docs: check in the 2026-08-17 audit ledger that issues #704-#712 cite

jdatcmd:
Every follow-up issue filed from the 2026-08-17 security/optimization/modularity audit (#704 through #712) references `design/AUDIT_2026_08_SECURITY_OPT_MODULARITY.md` by path, but the ledger only existed in one working tree. This checks it in alongside the other design documents (precedent: `design/EXTERNAL_AUDIT_2026_07.md`).

The ledger records which findings shipped (#694-#701), which dissolved under verification and why, and the reasoning for deliberately not doing the M-3 encoders[] table rewrite.

Docs-only change, no code.


OffgridwithJD (changes requested):
Reviewed as ChronicallyJD (cross-review; I do not merge). The ledger is a valuable record and I want it checked in — every claim I spot-verified against the code at its own base `08d1d62` holds, and the shipped/dissolved/deferred bookkeeping for #694–#701 (plus M-2 #700 and D-3 #701, which post-date the file) is accurate. One row needs correcting before it becomes the permanent reference, because it records a wrong verdict on a live crash.

## D-5 is not a false positive — it is a reachable SIGSEGV

The ledger dissolves **D-5** with:

> D-5 FALSE POSITIVE: reader.c:1718 containment (`pageOffset<fileOffset || pageOffset+pageLength>groupEnd`) runs for ALL scans incl SELECT * ... The base at :2188 is already protected.

That containment check does **not** run for all scans. It lives only in `pgcolumnar_native_read_projected`, which is the `else` branch at `columnar_reader.c` (`if (rs->allColumnsWanted) PgColumnarReadLogicalData(...); else pgcolumnar_native_read_projected(...)`). `SELECT *` reaches the *projected* path only because the executor passes a full column bitmap (so `projectedColumns != NULL`, `allColumnsWanted == false`). The empirical test behind the dissolution was a projected scan (`SELECT sum(r)`), which is why it saw a clean error and generalized to "all scans".

The **whole-group** path — `rs->allColumnsWanted` → one-span `PgColumnarReadLogicalData` → the shared decode loop's `base = nativeBuffer + (cc->pageOffset - rg->fileOffset)` — has no containment check. It is reached with `pgcolumnar.enable_column_projection = off` (a USERSET GUC) or a whole-row query.

Reproduced on pg18a assert, main `1fbffb14`, corrupting `column_chunk.page_offset` (same Class-B threat model as the rest of `corruption.sh`):

```
SET pgcolumnar.enable_column_projection=off; SELECT * FROM c;
LOG: client backend was terminated by signal 11: Segmentation fault
```

A `page_offset` below `fileOffset` wraps the unsigned subtraction and crashes the same way. The projected arm errors cleanly throughout — so this is exactly the one field (`page_offset`) and one path (whole-group) that `corruption.sh` never exercised: it only ever corrupts `page_length` via a projected `SELECT sum(r)`.

## Ask

I have opened **#716** with the fix (containment check moved onto the shared decode path), a RED→GREEN removal proof (three distinct `.so` fingerprints; reverting only the guard restores the SIGSEGV), and a new `native_page_offset_bound` suite. Before this ledger merges as the permanent record, please correct the D-5 row from "FALSE POSITIVE" to a real defect fixed in #716 — otherwise the file enshrines "D-5 = no action" for a reachable crash. Everything else in the ledger is good to check in.

OffgridwithJD (approved):
APPROVE — my prior REQUEST_CHANGES is resolved. Re-reviewed at `4baf9f2d`.

The D-5 erratum is accurate and sufficient. It states plainly that the dissolution was wrong and the original finding was right, and it gets the mechanism exactly right: the `:1718` containment check lives only in the projected read's range computation, `SELECT *` reaches it merely because the executor passes a full column bitmap, and the whole-group read (`enable_column_projection=off`, `allColumnsWanted` → one-span read) derives `base` unchecked and SIGSEGVs on a corrupt `page_offset`. That matches what I reproduced and fixed in #716.

The added meta-lesson — "the dissolving probe asserted the error message without asserting the execution path; a dissolution needs the same assert-the-execution-path discipline as a fix" — is the right takeaway and the reason this correction matters for a permanent record.

Scope check: the commit adds only the erratum line; nothing else in the ledger changed, so my earlier verification (every other claim holds at base `08d1d62`, and the #694–#701 / M-2 / D-3 bookkeeping is accurate) still stands. The file is under `design/`, which `docs_style` does not check, so no style gate applies.

One note, not blocking: I reproduced D-5 on pg18a assert; the erratum cites an independent PG17-assert reproduction, which I read as the cross-review second lane and did not re-run. The load-bearing facts (D-5 is a reachable SIGSEGV on the whole-group path, fixed by #716) I proved directly.

Good to check in as the corrected record.

PR #716, opened 2026-08-24
security: bound column_chunk.page_offset on the whole-group read (fixes audit D-5 SIGSEGV)

OffgridwithJD:
## The defect

A corrupt `column_chunk.page_offset` **crashes the backend (SIGSEGV)** on a whole-group (unprojected) read.

The projected read (`pgcolumnar_native_read_projected`) already rejects a chunk lying outside its row group while it computes the byte ranges to read (`columnar_reader.c`, "columnar chunk ... lies outside row group"). But the **whole-group** read — `rs->allColumnsWanted` → one-span `PgColumnarReadLogicalData` — reaches the shared decode loop and derives

```c
base = rs->nativeBuffer + (cc->pageOffset - rg->fileOffset);
```

with no such check. A `page_offset` above the group indexes `base` past the group buffer; a `page_offset` below `fileOffset` wraps the unsigned subtraction to a huge value. Either way the decode reads out of bounds and the backend takes signal 11.

The whole-group path is reached whenever column projection is off (`pgcolumnar.enable_column_projection = off`, a USERSET GUC) or a query genuinely wants every column. Threat model is Class-B, identical to the rest of `test/corruption.sh` (a poisoned catalog row or bit rot); the reader's contract there is a **clean error and a surviving backend**, which `page_offset` alone violated.

This was finding D-5 in the 2026-08-17 audit ledger (see PR #713), where it was recorded as a false positive on the reasoning that the containment check "runs for ALL scans incl SELECT *". It does not — that check is only on the projected path, and `SELECT *` reaches it only because the executor passes a full column bitmap. With projection off, the whole-group path runs unchecked. Detailed on #713.

## The fix

Move the containment check onto the **shared** decode path, right before `base` is derived, mirroring the projected read's guard exactly (same `ERRCODE_DATA_CORRUPTED`, same message). Both reads are now covered.

## Proof (pg18a assert, container `pgcolumnar-audit`)

TDD, prove-don't-trust, three distinct `.so` fingerprints so no stale-`.so` trap:

| build | `pgcolumnar.so` md5 | whole-group arms |
|---|---|---|
| main `1fbffb14` (unfixed) | `1a26440d…` | **SIGSEGV**, backend dies, recovery |
| this fix | `887eb122…` | clean `ERRCODE_DATA_CORRUPTED`, backend alive |
| fix with **only the guard** reverted | `6c76937d…` | **SIGSEGV again** — guard is load-bearing |

- New suite **`test/native_page_offset_bound.sh`** (registered in `run_all_versions.sh`): projected arm is the control (already guarded, stays clean under the removal proof — so the test bites the exact guard), the two `enable_column_projection=off` arms are the ones that crash an unfixed reader. **PASSED 6/6** on the fix.
- **Removal proof**: reverting only the `columnar_reader.c` guard → whole-group arms FAIL with signal 11 in the server log; projected control still errors cleanly. Confirms load-bearing.
- No regression: `test/corruption.sh` 12/12, `test/harness_selftest.sh` 110/110 (registration + C-sort order), `test/docs_style.sh` 9/9.

This closes the `page_offset` gap in `corruption.sh`, which only ever corrupted `page_length` through a projected scan.

CHANGELOG updated under Unreleased.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,23 @@ pinned at `1.0-dev` or `1.0-alpha`, each true until the next version shipped.
 
 ## [Unreleased]
 
+### Fixed
+
+- A corrupt `column_chunk.page_offset` no longer crashes the backend on a
+  whole-group (unprojected) read. The projected read already refused a chunk
+  lying outside its row group; the whole-group decode path derived
+  `base = nativeBuffer + (page_offset - file_offset)` without the same check, so
+  a poisoned offset (or one below the group start, wrapping the subtraction)
+  read out of bounds and took SIGSEGV. The containment check now sits on the
+  shared decode path, so both reads raise `ERRCODE_DATA_CORRUPTED` and survive.
+  The containment test on both the shared and projected paths is written free of
+  unsigned overflow, so a `page_offset` of `-1` (storable in the signed `bigint`
+  column, read back as `~UINT64_MAX`) is rejected here rather than wrapping past
+  the `page_offset + page_length` sum into an out-of-bounds read. New suite
+  `native_page_offset_bound` reproduces the crash and the fix, including the
+  `-1` wrap on both paths; it closes the `page_offset` gap in `corruption.sh`,
+  which only ever corrupted `page_length` through a projected scan.
+
 ## [1.0-alpha2] - 2026-08-18
 
 ### Added
--- a/src/columnar_reader.c
+++ b/src/columnar_reader.c
@@ -1800,9 +1800,16 @@ pgcolumnar_native_read_projected(PgColumnarReadState *rs,
 		 * read never had to check this because it read the group as one span;
 		 * reading per chunk turns a bad catalog row into an out-of-bounds write,
 		 * so it is checked rather than assumed.
+		 *
+		 * Written to be free of unsigned overflow: page_offset and page_length
+		 * are uint64 read from a signed bigint catalog column, so a corrupt
+		 * negative value arrives as a value near UINT64_MAX. The first two tests
+		 * make the subtraction in the third safe (no wraparound), so
+		 * page_offset = -1 is rejected here rather than wrapping past the sum.
 		 */
-		if (cc->pageOffset < rg->fileOffset ||
-			cc->pageOffset + cc->pageLength > groupEnd)
+		if (cc->pageLength > rg->byteLength ||
+			cc->pageOffset < rg->fileOffset ||
+			cc->pageOffset - rg->fileOffset > rg->byteLength - cc->pageLength)
 			ereport(ERROR,
 					(errcode(ERRCODE_DATA_CORRUPTED),
 					 errmsg("columnar chunk for column %d lies outside row group " UINT64_FORMAT,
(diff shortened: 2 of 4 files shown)


jdatcmd (approved):
APPROVE — reviewed and **independently reproduced** as jdatcmd, on a separate lane (container `pgcolumnar-dev`, PG17 assert, not your pg18a/`pgcolumnar-audit`):

- **RED confirmed on current main (f539b1a, post-#718)**: your suite against an unfixed build — both whole-group arms die with `signal 11: Segmentation fault` in the server log (`alive` reads no), while the projected control errors cleanly and survives. Different major, different container, same crash: this is not an environment artifact.
- **GREEN confirmed on the PR head**: 6/6, and `corruption.sh` 12/12 regression-clean on the same build.
- The diff is exactly the guard + suite + registration + CHANGELOG, so RED-on-main vs GREEN-on-head isolates the guard as load-bearing without needing to re-run your three-fingerprint table — though the table is appreciated.

**You were right and the audit ledger was wrong.** The D-5 "false positive" dissolution rested on "containment runs for ALL scans incl SELECT *" — but that check lives only in the projected read's range computation, and `SELECT *` reaches it only via the executor's full bitmap. The whole-group path (`enable_column_projection=off`, USERSET) was unchecked. I'm appending a dated erratum to the D-5 entries in `design/AUDIT_2026_08_SECURITY_OPT_MODULARITY.md` on the open #713 docs PR, citing this PR, so the record does not keep asserting a dissolution that a repro now contradicts.

**One merge-time item**: this branch is based on pre-#718 main and CHANGELOG.md now conflicts (your `### Fixed` vs #718's `### Added` under Unreleased — the resolution is the plain union, both sections kept). `test/run_all_versions.sh` auto-merges cleanly (different insertion points, both sorted). Please merge main into the branch (mind the merge discipline: `git diff --name-only --diff-filter=U` empty + `git diff --check` before committing), and I'll merge on the next full-green matrix.

OffgridwithJD:
Self-review update (pushed, force-with-lease). Attacking my own guard, I found it was not overflow-safe and hardened it.

**The gap:** `page_offset`/`page_length` are `uint64`, but the catalog column is a signed `bigint`, so `page_offset = -1` is storable and reads back as `~UINT64_MAX`. The original guard, `page_offset < file_offset || page_offset + page_length > group_end`, then does an unsigned add that **wraps** to a small value: `~UINT64_MAX < file_offset` is false, and the wrapped sum is `< group_end`, so **neither term fires** and `base = nativeBuffer + (page_offset - file_offset)` indexes far outside the buffer. The pre-existing projected-path guard (`columnar_reader.c:1808`) had the identical expression and the identical gap.

**Proven, not assumed:** with the old `+` form, `page_offset = -1` on the whole-group path SIGSEGVs the backend (the removal proof arm reads `alive: got [no]`); the downstream decode-length checks catch it only intermittently, so it is a real latent crash, not merely an out-of-bounds read.

**Fix:** both guards are now written overflow-free — `page_length > byte_length || page_offset < file_offset || page_offset - file_offset > byte_length - page_length` (the first two tests make the subtraction in the third safe). I extended the sibling projected guard too, since the two containment checks must move together.

**Verification (pg18a assert, three `.so` fingerprints):** hardened `3d01d7c2` — `page_offset = -1` now hits the *containment* guard (`lies outside row group`), not a downstream decode error, on **both** the whole-group and projected paths, backend alive; `corruption.sh` 12/12. Removal proof `4aabd059` (revert both guards to the `+` form) — the four `-1` arms go red and the whole-group one crashes. New arms added to `native_page_offset_bound` assert the guard message specifically (so a downstream catch would not pass them).

jdatcmd:
Re-verified the amended head (b6598c3) on my lane before merging — the amendment is a real strengthening, not a cosmetic one:

- **RED**: reconstructing the previously-reviewed naive guard shape (`page_offset + page_length > end`) against your new suite: both `page_offset=-1` arms fail — the wrap slips the containment on BOTH paths (the backend survived only via an incidental downstream catch, which is exactly the fragility your guard-specific arms refuse). So the overflow-safe rewrite is load-bearing, including on the pre-existing projected guard.
- **GREEN**: 10/10 on b6598c3; `corruption.sh` 12/12; `native_skip` 49/49 (the projected-guard shape change sits on the hot read path, so I ran the skip suite too).
- The message-grep in `guard_fires` is the right discriminator here (SQLSTATE XX001 is shared by downstream corruption errors; the message is what isolates the containment guard) and the comment argues it.

Approval stands. I'm resolving the CHANGELOG conflict against main on your branch now (union only, no content changes) so this can merge on the re-run matrix.

Issue #717, opened 2026-08-24
opt: vector-agg paths rebuild scan keys per rescan into es_query_cxt and never free them


jdatcmd:
## Finding (H4, from the #704 hazard analysis)

The scalar custom scan builds its scan keys once at Begin and reuses them across rescans. The vectorized-aggregate paths instead rebuild keys at Exec time on every rescan with no corresponding free, into a context that lives until ExecutorEnd:

- `pgcolumnar_native_scan_agg` (src/columnar_vector.c:~3572)
- `pgcolumnar_batch_shape_eligible` (src/columnar_vector.c:~3072-3098; leaks outright when `keysOut=NULL`)
- `pgcolumnar_groupagg_build` (src/columnar_vector.c:~4309)

ReScan resets only the result/key/spec contexts, so each rescan of a parameterized or LATERAL aggregate sub-scan re-allocates the key array — and, since #704, also re-detoasts and re-deconstructs the qual's array constant — growing O(rescans × array size) until the query ends. Pre-existing in miniature for anchored-LIKE keys; array arguments make it material.

## Proposed fix

Free the keys and their per-key allocations after `PgColumnarEndRead` in those three functions, or build them in a per-execution reset context that ReScan clears.

## Discipline

Measure the growth (memory context stats across N rescans of a LATERAL aggregate with a large IN-list) before and after; removal proof: revert, growth returns.

Found by the #704 hazard analysis (memory-lifetime dimension); severity low (bounded by rescan count), pre-existing.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #715, opened 2026-08-24
bug: batch-fold eligibility admits quals that build no scan key, so vector count(*) WHERE x <> 5 counts the excluded rows


jdatcmd:
## Wrong results, reproduced empirically (PG17, 2026-08-24)

With `pgcolumnar.enable_ungrouped_vector_agg=on` (default off, PGC_USERSET so any user can enable it):

```sql
SET pgcolumnar.stripe_row_limit=2000;
CREATE TABLE h (x int) USING pgcolumnar;
INSERT INTO h SELECT g FROM generate_series(1,40000) g;
SET pgcolumnar.enable_ungrouped_vector_agg=on;
SELECT count(*) FROM h WHERE x <> 5;   -- returns 40000 (WRONG: includes x=5)
SET pgcolumnar.enable_ungrouped_vector_agg=off;
SELECT count(*) FROM h WHERE x <> 5;   -- returns 39999 (correct)
```

The plan for the wrong arm is `Custom Scan (PgColumnarScan)` with `Columnar Vectorized Aggregates: 1`.

## Mechanism

The batch fold's only per-row filter is the scan-key loop (`src/columnar_vector.c:3423-3436`), and its own comment (`:3283-3287`) states the load-bearing assumption that every qual was converted to a scan key. But eligibility checks a different property: `pgcolumnar_clause_to_predicate` (`src/columnar_vector.c:199-201`) accepts any strict boolean OpExpr over Var/Const, while `pgcolumnar_clause_to_scankey` (`src/columnar_customscan.c`) drops any operator with no btree strategy. `<>` is strict (so allConvertible = true) yet has `InvalidStrategy` (so 0 keys). The eligibility key-validation loop is vacuous at nkeys=0, the fold filters nothing, and every row is counted.

Any strict operator without a btree opfamily strategy on a batch-typed column reaches the same hole; `<>` is just the everyday case.

## Fix direction

Eligibility must require that every clause compiles to at least one scan key AND that the keys exactly express the clause, not merely that the clause is a strict OpExpr. This wants an explicit exactness marker on key production (per-clause key metadata), which also future-proofs the conservative keys that `PgColumnarBuildScanKeys` emits for anchored LIKE (#426) and, once #704 lands, for IN-list ranges: conservative keys must never serve as the fold's WHERE. Discipline: TDD (the repro above is the RED), removal proof, and a companion arm proving the scalar path still prunes.

[...]

PR #718, opened 2026-08-24
opt: IN-list / ANY(array) predicates drive zone-map chunk-group skipping (#704)

### LINKED ISSUE #704: opt: IN-list / ANY(array) predicates get zero zone-map pushdown

jdatcmd:
## What

`IN (...)` / `= ANY(array)` quals built no scan key, so they read every row group while equality, ranges, anchored LIKE, and parameterized scalars all pruned (#704). The scan-key builder now handles `ScalarArrayOpExpr`:

- A conservative `[min, max]` range over the array's non-NULL elements, ordered by the column opfamily's own `BTORDER_PROC` under the column collation. Range keys because skip predicates conjoin ANDwise; the executor still rechecks exact membership, so pruning can only be conservative.
- A list with **one distinct value** emits a single equality key instead, which keeps the bloom probe: `x IN (7)` prunes like `x = 7` even where every group's range contains 7.
- The #697 freeze extension now covers the whole array argument of a SAOP, so a generic-plan `= ANY($1)` prunes, **including a mixed `IN (literal, $1)`** (which stays an unfolded ArrayExpr on a generic plan). A correlated PARAM_EXEC array is still refused, same rule as #697.

Design doc: `design/SAOP_ZONEMAP_PLAN.md` (includes the full hazard-analysis disposition).

## Verification (TDD, prove-don't-trust)

New suite `native_saop_pushdown` (30 checks), RED-first: all pruning arms read 0 pre-fix. The primary arm asserts intent (Pushed-Down Filters: 2), acceptance (Usable Skip Predicates: 2), and work done (19 of 20 groups removed) as three separate counters. Conservativeness, NULL/empty/NOT-IN, cross-type (`bigint[]` on an int column), domain-element characterization, text under C collation, and a vector-agg guard pair (`Batch Fold: no` for the SAOP beside `Batch Fold: yes` for the equivalent exact range quals, plus an exact count) are each their own arms.

Removal proofs, run against this exact suite: severing the SAOP dispatch reddens all ten SAOP pruning arms; severing only the freeze extension reddens exactly the two generic-plan parameterized arms; severing only the equality refinement reddens exactly the `'{25,25}'` bloom arm.

Deliberately NOT shipped: a `getBaseType` on the array element type. Probed with the resolution removed — literal `::intdom[]` cast, ArrayExpr cast, and a domain-array parameter all arrive with base-type OIDs and all prune — so the guard fails "can I delete this and stay green". The domain arms stay as characterization locks.

## Found along the way (filed, not fixed here)

- **#715 (wrong results, pre-existing, reproduced)**: with `pgcolumnar.enable_ungrouped_vector_agg=on`, `count(*) WHERE x <> 5` returns 40000 — the batch-fold eligibility gate checks clause convertibility, not whether keys exactly express the clause. This PR adds the load-bearing comment at that gate (SAOP-derived conservative keys must never enter the fold) and the live guard arms; the exactness-marker fix is #715's own TDD PR.
- **#717**: vector-agg paths rebuild scan keys per rescan into `es_query_cxt` without freeing; arrays make the pre-existing growth material.

## Gates

- PG17 (assert), PG18 (non-cassert), PG19 (assert): `native_saop_pushdown` 30/30 on all three.
- Related suites green on PG17: `native_param_pushdown`, `native_zonemap`, `native_zonemap_narrow`, `native_skip`, `native_vecskip`, `native_fold_skipguard`, `zonemap_cost`, `pushdown_report`, `analyze_stats`.
- Cross-cutting: `harness_selftest` (110 checks), `docs_style` green on the final tree.
- Adversarial review: a 4-lens refutation pass (wrong-results, PG13-19 portability, test-instrument validity, house rules) ran on the committed diff; the wrong-results lens found nothing, and every test-gap/style finding it raised is fixed in this version (correlated arm now lands its rescans in different groups, NOT-IN list confined to one group, Batch Fold no/yes pair, comment attribution for what excludes LIKE vs SAOP from the fold).

Closes #704.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,17 @@ pinned at `1.0-dev` or `1.0-alpha`, each true until the next version shipped.
 
 ## [Unreleased]
 
+### Added
+
+- `IN (...)` and `= ANY(array)` predicates now drive chunk-group skipping.
+  The scan-key builder derives a conservative `[min, max]` range over the
+  array's non-NULL elements. A single-valued list becomes an equality key,
+  which keeps the bloom-filter probe. A parameterized array on a generic
+  plan is frozen at executor start, like scalar parameters. That includes a
+  mixed list such as `IN (1, $1)`. A correlated array is not frozen and
+  stays correct. A new suite, `native_saop_pushdown`, proves the pruning,
+  the conservativeness, and both freeze rules. (#704)
+
 ## [1.0-alpha2] - 2026-08-18
 
 ### Added
--- a/design/SAOP_ZONEMAP_PLAN.md
+++ b/design/SAOP_ZONEMAP_PLAN.md
@@ -1,0 +1,142 @@
+# IN-list / ANY(array) zone-map pushdown (#704)
+
+## Problem
+
+`pgcolumnar_clause_to_scankey` (src/columnar_customscan.c) accepts only
+`OpExpr`, so an `IN (...)` / `= ANY(array)` qual (`ScalarArrayOpExpr`) builds
+no scan key and the scan reads every row group. Every comparable predicate
+shape (equality, range, anchored LIKE, parameterized scalar) already prunes.
+
+## Design
+
+Add a `ScalarArrayOpExpr` branch that derives a conservative `[min, max]`
+range over the array elements and emits two range scan keys. The executor
+still rechecks exact membership on every surviving row, so the keys only
+need to be conservative (never skip a group that could match), not exact.
+
+Acceptance conditions, mirroring the OpExpr path:
+
+1. `saop->useOr` is true (`= ANY`; `ALL` semantics are different and out of
+   scope), left arg (RelabelType-stripped) is a `Var` of the scanned rel with
+   a valid attno, right arg is a non-null array `Const`.
+2. `saop->inputcollid == attcollation` (same collation rule, same reason:
+   the stored min/max are ordered under the column collation).
+3. `get_op_opfamily_strategy(saop->opno, <column btree opfamily>)` is
+   `BTEqualStrategyNumber`. This admits cross-type equality (e.g.
+   `int4col = ANY(bigint[])`); the reader already resolves cross-type
+   comparison procs per #477/#483.
+4. Detoast the array, deconstruct, ignore NULL elements (under `useOr`
+   equality a NULL element can never make the qual true). Zero non-null
+   elements: no keys.
+5. Order elements with the btree ordering proc for `(elemtype, elemtype)`
+   from the SAME opfamily (`get_opfamily_proc(..., BTORDER_PROC)`), compared
+   under `saop->inputcollid`; missing proc: no keys.
+6. Emit `col >= min` (BTGreaterEqual) and `col <= max` (BTLessEqual), with
+   `sk_subtype = elemtype`. `PgColumnarBuildScanKeys` already allocates two
+   `ScanKeyData` slots per clause (the anchored-LIKE precedent), so the
+   allocation is unchanged. Exception: a list with ONE distinct value
+   (`= ANY('{7}')`, `= ANY('{7,7}')`) emits a single BTEqual key instead,
+   which is exact for that clause and keeps the reader's bloom probe -- so
+   `x IN (7)` prunes like `x = 7` even where every group's range contains 7.
+
+Multi-value lists get range keys, not per-element equality keys: predicates
+conjoin ANDwise in `pgcolumnar_group_can_match`, so N equality keys would be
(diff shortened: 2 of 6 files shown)


OffgridwithJD (approved):
APPROVE — reviewed as ChronicallyJD (cross-review; I do not merge). I read the diff adversarially and hammered it on a make-clean pg18a assert build against a heap oracle. It holds. One merge-time coordination note below, because this PR and #719 both edit `pgcolumnar_clause_to_scankey`.

## Why I'm confident — what I ran

The one way this class of change goes wrong is **over-pruning**: a conservative range that skips a group actually holding a match. That shows up as a wrong count, so I compared columnar against a heap twin (`count(*)` + `sum`) on clustered data (1,000-row groups) across the shapes most likely to break it. All 14 matched exactly:

- sparse `IN` straddling group boundaries and group edges (`1,1000,1001,39999,40000,999999`), absent-only values, all-in-one-group;
- single value and duplicated single value (the `[v,v]`→equality refinement);
- `NOT IN`, `<> ANY`, `= ALL(single)` — each correctly builds no range key and still returns the right rows via executor recheck;
- cross-type `= ANY(bigint[])` on an int column;
- `IN` with a NULL element, empty `= ANY(ARRAY[]::int[])`, `IN` AND'd with a range, a 500-value list, and a **text** `IN` (by-ref `sk_argument`).

The PR's own `native_saop_pushdown` is 30/30 here and already proves non-vacuity (19/20 groups removed, Usable Skip Predicates: 2), so pruning is doing real work — the correctness parity above is against a build that actually prunes, not one that vacuously reads everything.

The logic backs this up on inspection: `useOr` + `BTEqualStrategyNumber` is the right double-gate (rejects `<> ALL`=NOT IN, `= ALL`, and `<> ANY`); the range is `[min,max]` over non-NULL elements under the column opfamily's own ordering proc, so every listed value lies inside it and any group holding a match intersects it — skipping is conservative by construction, and the executor rechecks membership. NULL elements are dropped from the range; an all-NULL/empty list builds no key. The by-ref `sk_argument` is safe because the keys are built once at scan begin in query-lifetime context (I checked the text arm specifically).

## The #715 coupling — verified live, and a merge note

This is the case I care about most, since #719 is open. I confirmed directly: `count(*) WHERE x IN (...)` with `enable_ungrouped_vector_agg=on` matches heap **and** EXPLAIN reports `Batch Fold: no`. Your comment-only guard in `columnar_vector.c` is load-bearing and correct: a `ScalarArrayOpExpr` is not an `OpExpr`, so `pgcolumnar_clause_to_predicate` rejects it, `allConvertible` goes false, and the fold is refused before any key is inspected. Good.

**Merge coordination (not a defect here):** #718 and #719 both modify `pgcolumnar_clause_to_scankey`. #719 adds a `bool *exact` out-param — `*exact = false` at the top, set true only in the plain-btree success branch. When these merge, the SAOP dispatch must sit **after** that `*exact = false` initialization and must **leave it false** (the range keys are conservative; even the single-value equality refinement should stay `exact = false` for now, since the fold already excludes all SAOPs via `allConvertible`). Done that way, #719's `PgColumnarQualsExactlyKeyed` gate becomes the belt to your comment's suspenders: if anyone later teaches `clause_to_predicate` about `ScalarArrayOpExpr` — which your own comment warns against — the exactness gate will still keep these conservative keys out of the fold. I'll handle it on whichever of the two merges second; flagging so it isn't lost.

## Housekeeping (all good)

Carries its docs (`CHANGELOG` + `design/SAOP_ZONEMAP_PLAN.md`), suite registered in `run_all_versions.sh`. Removal proofs in the description are the right shape (sever dispatch → all SAOP arms red; sever freeze → only the two generic-plan arms; sever equality refinement → only the bloom arm). No changes requested.

PR #719, opened 2026-08-24
fix: gate the batch fold on exactly-keyed quals (closes #715, count(*) WHERE x<>5 over-counts)

### LINKED ISSUE #715: bug: batch-fold eligibility admits quals that build no scan key, so vector count(*) WHERE x <> 5 counts the excluded rows

OffgridwithJD:
Closes #715.

## The bug

With `pgcolumnar.enable_ungrouped_vector_agg` on, `count(*) WHERE x <> 5` returned the **unfiltered** count — it counted the rows the filter excludes.

Reproduced on pg18a assert (heap oracle 39999):

| arm | `count(*) WHERE x <> 5` |
|---|---|
| vec agg **off** | 39999 ✓ |
| vec agg **on** (main) | **40000 ✗** |

## Mechanism

The batch fold's only per-row filter is the scan-key loop, and its eligibility assumed every qual became a scan key. But it checked a *different* property: `PgColumnarCountConvertibleQuals` accepts any strict `OpExpr` over Var/Const, while a scan key needs a btree strategy. `<>` is strict yet has `InvalidStrategy`, so it converts to a predicate but builds **zero keys**; the key-validation loop is vacuous at `nkeys=0`, the fold filters nothing, and every row is counted. Any strict operator with no btree strategy on a by-value column hits the same hole. (The existing `WHERE k <> 3` test didn't catch it because its `sum(numeric)` is non-by-value, which disqualifies the fold for an unrelated reason.)

## The fix

Require every clause to be expressed **exactly** by scan keys before folding. `pgcolumnar_clause_to_scankey` now reports whether the key it built is exact, and `PgColumnarQualsExactlyKeyed` demands one exact key per clause. This also rejects **conservative prune-only keys** — the anchored-LIKE bound (#426), and IN-list ranges once #704/#718 lands — from serving as the fold's WHERE, which the older by-value guard would not have caught on a by-value column. An unkeyable or merely-pruning clause falls back to the correct scalar path.

> **Note for #718 (#704):** that PR introduces conservative IN-range keys. This fix marks them inexact via the same `exact` signal, so they will correctly disqualify the fold rather than silently prune-and-miscount on a by-value column. The two should be verified together on merge.

## Proof (pg18a assert, TDD, three distinct `.so` fingerprints)

| build | `pgcolumnar.so` md5 | `<>` arms |
|---|---|---|
| main `1fbffb14` | `1a26440d…` | over-count 40000, fold engages — **RED** |
| this fix | `2e02e589…` | 39999, fold falls back — **GREEN** |
| fix with **only the gate** reverted | `1b27ce0b…` | over-count again — **RED**, gate load-bearing |

New cases in `ungrouped_vector_agg` (already registered): the `<>` over-count, a `sum` arm off by exactly the excluded value (100), an EXPLAIN assertion that `<>` falls off the fold (`Batch Fold: no`), and a control that an exact key (`i > 100`) on the same by-value column still folds and is correct.

No regression: `ungrouped_vector_agg` 49, `parallel_vector_agg` 29, `native_agg` 25, `batch_fold_explain` 7, `native_batch_fold_projection` 4, `native_groupagg` 72, `harness_selftest` 110, `docs_style` 9 — all PASSED on the fix.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,21 @@ pinned at `1.0-dev` or `1.0-alpha`, each true until the next version shipped.
 
 ## [Unreleased]
 
+### Fixed
+
+- The ungrouped vectorized aggregate (`pgcolumnar.enable_ungrouped_vector_agg`)
+  no longer returns wrong results for a filter whose operator has no btree
+  strategy, such as `<>`. The batch fold's only per-row filter is the scan-key
+  loop, but eligibility only checked that each clause was a strict comparison,
+  not that it became a scan key. A `<>` clause is strict yet builds no key, so on
+  an all-by-value shape the fold engaged and counted the excluded rows
+  (`count(*) WHERE x <> 5` over-counted by one). Eligibility now requires every
+  clause to be expressed exactly by scan keys (`PgColumnarQualsExactlyKeyed`);
+  conservative prune-only keys (anchored LIKE, and IN-list ranges once #704
+  lands) are marked inexact and also disqualify the fold, which then falls back
+  to the correct scalar path. Covered by new cases in `ungrouped_vector_agg`
+  (#715).
+
 ## [1.0-alpha2] - 2026-08-18
 
 ### Added
--- a/src/columnar.h
+++ b/src/columnar.h
@@ -911,6 +911,8 @@ typedef struct PgColumnarGroupStats
 
 extern ScanKey PgColumnarBuildScanKeys(List *qual, Index scanrelid,
 									 TupleDesc tupdesc, int *nkeys);
+extern bool PgColumnarQualsExactlyKeyed(List *qual, Index scanrelid,
+									  TupleDesc tupdesc);
 
 /*
  * How many chunk groups the planner samples when estimating how much a
--- a/src/columnar_customscan.c
+++ b/src/columnar_customscan.c
@@ -407,7 +407,7 @@ pgcolumnar_like_prefix_scankey(OpExpr *op, Var *var, Const *con,
 
 static int
 pgcolumnar_clause_to_scankey(Node *clause, Index scanrelid, TupleDesc tupdesc,
-						   ScanKey key)
+						   ScanKey key, bool *exact)
 {
 	OpExpr	   *op;
 	Node	   *leftop;
@@ -417,6 +417,16 @@ pgcolumnar_clause_to_scankey(Node *clause, Index scanrelid, TupleDesc tupdesc,
 	bool		varOnLeft;
 	TypeCacheEntry *tce;
 	StrategyNumber strat;
+
+	/*
+	 * An EXACT key expresses its clause completely: a row satisfies the clause
+	 * iff it satisfies the key. Only the plain btree comparison below is exact.
+	 * Conservative keys that merely PRUNE (the anchored-LIKE bound, #426) leave
+	 * exact = false, so a caller that re-checks keys as its whole filter (the
+	 * batch fold, #715) can refuse them. Default false; a dropped clause (return
+	 * 0) is not exact either.
+	 */
+	*exact = false;
 
 	if (!IsA(clause, OpExpr))
 		return 0;
@@ -503,6 +513,7 @@ pgcolumnar_clause_to_scankey(Node *clause, Index scanrelid, TupleDesc tupdesc,
 	key->sk_collation = con->constcollid;
 	key->sk_argument = con->constvalue;
 
+	*exact = true;
 	return 1;
 }
 
@@ -530,11 +541,50 @@ PgColumnarBuildScanKeys(List *qual, Index scanrelid, TupleDesc tupdesc,
 	 */
 	keys = (ScanKey) palloc0(sizeof(ScanKeyData) * 2 * list_length(qual));
 	foreach(lc, qual)
+	{
+		bool		exact;	/* not needed here; skipping is conservative-safe */
+
 		n += pgcolumnar_clause_to_scankey((Node *) lfirst(lc), scanrelid, tupdesc,
-										&keys[n]);
+										&keys[n], &exact);
+	}
 
 	*nkeys = n;
 	return keys;
+}
+
+/*
+ * PgColumnarQualsExactlyKeyed
+ *		Whether the entire restriction list is expressed EXACTLY by scan keys:
+ *		every clause produces at least one key AND every one of those keys is
+ *		exact (a plain btree comparison), not a conservative prune-only key.
+ *
+ *		PgColumnarBuildScanKeys silently drops a clause it cannot turn into a key
+ *		(for example `<>`, which has no btree strategy) and emits conservative
+ *		bounds for others (anchored LIKE, #426; IN-list ranges once #704 lands).
+ *		That is correct for chunk-group skipping, where the executor still
+ *		re-applies the clause as a filter. It is WRONG for the batch fold, whose
+ *		only per-row filter IS the scan-key loop: a dropped or merely-pruning
+ *		clause lets non-matching rows be counted (issue #715). The fold gates on
+ *		this; an empty qual list is exactly keyed (there is nothing to filter).
+ */
+bool
+PgColumnarQualsExactlyKeyed(List *qual, Index scanrelid, TupleDesc tupdesc)
+{
+	ListCell   *lc;
+
+	foreach(lc, qual)
+	{
+		ScanKeyData scratch[2];	/* an anchored LIKE writes two keys */
+		bool		exact = false;
+		int			n;
+
+		n = pgcolumnar_clause_to_scankey((Node *) lfirst(lc), scanrelid, tupdesc,
+									   &scratch[0], &exact);
+		if (n < 1 || !exact)
+			return false;
+	}
+
+	return true;
 }
 
 /*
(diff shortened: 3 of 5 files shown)


jdatcmd (changes requested):
The fix is right and I verified it independently — requesting changes only for a real compiler warning and the now-mandatory merge with main (#718 landed first). Substance first:

## Verified on my lane (container `pgcolumnar-dev`, PG17 assert)

- **RED**: I reproduced the over-count on main myself this morning while filing #715 (`count(*) WHERE x <> 5` → 40000 with the GUC on, 39999 off), so the defect is confirmed on two lanes.
- **GREEN**: your head builds and `ungrouped_vector_agg` runs 49/49 here, including all four new #715 arms (count parity, sum parity, `Batch Fold: no` for `<>`, and the `i > 100` still-folds control — that control is the arm I most wanted to see, and it's there).
- The gate design (exactness reported at key production, `PgColumnarQualsExactlyKeyed` demanding one exact key per clause) is exactly the fix direction #715 asked for, and marking conservative keys inexact makes it the belt to #718's comment-and-test suspenders.

## Itemized asks

1. **`-Wdeclaration-after-statement` warning, reproduced here**: `src/columnar_customscan.c:423:9: warning: ISO C90 forbids mixed declarations and code`. The `*exact = false;` statement sits between the declarations (`OpExpr *op;` above, `Node *leftop;` etc. below). Move the assignment below the last declaration. Every build of the file warns until then.

2. **Merge main into the branch** (#718 landed, so this PR merges second — the coordination case from my #718 review, which you called). What I see from a local test-merge:
   - `src/columnar_customscan.c` **auto-merges, and the auto-merge is semantically correct**: the SAOP dispatch ends up after `*exact = false;` and never sets it, so `PgColumnarQualsExactlyKeyed` refuses IN-list keys (n=2, exact=false) — the agreed resolution, including keeping the single-distinct-value equality key inexact for now. Please still read the merged function rather than trusting the auto-merge, and note `scratch[2]` in `PgColumnarQualsExactlyKeyed` stays sufficient (SAOP also writes at most two keys).
   - `src/columnar_vector.c` **conflicts**: your gate insertion vs #718's LOAD-BEARING comment at the same spot. Keep both, and please reword the comment's "you must first give keys an exactness marker" clause — that marker now exists directly below, so the comment should say the gate underneath is that marker (belt), with `native_saop_pushdown`'s no/yes arms and `ungrouped_vector_agg`'s #715 arms as the live checks.
   - `CHANGELOG.md` conflicts: plain union of the two Unreleased sections. Your entry's "IN-list ranges once #704 lands" phrasing should drop the "once ... lands" now that it has.
   - Merge discipline per the house rule: `git diff --name-only --diff-filter=U` empty and `git diff --check` before committing.

3. **Post-merge, run the two suites together on the merged tree**: `ungrouped_vector_agg` (49) and `native_saop_pushdown` (30). That is the "verified together on merge" your own note asks for — the SAOP guard arms and the exactness gate now check each other.

Happy to approve on the next push with those in.

OffgridwithJD:
Self-review update (no code change — the gate holds under attack). I tried to make `PgColumnarQualsExactlyKeyed` produce a wrong result two ways and could not.

Ran an 11-shape matrix on an all-by-value table against a heap oracle (count + sum), each with the fold decision printed:

- **Admits every exact shape** (fold=yes, matches heap): `x = 5`, `x > 100 AND x < 200`, commuted `5 < x`, `BETWEEN`, `>=` boundary, `= no-match`. So the gate is not over-strict — exact keys still fold.
- **Rejects every inexact shape** (fold=no, matches heap): `x <> 5`, `x = 5 OR x = 6` (BoolExpr), `x > 100 AND y <> 3`, `x > 100 AND x IN (...)`.
- The decisive one: `x >= 100 AND x <= 39900 AND x <> 500 AND y > 2` — three exact clauses and one `<>` — correctly falls **off** the fold (one inexact clause disqualifies the whole WHERE) and still returns the right count.

No false-exact: `exact = true` is set only in the plain-btree `Var op Const` success branch (one key, real btree strategy, re-checked exactly per row), commuted operators included; LIKE, SAOP, BoolExpr, and no-key clauses all stay inexact. The only cost is that a WHERE mixing exact clauses with an unkeyable one folds nowhere — a conservative performance choice, never a wrong answer.

Still stands: the #718 merge note (both PRs edit `pgcolumnar_clause_to_scankey`; the SAOP dispatch must sit after this PR's `*exact = false` init and leave it false). I'll handle it on whichever merges second.

OffgridwithJD:
Pushed — all three asks are in (head is the merge commit `93736f7`, over the reworded fix `a2a09dd`).

**1. `-Wdeclaration-after-statement`.** Fixed: `*exact = false;` now sits below the last declaration in `pgcolumnar_clause_to_scankey`, not between `OpExpr *op;` and the rest. Verified: `grep -c` for the warning in a clean `make` log is **0**.

**2. Merged main (#718 landed first).** Read the merged functions rather than trusting the auto-merge, as you asked:
- `columnar_customscan.c` did in fact conflict (my warning-fix moved the same region your SAOP dispatch landed in), but resolves to exactly the agreed shape: `*exact = false;` first, then the SAOP dispatch returns early leaving it false. So `PgColumnarQualsExactlyKeyed` sees `n = 2, exact = false` for an IN-list and refuses it — the single-distinct-value equality key stays inexact too. `scratch[2]` is sufficient (SAOP writes at most two keys; I confirmed the `return 2` path).
- `columnar_vector.c` conflicted as predicted. Kept both, and reworded the "you must first give keys an exactness marker" clause: the comment now says the gate directly below **is** that marker (belt), with the byval gather guard as the independent second reason LIKE/varlena filters fall back, and names both `native_saop_pushdown`'s and `ungrouped_vector_agg`'s arms as the live checks.
- `CHANGELOG.md`: unioned under one `[Unreleased]` (Added: your #704 entry; Fixed: my #715 entry), and dropped the "once #704 lands" phrasing — it now reads "the IN-list ranges added above for #704".
- `git diff --check` clean, no conflict markers remain.

**3. Both suites together on the merged tree (pg18a assert):** `ungrouped_vector_agg` **49/49** and `native_saop_pushdown` **30/30**, plus `harness_selftest` 110 and `docs_style` 9. The SAOP guard arms and the exactness gate now check each other and both stay green.

Separately, self-reviewing my own fix I ran an 11-shape heap-oracle matrix (posted above): the gate admits every exact shape and rejects every inexact one, including a compound WHERE where one `<>` among exact clauses correctly drops the whole fold. No false-exact. Ready for another look.

jdatcmd (approved):
APPROVE — all three asks verified on my lane (PG17 assert, `pgcolumnar-dev`) at the merged head `93736f7`:

1. **Warning gone**: clean build, zero `warning` lines; `*exact = false;` now sits below the declarations.
2. **Merge resolution read, not trusted**: `pgcolumnar_clause_to_scankey` has the agreed shape (init first, SAOP dispatch returns with exact still false, single-distinct-value equality key included), and the reworded LOAD-BEARING comment in `columnar_vector.c` is now accurate — the gate is named as the exactness marker, the byval gather guard as the independent second reason for LIKE/varlena fallback, both suites cited as the live checks.
3. **Joint suites on the merged tree, my lane**: `ungrouped_vector_agg` 49/49 and `native_saop_pushdown` 30/30 — the belt and the suspenders check each other on a second lane and major.

The 11-shape heap-oracle matrix in your self-review is the right instrument for a gate like this (admits every exact shape, rejects every inexact one, compound-WHERE drop included). Merging on full green; if #716's merge re-conflicted the CHANGELOG I'll push the mechanical union to the branch as with the others.

Issue #720, opened 2026-08-24
bug: merge join with multi-key merge cond over index-fetched columnar rows errors 'pfree is not supported by the bump memory allocator' (PG17)


jdatcmd:
## Repro (PG17 assert, main d016e23, container `pgcolumnar-dev`)

```sql
SET pgcolumnar.stripe_row_limit=2000;
CREATE TABLE f (id int, v text) USING pgcolumnar;
INSERT INTO f SELECT g, 'v'||g FROM generate_series(1,100000) g;
CREATE INDEX f_id ON f (id);
CREATE TABLE h (id int, v text);          -- heap
INSERT INTO h SELECT g, 'v'||g FROM generate_series(1,2000) g;

SET enable_seqscan=off; SET enable_bitmapscan=off;
SET pgcolumnar.enable_custom_scan=off;
SET enable_nestloop=off; SET enable_hashjoin=off;

SELECT count(*) FROM h JOIN f ON f.id = h.id AND f.v = h.v;
-- ERROR:  pfree is not supported by the bump memory allocator
```

The GUCs only force the plan shape (Merge Join over a Sort over `Index Scan using f_id`); a planner can choose it naturally, so this is a reachable wrong-error/failed-query bug, not a corner-case footgun.

## Variants that narrow it

- `JOIN ... ON f.id = h.id` (single int key): works.
- `JOIN ... ON f.v = h.v` (single text key, still sorts the fetched text): works.
- Both keys (`id` and the fetched text `v`) in the merge condition: errors.
- Plain `SELECT v FROM f WHERE id < 5000 ORDER BY v` through the same forced index scan + sort: works.

So the failure needs a multi-key merge sort whose second key is a by-reference datum fetched through the columnar index-fetch (deferred-decode) path. PG17's tuplesort keeps sort tuples in a bump context (bump allocator is new in 17); something on the columnar fetch/slot side `pfree`s a chunk that landed in that context. Found while building #709's fetch suite — the failing arm predates any #709 code (unfixed main .so).

## Discipline

TDD: the repro above is the RED; the fix must identify WHO pfrees the bump chunk (slot copy path? deferred-decode re-decode freeing a stale datum?) rather than wrapping the pfree; removal proof; run on PG16 (no bump allocator) and PG18/19 to map the affected majors.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #157, opened 2026-07-26
Fetch by row number decodes every column, so a wide table falls off the fetch cache and back to per-row group decode (#143 option D)


OffgridwithJD:
Follow-up to #143, which was closed by #151 and #152. Options B and C are genuinely done — the quadratic behaviour that issue was titled for is gone. Option D, the needed-columns bitmap, was never implemented, and this is what it costs.

`ColumnarReadRowByNumber` takes no projection:

```c
ColumnarReadRowByNumber(Relation rel, Snapshot snapshot, uint64 rowNumber,
                        Datum *values, bool *nulls)
```

It loops `for (c = 0; c < natts; c++)` and decodes every column of the row group, whatever the caller needs. On a narrow table that is mild waste. On a wide one it is a cliff, because the decoded bytes are measured against the fetch cache's 32 MB cap and decoding *all* columns is what pushes an entry over it. Once over, the entry is dropped after every fetch and the next fetch decodes the whole group again — back to pre-#148 behaviour.

## Measured

18.4 at `1be027b`. 200,000 rows, 2,000 rows fetched by index, reading **one** column (`SELECT sum(c1) ... WHERE id BETWEEN ...`), timed server-side:

| table | fetch of the same 2,000 rows |
| --- | --- |
| 3 columns | 10.8 ms |
| 11 columns | 56,432.8 ms |
| 41 columns | 225,929.6 ms |

Same rows, same index, same single column read. The only thing that changed is how many columns the table has that the query does not touch.

That it is the cap, and not simply per-column decode cost, shows when the row group is made smaller so that decoding every column still fits. 41 columns throughout, 2,000 fetches:

| rows per row group | fetch time |
| --- | --- |
| 200,000 | 300,031.4 ms |
| 50,000 | 77,768.2 ms |
| 10,000 | **19.9 ms** |

A 20x smaller row group is 15,000x faster for the same work. Nothing about the query changed; the entry simply fits under the cap and stays cached.

(My MB estimates while testing undercounted, because the entry also holds the group's raw bytes and the #152 position indexes, not just the decoded columns. So the crossover sits at a smaller nominal size than a naive `natts * width * rows` suggests. The direction and the cliff are what the measurements establish; the exact threshold I have not pinned down.)

## The worst caller does not want values at all

`columnar_index_delete_tuples` in `columnar_tableam.c` calls it in a loop purely to learn whether a row is live:

```c
[...]

PR #721, opened 2026-08-24
opt: memoize the by-row-number fetch's row-group lookup (#709)

### LINKED ISSUE #709: opt: index/bitmap fetch does O(K·G) linear row-group search
## Finding (O-7, from the 2026-08-17 aggressive security/opt/modularity audit — `design/AUDIT_2026_08_SECURITY_OPT_MODULARITY.md`)

`columnar_reader.c:3348-3359` — index/bitmap fetch re-reads the full
row-group list and does a linear O(G) search per row, giving O(K·G) overall
for K fetched rows across G row groups.

## Proposed fix

Binary search over the row-group list (already sorted by
`firstRowNumber`), or cache the last-resolved group the way
`columnar_delete_vector.c`'s `lastGroup`/`lastChunk` cache already does for
the sequential-delete case.

med, complexity — EXPLAIN/timing-provable (K·G scaling should collapse to
K·log(G) or better).

## Discipline

Benchmark with G large enough that O(K·G) vs O(K·log G) is measurable;
removal-proof.

jdatcmd:
## What

`pgcolumnar_fetch_row` read the ENTIRE row-group list out of the catalog on every call — twice per index-fetched row (liveness settle + deferred decode, #157) — so K fetched rows cost ~2K catalog index scans: **measured 6001 scans of `row_group_pkey` for 3000 rows**. The lookup is now memoized per (storage, command, **snapshot content** — xmin/xmax/curcid/xip/subxip compared by value, so pointer reuse cannot alias), with the delete-vector reader's proven last-hit + linear-walk + **refresh-on-miss** shape. The refresh re-reads under the fetch's own snapshot, preserving the old contract that a group flushed earlier in the same statement is visible. Post-fix: the same 3000 fetches cost **2** scans. Only `SNAPSHOT_MVCC` memoizes; SnapshotDirty (unique checks) and SnapshotAny (trigger re-fetches) keep the per-fetch read. No binary search, deliberately: the catalog read is the whole win and a comparator would be code no test can turn red.

Design + full hazard ledger: `design/FETCH_GROUP_LOOKUP_PLAN.md`.

## The two resets, and how each earned its keep

A HIT is sound because within one (cid, snapshot content) the visible group set can only **gain** rows — with two exceptions, both reset and both removal-proven:

- **Subtransaction abort** flips a savepoint-flushed group's xmin verdict with NO key change. Severed: the aborted rows resurrect through an open cursor (`MOVE ALL` reads 102990 for 99990).
- **Row-group retirement.** This one has a story: PG17 probing showed the rewrite path crossing an incidental CCI that masks a missing reset, so it was briefly removed under the deletion-proof rule with a sentinel arm left behind — and the sentinel **fired on PG18** (no such CCI there: one doubled all-NULL row, 501). Adversarial review then found the stronger shape: `pgcolumnar.compact()` retires FULLY deleted groups mid-statement with no flush and no CCI on any major. Severed on a non-assert build, its arm resurrects **all 2000 rows** of the retired group (`10|t|2000`). The reset ships with both arms as its proof (assert builds mask the compact arm via an assert-only CCI; the matrix's non-assert majors carry it).

Poisoning manifests as **all-NULL rows** (group metadata from the memo, fresh `column_chunk` read sees nothing), so every hazard arm counts ROWS, never values — the value-counting first drafts of these arms sat green over 3000 resurrected rows.

## Verification

- `native_fetch_group_memo` (14 checks) **green on PG17 (assert), PG18 (non-assert), PG19 (assert)**. Instrument discipline: xact-numscans BASELINE/AFTER delta in one session, heap-scan delta 0 (a broken index resolution cannot fake the bound), nested-loop plan shape pinned with join methods forced (a merge join turns the probe into a full scan — and separately trips pre-existing #720), 23505 asserted by SQLSTATE.
- Removal proofs, each exact: memo severed → only the scan-delta arm red; subxact reset severed → only the cursor arm red; retirement reset severed → 501 + 2000-resurrected on pg18_nc.
- Regression battery green on PG17: `native_fetch_cache`, `native_fetch_bigcap`, `native_fetch_interrupt` (placement arm updated to pin the check before the lookup CALL and that the lookup owns the catalog read), `native_fetch_position`, `native_fetch_projection`, `index_only`, `unique_conc`, `corruption`, `native_reclaim`, `native_recluster`, `autovacuum`, plus `harness_selftest` (110) and `docs_style`.
- Two multi-agent adversarial passes (design-stage hazards; finished-diff review). The review's material findings are all folded in: the compact() retire path, the `compact_rewrite >= 0` tautology (now `= 1`), a PREPARE-path dangling-context window (memo now resets there), `SNAPSHOT_HISTORIC_MVCC` excluded from the memoizable predicate, an inert forced GUC dropped, and the wrong-fact comments about xact-callback context lifetime corrected.

## Found along the way

- **#720** (filed): merge join with a multi-key merge cond over index-fetched columnar rows errors with `pfree is not supported by the bump memory allocator` on PG17 — pre-existing, minimal repro in the issue.

Closes #709.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,6 +32,18 @@ pinned at `1.0-dev` or `1.0-alpha`, each true until the next version shipped.
   assert build. `bench/provision.sh check` continues to report their presence;
   installing them is now a scripted, pinned step instead of a manual one (#702).
 
+### Changed
+
+- An index or bitmap fetch no longer reads the whole row-group list out of
+  the catalog for every row. The list is memoized per command and snapshot,
+  with a refresh on any miss, so a group flushed earlier in the same
+  statement stays visible. Measured on 3000 index-fetched rows: 6001 catalog
+  index scans before, 2 after. The memo resets on subtransaction abort and
+  on row-group retirement, and a new suite, `native_fetch_group_memo`,
+  proves both by resurrection: through an open cursor after ROLLBACK TO
+  SAVEPOINT, and through stale index entries after a same-statement
+  compaction. (#709)
+
 ### Fixed
 
 - The ungrouped vectorized aggregate (`pgcolumnar.enable_ungrouped_vector_agg`)
--- a/design/FETCH_GROUP_LOOKUP_PLAN.md
+++ b/design/FETCH_GROUP_LOOKUP_PLAN.md
@@ -1,0 +1,164 @@
+# O(K·G) index/bitmap fetch row-group lookup (#709)
+
+## Problem
+
+`pgcolumnar_fetch_row` (columnar_reader.c) runs once per fetched TID: per row
+of an index or bitmap scan, per duplicate probe of `_bt_check_unique`, and per
+UPDATE re-fetch. Each call reads the ENTIRE row-group list out of the catalog
+(`PgColumnarReadRowGroupList`: a `systable` index scan over
+`pgcolumnar.row_group` plus one palloc'd node per group) and then walks it
+linearly. K fetches over G groups cost K catalog index scans and O(K·G)
+comparisons; the list read dominates. The code's own comment records the
+reason it was left uncached: a group flushed earlier in the same statement
+must become visible to a later fetch (read-your-writes through
+`PgColumnarCatalogSnapshot`, which advances `curcid`).
+
+## Design
+
+A backend-global row-group-LIST memo beside the existing decoded-group cache
+(`columnarFetchCache`, which already memoizes per `(storageId, groupNumber,
+cid)`), reusing the delete-vector reader's refresh-on-miss discipline
+(`delete_vector_find_row_group`), with these rules:
+
+1. **MVCC snapshots only.** A `SnapshotDirty` probe (`_bt_check_unique`) or
+   `SnapshotAny` fetch (trigger OLD-row re-fetch) keeps today's per-fetch
+   read. Dirty visibility changes as other transactions commit; the unique
+   path relies on seeing current state per probe.
+2. **Key = (storageId, CommandId, full snapshot content).** The cached list
+   is valid only for a snapshot that judges every catalog tuple identically
+   to the one that built it: compare `xmin`, `xmax`, advanced `curcid`,
+   `xcnt` + `xip[]`, `subxcnt` + `subxip[]`, `suboverflowed`,
+   `takenDuringRecovery`. Content comparison, not pointer identity: pointer
+   reuse cannot alias, and two references to one logical snapshot hit. A
+   mismatch on anything rebuilds. A handful of slots (LRU, like the group
+   cache's 4) covers interleaved cursors.
+3. **Refresh on miss.** A rowNumber no cached group covers triggers ONE
+   re-read under a fresh `PgColumnarCatalogSnapshot`, then a second lookup
+   (the delete-vector two-attempt shape). This is what preserves the
+   same-statement-flush contract: a group flushed after the memo was built
+   is a miss, and the refresh sees it (same cid, `curcid` advanced). A group
+   never covers new row numbers after creation, and within one command the
+   own-transaction catalog state visible at `curcid+1` can gain groups but
+   not lose them (reclaim/rewrite runs as its own statement, and cross-
+   transaction deletions are invisible to the same MVCC snapshot by
+   definition), so a HIT cannot be stale.
+4. **Last-hit fast path + linear walk, NO binary search.** The adversarial
+   pass showed a sorted-array binary search is unfalsifiable here: the
+   instrument counts catalog reads, which the memo alone removes, and the
+   residual in-memory walk over G entries (fixtures 20-40 groups; 6M rows at
+   the default stripe is 40) is microseconds. The proven sibling
+   (`delete_vector_find_row_group`) is exactly last-hit + linear + two
+   attempts, unsorted. Shipping a comparator with no removal proof fails the
+   house rule; the memo keeps the sibling's shape.
+5. **Reset with the group cache, PLUS two discards the group cache gets away
+   without.** The memo drops where `columnarFetchCache` drops (top-level
+   commit/abort, executor end), and additionally:
+   - **on subtransaction abort** (the existing subxact callback): a group
+     flushed by a live savepoint is visible when the memo is built, and
+     `ROLLBACK TO SAVEPOINT` flips its `xmin` verdict with NO change in
+     snapshot content, cid, or any reset event the group cache watches.
+     Today's per-fetch read re-judges `xmin` every time; a memo HIT would
+     serve the aborted group's rows off its durably-written file bytes. The
+     decoded-group cache survives this only BECAUSE the per-fetch list read
+     answers first — the memo takes over that role, so it must reset.
+   - **on row-group retirement** (`PgColumnarRetireGroup`) — and the
+     journey of this reset is the record worth keeping. The hazard analysis
+     demanded it; probing on PG17 then showed the mid-statement rewrite path
+     crossing a `CommandCounterIncrement` (the flush's storage-row update),
+     which rebuilds the memo anyway, so the reset failed every PG17 removal
+     proof and was REMOVED under the deletion-proof rule, with the suite's
+     same-statement rewrite arm left as a sentinel. The sentinel then fired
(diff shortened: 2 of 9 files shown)


OffgridwithJD (changes requested):
Reviewed as ChronicallyJD (cross-review; I do not merge). This is strong work and the correctness scaffolding is unusually good. I verified the load-bearing claims independently and I'm requesting one change: the hazard ledger has a third loss path it enumerates as complete, and the safety there rests on an unstated command-id invariant of exactly the class that already bit this PR once.

## What I verified (independently, prove-don't-trust)

- Builds clean (0 warnings). `native_fetch_group_memo` (14) plus `native_fetch_cache/bigcap/interrupt/position/projection`, `index_only`, `unique_conc`, `corruption`, `harness_selftest` (110), `docs_style` (9) all green on PG17 assert.
- The retirement reset is genuinely load-bearing **on a non-assert build**, which is the point that matters since (as you note) assert builds mask it via an incidental CCI. On pg18n I severed only `PgColumnarGroupMemoReset(true)` in `PgColumnarRetireGroup` (metadata.c:798): the compact arm resurrects all 2000 rows of the retired group (`10|t|2000`) and the retire arm doubles a row (`501`), while the subxact arm stays green. So the reset is proven, the suite counts rows, and the non-assert matrix carries the proof the assert matrix cannot. Nicely done.
- The snapshot key compares a superset of `XidInMVCCSnapshot`'s inputs (xmin/xmax/xip/subxip/suboverflowed/takenDuringRecovery, plus curcid conservatively), and stores a content copy, so a false HIT from an omitted field or pointer reuse isn't possible. `SNAPSHOT_HISTORIC_MVCC` correctly excluded.
- The "only gains except two losses" invariant hinges on `row_group` rows never mutating in place. Confirmed structurally: the only `row_group` writers are `PgColumnarInsertRowGroupRow` and the two deleters; the in-file `CatalogTupleUpdate`s are on `delete_vector`/`storage`. So the "rewrite mutates bounds in place" hazard class can't occur — recluster/compact retire-and-add at a higher group number. Good.

## The change I'm requesting — the ledger's loss cases are incomplete

The design doc and the `reader.c` comment state the mid-statement loss cases are exactly **subxact-abort** and **retirement**. There is a third: **`PgColumnarDeleteMetadata`** (metadata.c:1639) deletes all `row_group` rows and does **not** reset the memo, and one caller — `pgcolumnar_relation_nontransactional_truncate` (tableam.c:742) — wipes the metadata **while keeping the same storageId** (the vacuum-path callers use a *new* `oldStorageId`, so their memo is naturally dead; only the truncate path is same-storage). So a memo keyed on `(storageId, cid, snapshot)` could, in principle, survive a wipe of its own storage's groups.

It is **not exploitable today** — I checked. The memo is backend-global and TRUNCATE marks `currentCommandIdUsed`, so `CommandCounterIncrement` advances the cid before any subsequent fetch; a same-session `SELECT (build memo); TRUNCATE t; INSERT; SELECT` returns the new data (`id=1 -> w1`), count 1499, 0 null rows — the post-wipe fetch misses the stale memo and refreshes. The safety is entirely the "any metadata-losing write advances the command id before the next fetch" invariant.

But that invariant is **unstated and untested**, and it is the same shape as the one the retirement reset already tripped over: that reset was removed on a cid/CCI theory that held on PG17 and broke on PG18. Leaving a second cid-dependent-safety path undocumented in a ledger that claims its loss cases are complete is the trap this PR otherwise works hard to avoid.

**Ask (either resolves it):** add `PgColumnarGroupMemoReset(true)` to `PgColumnarDeleteMetadata` — one line, symmetric with the `RetireGroup` reset, and it makes the memo's correctness independent of the truncate/cid path — **or**, if you'd rather not, state the command-id-advance invariant explicitly in the ledger and add a note at `PgColumnarDeleteMetadata` so a future refactor of the truncate/cid path can't silently break it. I'd prefer the reset.

## One coordination note (not part of this PR)

This PR leaves #720 unfixed by design, and my #722 (the #720 fix) refactors the same `pgcolumnar_fetch_row` region (it moves the version check into the fetch's private context and builds the needed-column Bitmapset in a scratch context). Whichever of #721/#722 merges second will conflict there; both changes must survive — the memo lookup and the two pfree-context fixes are independent and compatible. I'll handle the merge on #722's side if #721 lands first.

jdatcmd:
Pushed `b116e5e` — your ask taken, preferred option: `PgColumnarGroupMemoReset(true)` now sits at the end of `PgColumnarDeleteMetadata`, symmetric with the retirement reset, so the memo's correctness no longer leans on the truncate path's cid accident.

Also carried into the record, since the enumeration claimed completeness and was wrong:
- the ledger (`design/FETCH_GROUP_LOOKUP_PLAN.md`) now lists the metadata wipe as the THIRD loss case, with your probe result and the explicit statement that no removal proof for this reset is constructible on current majors (TRUNCATE's cid advance masks it) — recorded as exactly the retirement-reset shape, which is why it ships anyway;
- the `reader.c` memo header's loss list gains the third bullet;
- the suite gains a characterization arm (same-transaction TRUNCATE + reinsert → fetches see exactly the new rows, `count(*) == count(v)` so no NULL resurrection), 15 checks now, green on PG17 assert and pg18_nc.

On the #722 coordination: agreed the changes are independent and compatible; I'm reviewing #722 next, and whichever lands second takes the `pgcolumnar_fetch_row`-region merge (I'll handle it on this side if #722 goes first).

jdatcmd:
#722 merged first, so the composition merge landed on this side as agreed (`9f9fa83`): the composed `pgcolumnar_fetch_row` order is interrupt check → fetch-context switch → version check (your #720 fix, frees in the AllocSet) → memo lookup (#709, contexts under TopTransactionContext regardless of ambient). Read, not trusted: verified by eye and by the joint run — `native_fetch_group_memo` (15), `native_fetch_sort_context` (7), `native_fetch_interrupt`, `native_fetch_cache` all green on the merged tree on PG17 assert AND pg18_nc, plus `harness_selftest` 110.

Ready for your re-review: the b116e5e commit carries the `PgColumnarDeleteMetadata` reset you asked for plus the ledger's third loss case and the truncate characterization arm.

OffgridwithJD (approved):
APPROVE — my REQUEST_CHANGES is resolved. Re-reviewed at `9f9fa83` (b116e5e + the #722 composition merge).

## The requested change — done, preferred option

`PgColumnarGroupMemoReset(true)` now sits at the end of `PgColumnarDeleteMetadata`, after the `row_group` delete, symmetric with the `RetireGroup` reset. The memo's correctness no longer leans on the truncate path's command-id accident. And the ledger's completeness gap is closed the right way:

- the `reader.c` memo header gains the third loss bullet (the same-storage metadata wipe);
- the design doc lists it as the third loss case, records *why* the vacuum callers are exempt (they wipe an old storage id), and states plainly that no removal proof is constructible on current majors because TRUNCATE's cid advance masks it — with my probe result as the evidence. Recording "this reset ships without a removal proof, and here is exactly why" is the honest form, and it matches the retirement-reset precedent where the cid mask was real and major-dependent.
- the suite gains a same-transaction TRUNCATE+reinsert characterization arm (15 checks now).

## The #722 composition — independently verified

#722 landed first, so the `pgcolumnar_fetch_row`-region merge came out on this side. I read it, not trusted it: the composed order is `CHECK_FOR_INTERRUPTS` → `AllocSetContextCreate("columnar fetch")` + switch → version check (my #720 fix, so its catalog-scan frees land in the AllocSet) → `pgcolumnar_lookup_row_group` (the memo). My `decode_upto` scratch-context change is intact in `columnar_tableam.c`. Both fixes coexist as intended — the memo lookup's own contexts hang off `TopTransactionContext` regardless of the ambient, so running it inside the fetch context is fine.

Joint run, built from `9f9fa83` on my side:
- **PG17 assert**: `native_fetch_group_memo` 15, `native_fetch_sort_context` 7 (my #720 suite — the merge-join repro no longer errors *and* the memo is engaged), `native_fetch_interrupt`, `native_fetch_cache`, `harness_selftest` 110, `docs_style` 9 — all green, 0 warnings.
- **pg18n non-assert** (where the memo hazards actually show): same six suites all green, 0 warnings.

So #720's pfree fix and #709's memo are confirmed independent and compatible on both an assert and a non-assert build. Good to merge.

PR #722, opened 2026-08-24
fix: keep deferred-decode fetch pfrees out of the caller's context (closes #720)

### LINKED ISSUE #720: bug: merge join with multi-key merge cond over index-fetched columnar rows errors 'pfree is not supported by the bump memory allocator' (PG17)

OffgridwithJD:
Closes #720.

## The bug

A merge join over index-fetched columnar rows aborts with `pfree is not supported by the bump memory allocator` on PostgreSQL 17+. The plan is `Merge Join → Sort → Index Scan`, which a planner can choose on its own.

## Root cause (the issue's hypothesis was close but the site was different)

A backtrace (`backtrace_functions='BumpFree'`) + `addr2line` pinned it. A columnar index scan returns a **deferred** slot; when tuplesort materialises it via `copy_minimal_tuple` → `pgcolumnar_slot_decode_upto`, `CurrentMemoryContext` is PG17's **bump context**, which forbids `pfree`. The deferred decode pfreed there **twice**, and the first masked the second:

1. `pgcolumnar_slot_decode_upto` (`columnar_tableam.c`) builds the needed-column `Bitmapset` with `bms_add_member` and frees it with `bms_free` — in the ambient (bump) context.
2. `pgcolumnar_fetch_row` (`columnar_reader.c`) runs the per-command format-version check *before* switching into its private `"columnar fetch"` AllocSet; that check opens a catalog table and its btree scan frees a search stack (`_bt_freestack`).

The by-reference key is incidental (it forces the tuple to be materialised through the deferred slot); the illegal operations are the two frees, not any datum. This is why the single-key variants worked — they never defer a second key through the sort.

## The fix

Build the `Bitmapset` in a private AllocSet in `decode_upto`, and move the version check below the context switch in `fetch_row`. Both transient frees now land in a pfree-supporting context. The decoded values are unaffected — they were already copied into the caller's context, where `palloc` is legal — so only the frees were ever the problem. (This also covers a wide table's multi-word `Bitmapset`, whose `repalloc` the bump allocator equally forbids, since the whole set now lives in the private context.)

## Verification (TDD, prove-don't-trust)

New suite **`native_fetch_sort_context`** drives the exact plan and checks the count and `sum(id)` against a heap mirror (an errored query returns no value and fails the numeric check, so the red is the abort itself).

- **RED → GREEN** on PG17 assert: the two-key and three-key arms fail before the fix, all pass after.
- **Per-site removal proof**: reverting *only* the `decode_upto` change reddens (the `bms_free` site returns); reverting *only* the `fetch_row` change reddens (the version-check site returns). Both halves are load-bearing.
- **Majors mapped**: PG15.18 and PG16.14 (no bump allocator) PASS even unfixed — never affected; PG18.4 and PG19 FAIL before, PASS after. So the reachable majors are 17/18/19.
- No regression: `harness_selftest` 110 (registration + C-sort), `docs_style` 9, clean build with zero warnings.

The single-key controls (which always worked) stay green, confirming the fix does not disturb the working paths.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,19 @@ pinned at `1.0-dev` or `1.0-alpha`, each true until the next version shipped.
 
 ## [Unreleased]
 
+### Fixed
+
+- A merge join over index-fetched columnar rows no longer aborts with `pfree is
+  not supported by the bump memory allocator` on PostgreSQL 17 and later. A
+  columnar index scan returns a deferred slot, and when a sort materialises it
+  the current context is PostgreSQL 17's bump context, which forbids `pfree`. The
+  deferred decode pfreed there twice: the needed-column `Bitmapset`
+  (`bms_add_member`/`bms_free`) and, on a storage's first fetch, the
+  format-version check's catalog scan (which frees a btree search stack). Both
+  now run in a private pfree-supporting context; the decoded values are
+  unaffected because they were already copied into the caller's context, where
+  `palloc` is legal. Covered by a new suite, `native_fetch_sort_context` (#720).
+
 ### Added
 
 - `IN (...)` and `= ANY(array)` predicates now drive chunk-group skipping.
--- a/src/columnar_reader.c
+++ b/src/columnar_reader.c
@@ -3416,6 +3416,18 @@ pgcolumnar_fetch_row(Relation rel, Snapshot snapshot, uint64 rowNumber,
 	 * cannot misread. Memoized per command (see the slot above) so a large index
 	 * scan pays one catalog check rather than one per row.
 	 */
+	tmp = AllocSetContextCreate(CurrentMemoryContext, "columnar fetch",
+								ALLOCSET_SMALL_SIZES);
+	oldContext = MemoryContextSwitchTo(tmp);
+
+	/*
+	 * The format-version check opens a catalog table and scans it, and that
+	 * scan frees a btree search stack with pfree. It must run inside this
+	 * AllocSet context, not the caller's: this path is reached from a slot
+	 * materialization during tuplesort_puttupleslot, whose current context on
+	 * PG17 is a bump context that does not support pfree (#720). Running it here
+	 * keeps the transient catalog allocations in a pfree-supporting context.
+	 */
 	if (wantValues &&
 		!(columnarFetchFmtOk && columnarFetchFmtOkStorageId == storageId))
 	{
@@ -3423,10 +3435,6 @@ pgcolumnar_fetch_row(Relation rel, Snapshot snapshot, uint64 rowNumber,
 		columnarFetchFmtOkStorageId = storageId;
 		columnarFetchFmtOk = true;
 	}
-
-	tmp = AllocSetContextCreate(CurrentMemoryContext, "columnar fetch",
-								ALLOCSET_SMALL_SIZES);
-	oldContext = MemoryContextSwitchTo(tmp);
 
 	metaSnapshot = PgColumnarCatalogSnapshot(snapshot);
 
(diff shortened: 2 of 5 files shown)


jdatcmd (approved):
APPROVE — reviewed and independently verified on my lane (`pgcolumnar-dev`, PG17 assert):

- **GREEN**: head builds with zero warnings; `native_fetch_sort_context` 7/7 (both plan-shape premises, both oracle arms, the widened three-key arm, both controls).
- **Both per-site removal proofs reproduced here**: reverting only the `decode_upto` scratch context reddens the three #720 arms; reverting only the `fetch_row` version-check move reddens them too — each half is load-bearing, exactly as claimed. The failure mode reads as `got [SET]` against the numeric oracle, so the error-is-the-red instrument works as documented.
- **Regressions green**: `native_fetch_cache`, `native_fetch_interrupt`, `index_only`, `corruption`, `harness_selftest`, `docs_style`.
- The diagnosis is better than my issue's hypothesis — the backtrace-pinned root cause (two transient `pfree`s: the Bitmapset teardown and `_bt_freestack` from the pre-switch version check, with the by-ref key merely forcing materialization) explains every variant in #720, including why the single-key shapes never failed. The fix shape is right: only the frees move; values still land in the caller's context where `palloc` is legal, which also covers the multi-word-Bitmapset `repalloc` case.
- Suite registered in sorted position; CHANGELOG under Fixed; majors mapped with 15/16 as never-affected controls.

On coordination: #721 will merge second (it's back in your court for re-review), so I'll take the `pgcolumnar_fetch_row`-region merge on #721's side — the composed order there will be interrupt check → tmp context switch → version check (yours) → memo lookup (mine), which keeps both properties: the version check's frees in the AllocSet, and the memo contexts under TopTransactionContext regardless of ambient context.

Merging on full green.

Issue #708, opened 2026-08-18
opt: grouped aggregation is row-at-a-time; ungrouped has batch_fold, grouped doesn't


jdatcmd:
## Finding (O-3, from the 2026-08-17 aggressive security/opt/modularity audit — `design/AUDIT_2026_08_SECURITY_OPT_MODULARITY.md`)

`columnar_vector.c:4324-4371` — grouped (`GROUP BY`) aggregation is fully
row-at-a-time today. Ungrouped aggregation already has `batch_fold`; grouped
has no equivalent batch path.

## Proposed fix

Extend the batch-fold approach to the grouped case (e.g. batch-partition by
group key, then fold per group), matching the ungrouped aggregation's
existing batching strategy.

## Severity / confidence (audit's own rating)

med value, complexity — high effort.

## Discipline

Measure grouped-aggregate throughput before/after with a fixed benchmark
query; removal-proof (revert → measurement regresses back to row-at-a-time
baseline).

jdatcmd:
## Measured profile before implementing, plus a trap for whoever builds this

I sized the prize with a deterministic instrument (instructions retired by the backend via `perf stat -p`, contention-immune — wall clock is unusable on this host, a competing process pegs a core and within-arm scatter hit 12%). Fixture: 4M rows, `k int` with 8 distinct groups, `v int`, PG17 non-assert, 5 reps, warm scan excluded.

**Grouped aggregate today: 18,397,240,147 instructions for 5 × 4M rows = ~920 instructions per row.**

For what is fundamentally "decode an int, hash-probe among 8 groups, add", that is enormous — a tight batch loop should be one to two orders of magnitude less. The prize here is real and large. Per-row costs visible in `pgcolumnar_groupagg_build`'s loop, each paid 4M times: `ResetExprContext`; the `bms_next_member` slot-staging loop; `ExecStoreVirtualTuple`; `ExecQual` when quals exist; `ExecEvalExprSwitchContext` per group key (fmgr + expression machinery even for a plain Var); `pgcolumnar_groupagg_lookup`'s `FunctionCall1Coll` hash per key plus fmgr equality on probe; and `pgcolumnar_apply_one` per aggregate. That is roughly six to eight fmgr calls per row, which matches ~920.

### Trap for the implementer: the ungrouped path is NOT a baseline

[...]

Issue #707, opened 2026-08-18
sec: debug metadata mutators lack owner check (advance_reserved_offset, set_metapage_version)


jdatcmd:
## Finding (S-F2, from the 2026-08-17 aggressive security/opt/modularity audit — `design/AUDIT_2026_08_SECURITY_OPT_MODULARITY.md`)

`columnar_vacuum.c:2102,2137` — `pgcolumnar_debug_advance_reserved_offset`
and `pgcolumnar_debug_set_metapage_version` mutate storage metadata with no
owner check.

Traced: these functions are currently unbound in the catalog (not exposed
as callable SQL functions today), so this is a latent footgun only — not
presently exploitable.

## Proposed fix

Add an owner check (or equivalent ACL gate) before these functions ever get
bound in the catalog, so the footgun can't surface later without a
deliberate decision to add the check.

## Severity / confidence (audit's own rating)

low, trace.

## Discipline

If/when these are bound, gate on `SQLSTATE 42501` (aclcheck_error), not
error text.

Issue #705, opened 2026-08-18
opt: Gorilla float decode is bit-at-a-time; apply the #501 word-at-a-time rewrite


jdatcmd:
## Finding (O-5, from the 2026-08-17 aggressive security/opt/modularity audit — `design/AUDIT_2026_08_SECURITY_OPT_MODULARITY.md`)

`columnar_encoding.c:640-660` (`br_get`) — Gorilla float decode reads the
bitstream one bit at a time. #501 already did the word-at-a-time rewrite for
integer bit-unpacking and measured a 19-21% improvement, but the same
technique was never applied to the Gorilla decoder.

## Proposed fix

Apply the #501 word-at-a-time bit-reading technique to `br_get` in the
Gorilla decode path.

## Severity / confidence (audit's own rating)

med, complexity.

## Discipline

Measure before/after with the same methodology #501 used (so the improvement
is comparable), removal-proof the fix (revert → measurement regresses).

jdatcmd:
Claiming #705-#708 (jdatcmd lane, PG17 in pgcolumnar-dev). Order: #706, #707, #705, #708 — one TDD PR each.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

jdatcmd:
## Measured and dispositioned: the #501 technique does NOT transfer to Gorilla. Shipping zero code.

I implemented the word-at-a-time `br_get` (with a per-bit reference oracle wired into `debug_encoding_selftest`, removal-proven: severing the high-byte fold or the width mask makes the oracle fail), then measured it. The measurement refutes the premise, so the rewrite is not being shipped.

### The instrument, and why the obvious one was unusable

Wall-clock A/B first said "10–43% faster", then said the opposite depending on arm ORDER. The cause was environmental: a process on the host was pegging a core, and within-arm scatter reached 12% — wider than the effect being hunted. A sequential A/B would have shipped a fabricated number.

So the decision was made on **instructions retired by the backend** (contention-immune, attached via `perf stat -p` to a single long-lived session, warm scan excluded, `count(*)` baseline subtracted at ~1.3M instructions i.e. negligible). Fixture premise asserted every run: 200/200 vectors actually GORILLA-encoded, on `random()::float8` — the shape with ~8.4 stored bytes/row, i.e. **long** meaningful fields, deliberately chosen because the smooth pattern the invariants suite uses (`1.0 + g*1e-15`, 0.44 bytes/row) is the *worst* case for this hypothesis.

[...]

PR #723, opened 2026-08-24
sec: refuse userinfo on the objstore write path, matching the read path (#706)

### LINKED ISSUE #706: sec/mod: userinfo '@' guard missing on objstore write path (os_write_handle)
## Finding (M-5, from the 2026-08-17 aggressive security/opt/modularity audit — `design/AUDIT_2026_08_SECURITY_OPT_MODULARITY.md`)

`columnar_objstore_module.c:1691-1715` — the userinfo `@` guard (SSRF
hardening on URL parsing) is present in `os_open` (read) and `http_request`,
but missing in `os_write_handle` (write/delete/list).

Traced: the write path is still fail-closed today via the single
`os_connect` choke point (link-local + CRLF guards apply regardless), so
this is not currently exploitable. It's a sibling-divergence /
defense-in-depth gap: a diagnostic-message divergence, and userinfo bytes
end up in `h->url` on the write path where the read path scrubs them.

## Proposed fix

Add the same userinfo `@` guard to `os_write_handle` that `os_open` and
`http_request` already have, so all three entry points are consistent.

low, trace (not currently reachable as an exploit — consistency/DiD fix).

## Discipline

Characterization test showing the current diagnostic-message /
userinfo-in-URL divergence between read and write paths, then the guard,
then removal proof.

jdatcmd:
## What

`os_open` (read) and `http_request` refused `user@host` in an http(s) authority from the start; `os_write_handle` (export sink, delete, list) parsed the same authority **without** the guard. The write path still failed closed — an `@`-carrying host cannot match the allow-list (42501), and a `user:pass@` URL split at the userinfo colon into a rejected port — but by accident, with the wrong diagnostic, and with the userinfo bytes carried in the handle's URL. This adds the siblings' guard to `os_write_handle`, before the colon parse.

## Verification (TDD)

New suite `objstore_userinfo` (7 checks, **no server needed**: both the guard's 22023 and the allow-list's 42501 fire pre-network). RED on unfixed was instructive: `export_parquet` with `user:pass@` read 22023 *even unfixed* — via the **port** check (the parse splits at the userinfo colon), the wrong guard — so the suite pins SQLSTATE **and** the guard's own message; the message arm is load-bearing. `export_arrow` with bare `user@` showed the true divergence (42501).

- GREEN: 7/7 on PG17 assert, zero warnings.
- Removal proof: severing only the `os_write_handle` guard (the second of the three occurrences) reddens exactly the two write arms — 42501 returns on `user@`, and the accidental 22023 loses its userinfo message — while the read arms stay green.
- A clean-URL arm proves the guard did not swallow the allow-list's own 42501.
- Regression green: `objstore_crlf`, `objstore_allowlist`, `objstore_sink_write`, `harness_selftest` (110), `docs_style`.

First of the #705–#708 batch (claimed on the issues); #707 next.

Closes #706.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,14 @@ pinned at `1.0-dev` or `1.0-alpha`, each true until the next version shipped.
 
 ### Fixed
 
+- The object-store write path (export sink, delete, list) now refuses a URL
+  with userinfo (`user@host`) with the same error the read path always gave.
+  Before, the write parse admitted the URL and failed closed downstream by
+  accident: the allow-list cannot match a host with an embedded `@`, and a
+  `user:pass@` URL split at the wrong colon into a rejected port. A new
+  suite, `objstore_userinfo`, pins all three entry points on the SQLSTATE
+  and the guard's own message. (#706)
+
 - A merge join over index-fetched columnar rows no longer aborts with `pfree is
   not supported by the bump memory allocator` on PostgreSQL 17 and later. A
   columnar index scan returns a deferred slot, and when a sort materialises it
--- a/objstore/columnar_objstore_module.c
+++ b/objstore/columnar_objstore_module.c
@@ -1698,6 +1698,21 @@ os_write_handle(const char *url, const PgColumnarObjStoreConfig *cfg)
 			ereport(ERROR,
 					(errcode(ERRCODE_INVALID_PARAMETER_VALUE),
 					 errmsg("columnar: \"%s\" has no object path", url)));
+
+		/*
+		 * Same userinfo refusal as os_open and http_request (#706). Without
+		 * it a `user@host` write target parsed with the '@' embedded in the
+		 * host (or, with `user:pass@`, split at the userinfo colon into a
+		 * garbage port). Both shapes still failed closed downstream -- the
+		 * allow-list cannot match an '@'-carrying host, and port 0 is
+		 * refused -- but by accident and with the wrong diagnostic, and the
+		 * userinfo bytes rode along in the handle's url.
+		 */
+		if (memchr(authority, '@', slash - authority) != NULL)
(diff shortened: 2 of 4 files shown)


OffgridwithJD (approved):
APPROVE — reviewed as ChronicallyJD (cross-review; I do not merge). Small, correct, well-tested sibling-divergence fix.

## Why it's right

The new guard in `os_write_handle` is byte-identical to the two read-path siblings — same `memchr(authority, '@', slash - authority)`, same `ERRCODE_INVALID_PARAMETER_VALUE`, same message as `os_open` (:1527) and `http_request` (:2647). Coverage is now complete and I checked there's no fourth hole: the file has exactly three authority parses and now three `@` guards, and at each site the guard precedes the `:` colon parse (1527<1534, 1711<1718, 2647<2654), so userinfo can't reach `h->url` or be mis-split into a port.

## Verified

- Built on PG17 assert, 0 warnings. `objstore_userinfo` 7/7 green; regressions `objstore_crlf` (5), `objstore_allowlist` (16), `objstore_sink_write` (30), `harness_selftest` (110), `docs_style` (9) all green.
- Removal proof reproduced: severing only the `os_write_handle` guard (leaving the other two intact) reddens the suite — the userinfo-message arm fails (`got [0] want [1]`). That's the load-bearing point, because the write path already fails closed without the guard (the allow-list rejects an `@`-host, and `user:pass@` mis-splits into a refused port), so a plain does-it-error test would pass even unfixed. Pinning the guard's own message is what makes the arm non-vacuous — good test design, and exactly the trap this kind of "already fails closed by accident" fix invites.

No changes requested. This is defense-in-depth (correct diagnostic + no userinfo bytes carried in the handle URL) rather than a reachable exploit, which matches the audit's M-5 rating.

Issue #568, opened 2026-08-10
vacuum, vacuum_sorted and cluster take AccessExclusiveLock before checking privileges, so an unprivileged role can block readers on any columnar table


jdatcmd:
## Summary

`pgcolumnar.vacuum`, `pgcolumnar.vacuum_sorted` and `pgcolumnar.cluster` call
`table_open(relid, AccessExclusiveLock)` before any privilege check. The
ownership check runs afterwards, on the already-opened relation. A role that will
be refused therefore joins the lock queue first and is told no second.

PostgreSQL's lock queue is FIFO, so a pending `AccessExclusiveLock` request
blocks every lock request that arrives behind it. An unprivileged role with
`USAGE` on the `pgcolumnar` schema can use this to stall all access to a columnar
table it does not own, for as long as it is willing to keep calling.

`EXECUTE` on these functions is `PUBLIC`, because the install script contains no
`GRANT` or `REVOKE` at all. See #559 for that root cause.

## Measured

PostgreSQL 17 assert build, clean `git archive main` at 744b7ca, built and
installed in isolation. `t_lock` is `NOSUPERUSER`, holds `USAGE` on the schema
and nothing on `victim`, which has had `REVOKE ALL` applied.

Session A holds a plain `AccessShareLock`:

```sql
BEGIN; SELECT count(*) FROM victim; SELECT pg_sleep(12);
```

Session B, as `t_lock`, calls `pgcolumnar.vacuum('victim')`. While A still holds
its lock, from a third session:

```
AccessExclusiveLock granted=false by=t_lock on=victim
```

`t_lock` is in the queue, ungranted, on a table it does not own. The refusal
arrives only once A releases and the lock is actually granted:

```
t_lock result after A released: ERROR:  must be owner of table victim
```

So the check works. It just runs too late to matter for availability. Every
reader arriving between B's request and B's refusal waits behind an
`AccessExclusiveLock` that was never going to be permitted.

## Scope, counted rather than described

Buckets over every C entry point on main that opens a relation, symbols resolved
from the `AS 'MODULE_PATHNAME','<symbol>'` clause:

```
entry points that open a relation: 15
LOCK BEFORE CHECK: 4   NO GATE: 10   check first: 1   sum = 15
```

```
LOCK BEFORE CHECK  AccessExclusiveLock  pgcolumnar_cluster           columnar_vacuum.c
[...]

PR #724, opened 2026-08-24
sec: make the debug metadata mutators owner-only (#707)

### LINKED ISSUE #707: sec: debug metadata mutators lack owner check (advance_reserved_offset, set_metapage_version)

jdatcmd:
## What

`pgcolumnar_debug_advance_reserved_offset` and `pgcolumnar_debug_set_metapage_version` mutate storage metadata with no gate — one burns reserved file space, the other overwrites the metapage format version and **bricks every later read of the table, including its own DROP** (the drop path re-reads the metapage; the suite found that empirically). They ship unbound, but `native_gap.sh` and `native_format.sh` already bind them with one `CREATE FUNCTION`, which is all any operator needs to do. Both now check `PgColumnarRequireTableOwnerByOid` before `table_open` — the #568 refuse-before-the-lock idiom the other maintenance verbs use.

## Verification (TDD)

New suite `debug_hook_privilege` (10 checks): binds the hooks the way the suites do (in `public`, so PUBLIC's default EXECUTE leaves the owner gate as the only gate under test), probes as a **non-owner role running its own statements**, and asserts **SQLSTATE 42501 plus the `must be owner` message** — the message arm exists because an EXECUTE denial is also 42501 and must not impersonate the gate. Mirror arms prove the gate is ownership, not superuser-ness: the peon may still mutate (and brick) its own table.

- Per-check removal proofs: severing only the advance gate reddens exactly its two arms; severing only the setver gate reddens its two arms **plus the table-intact arm** — the ungated call really does destroy the other owner's table (`count(*)` errors on the bricked version).
- The issue's discipline note followed: gated on `SQLSTATE 42501` (`aclcheck_error`), never error text alone.
- Regression green: `native_gap`, `native_format` (the superuser binder suites, unaffected), `vacuum_lock_privilege`, `harness_selftest` (110), `docs_style`.

Second of the #705–#708 batch. #705 (Gorilla word-at-a-time) next.

Closes #707.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,15 @@ pinned at `1.0-dev` or `1.0-alpha`, each true until the next version shipped.
 
 ### Fixed
 
+- The debug metadata mutators (`pgcolumnar_debug_advance_reserved_offset`,
+  `pgcolumnar_debug_set_metapage_version`) are now owner-only, checked
+  before the table is opened like the other maintenance verbs. They ship
+  unbound, but a binding is one `CREATE FUNCTION` away, and one of them can
+  overwrite a table's metapage version and brick every later read. A new
+  suite, `debug_hook_privilege`, binds them the way the test suites do and
+  proves the gate as a non-owner role, by SQLSTATE and the owner message.
+  (#707)
+
 - A merge join over index-fetched columnar rows no longer aborts with `pfree is
   not supported by the bump memory allocator` on PostgreSQL 17 and later. A
   columnar index scan returns a deferred slot, and when a sort materialises it
--- a/src/columnar_vacuum.c
+++ b/src/columnar_vacuum.c
@@ -2105,6 +2105,14 @@ pgcolumnar_debug_advance_reserved_offset(PG_FUNCTION_ARGS)
 	int32		npages = PG_GETARG_INT32(1);
 	Relation	rel;
 
+	/*
+	 * Owner-only, checked before table_open like the other maintenance verbs
+	 * (#568, #707). This function ships unbound, but a binding is one CREATE
+	 * FUNCTION away (the test suites do exactly that), and a storage-metadata
+	 * mutator with no gate is a footgun waiting for that binding.
+	 */
+	PgColumnarRequireTableOwnerByOid(relid);
+
 	if (npages < 0)
 		ereport(ERROR,
 				(errcode(ERRCODE_INVALID_PARAMETER_VALUE),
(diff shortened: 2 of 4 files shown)


OffgridwithJD (approved):
APPROVE — reviewed as ChronicallyJD (cross-review; I do not merge). Clean, correct, and the test is a model of how to prove a privilege gate.

## Why it's right

Both — and only — `pgcolumnar_debug_*` mutators now call `PgColumnarRequireTableOwnerByOid(relid)` **before** `table_open` (advance at :2114 before the open at :2121; setver at :2153 before :2155), matching the `#568` refuse-before-the-lock idiom the other maintenance verbs use. Refusing before the open is strictly safer here (a non-owner never takes even the RowExclusiveLock), and the ByOid variant is the right one for that. No third debug mutator exists, so coverage is complete.

## Verified on PG17 assert (0 warnings)

- `debug_hook_privilege` 10/10; regressions `native_gap` (8), `native_format` (16, the superuser binder suites, unaffected), `vacuum_lock_privilege` (9), `harness_selftest` (110), `docs_style` (9) all green.
- Both removal proofs reproduced, per-gate exact: severing only the advance gate reddens exactly its two arms; severing only the setver gate reddens its two arms **plus** the table-intact arm — the ungated call really does brick the other owner's table (`count(*)` on the bricked version errors). Each gate is independently load-bearing.

## What makes the test non-vacuous (the part that matters)

This is exactly the shape a privilege test needs and often lacks: the deny arm probes as a **non-owner running its own statements** (not a definer context that would false-pass), it pins the **`must be owner` message** and not bare 42501 (an EXECUTE denial is also 42501 and would otherwise impersonate the gate), and the **positive-control mirror arms** prove the gate is ownership rather than superuser-ness by showing the peon can still brick its *own* table. The setver impact arm goes one better and demonstrates the actual damage the gate prevents. No changes requested.

Issue #289, opened 2026-07-31
Vectorized decompression + aggregation for full-scan aggregates (~4× behind TimescaleDB)


OffgridwithJD:
### Motivation
In the 100M-row TSBS-cpu benchmark (PG18), even on full-scan aggregates where pgColumnar
reads *less* data than TimescaleDB (2.67 GB vs 7.79 GB on disk), TimescaleDB is **~4×
faster**:

| query | pgColumnar | TimescaleDB | heap |
|---|---:|---:|---:|
| q4 — avg/hour/host, 12 h | 11,247 ms | **2,648 ms** | 13,009 ms |
| q5 — avg of 10 metrics/hour/host | 22,131 ms | **5,747 ms** | 24,518 ms |
| q6 — full scan + filter | 32,470 ms | **6,131 ms** | 10,304 ms |
| q8 — top-20 host-hours | 21,608 ms | **4,637 ms** | 15,315 ms |

(serial warm median.) pgColumnar beats heap on the wide aggregates (q4/q5) — the columnar
projection pays off — but is well behind TimescaleDB's vectorized execution.

### Root cause
TimescaleDB decompresses its columnstore in batches and aggregates over those batches with
vectorized operators. pgColumnar's scan decodes each vector but then feeds rows into
PostgreSQL's row-wise executor for aggregation, paying per-row `ExecAgg` overhead over
100M rows. Correct, but the per-row path is the bottleneck once the data is already in
column-major form.

### Proposed direction (cleanroom, larger effort)
Push aggregation down into the columnar scan and operate on decoded column batches:

- decode into column-major batches (already available inside `ColumnarScan`) and run
  **batch aggregate kernels** (`sum`/`avg`/`min`/`max`/`count`, filter predicates) over a
  vector at a time instead of row-at-a-time;
- evaluate the `WHERE` predicate on the batch to produce a selection vector before
  aggregating (helps q6);
- structure kernels to be SIMD-friendly (contiguous typed arrays, no per-row function-call
  dispatch);
- integrate with grouped aggregation (`GROUP BY date_trunc(...), hostname`) via a
  vectorized hash-aggregate, or at minimum vectorize the per-group accumulation.

Reference the public literature and systems concepts only (MonetDB/X100 vectorized
execution, DuckDB vectorized operators, the Gorilla/columnar aggregation papers); the
implementation must be ours alone.

### Impact
[...]

PR #725, opened 2026-08-24
opt: fold the grouped vectorized aggregate column at a time (#708)

### LINKED ISSUE #708: opt: grouped aggregation is row-at-a-time; ungrouped has batch_fold, grouped doesn't

jdatcmd:
Closes #708.

The grouped vectorized aggregate was fully row-at-a-time. This gives it the
column-at-a-time fold the ungrouped node has had since #289, and measures the
result with an instrument that is fair on this host.

## What it does

`pgcolumnar_groupagg_batch_fold` walks row groups, gathers each needed column's
packed values directly, applies the delete mask, the skipped-vector map and the
scan keys inline, then probes the group hash table and folds. The row path paid,
for every row scanned: `PgColumnarReadNextRow`, `ResetExprContext`, the
`bms_next_member` slot-staging loop, `ExecStoreVirtualTuple`, `ExecQual`, and one
`ExecEvalExprSwitchContext` per group key. The fold pays none of them. What was
left of the fmgr traffic, the hash and the probe equality, goes inline for key
types whose equality **is** bitwise equality of the Datum.

The shape gates are now single-sourced in `pgcolumnar_batch_gates_ok`, shared
with the ungrouped fold, rather than copied. Every one of those gates was added
to fix a wrong answer or a crash, and a second fold arriving with a stale copy of
the list is precisely how #715 comes back. The grouped node adds one gate of its
own: every group key must be a plain `Var` it can read out of a gathered column.

## Measurement

Instructions retired by the backend (`perf stat -e instructions -p`), which host
contention does not move. 4,000,000 rows, `k int` with 8 distinct groups, `v int`,
PG17, 5 repetitions, warm scan excluded, the plan asserted on **both** arms
(`Columnar Batch Fold: yes` on the new arm; the baseline build has no such line
at all). Same cluster and same fixture for both arms, with only the `.so` swapped.

| query | `main` (row path) | #708 (fold) | ratio | per row |
| --- | ---: | ---: | ---: | --- |
| `SELECT k, count(*), sum(v) ... GROUP BY k` | 15,850,896,522 | 7,352,675,551 | **2.16x** | 793 to 368 |
| the same with `WHERE v < 3000000` | 13,899,374,785 | 6,292,504,808 | **2.21x** | 695 to 315 |

Run again with the arm order reversed, the same four figures came back within
0.68% (base plain 15,853,308,490; base filtered 13,899,612,015; fold plain
7,402,825,977; fold filtered 6,293,789,323), so the ordering effect that makes
wall clock unusable here is absent. One warm repetition moved 227ms to 116ms,
which corroborates the direction independently.

### Three instrument defects, each of which produced a confident wrong number

Worth recording, because two of them looked like results.

1. **This is a hybrid CPU, so `perf` prints two `instructions` lines**, one per
   PMU, each scaled by the fraction of the window that PMU was counting.
   `head -1` took the E-core line alone (6.5e9 of a 14.1e9 backend). **Summing
   them is worse**: one run showed `cpu_core` 15.85e9 at 99.70% beside
   `cpu_atom` 11.07e9 at **0.30%**, an eleven-billion "count" extrapolated 333x
   from three-tenths of one percent, which reported 26.9e9 for an arm two other
   runs put at 15.85e9. Read the PMU that held the window; refuse the run if
   none did.
2. A non-interactive shell starts async children with `SIGINT` **ignored**, and
   the disposition is inherited, so `kill -INT $perfpid` was silently a no-op.
3. `perf stat ... -- sleep N` runs the sleep as a child that inherits the
   command FIFO. On interrupt it was orphaned, kept the write end open, `psql`
   never saw EOF, and the measurement hung in `wait()`. `--timeout` has neither
   problem.

## Verification

`test/native_groupagg_batch.sh`, 62 checks, built RED before the code existed.
It separates the two claims, because an optimisation that stops optimising still
returns correct answers:

- **Answers.** A heap mirror for every shape, plus a toggle-differential against
  the same rows with the path off, which is byte-exact.
- **Work.** `Columnar Batch Fold` on the grouped node under `EXPLAIN ANALYZE`,
  which is the only assertion that can go red when the fold stops folding.

Eligibility is asserted in both directions: a by-reference key, an expression
key, `min`/`max`, and a `<>` filter each have an arm proving the fold **declines**
and an oracle proving the row path still answers correctly.

### Removal proofs

| mutation | result |
| --- | --- |
| delete the `pgcolumnar_groupagg_batch_fold` call | **10 work-done arms red, all 52 correctness arms green** |
| add `FLOAT8OID` to the bitwise-equality key list | **red**: `-0.0` and `0.0` become 3 groups where the heap has 2 |
| drop `PgColumnarQualsExactlyKeyed` from the gates | **red**: the `<>` arm folds and returns a wrong answer |

The float8 proof needed the fixture fixed first: written `(-0.0, 2)` in a
`VALUES` list the literal goes through `numeric`, which has no signed zero, and
reached the column as `+0.0`. The arm passed the mutation while proving nothing.
It now asserts `float8send` is `8000000000000000` before it compares anything.

Two arms were also weaker than they read. Wrapping the grouped query in an outer
`count(*)` changes the plan so the fold never runs underneath it, and the
parallel value arms compared md5s that differed only by `psql`'s `SET` command
tags, which reads exactly like a parallel double-count and was not one.

### Suites

`native_groupagg_batch` 62/62, and green on `native_groupagg`,
`native_groupagg_wide_cost`, `ungrouped_vector_agg`, `batch_fold_explain`,
`parallel_vector_agg`, `native_agg`, `native_agg_addcolumn`, `native_agg_deletes`,
`native_fold_deferral`, `native_fold_skipguard`, `native_vecdecode`,
`native_saop_pushdown`, `rewrite_group_scan`, `native_fetch_group_memo`,
`native_late_materialization`, `column_projection`.

Builds clean with zero warnings on PostgreSQL 15, 16, 17, 18 and 19.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -63,6 +63,28 @@ pinned at `1.0-dev` or `1.0-alpha`, each true until the next version shipped.
 
 ### Changed
 
+- The grouped vectorized aggregate (`pgcolumnar.enable_group_vectorization`)
+  now folds column at a time instead of row at a time. Where every group key is
+  a plain column, every aggregate is one the fold accumulates (`count`, `sum`
+  and `avg` over integer and float columns), and the whole `WHERE` is expressed
+  exactly by scan keys, the node walks row groups and reads each column's
+  packed values directly. The row path paid, for every row scanned,
+  `PgColumnarReadNextRow`, an expression context reset, staging the row into a
+  virtual slot, `ExecQual`, and one expression evaluation per group key; the
+  fold pays none of those, and the group hash and the probe equality are inline
+  for integer, `oid`, `date`, `time` and `timestamp` keys. Float keys keep the
+  type's own hash and equality functions, because `-0.0` and `0.0` are one
+  group with two bit patterns. Measured on 4,000,000 rows with 8 groups, 5
+  repetitions, counting instructions retired by the backend: 15,850,896,522
+  before and 7,352,675,551 after without a filter (2.16x), and 13,899,374,785
+  before and 6,292,504,808 after with one (2.21x). `EXPLAIN` reports
+  `Columnar Batch Fold` on the grouped node, as it already did on the ungrouped
+  one, so a plan states whether the fold ran. Shapes the fold declines, among
+  them a by-reference key such as `text`, an expression key, `min` and `max`,
+  and a filter no scan key expresses exactly, run the row path and return the
+  same results. A new suite, `native_groupagg_batch`, proves the results
+  against a heap mirror and proves the work separately from them. (#708)
+
 - An index or bitmap fetch no longer reads the whole row-group list out of
   the catalog for every row. The list is memoized per command and snapshot,
   with a refresh on any miss, so a group flushed earlier in the same
--- a/docs/features.md
+++ b/docs/features.md
@@ -56,7 +56,11 @@ settings see the [configuration reference](configuration.md); for constraints se
   data. Set `pgcolumnar.enable_vectorization` to `off` to force an ordinary
   aggregate over the scan instead. An opt-in path vectorizes `GROUP BY` too. Set
   `pgcolumnar.enable_group_vectorization` to `on` to group and aggregate in one
-  pass over the reader. It is off by default.
+  pass over the reader. It is off by default. Where the query groups by plain
+  columns, aggregates with `count`, `sum` or `avg`, and filters only with
+  comparisons the scan keys express, that pass reads each column's values
+  directly instead of one row at a time. `EXPLAIN` reports `Columnar Batch Fold`
+  on the node, and other shapes read row at a time and return the same results.
 - Column statistics. `ANALYZE` samples rows from across the row groups. It stores
   the null fraction, the distinct counts, the most-common values, the histograms
   and the correlation. The planner then estimates predicate selectivity from that
--- a/docs/features.md
+++ b/docs/features.md
@@ -56,11 +56,12 @@ settings see the [configuration reference](configuration.md); for constraints se
   data. Set `pgcolumnar.enable_vectorization` to `off` to force an ordinary
   aggregate over the scan instead. An opt-in path vectorizes `GROUP BY` too. Set
   `pgcolumnar.enable_group_vectorization` to `on` to group and aggregate in one
-  pass over the reader. It is off by default. Where the query groups by plain
-  columns, aggregates with `count`, `sum` or `avg`, and filters only with
-  comparisons the scan keys express, that pass reads each column's values
-  directly instead of one row at a time. `EXPLAIN` reports `Columnar Batch Fold`
-  on the node, and other shapes read row at a time and return the same results.
+  pass over the reader. It is off by default. That pass applies where the
+  query groups by plain columns, aggregates with `count`, `sum` or `avg`, and
+  filters only with comparisons the scan keys express. There it reads each
+  column's values directly instead of one row at a time. `EXPLAIN` reports
+  `Columnar Batch Fold` on the node. Other shapes read row at a time and return
+  the same results.
 - Column statistics. `ANALYZE` samples rows from across the row groups. It stores
   the null fraction, the distinct counts, the most-common values, the histograms
   and the correlation. The planner then estimates predicate selectivity from that
--- a/src/columnar_vector.c
+++ b/src/columnar_vector.c
@@ -4244,8 +4244,23 @@ PgColumnarBeginGroupAggScan(CustomScanState *node, EState *estate, int eflags)
  *		The HASH is private to this hash table and never leaves it, so any hash
  *		consistent with the equality relation is correct. The EQUALITY is the
  *		part that has to match how core groups, and that is what this list is
- *		asserting -- one entry at a time, for types whose equality function is a
- *		plain integer compare of the stored width.
+ *		asserting -- one entry at a time.
+ *
+ *		STATE THE INVARIANT PRECISELY, because the next person to extend this
+ *		list will read this sentence instead of re-deriving it. The requirement
+ *		is NOT "the type's equality is an integer compare of the stored width".
+ *		It is that two values this type calls EQUAL must produce identical Datum
+ *		BITS *after fetch_att* -- which is a property of the GATHER as much as of
+ *		the equality operator.
+ *
+ *		oid is the entry that shows the difference. fetch_att SIGN-EXTENDS a
+ *		len-4 column, so an oid above 2^31 arrives with the high half set, which
+ *		the narrower rule does not describe at all. It is still correct here,
+ *		and only because the sign extension is CONSISTENT: the same oid always
+ *		yields the same bits, so equal values stay equal and unequal values stay
+ *		unequal. A type for which that did not hold would satisfy the narrower
+ *		rule and still group wrongly. (Found in review, by probing oid values
+ *		above 2^31 against a heap oracle rather than by reading the list.)
  *
  *		float4 and float8 are deliberately ABSENT even though they are passed by
  *		value and the fold reads them happily: -0.0 and 0.0 are ONE group and
--- a/src/columnar_vector.c
+++ b/src/columnar_vector.c
@@ -650,6 +650,19 @@ typedef struct PgColumnarGroupKey
 	bool		byval;
 	FmgrInfo	hashFn;			/* type hash function */
 	FmgrInfo	eqFn;			/* type equality operator function */
+
+	/*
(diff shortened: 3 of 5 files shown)


jdatcmd:
## PG18 + PG19 gate: green

The per-PR cadence run, 13 suites each on PostgreSQL 18.4 and 19beta2. 26 of 26
passed, with no skips and no empty verdicts.

```
native_groupagg_batch  native_groupagg          native_groupagg_wide_cost
ungrouped_vector_agg   batch_fold_explain       parallel_vector_agg
native_agg             native_agg_addcolumn     native_agg_deletes
native_fold_deferral   native_fold_skipguard    native_vecdecode
native_saop_pushdown
```

PG17 was green earlier on the same 13 plus `rewrite_group_scan`,
`native_fetch_group_memo`, `native_late_materialization` and
`column_projection`. The extension builds with zero warnings on 15, 16, 17, 18
and 19.

OffgridwithJD (changes requested):
Reviewed as ChronicallyJD (cross-review; I do not merge). The correctness core is sound — I went hardest at the two things you flagged, and both hold. One blocking item: the PR's own docs change reds `docs_style`.

## #2 — the `simple_key` bitwise-equality whitelist (the one you most wanted broken): holds

I tried to break it empirically, per type, on PG17 with the fold confirmed engaged (`Columnar Batch Fold: yes` on every arm). For each whitelisted type I built a columnar table and a heap twin over adversarial key values and compared the *full grouped result* (md5 of the ordered `k | count | sum` rows), so a false split or false merge shows as a hash mismatch:

- `int2` (−32768, −1, 0, 32767, NULL, dups), `int4` (INT32_MIN, −1, …), `int8` (INT64_MIN/MAX): the negatives are the sign-extension stress, and they group correctly. `fetch_att` at columnar_vector.c:4718 sign-extends len-2/4 deterministically, so equal values give equal Datum bits — bitwise equality is sound.
- `oid` with values > 2³¹ (2147483648, 4294967295): `fetch_att` sign-extends the unsigned oid, but *consistently*, so grouping is still correct — matched heap.
- `date` (pre/post 2000-01-01, i.e. negative and positive int4), `time` (0 / 23:59:59.999999), `timestamp`/`timestamptz` including `infinity`/`-infinity`: all matched heap.
- NULL keys in every type: grouped as one group, matched heap (the fold tracks `keynulls` separately, not the garbage Datum).

So the list is right, and the float exclusion for −0.0 is the correct line. No wrong-grouping found.

## #1 — the `pgcolumnar_batch_gates_ok` extraction: no ungrouped regression

Verified the extraction is a pure refactor for the already-shipped ungrouped path: same five gates, same order, same short-circuits, and each parameter is bound to the identical `state->` field the inline code read (`specs/naggs/quals/scanrelid/projected/tupdesc`). The ungrouped `Begin` (where `state->projected` is built, the #423 gate's input) is byte-for-byte unchanged, so the by-value gate still sees the same set built at the same point before the EXPLAIN return. The grouped caller only *tightens* (adds the every-key-a-plain-Var requirement). No previously-rejected shape is now accepted on the ungrouped path.

## Suites

`native_groupagg_batch` 62, `native_groupagg` 72, `ungrouped_vector_agg` 49, `parallel_vector_agg` 29, `native_agg` 25, `batch_fold_explain` 7, `harness_selftest` 110 — all green on PG17 assert, 0 warnings. I leaned on the 62-check suite for the cursor discipline (#3), the parallel fall-back error arm (#4), and the EXPLAIN-only Begin (#5); those arms pass and I spot-confirmed the fold engages under EXPLAIN ANALYZE.

## The change I'm requesting — `docs_style` is red on this PR's own sentence

`test/docs_style.sh` fails: the new sentence in `docs/features.md` is **36 words**, over the STE ≤25-word limit (main passes; this sentence is PR-introduced):

> Where the query groups by plain columns, aggregates with `count`, `sum` or `avg`, and filters only with comparisons the scan keys express, that pass reads each column's values directly instead of one row at a time.

docs_style is in the gate, so the PR can't merge green as-is. Split it, e.g.:

> That pass applies where the query groups by plain columns, aggregates with `count`, `sum` or `avg`, and filters only with comparisons the scan keys express. There it reads each column's values directly instead of one row at a time.

Both halves are ≤25 words. Re-run `docs_style` after (9/9) and I'll clear the request — nothing else is outstanding.

OffgridwithJD (approved):
APPROVE — my REQUEST_CHANGES is resolved (re-reviewed at `fcf6a27`), and I went and closed the #3 question empirically rather than leaving it on the suite.

## docs_style — fixed

9/9 on `fcf6a27`. The split reads cleanly and both halves are under the limit.

## #3 cursor discipline — I exercised it, and it's correct

You were right that neither the suite nor my first fixture actually skipped a vector (group-level pruning removed whole chunk groups but `Columnar Vectors Skipped` stayed 0). So I built the needle-in-haystack shape `native_decode_gating` uses — big stripe, `chunk_group_row_limit => 1024`, five payload columns, a `sel` column whose match lives in a handful of vectors — and ran a folded `GROUP BY k ... WHERE sel = 1`:

```
Columnar Batch Fold: yes
Columnar Vectors Skipped: 287
Columnar Vectors Decoded: 6
grouped result == heap oracle  (6 groups, exact counts and sums)
```

287 vectors skipped, 6 decoded, and the grouped aggregate matches heap to the byte. A `cpresent` misalignment across a skipped vector would have mis-attributed rows and shown as wrong per-group counts/sums; it doesn't. So the copied cursor discipline is faithful — exercised and correct, not merely present.

**Recommendation (non-blocking): add the arm you offered.** An arm that asserts `Columnar Vectors Skipped > 0` on a folded grouped query AND checks the result against a heap mirror is worth landing, because the premise is the fragile part — the current arms pass at 0 skips, so a future change that stops building the grouped skipvec would leave this silent path green-but-unexercised again. I've proven correctness now, so it's a regression lock, not a merge gate. Your call on this-PR vs follow-up.

## oid comment — yes, please tighten it

Agreed, and I think it matters more than a nit. The real invariant your whitelist depends on is not "the type's equality is an integer compare of the stored width" — the oid case breaks that reading, since `fetch_att` sign-extends an oid above 2^31 and grouping stays correct only because the gather does so *consistently*. The next person extends the list by reading that sentence. State the actual requirement: **equal values must produce equal Datum bits after `fetch_att` — a property of the gather, not only of the equality operator** (which is why a fixed-width byval type with a canonical representation qualifies and float does not). Comment-only; push it on top and I won't re-review.

## The rest (unchanged from my first pass)

Whitelist proven correct across all 8 types vs heap with the fold engaged (incl. the sign-extension and NULL-key breakers); the `pgcolumnar_batch_gates_ok` extraction is a pure refactor with no ungrouped regression; 62-check suite + neighbours green, 0 warnings. Good to merge once your CI confirms both suites legs green on `fcf6a27`.

OffgridwithJD:
Re-reviewed `d22e60f` (approval stands — the delta since `fcf6a27` is comment + test only, verified: `git diff fcf6a27..d22e60f -- src/*.c src/*.h` has no non-comment added lines).

Independently reproduced your removal proof, because closing #3 with an arm that catches the misalignment is exactly the bar this repo holds: built `d22e60f` (native_groupagg_batch 69/69, 0 warnings), then moved `cpresent[col]++` inside the `!vecSkipped` branch — the silent cursor misalignment — rebuilt, and **exactly the two new arms went red** (`skipped vectors: answers match the heap mirror`, `... and their sums match`), the other 67 green. That matches your result, and it's the complete closure: before the arm the misalignment was invisible to the whole suite; now it isn't. Putting it in this PR rather than deferring was the right call.

The oid comment now states the real invariant (equal values → equal Datum bits after `fetch_att`, a gather property, with oid as the case that shows sign-extension must be *consistent*). Good.

Nothing outstanding. Merge when CI on `d22e60f` is green.

Issue #727, opened 2026-08-24
opt: the vectorized aggregate rescan path grows ~655 bytes of query memory per rescan, independent of scan keys


jdatcmd:
## Finding

Found while fixing #717, and deliberately **not** folded into it, because it is a
different allocation with a different cause.

#717 was scan keys being rebuilt per rescan into `es_query_cxt`. With that
fixed, the ungrouped vectorized aggregate still grows query memory linearly in
the rescan count, and the growth is **the same with and without an IN-list**, so
it is not key-related at all.

## Measurement

`ExecutorState` peak, sampled with `pg_log_backend_memory_contexts()` from a
second session, LATERAL correlated aggregate over a 50k-row columnar table,
`SMALL`=200 and `LARGE`=3200 rescans:

| arm | 200 rescans | 3200 rescans | slope |
| --- | ---: | ---: | ---: |
| with a 20000-element `IN` list | 655,424 | 2,621,504 | **655 B/rescan** |
| no `IN` list at all | 131,072 | 2,097,152 | **655 B/rescan** |

Identical slopes. The IN-list moves the intercept (the keys themselves are live
during a scan) but not the growth. Confirmed against the pre-#717 build too,
where the no-IN-list arm reads byte-identical (131,072 -> 2,097,152), so this
predates #717's fix and is untouched by it.

Roughly 2 MB over 3200 rescans. Low severity, bounded by the rescan count and
freed at `ExecutorEnd`, exactly like #717 was.

## Where to look

Per-execution allocations in `pgcolumnar_native_scan_agg` and the grouped
equivalent that are not freed and not in a reset context. Candidates, unverified:

- the `PgColumnarReadState` and whatever `PgColumnarBeginRead` allocates that
  `PgColumnarEndRead` does not release
- the `values`/`nulls` arrays palloc'd per call at the top of the scan-fold
  functions
- the `projected` `Bitmapset` where it is rebuilt per execution rather than
  reused from Begin

655 bytes is small enough that it is probably one or two of these rather than a
single large object; the way to find it is a memory-context dump with the
contexts named, not inspection.

## Discipline

`test/vector_agg_rescan_memory.sh` already has the instrument and the control
arm; this needs its own assertion on the **no-IN-list** slope, which that suite
currently uses only as a control and does not gate. Removal proof: the slope
returns.

Note the differential in that suite is deliberately immune to this leak, so
fixing this must not change its result. If it does, one of the two measurements
is wrong.

[...]

Issue #726, opened 2026-08-24
explain: plain EXPLAIN reports `Columnar Pushed-Down Filters: 0` for the ungrouped vector aggregate whatever the real count


jdatcmd:
## Finding

Found while chasing an anomaly during #717: a plain `EXPLAIN` of the ungrouped
vectorized aggregate reports `Columnar Pushed-Down Filters: 0` even for a query
whose quals genuinely build scan keys.

`PgColumnarBeginAggScan` returns early for `EXEC_FLAG_EXPLAIN_ONLY`
(`src/columnar_vector.c:2154`), and `state->nscankeys` is assigned after it
(`:2204`). So on a plain EXPLAIN the field is still its `palloc0` zero when
`PgColumnarExplainPushedDown` prints it (`:3889`).

The **grouped** node does it correctly: its `nscankeys` assignment (`:4157`) is
before its EXPLAIN-only return (`:4224`). So the two vectorized nodes disagree
about the same label, which is exactly the failure `CONTEXT.md` warns about when
one label has to mean one thing on every node (#493).

Concretely, a LATERAL aggregate over a columnar table with a 20000-element
`IN (...)` and a correlated residual reports:

```
Custom Scan (PgColumnarScan)
  Columnar Vectorized Aggregates: 1
  Columnar Pushed-Down Filters: 0        <-- keys ARE built at execution
```

The keys are real: they are what #717 measured being rebuilt per rescan, and the
growth scales with the array size (9.8 MB at 2000 elements, 90.8 MB at 20000,
0.5 MB with no IN-list), so the build is certainly happening.

## Why it matters more than a cosmetic zero

This is the same class as #602: a plain-EXPLAIN line reporting a value that was
never computed. `Columnar Pushed-Down Filters` is a line people read to decide
whether pushdown is working, and a hard zero reads as "pushdown is not
happening" rather than "this number was not filled in". It is also the line a
pushdown measurement would pin its premise on, and a premise that is always zero
cannot fail for the right reason.

## Proposed fix

Move the `PgColumnarCountConvertibleQuals` / `PgColumnarCountScanKeys` pair
above the EXPLAIN-only return in `PgColumnarBeginAggScan`, matching the grouped
node. Both are catalog-and-plan only and do no I/O, which is the same argument
the scalar node already makes for computing its keys in EXPLAIN-only mode
(`columnar_customscan.c`, "they do no I/O, so EXPLAIN can report them").

## Discipline

[...]

PR #728, opened 2026-08-25
opt: give the vectorized aggregates a scan-key context reset per rescan (#717)

### LINKED ISSUE #717: opt: vector-agg paths rebuild scan keys per rescan into es_query_cxt and never free them

jdatcmd:
Closes #717.

The scalar custom scan builds its pushdown scan keys once at Begin and reuses
them. The vectorized aggregate paths rebuild them at execution, into
`es_query_cxt`, which lives until `ExecutorEnd`, so every rescan of a LATERAL or
parameterized aggregate sub-scan stacked another set. Since #704 the build of an
`IN (...)` key also detoasts the array constant and deconstructs it into `elems`
and `nulls`, which is what makes the growth material rather than academic.

**Five sites, not the three the issue names.** #708 added two when the grouped
node gained its own fold and its own eligibility check, so this cleans up after
my own change as well as the original three. They now build into a per-node
context reset at the top of each build. No two live key sets ever coexist in one
node, because the fold ends its read before returning false, so the row path
that follows is free to reset.

`PgColumnarCountScanKeys` is fixed too: it `pfree`d the `ScanKeyData` array but
left the deconstructed array behind. Every caller is a Begin, so it is bounded
rather than growing and it is **not** the O(rescans) defect. Same defect in
miniature, cheaper to fix than to explain.

## Measurement

`pg_log_backend_memory_contexts()` sampled from a second session against the
running backend, MAX of `ExecutorState` across samples. LATERAL correlated
aggregate over a 50k-row columnar table, 20000-element `IN` list, peak taken at
200 and 3200 rescans so the fixed setup cost cancels and a slope is left.

| arm | per-rescan growth |
| --- | ---: |
| with the `IN` list, fix reverted | **179,877 B** |
| without the list, fix reverted | 655 B |
| with the `IN` list, fixed | **655 B** |
| without the list, fixed | 655 B |

A prediction from the mechanism alone, one `Datum` and one `bool` per element,
is 180,000 B. It and the measurement agree to **0.07%**.

**The mechanism was verified, not inferred from that agreement.** An arithmetic
coincidence reads exactly like proof, so I checked it: the growth scales
linearly with the list size (9.8 MB at 2,000 elements, 90.8 MB at 20,000, both
over 500 rescans) and **disappears entirely without a list** (0.5 MB). That is
what identifies the array deconstruction specifically.

## Why the test asserts a difference and not a bound

An absolute byte limit has to be picked against one machine's block sizes, and
here it would also charge this defect for memory it does not cause: the
aggregate rescan path grows about 655 B per rescan **with or without** an
IN-list, at slopes that measure identical. That is a separate leak, filed as
#727 and untouched by this PR. Subtracting a no-IN-list control removes it
exactly and leaves only what adding the list costs.

The control arm reads byte-identical before and after the fix
(131,072 -> 2,097,152 either way), which is what says the fix moved only the
thing it claims to.

## Removal proof

Revert the context switches, keep the suite: `an IN-list does not add
per-rescan query memory (#717)` goes red at 179,222 B excess, with every
premise arm still green. Fixed, the excess is 0.

## Filed while working this, deliberately not folded in

- **#726** — plain `EXPLAIN` reports `Columnar Pushed-Down Filters: 0` for the
  ungrouped node whatever the real count, because `nscankeys` is assigned after
  the EXPLAIN-only return. The grouped node does it correctly, so one label
  means two things depending on the node. Found chasing an anomaly in my own
  measurement, and it is why the suite's scan-key premise reads under ANALYZE.
- **#727** — the ~655 B/rescan residual above.

## Suites

14 green on PG17: `vector_agg_rescan_memory`, `native_groupagg_batch`,
`native_groupagg`, `ungrouped_vector_agg`, `parallel_vector_agg`, `native_agg`,
`native_agg_addcolumn`, `native_agg_deletes`, `native_saop_pushdown`,
`native_param_pushdown`, `batch_fold_explain`, `native_fold_deferral`,
`docs_style`, `harness_selftest`.

`docs_style` is in that list on purpose: last PR I ran the suites my *code*
touched and not the one that owns the *files* I touched, and it cost a
round trip.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -63,6 +63,18 @@ pinned at `1.0-dev` or `1.0-alpha`, each true until the next version shipped.
 
 ### Changed
 
+- The vectorized aggregates no longer stack a fresh set of pushdown scan keys
+  in query memory on every rescan. The scalar scan builds its keys once at
+  Begin; these paths rebuild them at execution, and did so into a context that
+  lives until the query ends, so a LATERAL or parameterized aggregate sub-scan
+  grew with the rescan count. An `IN (...)` key made it material, because
+  building one also detoasts the array constant and deconstructs it. Measured on
+  a LATERAL aggregate with a 20000-element list: adding the list cost 179,222
+  bytes of query memory per rescan before, and nothing measurable after, against
+  a no-list control that is unchanged. The keys now live in a per-node context
+  that is reset at each build. A new suite, `vector_agg_rescan_memory`, measures
+  the two arms and asserts the difference. (#717)
+
 - The grouped vectorized aggregate (`pgcolumnar.enable_group_vectorization`)
   now folds column at a time instead of row at a time. Where every group key is
   a plain column, every aggregate is one the fold accumulates (`count`, `sum`
--- a/src/columnar_customscan.c
+++ b/src/columnar_customscan.c
@@ -2669,12 +2669,28 @@ PgColumnarReScanCustomScan(CustomScanState *node)
 int
 PgColumnarCountScanKeys(List *qual, Index scanrelid, TupleDesc tupdesc)
 {
-	ScanKey		keys;
+	MemoryContext tmp;
+	MemoryContext old;
 	int			nkeys = 0;
 
-	keys = PgColumnarBuildScanKeys(qual, scanrelid, tupdesc, &nkeys);
-	if (keys != NULL)
-		pfree(keys);
+	/*
+	 * This builds keys only to count them, and throws them away. pfree'ing the
+	 * ScanKeyData array was not enough: for a SAOP the build also detoasts the
+	 * array constant and deconstructs it into elems and nulls, and those were
+	 * left behind in whatever context the caller was in (#717). Build in a
+	 * scratch context and delete the whole thing.
+	 *
+	 * Bounded rather than growing -- every caller is a Begin, which runs once
+	 * per execution -- so this is not the O(rescans) growth #717 is about. It is
+	 * the same defect in miniature, and cheaper to fix than to explain.
+	 */
+	tmp = AllocSetContextCreate(CurrentMemoryContext,
+								"columnar scan key count",
+								ALLOCSET_SMALL_SIZES);
+	old = MemoryContextSwitchTo(tmp);
+	(void) PgColumnarBuildScanKeys(qual, scanrelid, tupdesc, &nkeys);
+	MemoryContextSwitchTo(old);
+	MemoryContextDelete(tmp);
 	return nkeys;
 }
 
--- a/src/columnar_vector.c
+++ b/src/columnar_vector.c
@@ -560,6 +560,18 @@ typedef struct PgColumnarAggScanState
 	int			nscankeys;		/* scan keys, the shared label's quantity (#493) */
 
 	MemoryContext resultContext;	/* holds min/max running values */
+
+	/*
+	 * Scan keys are rebuilt on every execution of this node, not once at Begin
+	 * like the scalar custom scan's (#717). They live here rather than in
+	 * es_query_cxt so that a rescan reclaims the previous build instead of
+	 * stacking it: for a SAOP the build also detoasts the array constant and
+	 * deconstructs it, so the growth is O(rescans x array size) and reached
+	 * 346 MB over 2000 rescans of a LATERAL aggregate with a 20000-element
+	 * IN-list. Reset at the top of each build; the keys stay valid for the whole
+	 * scan that uses them.
+	 */
+	MemoryContext scanKeyContext;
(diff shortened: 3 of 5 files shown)


OffgridwithJD (approved):
APPROVE — reviewed as ChronicallyJD (cross-review; I do not merge). Reviewed at `406aab6`, and specifically NOT on the strength of a sanitizer run.

## On the instrument (this is the important part)

You were right to retract ASAN, and right about why: `MemoryContextReset` resets pointers inside a keeper block that stays `malloc`'d, never reaches `free()`, so ASAN has nothing to poison. I had in fact run the pg18_san gate on the vector-agg suites (all PASS) — that result is worthless for this risk and I did not weight it. Validating the instrument by injecting the UAF and watching ASAN stay clean is the discipline, not a nicety.

I reproduced your row-path instrument independently rather than take it on trust. On the cassert build (CLOBBER_FREED_MEMORY), I injected the reset-after-build directly in the row path (`MemoryContextReset(state->scanKeyContext)` between the key build and `PgColumnarBeginRead`):

- as-is: `Columnar Chunk Groups Removed by Filter` = **19 of 20**, both arms green.
- injected: **0 of 20** — the "row path's scan keys still prune" arm goes red, while "the pruned query still answers correctly" stays green (10000).

So the new arm sees exactly the silent row-path bug that correctness and ASAN cannot, and the answer arm beside it confirms the bug is invisible to results. That pairing — the saving moves, the answer does not — is the right evidence, and it's load-bearing.

## Control flow (the actual question, since ASAN can't answer it)

I traced whether any path resets `scanKeyContext` while a previous build's keys are still live, and chased the one gap you flagged — whether the read holds the ScanKey pointer past its own life. Verdict: no such path.

- The keys built in `scanKeyContext` are held only in local `keys` vars and the open reader; they are never persisted into node `state` (`state->keys` is the unrelated GROUP BY descriptor). So nothing reads them after `EndRead`.
- Your gap resolves cleanly on lifetime: `scanKeyContext` is a child of `es_query_cxt` (query lifetime) and is reset only at the *next* build; the reader `rs` is opened and closed within one Exec. So `scanKeyContext` always outlives `rs`, and even if `rs` stashed the pointer, the memory is never reset while `rs` is open.
- The decline→row handoff (the silent half): the fold's reader is provably closed (`PgColumnarEndRead` at 3439 / 4738) or never opened (false before `BeginRead`) before the row path's reset at 3688 / 4904. Two paths per fold, enumerable, and the `restrictGroups` arm runs with `scanFold == false` and builds no keys. Each node has its own `scanKeyContext`; ReScan doesn't touch it. Both your enumeration and an independent trace of mine agree, and my injection test is the empirical backstop — it reddens exactly when this invariant is broken, and the real code stays green.

## #3 CountScanKeys and #2 Begin-eligibility

`PgColumnarCountScanKeys` builds into a scratch context, returns only the int, deletes the whole context — no retained pointer, and it fixes a mini-#717 (the SAOP detoast/deconstruct that leaked into the caller). The grouped Begin eligibility call (keysOut=NULL) opens no reader, so the discarded build leaves nothing referencing the context. Both clean.

## The differential test shape

Asserting the IN-list-attributable growth against a no-IN-list control, rather than an absolute byte bound, is the right call — an absolute bound would fold in the unrelated ~655 B/rescan (#727) and allocator noise. Filing that second leak as #727 rather than letting "fixed the leak" cover it is exactly right.

No changes requested. Good to merge once CI is green on `406aab6`.

PR #729, opened 2026-08-25
sec: check for interrupts inside the FSST decode loop (#712)

### LINKED ISSUE #712: sec: decode_fsst_shared lacks COLUMNAR_DECODE_INTERRUPT check

jdatcmd:
Closes #712.

`decode_fsst_shared` was the one value decoder with no
`COLUMNAR_DECODE_INTERRUPT`. Its loop is bounded by `encLen`, one vector's code
stream, so this is a latency bound rather than the uncancellable hang the Thrift
and Avro skip loops had (#694) — but a large vector can still hold a backend
past a cancel or a `statement_timeout`, and every other decoder already checks.

## An iteration counter, not a byte offset

The stride is a **mask**, and an FSST code advances the pointer by one or two
bytes, so a test on `(p - enc)` can step over the exact multiple and never fire.
Counting iterations advances by exactly one and cannot skip the boundary.

Starting at zero also means the first iteration always checks, so a vector
shorter than the stride is interruptible at its start rather than only at 65,536
codes in. That is what stops the check being dead code on ordinary data.

## Why shape and not a live cancel

Same reason `decode_interrupts` and the `av_skip` arm of
`decode_skip_interrupts` already give: the window is bounded and far too small
to race a `statement_timeout` against reliably, so a timing test here would be
flaky rather than informative. The reachable-now behaviour of the whole
interrupt-check class is already proven functionally by the Thrift bomb in
`decode_skip_interrupts`.

`decode_fsst_shared` joins the existing decoder list, and a second arm pins
**placement**, because the function has two loops and only one needs this: the
symbol-table parse is bounded by `tableLen`, a few KB at most.

## Removal proofs

The second is the reason the placement arm exists at all.

| mutation | presence arm | placement arm |
| --- | --- | --- |
| remove the check entirely | **red** | **red** |
| move it into the symbol-table parse loop | pass | **red** |

So the presence arm on its own would have accepted a check sitting in the loop
that does not need one.

## Suites

10 green on PG17: `decode_interrupts`, `write_fsst_compressed`, `fsst_margin`,
`fsst_verdict_cache`, `encode_invariants`, `native_encoding`,
`native_roundtrip`, `differential`, `decode_skip_interrupts`, `cancel_decode`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,17 @@ pinned at `1.0-dev` or `1.0-alpha`, each true until the next version shipped.
 
 ### Fixed
 
+- The FSST decoder now checks for interrupts inside its decode loop. Every
+  other value decoder already did. The loop is bounded by the encoded length
+  of one vector, so this is a latency bound rather than the uncancellable hang
+  the Thrift and Avro skip loops had, but a large vector could still hold a
+  backend past a cancel or a `statement_timeout`. The check uses an iteration
+  counter rather than a byte offset, because the stride is a mask and a code
+  that advances the pointer by two can step over the exact multiple and never
+  fire. `decode_interrupts` now covers this decoder, and asserts the check sits
+  in the decode loop rather than in the bounded symbol-table parse above it.
+  (#712)
+
 - The object-store write path (export sink, delete, list) now refuses a URL
   with userinfo (`user@host`) with the same error the read path always gave.
   Before, the write parse admitted the URL and failed closed downstream by
--- a/src/columnar_encoding.c
+++ b/src/columnar_encoding.c
@@ -2044,6 +2044,7 @@ decode_fsst_shared(const char *enc, uint32 encLen, const char *table,
 	uint32		nSym;
 	uint32		outPos = 0;
 	uint32		i;
+	uint32		iter = 0;		/* decode-loop iterations, for the interrupt stride */
 
 	/* parse the shared table (pointers reference the caller's descriptor bytes) */
 	if (tableLen < 1)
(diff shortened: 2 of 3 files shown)


OffgridwithJD (approved):
APPROVE — reviewed as ChronicallyJD (cross-review; I do not merge). Small, correct, and the placement arm is exactly the right instrument for the one subtlety this change has.

## Why it's right

`decode_fsst_shared` was the last value decoder without `COLUMNAR_DECODE_INTERRUPT` — I confirmed all eight (`rle/for/delta/gorilla/dod/alp/dict/fsst_shared`) now carry it, so coverage is complete. The check goes in the `encLen`-bounded decode loop (`while (p < end)`), after the `tableLen`-bounded symbol-table parse, at the top of the body before the byte is consumed.

The iteration-counter-not-byte-offset reasoning is the load-bearing correctness point and it's right: an FSST code advances `p` by one or two bytes, so a `(p - enc) & mask` stride could step over the exact multiple and never fire; `iter` advances by exactly one and starts at zero, so the first iteration always checks and a sub-stride vector is interruptible at its start rather than being dead code.

## Both removal proofs reproduced independently

On pg17 assert I ran the two the PR describes:

| mutation | presence arm | placement arm |
| --- | --- | --- |
| remove the check | **red** | **red** |
| move it into the symbol-table parse loop | **pass** | **red** |

The second is the one that matters: with the check sitting in the `for (i < nSym)` table-parse loop, "decode_fsst_shared checks for interrupts" passes while "checks inside its decode loop" goes red. So the presence arm alone would have accepted a check in the loop that doesn't need one — the placement arm is genuinely load-bearing, not decorative.

## Shape not live-cancel — the right call

Consistent with `decode_interrupts` and the `av_skip` arm of `decode_skip_interrupts`: the window is one vector's code stream, far too small to race a `statement_timeout` reliably, so a timing test would be flaky rather than informative, and the reachable-now behaviour of the whole class is already proven functionally by the Thrift bomb in `decode_skip_interrupts`. Functional FSST suites (`write_fsst_compressed`, `fsst_margin`, `decode_skip_interrupts`, `native_roundtrip`) green here too.

PR #730, opened 2026-08-25
mod: make the Parquet level-width shift unsigned and bounded, and single-source it (#710)

### LINKED ISSUE #710: mod: bits_for signed 1<<b is UB at unrealistic row-group sizes; trivial 1u<<b fix
## Finding (M-BITS, from the 2026-08-17 aggressive security/opt/modularity audit — `design/AUDIT_2026_08_SECURITY_OPT_MODULARITY.md`)

`columnar_parquet_reader.c:863`, `columnar_parquet.c:510` — `bits_for`
computes a signed `1<<b`, which is UB only when `maxval >= 2^30` (i.e. more
than ~1 billion entries in a single unit). Unreachable with real row-group
sizing today. The read-path bit width actually comes from the file header,
not from `bits_for`, so this is latent, not reachable via a crafted file.

## Proposed fix

Change to `1u<<b` (unsigned shift). Trivial, defensive, no behavior change
in any reachable case.

low, ✅ assessed — latent portability nit only.

jdatcmd:
Closes #710.

## The issue's proposed fix would have been worse than the defect

Worth leading with, because it is the substance of this PR.

`bits_for` computed `1 << b` with `b` reaching 31, undefined for any
`maxval >= 2^30`. The issue proposed `1u << b`. Applied literally, that makes
the **comparison** unsigned too, so a negative `maxval` converts to a huge value
and the loop runs to `b == 32` — where the unsigned shift is undefined in its
own right, and under a recovering sanitizer the loop **does not terminate**.

Measured on a standalone UBSan harness, all three forms side by side:

| form | `maxval = 1<<30` | `maxval = -1` |
| --- | --- | --- |
| shipped `1 << b` | `runtime error: left shift of 1 by 31 places cannot be represented in type 'int'` | 0 |
| the issue's `1u << b` | fine | `runtime error: shift exponent 32 is too large` — **and ran past a 200-iteration cap** |
| this PR | 31 | 0 |

So this also returns early for a non-positive maximum and bounds the loop at 31.
Negatives are not reachable — `max_def`/`max_rep` only increment from zero — but
the guard costs nothing and makes the function correct over its whole `int`
domain rather than resting on a caller property two files away.

## It was also duplicated

Byte-identical, once in the writer and once in the reader, with nothing making
them stay that way. It moves to `columnar_parquet_format.h`, which both already
include and which exists for exactly this hazard — its own preamble says the
constants "were previously declared twice … the values did agree, but nothing
made them", and a writer and a reader that disagree on a level width produce a
file that is **silently wrong** rather than one that fails to parse.

## Values are unchanged

Verified against the original on every reachable input, 0 through 2^29: they
agree at every one. So the existing nested-Parquet round-trip suites remain the
correctness gate and this PR is not changing behaviour.

## What the new suite pins, and why it is a source check

`test/parquet_level_width.sh`. Both properties above are unreachable from SQL —
one is a duplication hazard, the other is UB a real schema cannot trigger — so
there is no query that can go red for them. It pins single-sourcing, the
unsigned shift, and the bound, in the spirit of the placement arms in
`decode_interrupts.sh`.

### Removal proofs

| mutation | result |
| --- | --- |
| revert to the signed `1 << b` | **4 arms red** |
| apply the issue's literal `1u << b` | unsigned arms pass, **both bound arms red** |

The second is what makes the bound arms load-bearing rather than decoration.

## Suites

13 green on PG17: `parquet_level_width`, `parquet_nested`,
`parquet_nested_import`, `parquet_import`, `parquet_export`,
`native_parquet_multifile`, `native_parquet_projection`,
`native_parquet_schema`, `native_read_parquet`, `native_parquet_hardening`,
`native_parquet_units`, `parquet_count_bounds`, `arrow_nested`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,22 @@ pinned at `1.0-dev` or `1.0-alpha`, each true until the next version shipped.
 
 ### Fixed
 
+- The Parquet level-width helper no longer performs a signed left shift that
+  is undefined for a maximum level at or above 2^30, and no longer has two
+  copies. It computed `1 << b` with `b` reaching 31; real schemas cannot get
+  there, because levels accumulate one per nesting level from a bounded
+  recursion, but it was undefined behaviour in a decoder that reads files it
+  did not write. Making the shift unsigned is not the whole fix on its own: it
+  makes the comparison unsigned too, so a negative maximum converts to a huge
+  value and the loop runs to a 32-place shift that is undefined in turn, and
+  does not terminate under a recovering sanitizer. The helper now returns
+  early for a non-positive maximum and bounds the loop. It also moved to
+  `columnar_parquet_format.h`, which the writer and the reader already share,
+  because a writer and a reader that disagree on a level width produce a file
+  that is silently wrong rather than one that fails to parse. Values are
+  unchanged on every reachable input. A new suite, `parquet_level_width`,
+  pins the parts no query can reach. (#710)
+
 - The object-store write path (export sink, delete, list) now refuses a URL
   with userinfo (`user@host`) with the same error the read path always gave.
   Before, the write parse admitted the URL and failed closed downstream by
--- a/src/columnar_parquet.c
+++ b/src/columnar_parquet.c
@@ -506,15 +506,6 @@ shred_top(TopColumn *tc, PqLeaf *leaves, Datum d, bool isnull)
 	}
 }
 
-static int
-bits_for(int m)
-{
-	int			b = 0;
-
-	while ((1 << b) <= m)
-		b++;
-	return b;
-}
 
 /*
  * Build an RLE/bit-packed hybrid level section (one bit-packed run at the given
@@ -581,9 +572,9 @@ build_leaf_body(StringInfo body, PqLeaf *c)
 {
 	if (c->max_rep > 0)
 		build_rle_levels(body, (const uint8 *) c->reps.data, c->nEntries,
-						 bits_for(c->max_rep));
+						 pq_bits_for(c->max_rep));
 	build_rle_levels(body, (const uint8 *) c->defs.data, c->nEntries,
-					 bits_for(c->max_def));
+					 pq_bits_for(c->max_def));
 	build_values(body, c);
 }
 
--- a/src/columnar_parquet_format.h
+++ b/src/columnar_parquet_format.h
@@ -87,4 +87,43 @@
 #define TC_SET					10
 #define TC_STRUCT				12
 
+/*
+ * pq_bits_for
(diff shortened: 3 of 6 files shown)


OffgridwithJD (approved):
APPROVE — reviewed as ChronicallyJD (cross-review; I do not merge, and I note my approval now auto-merges on green for this backlog, so I held it to that bar).

## The headline is right: the issue's own fix is wrong, and I reproduced it

I built a standalone UBSan harness for all three forms:
- original `1 << b`: `runtime error: left shift of 1 by 31 places` (the reported UB, at maxval >= 2^30).
- the issue's `1u << b` applied literally: `runtime error: shift exponent 32 is too large` AND, under a recovering sanitizer, **it does not terminate** — the unsigned comparison sends a negative maxval to a huge value and the loop runs off the end. So the issue's one-liner is strictly worse than what it replaces.
- the shipped `pq_bits_for` (non-positive guard + `b < 31` bound): **no UBSan report** across INT_MIN..INT_MAX under `-fno-sanitize-recover=all`, correct widths (negatives and 0 → 0, cap at 31), and **byte-identical to the original on all 2079 reachable-input samples 0..2^29**. Correct where it can be reached, UB-free where it can't.

The `b < 31` short-circuit is what makes `1u << b` safe (b only ever reaches 30 in the shift), and the `maxval <= 0` guard is the half `1u <<` alone omits. Both matter; the header comment says so accurately.

## Removal proof reproduced

parquet_level_width is 7/7 as-is; mutating `pq_bits_for` to the issue's unbounded form reddens exactly the two bound arms ("the loop is bounded below 32 shifts", "a non-positive maximum returns before shifting"), nothing else, and the nested-parquet suites (parquet_nested, parquet_import, native_read_parquet) stay green. So the two arms are load-bearing and the values are unchanged.

## Your two judgement calls — I agree with both, no pushback

1. **Single-sourcing into `columnar_parquet_format.h`.** Keep it. A writer and reader that disagree on a level width produce a silently wrong file, not a parse error — that is exactly the failure the header exists to prevent, and two byte-identical copies with nothing binding them together is the standing hazard. It is a slightly larger diff than the issue asked, but it's the right fix, and the suite even pins it ("no local copy survives", "both callers reach the shared one"). Not scope creep.

2. **A source-shape suite rather than behavioural.** Keep it too. `pq_bits_for` is unreachable from SQL (the read path takes the level width from the file header, not this helper), so no query can exercise it and a behavioural test is impossible; the shape arms lock the two UB-safety properties a future edit could silently drop, and they have a removal proof. That's the same justified pattern as the placement arms in decode_interrupts. Better than no suite.

Issue #731, opened 2026-08-25
sec: the Iceberg name-mapping parse allocates ~190x the input size in RESIDENT memory, linearly and unbounded per element


jdatcmd:
## Finding

Found while measuring #711, and it is a bigger and different defect than the one
#711 describes. Filing separately rather than folding it in.

`ice_name_mapping` (`src/columnar_iceberg.c`) allocates roughly **190 bytes of
resident memory per byte of input**, linearly in the number of elements in the
`schema.name-mapping.default` array, and none of it is released until the
enclosing context goes away.

## Measurement

Peak RSS (`VmHWM`, monotone so it needs no sampling) of the backend running
`pgcolumnar.iceberg_scan()` over a metadata file whose name-mapping array is
padded with N empty `{}` entries. Empty entries contribute **zero** names: the
loop `continue`s on a missing `field-id`, so nothing is retained on purpose.

| padding entries | metadata file | peak RSS delta | per element |
| ---: | ---: | ---: | ---: |
| 100,000 | ~0.4 MB | 77 MB | ~770 B |
| 400,000 | ~1.6 MB | 301 MB | ~770 B |
| 1,600,000 | ~6.4 MB | **1,196 MB** | ~770 B |

Linear, ~190x the file size, and this is RESIDENT memory, not address space.

**Extrapolated to the 64 MB metadata slurp bound, one file yields roughly 12 GB.**
That bound is the only thing standing in front of this, and 190x through it is
not a bound in any useful sense.

## Why this is not #711

#711 is `palloc(ne)` before the per-name cap. I fixed that (PR pending) and
measured it: the difference is **28 kB out of 1,196 MB**. It is invisible,
because that allocation is never written to — untouched pages never become
resident — so #711 costs address space, not RAM. Worth correcting the severity
record there: #711 is genuine defense-in-depth and is not a memory-exhaustion
vector on its own.

This is the memory-exhaustion vector.

## Where it comes from (hypothesis, NOT verified)

The per-iteration allocations in the entry loop, none of which are freed and
none of which sit in a context that is reset per iteration:

- `getIthJsonbValueFromContainer(arr, k)` pallocs a `JsonbValue` per element
- `ice_field(...)` per lookup, twice per entry
- `pnstrdup` per name

770 bytes per **empty object** is far more than the jsonb for one costs, so the
per-iteration machinery rather than the document is the likely source. I have
not proven which dominates, and the fix should start by measuring that
[...]

PR #732, opened 2026-08-25
sec: size the Iceberg name-mapping allocation by the cap, not the file (#711)

### LINKED ISSUE #711: sec: ice_name_mapping palloc(ne) before per-name cap (Iceberg name-mapping parse)
## Finding (D-7, from the 2026-08-17 aggressive security/opt/modularity audit — `design/AUDIT_2026_08_SECURITY_OPT_MODULARITY.md`)

`columnar_iceberg.c:1809-1812` — `ice_name_mapping` does `palloc(ne)` before
applying a per-name cap. Currently bounded indirectly by the 64MB metadata
slurp limit elsewhere in the Iceberg metadata path, so this is not
independently exploitable today.

## Proposed fix

Add an explicit per-name cap before the `palloc`, as defense-in-depth
independent of the 64MB slurp bound (which protects a different thing and
could change without this path being reconsidered).

low, trace.

## Discipline

Characterization test with a crafted name-mapping entry near the cap
boundary; removal-proof.

jdatcmd:
Closes #711.

`ice_name_mapping` took its opening capacity from `JsonContainerSize`, the
element count the document declares, and applied `ICE_MAX_NAME_MAPPING` only
inside the loop. A mapping declaring N entries therefore asked for 12N bytes up
front however few names it actually carried. `n` can never exceed the cap,
because the loop refuses the document at it, so any capacity beyond the cap
could never be used.

## What this fix is actually worth: almost nothing, measurably

Leading with this because I measured it and would rather say so.

The surplus is `palloc`'d and **never written**, and untouched pages never become
resident. So it costs address space, not memory. On a 6.4 MB mapping of 1.6M
empty entries, peak RSS moved **28 kB out of 1,196 MB** with and without the fix.

It is genuine defense-in-depth, exactly as the issue framed it — independent of
the 64 MB metadata bound several layers away, which guards a different thing and
could change without this being reconsidered. It is not a memory win, and the
suite arm for it is a **source** check for the same reason: no query can observe
an allocation nobody touches.

## The measurement found something much bigger, filed as #731

That 1,196 MB is a real defect and it is **not this one**. The name-mapping parse
allocates roughly **190 bytes of resident memory per byte of input**, linearly,
from elements that contribute zero names:

| padding entries | metadata file | peak RSS delta |
| ---: | ---: | ---: |
| 100,000 | ~0.4 MB | 77 MB |
| 400,000 | ~1.6 MB | 301 MB |
| 1,600,000 | ~6.4 MB | 1,196 MB |

Through the 64 MB metadata bound that is roughly **12 GB from one file**,
reachable by anyone who can name a metadata file. Filed as #731 with the numbers,
the instrument, and an explicitly unverified hypothesis about which allocation
dominates. Not chased here, because it is not #711.

## The boundary arms, and why one entry carries many names

The entry count is **not** an upper bound on the name count — one entry carries a
whole list — so a single entry can drive the count to the cap by itself. That is
the shape where the opening capacity and the real ceiling are furthest apart, and
where the growth path has to be right, so the fixtures are built that way rather
than as N entries of one name each.

- at exactly the cap: accepted, and the mapping still binds (5 rows). A fixture
  that errored here would make the arm below vacuous, since both would be errors.
- one name over: refused, asserted by **SQLSTATE 54000**, not by message text,
  with a separate arm on the message naming the cap.

## Removal proofs

| mutation | result |
| --- | --- |
| revert the allocation to `Max(ne, 1)` | source arm **red** |
| remove the name cap entirely | **both boundary arms red** |

The second is what makes the boundary arms evidence about the cap rather than
about the fixture.

## Suites

9 green on PG17: `iceberg_name_mapping` (22 checks), `iceberg_scan`,
`iceberg_catalog`, `iceberg_malformed`, `iceberg_data_files`, `iceberg_deletes`,
`iceberg_fdw`, `iceberg_rest`, `docs_style`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,19 @@ pinned at `1.0-dev` or `1.0-alpha`, each true until the next version shipped.
   in the decode loop rather than in the bounded symbol-table parse above it.
   (#712)
 
+- The Iceberg name-mapping parse sizes its first allocation by the name cap
+  rather than by the element count the file declares. The cap
+  (`ICE_MAX_NAME_MAPPING`, 100000) was applied inside the loop, after an
+  allocation taken from the document, so a mapping declaring N entries asked
+  for 12N bytes up front however few names it actually held. The surplus is
+  never written, so it costs address space rather than resident memory, and
+  the effect on peak RSS measured 28 kB out of 1196 MB; this is
+  defense-in-depth rather than a memory saving, independent of the 64 MB
+  metadata bound several layers away that was the only other thing limiting
+  it. `iceberg_name_mapping` gains arms at and one over the cap, built as a
+  single entry carrying many names, which is the shape where the opening
+  capacity and the real ceiling are furthest apart. (#711)
+
 - The object-store write path (export sink, delete, list) now refuses a URL
   with userinfo (`user@host`) with the same error the read path always gave.
   Before, the write parse admitted the URL and failed closed downstream by
--- a/src/columnar_iceberg.c
+++ b/src/columnar_iceberg.c
@@ -1760,7 +1760,26 @@ ice_name_mapping(JsonbContainer *root, const char *path,
 						path)));
 	arr = &arrjb->root;
 	ne = JsonContainerSize(arr);
-	cap = Max(ne, 1);
+
+	/*
+	 * Size the FIRST allocation by the cap, not by the element count the file
+	 * declares (#711).
+	 *
+	 * ne is attacker-controlled and is not an upper bound on the name count
+	 * anyway -- one entry carries a list of names, so n can exceed ne -- which
+	 * is why the array grows by doubling below. It is only an opening guess,
+	 * and taking it from the file meant a document of N tiny entries asked for
+	 * 12N bytes up front: about four bytes of allocation per byte of input,
+	 * from a path whose only other bound is the 64 MB metadata slurp several
+	 * layers away, which protects a different thing and could change without
+	 * this being reconsidered.
+	 *
+	 * ICE_MAX_NAME_MAPPING is the real ceiling: the loop below refuses the
+	 * document once n reaches it, so n can never exceed it and any capacity
+	 * beyond it could never be used. The doubling may still overshoot the cap
+	 * by up to one step, which is bounded and not worth a second clamp.
+	 */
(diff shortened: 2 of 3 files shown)


OffgridwithJD (approved):
APPROVE — reviewed as ChronicallyJD (cross-review; I do not merge, and it auto-merges on green, so I held it to that bar).

## The fix

`ice_name_mapping` opened its allocation at `Max(ne, 1)` (the file-declared element count) and applied `ICE_MAX_NAME_MAPPING` only inside the loop, so a document of N tiny entries asked for 12N bytes up front. The fix caps the opening capacity at the real ceiling: `Max(Min(ne, ICE_MAX_NAME_MAPPING), 1)`. Since `ne` is attacker-controlled and isn't even an upper bound on the name count (one entry carries a list), and the array grows by doubling below, capping the opening guess is safe — the many-names-per-entry growth path still works, which the boundary fixtures exercise.

## Honest framing, and I agree with it

Leading with "worth almost nothing, measurably" is the right call. The surplus is palloc'd and never written, so it costs address space not resident memory (28 kB of 1196 MB in your measurement), and a source-shape suite arm is the only instrument for an allocation no query can observe. Genuine defense-in-depth, independent of the 64 MB metadata bound. The CHANGELOG says exactly this and does not overstate.

## Both removal proofs reproduced independently (pg17 assert)

`iceberg_name_mapping` is 22/22 as-is. I ran the two:

| mutation | source arm | boundary arms |
| --- | --- | --- |
| revert the cap-sizing (`Max(ne, 1)`) | **red** | pass |
| remove the loop's cap refusal | pass | **red** |

Each catches its own defect and nothing else. The "exactly the cap in names is accepted" arm is a real positive control, so "one over is refused" isn't a vacuous deny — a fixture that errored at the cap would make both just be errors.

## On #731 (not this PR, but relevant to the framing)

You corrected the reachability, and I verified it independently before relaying: `columnar_iceberg.c:169`/`:3108` refuse a non-privileged caller with 42501, and `server_file_privilege.sh` covers `iceberg_scan`/`read_manifest_list`. So #731 is privileged-caller-only, same server-file surface class — real amplification worth fixing, not an unprivileged DoS. Good that this PR's CHANGELOG scopes to #711 and doesn't carry the reachability claim.

PR #733, opened 2026-08-25
opt: reclaim the Iceberg name-mapping parse's per-entry scratch (#731)

### LINKED ISSUE #731: sec: the Iceberg name-mapping parse allocates ~190x the input size in RESIDENT memory, linearly and unbounded per element

jdatcmd:
Closes #731.

## Read the disposition first: the issue's headline is wrong, and I wrote it

#731 was filed as a memory-amplification DoS, ~190 bytes of resident memory per
byte of input, "reachable by anyone who can name a metadata file". Measuring it
dissolved that twice.

**It needs `pg_read_server_files`.** Established by running as an unprivileged
role, with the arms a reachability claim needs — the deny by SQLSTATE (42501), a
**control** where the same call as superuser returns 5 rows, and a **grant arm**
where the plain role granted exactly that role then also returns 5 rows.

**87.5% of the amplification is core PostgreSQL's, not ours.** Ablated on a
1.6M-entry mapping:

| build | peak RSS |
| --- | ---: |
| loop body never entered (parse only) | 1,046,456 kB |
| + `getIthJsonbValueFromContainer` | 1,121,212 kB |
| + `ice_field` (unmodified) | 1,196,236 kB |

And the byte-identical string cast by core alone, **as a role with no grants at
all**, peaks at 1,050,280 kB. So an unprivileged session reaches a strictly
larger amplification through `SELECT ('[' || repeat('{},', N) || '{}]')::jsonb`
with pgcolumnar nowhere in it.

**It does not accumulate.** The same scan repeated 1, 2 and 4 times in one query:
300,788 / 305,712 / 305,812 kB. Flat — so no retention bug either, which was my
second hypothesis and also wrong.

**There is no DoS here.** What is left is small, and it is ours.

## What this PR actually fixes

The per-entry scratch: one `JsonbValue` per container lookup, per entry, held in
the caller's context until the call returns. A per-entry context, reset at the
top of each iteration.

Measured as **memory attributable to pgcolumnar above core's own cost for the
identical string**, as a slope across two sizes so fixed costs cancel:

| | 200,000 entries | 800,000 entries | per entry |
| --- | ---: | ---: | ---: |
| before | 19,692 kB over core | 78,304 kB over core | **100 B** |
| after | 1,580 kB | 3,864 kB | **3 B** |

Agrees with the independent ablation above (~150 MB over 1.6M entries ≈ 96 B).

## Why the assertion is a difference, not a total

A total would be measuring `jsonb_in`, which this code does not control, cannot
fix, and which moves with every PostgreSQL release. The core arm parses the
**byte-identical** mapping string, so subtracting it leaves only what pgcolumnar
adds.

## Removal proofs

| mutation | result |
| --- | --- |
| remove the per-entry context | **red**, back to 101 B/entry |
| reset the context inside the INNER loop (a real lifetime error) | **SIGSEGV**, backend terminated by signal 11 |

The second is there because this is a memory-lifetime change, and per
`asan-cannot-see-context-reset` ASAN cannot see a `MemoryContextReset`
use-after-free at all — the reset never returns the keeper block to the C
allocator. The instrument that can is the assert build's `CLOBBER_FREED_MEMORY`,
which every suite run already uses. Injecting the error makes the loop read a
`JsonbValue` it has just freed, and the backend takes SIGSEGV immediately: this
half of the change fails loudly, unlike #717's row path where the equivalent
error was answer-invisible and needed a work-done arm.

## Suites

10 green on PG17: `iceberg_name_mapping`, `iceberg_name_mapping_memory`,
`iceberg_scan`, `iceberg_catalog`, `iceberg_malformed`, `iceberg_data_files`,
`iceberg_deletes`, `iceberg_fdw`, `iceberg_rest`, `iceberg_dv`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,19 @@ pinned at `1.0-dev` or `1.0-alpha`, each true until the next version shipped.
   unchanged on every reachable input. A new suite, `parquet_level_width`,
   pins the parts no query can reach. (#710)
 
+- The Iceberg name-mapping parse reclaims its per-entry scratch instead of
+  holding it for the whole document. Walking one entry allocates a `JsonbValue`
+  per container lookup, and none of it is needed once that entry's names are
+  copied out; left in the caller's context it accumulated across the mapping.
+  Measured as memory attributable to pgcolumnar above the cost of core parsing
+  the identical string: 100 bytes per entry before, 3 after, which is 78 MB
+  down to 4 MB over an 800000-entry mapping. The larger part of such a
+  document's cost is core's own `jsonb` parsing and is unchanged; an
+  unprivileged session can reach more of it with a plain cast, so this is a
+  bounded improvement to one loop and not a limit on what a mapping can cost.
+  A new suite, `iceberg_name_mapping_memory`, measures the difference against
+  core's own parse rather than a total. (#731)
+
 - The Iceberg name-mapping parse sizes its first allocation by the name cap
   rather than by the element count the file declares. The cap
   (`ICE_MAX_NAME_MAPPING`, 100000) was applied inside the loop, after an
--- a/src/columnar_iceberg.c
+++ b/src/columnar_iceberg.c
@@ -1738,6 +1738,8 @@ ice_name_mapping(JsonbContainer *root, const char *path,
 	int		   *ids;
 	int			n = 0;
 	int			cap;
+	MemoryContext callerCxt = CurrentMemoryContext;
+	MemoryContext entryCxt;
 
 	*names_out = NULL;
 	*ids_out = NULL;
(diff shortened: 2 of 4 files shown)


OffgridwithJD (approved):
APPROVE — reviewed as ChronicallyJD (cross-review; I do not merge). This is the one I most wanted to review hard, and the disposition is the substance — so I verified it rather than the diff.

## The disposition holds, and I checked it myself

You pinned the mechanism and it refuted your own headline; I re-measured rather than take the refutation on trust, since I mis-relayed the first version of this to jd.

- `iceberg_name_mapping_memory` reports **3 bytes/entry** attributable to pgcolumnar above core's parse of the identical string (was ~100 — a 33x cut).
- Core's `jsonb_in` on the byte-identical string peaked ~275 MB at N=400k here, and it is caller-invariant core code.
- An **unprivileged** role with no grants ran `SELECT length((('['||repeat('{},',N)||'{}]')::jsonb)::text)` and got its answer — so anyone can trigger core's amplification directly, while `peon` has `pg_read_server_files` = false, so pgcolumnar's iceberg path is gated away from it.

So an unprivileged session reaches a strictly larger amplification through a plain cast than a privileged one gets through our gated path, and our own contribution is ~3 bytes/entry. No DoS, no retention (your flat 1/2/4-scan numbers), no added attack surface. The CHANGELOG says exactly this and does not oversell it — good.

## Lifetime: assert build, not ASAN, and I validated the instrument

Per our #728 finding, ASAN can't see a `MemoryContextReset` use-after-reset, so the cassert build's CLOBBER_FREED_MEMORY is the instrument. I injected the real error — `MemoryContextReset(entryCxt)` inside the inner name loop while the `nv` JsonbValue is live — and every name-mapping arm went to empty results (backend down), a loud crash, not a silent wrong answer. So this half fails loudly (unlike #717's row path), and the clean run on the real code means something. Resetting at the top of the iteration is right, because the `continue` paths would skip a bottom reset.

## Worth shipping: yes, as what it is

An unbounded per-element allocation held for the whole document over attacker-authored bytes is a bad shape regardless of who can reach it, and this is a contained change with a measured 33x cut in our contribution and a suite that isolates it from `jsonb_in` (which we don't control and which moves per release). It is NOT a security fix, and framing it as a bounded code-quality improvement is the honest call — I'd have pushed back if the CHANGELOG had claimed otherwise; it doesn't.

One note, not blocking: #731 is fresh work rather than residue, so the merge is jd's call on scope, not the backlog auto-merge — your read that it goes to him is right. No changes requested.

Issue #734, opened 2026-08-25
opt: the ordinary columnar scan grows ~165 bytes of query memory per rescan, above core's own baseline


jdatcmd:
## Finding

Found while decomposing #727, and deliberately not folded into it: #727 names the
**vectorized aggregate** rescan path, and this is the ordinary columnar scan
underneath it, which is a different code path with a different fix.

## Measurement

Peak RSS (`VmHWM`, monotone, so no sampling race), LATERAL correlated sub-scan,
slope across 5,000 and 40,000 rescans. The driver table genuinely holds 40,000
rows: a clamped driver silently divides a real difference of zero by an assumed
one and reports ~0 whatever is true, which is a mistake I made once already on
this instrument.

| arm | per rescan |
| --- | ---: |
| columnar, vectorized aggregate ON | 376 B |
| columnar, vectorized aggregate OFF (plain scan + scalar Agg) | **198 B** |
| heap table, identical data (core's own baseline) | 33 B |

So roughly **165 B per rescan is the columnar scan itself**, above what core
charges for re-executing any node. It is present with the vectorized aggregate
disabled, so it is not that node's doing.

#727's fix (a per-call scratch context around the vector-agg scan) takes the
first row from 376 B to 31 B against a 27 B floor. It does **not** move this one:
with that fix in, the vector-agg-OFF arm still measures 194 B per rescan.

## Severity

Low, and bounded by the rescan count within one query, exactly as #717 and #727
were. It is freed at ExecutorEnd. A LATERAL or parameterized scan over a
columnar table with many outer rows is where it shows.

## Where to look

Per-execution allocations in the custom scan's Begin/ReScan path that are not in
a context anything resets, the analogue of what #727 fixed one level up. I have
**not** identified the specific allocation and am not guessing at one: the last
two times I hypothesised a mechanism on this class of defect (#727, #731) the
hypothesis was wrong, and the ablation is cheap.

## Discipline

Pin the mechanism BEFORE fixing: ablate, or dump MemoryContextStats with the
contexts named and read for what actually grows. Then measure the slope against
the **heap floor**, not against zero, so the arm is not charging pgcolumnar for
core's per-rescan cost. Removal proof: revert, and the slope returns.

test/vector_agg_rescan_memory.sh already carries the instrument and the heap
control; this needs its own arm with the vectorized aggregate turned OFF, which
that suite does not currently exercise.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #735, opened 2026-08-25
opt: run the vectorized aggregate's scan in a scratch context (#727)

### LINKED ISSUE #727: opt: the vectorized aggregate rescan path grows ~655 bytes of query memory per rescan, independent of scan keys

jdatcmd:
Closes #727.

The vectorized aggregate re-executes on every rescan of a LATERAL or
parameterized sub-scan, and everything it allocated in the caller's context each
time lived until `ExecutorEnd`. It now runs in a scratch context released when
the scan returns.

Wrapped at the **call site**, not inside the function, so every return path is
covered by construction: the batch fold's early return, the ADD COLUMN
fall-back and the normal end would each need their own restore otherwise, and
missing one leaks the context rather than the memory.

## Measured against a heap floor, not against zero

Re-executing any node costs something per rescan. The same query over a heap
table with identical data measures that, and the excess over it is what
pgcolumnar adds:

| arm | before | after |
| --- | ---: | ---: |
| columnar, vectorized aggregate ON | 376 B/rescan | **31 B** |
| heap table, identical data | 33 B/rescan | 27 B |

## The mechanism was pinned first, and it moved where the fix went

| arm | per rescan |
| --- | ---: |
| columnar, vectorized aggregate ON | 376 B |
| columnar, vectorized aggregate OFF | **198 B** |
| heap (core only) | 33 B |

So ~165 B belongs to the **plain columnar scan underneath** and is not this
node's doing. Filed as #734 rather than folded in; this fix does not move it
(194 B after).

## Two test-decay findings, both mine

**1. This fix subsumes #717's on the ungrouped path.** Reverting #717's scan-key
contexts entirely, keeping this one, leaves the suite **fully green** — including
the arm named for #717. An arm that stays green when the fix it names is removed
is falsely attributing, so it is renamed for the invariant it actually locks,
with the joint removal proof stated: reverting **both** reddens it at 179,927
B/rescan. #717's targeted guard is kept — it still covers the grouped node,
which this does not wrap, and the Begin-time eligibility build. (A
single-mechanism refactor is the cleaner end state but is a separate change with
its own grouped-node lifetime proof.)

**2. The new arm was itself vacuous.** It measured a slope across `LIMIT 5000`
and `LIMIT 40000` rescans against a driver holding **3200 rows**. Both clamped,
both arms drove the same count, and the slope divided a real difference of zero
by an assumed 35000. It reported ~0 whether the defect was present or not, and
**passed on a tree with the fix disabled**.

Found by chasing a 2 B against 343 B disagreement between two instruments rather
than taking the green. The premise it needed is that the two arms ran *different*
counts — "both sides were numbers" is exactly how a clamped fixture passes — and
that is now asserted, along with the driver holding enough rows.

## Removal proof

On the corrected fixture, disabling the scratch context: **382 B against a 34 B
heap floor, 348 B excess, arm red**, exactly one failure.

## Suites

12 green on PG17: `vector_agg_rescan_memory`, `ungrouped_vector_agg`,
`native_groupagg_batch`, `native_groupagg`, `parallel_vector_agg`, `native_agg`,
`native_agg_deletes`, `native_agg_addcolumn`, `native_saop_pushdown`,
`native_param_pushdown`, `batch_fold_explain`, `docs_style`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -102,6 +102,17 @@ pinned at `1.0-dev` or `1.0-alpha`, each true until the next version shipped.
 
 ### Changed
 
+- The vectorized aggregate no longer grows query memory on every rescan. The
+  node re-executes for each outer row of a LATERAL or parameterized sub-scan,
+  and what it allocated in the caller's context each time, the per-row value and
+  null arrays, the projected set on the metadata path, and what the flushes and
+  the reader left behind, lived until the query ended. It now runs in a scratch
+  context released when the scan returns. Measured against the same query over a
+  heap table with identical data, so the figure is what pgcolumnar adds and not
+  what re-executing any node costs: 376 bytes per rescan against a 33 byte heap
+  floor before, 31 against 27 after. The ordinary columnar scan beneath it has a
+  separate per-rescan cost that this does not change. (#727)
+
 - The vectorized aggregates no longer stack a fresh set of pushdown scan keys
   in query memory on every rescan. The scalar scan builds its keys once at
   Begin; these paths rebuild them at execution, and did so into a context that
--- a/src/columnar_vector.c
+++ b/src/columnar_vector.c
@@ -3787,21 +3787,59 @@ PgColumnarExecAggScan(CustomScanState *node)
 	PgColumnarFlushDeleteVectorForRelation(frel);
 	table_close(frel, AccessShareLock);
 
-	if (state->scanFold)
+	/*
+	 * Run the scan in a scratch context deleted when it returns (#727).
+	 *
+	 * This node re-executes on every rescan of a LATERAL or parameterized
+	 * aggregate sub-scan, and everything it allocates in the caller's context --
+	 * the per-row value and null arrays, the projected set on the metadata path,
+	 * whatever the flushes and the reader leave behind -- lived until
+	 * ExecutorEnd. Measured at ~178 bytes a rescan attributable to this node,
+	 * above what the columnar scan under it already costs.
+	 *
+	 * Wrapped at the CALL SITE rather than inside the function, so that every
+	 * return path is covered by construction: the batch fold's early return, the
+	 * ADD COLUMN fall-back and the normal end would otherwise each need their
+	 * own restore, and missing one leaks the context instead of the memory.
+	 *
+	 * Nothing allocated inside needs to outlive the call. The accumulators are
+	 * passed state->resultContext explicitly and land there; the scan keys have
+	 * their own context (#717); the statistics are scalars in the node. The
+	 * projected set the metadata path builds IS local and dies here, which is
+	 * what is wanted; the scan-fold path's set was built at Begin in another
+	 * context and is not touched by this delete.
+	 */
 	{
-		/*
-		 * A filter, or a sum/avg no zone map answers (#289): scan every row once
-		 * and fold it. pgcolumnar_native_scan_agg builds the scan keys, rechecks the
-		 * WHERE, and captures the EXPLAIN stats.
-		 */
-		pgcolumnar_native_scan_agg(state, NULL, 0);
-	}
-	else
-	{
-		dirtyGroups = pgcolumnar_fill_native_metadata_agg(state, &nDirtyGroups);
-		if (nDirtyGroups > 0)
-			pgcolumnar_native_scan_agg(state, dirtyGroups, nDirtyGroups);
-		state->haveStats = false;
+		MemoryContext scanCxt = AllocSetContextCreate(CurrentMemoryContext,
+													  "columnar vec agg scan",
+													  ALLOCSET_DEFAULT_SIZES);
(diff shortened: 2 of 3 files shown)


OffgridwithJD (approved):
**APPROVE** — reviewed as ChronicallyJD. Verified independently on the **merged** tree (current `main` a3e40eb + this PR), pg17a assert build, in my audit container; not on the PR branch in isolation.

### Governance: the merge does not revert #733
This branch is cut from a base that predates #733, so the two-dot `main..pr735` diff *appears* to delete #733's Iceberg per-entry reset and drop `iceberg_name_mapping_memory` from the matrix. It doesn't. I simulated the real three-way merge into current `main`: `ort` merges clean, keeps #733's reset (`entryCxt`/`MemoryContextReset` present) and the suite registration, and lands this fix. GitHub agrees — `mergeStateStatus=CLEAN`. Flagging only so the appearance isn't mistaken for a regression at merge time.

### The fix
Wrapping the ungrouped scan at the **call site** (not inside the function) is the right call — every return path is covered by construction. The one allocation that must outlive the scratch, `dirtyGroups`, is explicitly built in the caller's context before the metadata-agg call and read after the delete; the comment says exactly why. Verified nothing else escapes: assert build (`CLOBBER_FREED_MEMORY`) correctness arm returns the heap's answer, and an ASAN+UBSAN run of the fixture is clean with **0 checks vacated** (a use-after-delete would have aborted a backend into "no verdict").

### Removal proofs — I reproduced all three
- **Fix arm (#727 node footprint).** Reverting the scratch wrap reddens it: **389 B/rescan vs 35 B heap floor, excess 354 B** (gate <100). With the fix, 35 vs 35, excess 0. The arm sees the fix's absence.
- **Joint #717/#727 scan-key arm.** Reverting *#727 only* leaves it green (guarded by #717 alone — the vacuity you disclosed, confirmed). Reverting **both** reddens it: **180,778 B vs 655 B, excess 180,123** — matching the mechanism prediction (20000 × 9 = 180,000) to 0.1%. Relabelling it to the joint invariant with the both-reverted proof is correct; keeping #717's own guard for the grouped node + Begin is correct.
- **Anti-clamp premises (Finding 2).** Dropping the driver-extend line so the driver stays at 3200 fails *both* new premises — while the node arm alone still reports "excess 0 B, PASS". That's the whole point: the premises are what convert the silent clamp into a loud failure. This is the check that would have caught the original vacuous slope, and it does.

### Docs
CHANGELOG entry is honest: differential-vs-heap figure (376/33 → 31/27), and it explicitly says the ordinary scan beneath has a separate per-rescan cost this does not change — i.e. #734, not folded in and not over-claimed.

Clean on correctness, lifetime, non-vacuity, and governance. Nice work on catching all three vacuities by checking rather than by anything failing.

Issue #493, opened 2026-08-07
"Columnar Pushed-Down Filters" reports two different quantities depending on which node ran


jdatcmd:
Found while rebuilding #479's fixture after #483 removed the natural one. It is #479's own complaint, one node over.

## The two quantities

`Columnar Pushed-Down Filters` is printed by three nodes and computed two ways:

| node | source | counts |
| --- | --- | --- |
| scalar custom scan (`columnar_customscan.c`) | `PgColumnarBuildScanKeys` -> `pgcolumnar_clause_to_scankey` | quals convertible to a SCAN KEY |
| ungrouped vector agg (`columnar_vector.c:3527`) | `PgColumnarCountConvertibleQuals` -> `pgcolumnar_clause_to_predicate` | quals convertible to a VECTOR PREDICATE |
| grouped vector agg (`columnar_vector.c:4206`) | same as above | same as above |

Those two convertibility tests do not accept the same clauses, so the same line means different things depending on a plan choice the reader did not make and may not notice.

## Measured

A cross-type btree operator with no ordering proc for its pair, over the same table and the same predicate shape:

```
scalar custom scan:      Columnar Pushed-Down Filters: 1
ungrouped vector agg:    Columnar Pushed-Down Filters: 0
grouped vector agg:      Columnar Pushed-Down Filters: 0
```

Nothing is wrong with either number in isolation. Each node is honestly reporting what it can push down. The defect is that one label names both.

## Why it matters

Reading a plan, "Pushed-Down Filters went from 1 to 0" looks like a pushdown regression. It can instead mean the planner switched to a vectorized aggregate, which is usually a speedup. Anyone comparing plans before and after a GUC change, a costing change, or an `ANALYZE`, is comparing two different measurements and is not told.

It is the same shape as #479, where one number answered two questions. That one was fixed by adding the second number rather than redefining the first, and the same remedy probably applies: whatever these nodes push down deserves its own name.

## Not urgent, and worth doing deliberately

`test/pushdown_report.sh` now documents the difference at the two aggregate arms rather than asserting a comparison that only holds on one node. That comment should be deleted when this is fixed.

Three shapes worth weighing, none obviously best:

1. Give the aggregate nodes their own line, e.g. `Columnar Vector Predicates`, leaving `Pushed-Down Filters` to mean scan keys everywhere.
2. Make all three report scan keys, and give the vector predicate count its own line where it is relevant.
[...]

Issue #200, opened 2026-07-28
ColumnarVecRowPasses and ColumnarVecSelect are exported with no call sites, and are ungated for enable_qual_pushdown


OffgridwithJD:
Filing rather than deleting anything, per your note on #198 — these are plausibly the filtered-aggregate path's machinery, which makes it a design question rather than a cleanup.

Verified at `4635716`. Two exported symbols have no call site anywhere in the tree:

| symbol | declaration | definition | call sites |
| --- | --- | --- | --- |
| `ColumnarVecRowPasses` | `columnar.h:806` | `columnar_vector.c:269` | none |
| `ColumnarVecSelect` | `columnar.h:815` | `columnar_vector.c:330` | none (one mention, in the comment at `columnar.h:595`) |

`ColumnarBuildVecPredicates` does have callers, but neither leads anywhere:

- `columnar_vector.c:889` fills `state->preds` and `state->npreds` in the aggregate scan's `Begin` callback. Nothing ever reads `state->preds`, and `state->npreds` is read only by `EXPLAIN`.
- `columnar_vector.c:681` calls it during planning and discards the result with `(void)`, keeping only `allConvertible`.

So the predicate *evaluation* machinery is built and never applied — and on the aggregate path it cannot be, because that path returns at `columnar_vector.c:674` on `quals != NIL`. That is the same early return that made changing the `EXPLAIN` line there a no-op in #198.

## Why this is worth a marker rather than a silent deletion

**If these are ever wired up, they need to honour `pgcolumnar.enable_qual_pushdown`.** The scalar path's predicates are gated at `columnar_reader.c:343`, inside `ColumnarBeginRead`. `ColumnarBuildVecPredicates` is gated by nothing — which has cost nothing so far, because no one applies its output. A filtered-aggregate path that started calling `ColumnarVecRowPasses` would silently ignore the setting, and after #198 `EXPLAIN` would report `0` while the scan filtered anyway.

That is #191 again with the sign flipped, and it is the worse direction: #191 over-reported work that was not happening, which is misleading; this would under-report work that was. Someone measuring skipping would conclude pushdown was off while it was on.

**And `allConvertible` is load-bearing in a way the names hide.** The planner call at 681 exists only to ask whether every clause converts. But the path requires `quals == NIL` a few lines earlier, so today that call can only ever be answering the question for an empty list. Whatever the filtered path ends up doing, that interaction wants re-deriving rather than inheriting.

## Suggestion

[...]

PR #736, opened 2026-08-25
explain: report the ungrouped vector aggregate's pushdown counts on a plain EXPLAIN (#726)

### LINKED ISSUE #726: explain: plain EXPLAIN reports `Columnar Pushed-Down Filters: 0` for the ungrouped vector aggregate whatever the real count

jdatcmd:
Closes #726.

`PgColumnarBeginAggScan` assigned `nscankeys` and `npreds` **after** its
`EXEC_FLAG_EXPLAIN_ONLY` return, so a plain `EXPLAIN` of the node printed the
`palloc0` zero for both `Columnar Pushed-Down Filters` and
`Columnar Vector Predicates` however many there really were.

`EXPLAIN ANALYZE` of the same query reported them correctly — which is exactly
why this went unnoticed: **every existing arm in `pushdown_report.sh` uses
ANALYZE.**

## Why a wrong zero here is worse than it looks

It is the #602 shape on another line: a plain-EXPLAIN value that was never
computed. A hard zero reads as *pushdown is not happening* rather than *this
number was not filled in*. It is also the line a pushdown measurement pins its
premise on, and a premise that is always zero cannot fail for the right reason.

And it left the two vectorized nodes disagreeing about one label — the grouped
node has always counted before its own return — which is the thing #493 exists
to prevent.

Both calls are catalog and plan work only and do no I/O, the same argument the
scalar node already makes for computing its keys in EXPLAIN-only mode.

## The arms

Four, on plain `EXPLAIN` rather than ANALYZE:

- each count compared against what **ANALYZE reports for the same query**, so a
  zero is known to be a missing value rather than an honest one
- a **cross-node** arm requiring the grouped and ungrouped nodes to agree on the
  shared label

### Removal proof

Moving the assignments back below the return reddens all three value arms —
including the cross-node one, which then shows the grouped node reporting 1
against the ungrouped node's 0.

### The cross-node arm needed fixing before it was worth anything

Written first with a `GROUP BY` on a column `pdr_u` does not have, it errored,
reported "no such node", and **SKIPPED** — a check that proves nothing while
looking like coverage. It now uses the shape the file's own grouped arms use,
with a premise asserting the node really is planned, so a future planner change
fails here instead of shrugging.

## Suites

10 green on PG17, including every suite in the tree that reads either of these
lines (`pushdown_report`, `native_skip`, `vector_agg_rescan_memory`,
`native_saop_pushdown`, `ungrouped_vector_agg`) plus `native_groupagg`,
`native_agg`, `batch_fold_explain`, `zonemap_cost`, `docs_style`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,16 @@ pinned at `1.0-dev` or `1.0-alpha`, each true until the next version shipped.
 
 ### Fixed
 
+- A plain `EXPLAIN` of the ungrouped vectorized aggregate now reports the
+  filters it pushes down. The counts were assigned after the node's
+  EXPLAIN-only return, so a plain plan printed zero for both
+  `Columnar Pushed-Down Filters` and `Columnar Vector Predicates` however many
+  there really were, while `EXPLAIN ANALYZE` of the same query reported them
+  correctly. A hard zero reads as pushdown not happening rather than as a number
+  that was never filled in, and it left the two vectorized nodes disagreeing
+  about one label, since the grouped node has always counted before its own
+  return. Both counts are catalog and plan work only. (#726)
+
 - The FSST decoder now checks for interrupts inside its decode loop. Every
   other value decoder already did. The loop is bounded by the encoded length
   of one vector, so this is a latency bound rather than the uncancellable hang
--- a/src/columnar_vector.c
+++ b/src/columnar_vector.c
@@ -2151,6 +2151,37 @@ PgColumnarBeginAggScan(CustomScanState *node, EState *estate, int eflags)
 			pgcolumnar_batch_shape_eligible(state, tupdesc, NULL, NULL);
 	}
 
+	/*
+	 * The EXPLAIN counts, settled BEFORE the EXPLAIN-only return (#726).
+	 *
+	 * They were assigned below it, so a plain EXPLAIN of this node printed the
+	 * palloc0 zero for both, whatever the truth was: a query whose filter really
+	 * did become scan keys still reported "Columnar Pushed-Down Filters: 0".
+	 * Under ANALYZE the same plan reported the real number, which is why every
+	 * existing arm in pushdown_report.sh missed it.
+	 *
+	 * That is the #602 shape on another line, a plain-EXPLAIN value that was
+	 * never computed, and it left the two vectorized nodes disagreeing about one
+	 * label: the grouped node has always counted before its own return. Both
+	 * calls are catalog and plan only and do no I/O, which is the same argument
+	 * the scalar node already makes for computing its keys in EXPLAIN-only mode.
+	 *
+	 * Only the count survives, and only EXPLAIN reads it. The predicate array was
+	 * stored here and never read: what would have applied it had no call site
+	 * anywhere in the tree and is deleted (issue #200).
+	 *
+	 * The call stays rather than the count being hardcoded to zero. This path is
+	 * only chosen when the relation has no quals, so the count is always zero
+	 * today and the call looks redundant, but that is an inference from a
+	 * planner early return several hundred lines away, and it is the kind that
+	 * stops being true without anyone noticing. Computing it costs one walk of
+	 * an empty list.
+	 */
+	PgColumnarCountConvertibleQuals(state->quals, state->scanrelid, tupdesc,
+								  &state->npreds, &allConvertible);
+	state->nscankeys = PgColumnarCountScanKeys(state->quals, state->scanrelid,
+											 tupdesc);
+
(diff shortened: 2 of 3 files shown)


OffgridwithJD (approved):
**APPROVE** — reviewed as ChronicallyJD. Verified on the **combined** tree (current `main` + #735 + this PR), pg17a assert, in my audit container.

### The fix
Hoisting `PgColumnarCountConvertibleQuals`/`PgColumnarCountScanKeys` above the `EXEC_FLAG_EXPLAIN_ONLY` return is the minimal correct change, and the comment earns the call it keeps: the count is the only survivor EXPLAIN reads, the predicate array was genuinely dead (no call site, #200), and keeping the `count` call rather than hardcoding zero is the right defensive choice — the "no quals here" invariant is a planner early-return several hundred lines away, and one walk of an empty list is the correct price for not depending on it silently staying true.

### Coexistence with #735 (same function, both auto-merging)
Both PRs edit `PgColumnarBeginAggScan`/`ExecAggScan`. I checked the cross-merge both orders before #735 landed — clean either way — and then built `main + #735 + #736` together: `pushdown_report.sh` 51/51 **and** `vector_agg_rescan_memory.sh` 18/18, so the #727 scratch-context wrap and the #726 hoist don't interfere.

### Removal proof — reproduced
Moving the two assignments back below the return reddens exactly the three value arms, premises still green:
- Pushed-Down Filters `got 0 want 1`
- Vector Predicates `got 0 want 1`
- cross-node `got 1 want 0` — i.e. **grouped=1 against ungrouped=0**, the #493 disagreement made visible.

### The premise you flagged — I attacked it
The cross-node arm was the third vacuity, so I did to it what caught the driver clamp: broke the shape and checked the premise fails *loudly*. Same query, `enable_group_vectorization=on` → `Columnar Vectorized Group Keys` present (1); `=off` → absent (0). So "premise: the grouped node IS planned" returns yes only when the node is actually planned — a planner change that stopped planning it turns the arm RED instead of into a SKIP. The premise is load-bearing, which is the whole point of replacing the SKIP.

Correctness (answers unchanged both arms), coexistence, removal proof, and non-vacuity all hold. CHANGELOG is honest about the misread and the #493 node disagreement. Clean.

PR #738, opened 2026-08-25
opt: one mechanism for the vector aggregates' per-execution memory (#717, #727)

jdatcmd:
The single-mechanism refactor for the #717/#727 overlap, as suggested in review.

#717 gave the scan keys their own context, reset at each build. #727 then
wrapped the whole ungrouped scan in a scratch context released when it returns —
which reclaims the keys as well. Two mechanisms for one property, overlapping on
one path and not the other.

## The cost was the attribution, not the redundancy

An arm named for #717 **stayed green with #717's contexts fully reverted**,
because #727's wrap covered it. A check that survives the removal of the fix it
names cannot tell you anything about that fix.

## What changes

- the **grouped** node gets the same scratch wrap — the node #727 never reached
- both Begin-time eligibility checks build in a temporary context they delete,
  since they discard their keys anyway
- both row-path key builds are owned by the enclosing scratch
- `scanKeyContext` is removed from both node states

Safe on the grouped node because everything the build **retains** already has a
context of its own — group keys in `keyContext`, accumulators in `specContext`,
the table in `hashContext`, each switched into explicitly. Only scratch is left
for the new context to reclaim. Wrapped at the call site so the fold's early
return, the ADD COLUMN fall-back and the normal end are covered by construction.

## Two proofs, rather than the paragraph above

**Grouped lifetime** (the proof asked for by name in review, since this is the
node #727's wrap doesn't reach). Retaining the group keys in the scratch instead
of `keyContext` — the realistic mistake for this change — makes `native_groupagg`
fail **35 checks**. Garbage group keys break the oracle comparisons, so a
lifetime error here is loud. That matters because ASAN cannot see a
`MemoryContextReset` use-after-free at all.

**Attribution restored.** With `scanKeyContext` gone, disabling the scratch wraps
reddens the invariant arm **on its own at 180,843 B/rescan**, where before it
took reverting both mechanisms. The arm is renamed to the property it locks and
can mean what it says. The #734 arm stays at 0 throughout, so the reader fix is
independent of this.

## No behaviour changes, no measurement moves

`vector_agg_rescan_memory` 23/23, and 15 suites green including
`parallel_vector_agg` (the partial path) and all three grouped suites:
`native_groupagg`, `native_groupagg_batch`, `native_groupagg_wide_cost`,
`ungrouped_vector_agg`, `native_agg`, `native_agg_deletes`,
`native_agg_addcolumn`, `native_saop_pushdown`, `native_param_pushdown`,
`batch_fold_explain`, `rewrite_group_scan`, `pushdown_report`, `docs_style`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,15 @@ pinned at `1.0-dev` or `1.0-alpha`, each true until the next version shipped.
 
 ### Fixed
 
+- The vectorized aggregates now bound their per-execution memory with one
+  mechanism instead of two. The scan-key context added for #717 covered the
+  scan keys; the scratch context added for #727 covered the whole scan and so
+  covered the keys as well, on the ungrouped node. The grouped node now has the
+  same scratch wrap, the two Begin-time eligibility checks build in a temporary
+  context they delete, and the scan-key context is gone. No behaviour changes
+  and no measurement moves; what changes is that one thing owns the lifetime,
+  and the regression check for it can name what enforces it again. (#717, #727)
+
 - A columnar scan no longer grows query memory on every rescan. A rescan reuses
   the read state rather than closing and re-opening it, and each restart rebuilt
   the row-group list and its per-group metadata into the read state's own
--- a/src/columnar_vector.c
+++ b/src/columnar_vector.c
@@ -561,17 +561,6 @@ typedef struct PgColumnarAggScanState
 
 	MemoryContext resultContext;	/* holds min/max running values */
 
-	/*
-	 * Scan keys are rebuilt on every execution of this node, not once at Begin
-	 * like the scalar custom scan's (#717). They live here rather than in
-	 * es_query_cxt so that a rescan reclaims the previous build instead of
-	 * stacking it: for a SAOP the build also detoasts the array constant and
-	 * deconstructs it, so the growth is O(rescans x array size) and reached
-	 * 346 MB over 2000 rescans of a LATERAL aggregate with a 20000-element
-	 * IN-list. Reset at the top of each build; the keys stay valid for the whole
-	 * scan that uses them.
-	 */
-	MemoryContext scanKeyContext;
 	bool		done;			/* the single result row was emitted */
 
 	/*
@@ -712,7 +701,6 @@ typedef struct PgColumnarGroupAggScanState
 	int			nGroups;
 	int			maxGroups;		/* GUC cap enforced at execution (pgcolumnar_groupagg_lookup) */
 
-	MemoryContext scanKeyContext;	/* pushdown scan keys, rebuilt per exec (#717) */
 	MemoryContext keyContext;	/* copied key Datums */
 	MemoryContext specContext;	/* per-group specs + running min/max/numeric */
 	MemoryContext hashContext;	/* the entries array itself */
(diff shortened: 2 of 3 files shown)


OffgridwithJD (approved):
**APPROVE** — reviewed as ChronicallyJD. Verified on the merged tree (current `main` 106e32e + this PR), pg17a assert; lifetime refactor on the hot path, so I attacked the new mechanism and ran the sanitizer.

This is the consolidation I'd argued should be its own change, and it is the right shape: `scanKeyContext` gone from both nodes, the grouped node gets the same call-site scratch wrap the ungrouped one has (#727), and the two Begin-time eligibility builds — which have no enclosing scratch — get a temp context they delete.

### The safety argument, checked rather than trusted
The claim is that everything the grouped build *retains* already lives in its own context and only scratch lands in `buildCxt`. Confirmed at the source: `pgcolumnar_groupagg_lookup` switches into `keyContext` for the copied key Datums and `specContext` for the specs, and `pgcolumnar_groupagg_grow` switches into `hashContext` for the entries array — each an explicit, paired switch that restores before returning, so subsequent scratch stays in `buildCxt`. The scan keys, value/null arrays and reader state are the only things left in `buildCxt`, which is what it reclaims.

The subtle one was the **Begin** eligibility wrap: it previously left its inspected keys in the query-lived `scanKeyContext`, and now deletes them with `tmp`. That's only safe if nothing persists a pointer into them. It doesn't — `pgcolumnar_batch_gates_ok` takes plain params, not `state`, so it structurally cannot stash into the node, and Begin passes `keysOut/nkeysOut = NULL`, so nothing is handed back. The `tmp` delete is safe by construction.

### The grouped-node lifetime — attacked, since it's the new risk
This is the node #727's wrap never reached, so it's where a real lifetime bug would hide. I made the realistic mistake — retained the group keys in the scratch (`keyContext` → the build's own context) — rebuilt, and `native_groupagg` fails loudly: garbage keys throw the oracle into hash mismatches, `QUERY_ERROR`s and "node fires: no" across the WHERE/GROUP BY, added-column, after-deletes and signed-zero arms. So a lifetime error in the new wrap is loud, not silent — which is exactly what has to be established before removing the targeted guard.

### Attribution is back, and independence holds
On the merged tree the invariant arm (relabelled off `#717, #727` to the property) sits at excess 0, and its removal proof is now a single mechanism — the peer measured 180,843 B/rescan removing the wrap alone, where before it took reverting both. The #734 plain-scan arm reads 33 vs 33 throughout, so the reader fix is independent and the three rescan-memory fixes are separable.

### No behaviour change
All four grouped suites green on the merged tree — native_groupagg 72, native_groupagg_batch 69, parallel_vector_agg 29, native_groupagg_wide_cost 3 — plus vector_agg_rescan 23. Nothing moved; one thing owns the lifetime now.

### Sanitizer
ASAN+UBSAN clean over native_groupagg, native_groupagg_batch, parallel_vector_agg and vector_agg_rescan_memory under `pg18_san`, 0 vacated. And `buildCxt` is a delete rather than a reset, so ASAN actually sees a use-after-free of the grouped scratch here — a real second instrument alongside the assert-build inject, not just insurance.

CHANGELOG frames it honestly as a refactor with no behaviour or measurement change. Retained state proven to switch out of scratch, the new grouped lifetime proven loud, attribution and independence confirmed. Clean.

Issue #403, opened 2026-08-04
Opportunities from the ClickHouse VLDB 2024 paper, ranked against what we already have


OffgridwithJD:
Read *ClickHouse - Lightning Fast Analytics for Everyone*, Schulze, Schreiber, Yatsishin,
Dahimene and Milovidov, PVLDB 17(12):3731-3744, 2024.
[doi:10.14778/3685800.3685802](https://doi.org/10.14778/3685800.3685802)

Recording what is worth exploring, ranked, with what we already have next to it. Working
from the paper, which is a description of the design. Anything we build is ours.

Two of these I would break out into their own issues if wanted. The rest are recorded here
rather than filed, because filing a reading list as tickets is how a tracker rots.

## 1. Preimage rewriting for monotonic functions on the sort key

**The strongest item, and it lands on a shape we measured today.**

> Monotonicity traits allow to infer if a function produces sorted results on sorted input
> key value ranges. Second, some functions can compute the preimage of a given function
> result. This is used to replace comparisons of constants with function calls on the key
> columns by comparing the key column value with the preimage. For example,
> `toYear(k) = 2024` can be replaced by `k >= 2024-01-01 && k < 2025-01-01`.

We have no equivalent. `grep -riE 'monoton|preimage' src/` finds two comments and no code.

Why it matters here: a predicate on `date_trunc('minute', time)` cannot use the zone map on
`time` today, because the zone map holds `time` and the predicate is about a function of it.
Rewritten to a range on `time` it prunes with the machinery we already have. This is the
q1 to q3 shape, and #391 showed time pruning is doing the real work on those queries.

It is also principled rather than a special case, which is the objection I raised against
myself on #369. The technique is a function property plus a rewrite, not a hardcoded rule
about `date_trunc`.

## 2. A sparse sort-key index over chunk groups

> ClickHouse additionally stores, for every part, a mapping from the primary key column
> values of each granule's first row to the granule's id, i.e. the index is sparse. The
> resulting data structure is typically small enough to remain fully in-memory, e.g. only
> 1000 entries are required to index 8.1 million rows.

We evaluate zone maps to decide which chunk groups survive. That is linear in the number of
chunk groups. On today's bench table that is 667, which is nothing. At a hundred times the
size it is 66,700 min/max comparisons before a single row is read.

This pairs directly with today's clustering measurement. Clustering is what makes pruning
possible; a sparse index is what makes **locating** the surviving groups cheap. It only
[...]

Issue #369, opened 2026-08-04
The grouped parallel arm loses to the serial node because the Finalize is priced off estimate_num_groups on an expression key


jdatcmd:
## Summary

Split out of #349, which is otherwise complete. #366 made the grouped vectorized path
parallel-aware and it works — but the planner declines it on exactly the shapes where
it wins most, and the reason is a costing asymmetry rather than anything about the
node.

20M-row TSBS-shaped fixture, PG18 assert, 4 workers:

| shape | core | serial node | parallel arm |
|---|---:|---:|---:|
| G1 (1 metric, 12h window) | 6,144 ms | 4,555 ms | 4,587 ms *(serial chosen)* |
| G2 (10 metrics, 12h window) | 12,410 ms | 9,225 ms | 7,493 ms *(serial chosen)* |
| G3 (full scan, group by host) | 892 ms | 896 ms | **600 ms** *(parallel chosen)* |

**Forced onto G1 the parallel arm runs in 1,184 ms against the serial node's 4,555** —
a 3.9x win the planner is declining.

## Mechanism

The planner's own numbers on G1:

```
partialScan=15,042   ppath=16,292   gather=16,292   final=44,324   serial=20,042
dNumGroups=200,000   (actual groups ~8,000)
```

Our partial node under its Gather costs **16,292** against the serial node's **20,042**
— it wins on everything it actually does. The core `Finalize HashAggregate` on top adds
**28,031**, and that term is priced off `dNumGroups`.

For a `date_trunc(...)` grouping key `estimate_num_groups` cannot estimate distinctness
and returns a count near the input row count: 200,000 estimated against ~8,000 actual
here, and 2,000,000 against 48,000 on a larger fixture. The finalize is overpriced by
that ratio and loses the comparison on its own.

## Why it is structural, not a tuning miss

The serial node emits **finished values**, so it pays no finalize at all. #350
deliberately priced it per input row with no per-output-group term — an earlier version
charged per group and autoanalyze's group estimate flipped the plan choice, so the node
sometimes did not run.

So any two-phase plan pays a group-count-driven finalize and the serial node does not.
When the group estimate is inflated, the serial node wins **by construction**, whatever
the parallel arm actually costs.

## The mechanism is confirmed, not inferred

G3's grouping key is a plain column, where `estimate_num_groups` is accurate, and there
the parallel arm **is** chosen and **is** faster. Accurate estimate → chosen; inflated
estimate → declined. Same code, same box, same session.

[...]

PR #739, opened 2026-08-25
opt: rewrite date_trunc equality into a range so it can prune (#403 item 1)

jdatcmd:
Implements item 1 of #403, the item that issue names as the strongest and
recommends breaking out. From the ClickHouse VLDB 2024 paper's monotonicity
traits: a predicate on a *function* of the sort key cannot use that key's zone
map, but the preimage of the function is a range the zone map **can** use.

## Measured before building anything

500,000 rows clustered by time, 50 chunk groups:

| predicate | groups read | removed | rows |
| --- | ---: | ---: | ---: |
| `ts >= '2024-02-01' AND ts < '2024-02-02'` | 2 | 48 | 10000 |
| `date_trunc('day', ts) = '2024-02-01'` — before | **50** | 0 | 10000 |
| `date_trunc('day', ts) = '2024-02-01'` — after | **2** | 48 | 10000 |

Same answer, 25x fewer groups read, now matching the explicit range exactly.

The issue is old, so its premises were re-checked first: `monoton|preimage`
still finds only comments and no code, and the existing `date_trunc` handling in
`columnar_vector.c` is a group-**count** estimate for the grouped node —
unrelated machinery, though it showed the funcid-matching pattern.

## Being straight about the generality

Dispatched on funcid in the scan-key builder beside the IN-list rewrite, so
adding a second function means adding its preimage rather than editing the
general comparison path — the objection #369 raised against a special case.

But this inverts **exactly one function**, and the step comes from a small table
of unit names, because a month is not a fixed number of seconds and has to be an
interval applied by calendar arithmetic. The generality is in *where* the rewrite
happens, not yet in how many functions it knows. Unlisted units decline.

`timestamptz` is deliberately absent: `date_trunc(text, timestamptz)` truncates
in the session `TimeZone`, so the interval depends on a GUC that can change
after the key is frozen at executor start. The suite asserts the timestamptz
answer under **two different zones** rather than asserting an absence.

## Two of my own design claims were refuted by their removal proofs

Recorded because the comments now say what is true rather than what I assumed.

**"A non-truncated constant would admit rows the original excludes."** It would
not. The keys only prune, the executor re-applies the clause, and the derived
range is never *narrower* than the true matching set — the empty set is
contained in every range. Removing the check left the suite **green**. It is
defence in depth, kept because it is two lines and becomes load-bearing if these
keys are ever marked exact.

**"The conservative marking is load-bearing against the batch fold."** Not
observable today: `pgcolumnar_batch_type_ok` admits only int and float, so the
fold refuses a `timestamp` column on the **type** gate before exactness is
consulted. Marking the keys exact left the suite green too. The arm that names
the fold now records that it passes for a different reason than its title
suggests.

## Removal proof

Reverting the rewrite returns the work arm to 0 removed while every answer arm
stays green — which is the only thing that *can* move. A predicate rewrite that
prunes correctly and one that does not prune at all return identical rows, so
the pruning measurement is the only check that can see this change at all.

## Suites

`preimage_rewrite` 24/24, plus 15 green on PG17 covering every pruning path:
`pushdown_report`, `native_skip`, `native_zonemap`, `native_zonemap_narrow`,
`zonemap_cost`, `native_saop_pushdown`, `native_param_pushdown`,
`native_exact_selection`, `native_bloom`, `ungrouped_vector_agg`,
`native_groupagg_batch`, `vector_agg_rescan_memory`, `native_decode_gating`,
`differential`, `docs_style`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,21 @@ pinned at `1.0-dev` or `1.0-alpha`, each true until the next version shipped.
 
 ## [Unreleased]
 
+### Added
+
+- A predicate on `date_trunc(unit, ts) = constant` now drives chunk-group
+  skipping. A zone map holds `ts`, so a predicate about a function of `ts` could
+  exclude nothing and the scan read every group. It is rewritten to a range on
+  `ts`, which prunes with the machinery that already exists. Measured on 500,000
+  rows clustered by time in 50 chunk groups: the equivalent explicit range read
+  2 groups, the `date_trunc` form read all 50, and it now reads 2. The derived
+  keys are conservative, so the executor still applies the original clause. Only
+  the `timestamp` form is rewritten: `date_trunc` on `timestamptz` truncates in
+  the session time zone, so a key frozen at executor start could be wrong if the
+  zone changed. Units outside a known list, and a constant that is not itself
+  truncated, are declined rather than approximated. A new suite,
+  `preimage_rewrite`, measures the pruning and pins each declined shape. (#403)
+
 ### Fixed
 
 - The vectorized aggregates now bound their per-execution memory with one
--- a/design/ISSUE_403_PREIMAGE_REWRITE.md
+++ b/design/ISSUE_403_PREIMAGE_REWRITE.md
@@ -1,0 +1,83 @@
+# #403 item 1: preimage rewriting for monotonic functions
+
+## The opportunity, measured before building anything
+
+Fixture: 500,000 rows, one day per ~10,000 rows, clustered in time, 50 chunk
+groups, `stripe_row_limit => 10000`.
+
+| predicate | chunk groups read | removed | rows |
+| --- | ---: | ---: | ---: |
+| `ts >= '2024-02-01' AND ts < '2024-02-02'` | **2** | 48 | 10000 |
+| `date_trunc('day', ts) = '2024-02-01'` | **50** | 0 | 10000 |
+
+Same answer, 25x the groups read. The zone map holds `ts`; the predicate is
+about a function of `ts`, so nothing can exclude a group.
+
+Verified before starting, because the issue is old: `grep -riE 'monoton|preimage'
+finds three comments and no code. The existing `date_trunc` handling in
+`columnar_vector.c` is a group-COUNT estimate for the grouped node, unrelated,
+though it shows the funcid-matching pattern.
+
+## What ships
+
+`f(col) OP const` gains derived scan keys on `col`, where `f` is a function we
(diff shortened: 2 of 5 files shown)


OffgridwithJD (approved):
**APPROVE** — reviewed as ChronicallyJD. Verified on the merged tree (current `main` 43c7319 + this PR through 1cbfe05), pg17a assert + pg18_san.

Preimage-rewriting `date_trunc(unit, timestamp) = const` into a range on the column is a clean, well-scoped first cut: dispatched on funcid so the mechanism is a function property plus a rewrite, not a special case (the #369 objection), and it declines anything it cannot inarguably invert — declining costs a scan, never an answer.

### The win, reproduced
On the time-clustered fixture the rewrite prunes **48 of 50 chunk groups — identical to the equivalent explicit range** (reads 2, the full 25×), for the same rows. All oracle arms, both timestamptz zones, and every unit green; the general pushdown path is unaffected; ASAN+UBSAN clean over the pruning path.

### The two self-corrected claims — reproduced independently
You flagged two design claims your own removal proofs had refuted, and asked me to check the corrections rather than the code. Both hold:
- **The non-truncated-constant guard is defence in depth, not correctness.** I removed it, rebuilt, and the suite stays green — "a non-truncated constant returns no rows" included — because the keys only prune and the executor re-applies the clause, so an empty true set is contained in any derived range.
- **The conservative marking is not load-bearing against the fold today.** `pgcolumnar_batch_type_ok` admits only int2/4/8 and float4/8, so a timestamp column is refused on the type gate before exactness is consulted.

### Keep both — because the cliff is named
On your question of whether to remove them: keep both. Check 1 guards a specific future — the moment these keys are marked exact, a non-truncated constant makes the fold over-count with no recheck behind it, and #715 is exactly the incident where that path was reached from the other side. Defence in depth against a future you can name is worth two lines; defence in depth in general is not. That distinction is the whole of it.

### Two things this review turned up, both fixed
- The claim-1 correction had landed in the C comment and the design doc but not the test, which still taught the refuted framing in two places. Fixed in 2d25d96; all three artifacts now say the same thing, and the decline arms state what they can and cannot show.
- `run_all_versions.sh` registered `preimage_rewrite` in the wrong C-order slot (after `pg19_vacuum_options`; it belongs after `planner_choice_quality`), which failed `harness_selftest` on the full matrix — present since the original commit, so the PR had never been green. Fixed in 1cbfe05, verified through the runner's own `--list-suites`. I own half of this one: I ran the feature's suites and not the suite that gates the file the PR edits.

Non-test code is byte-identical across both follow-up commits, so the feature and sanitizer verification stand. Feature correct, corrections honest, registration fixed, full matrix green.

PR #742, opened 2026-08-25
ci: run the extension-upgrade guard, and stop the coverage runner failing on it (#741)

### LINKED ISSUE #740: The nightly coverage report has never captured any coverage (lcov capture produced nothing)
The `nightly deep gate` workflow has failed **every night since 2026-07-31**, the
night the coverage job was added. Every one of those runs ends the same way:

```
-- collect
FAIL  lcov capture produced nothing
```

`test/run_coverage.sh:112-114`. The job has never once succeeded, so the report has
never measured anything.

## It has been broken since it was introduced, not regressed into

- Last green nightly: run `30592054946`, 2026-07-30. Its job list is
  `suites (PG 15/16/17/18 x86_64)`, `suites (PG 18 aarch64)`, `sanitizer gate` —
  **no coverage job**. It was green because the job did not exist yet.
- `ca6c4da "Add a nightly coverage report, as a report and not a gate"` (#281) lands
  in the range `e560304..824d516`, i.e. between that last green run and the first red
  one on 2026-07-31 (run `30611698765`), which already reported
  `FAIL  lcov capture produced nothing`.

So this is not a regression to bisect. The job has produced `coverage.info` from zero
`.gcda` on all 25 of its runs.

## Mechanism

`nightly.yml:428-429` runs the script as root:

```yaml
sudo -E env "PATH=$PATH" \
  bash test/run_coverage.sh /usr/lib/postgresql/18/bin/pg_config
```

Under root, `test/lib.sh:149-152` runs the server as a different OS user and chowns
only the workdir:

```sh
if [ "$(id -u)" = "0" ]; then
        PGC_RUNPG=(runuser -u postgres --)
        chown -R postgres "$PGC_WORKDIR"
```

gcov writes `.gcda` beside the objects, at the absolute path recorded at compile
time — `$SRCDIR/src`. That directory belongs to the workspace user at mode 755, so
the `postgres` backend cannot create files in it. No `.gcda` is ever written, and
`lcov --capture` has nothing to find.

`GCOV_PREFIX` / `GCOV_PREFIX_STRIP` appear nowhere in the tree, so nothing redirects
the counters anywhere writable either.

## Proof (merged main 43ee2b7, PG 18 assert, container)

Built exactly as the script does (`COPT="--coverage"`), then ran one suite twice,
changing one thing between the arms.

Premises asserted first, so a zero cannot come from the wrong binary or a suite that
never ran:

- instrumentation present: `.gcno` count = **33**
- the installed `.so` is the instrumented one: built `d35931cd07497c0c6b4548c599364332`
  == installed `d35931cd07497c0c6b4548c599364332`
- `smoke` returned rc=0 `SMOKE TEST PASSED` in **both** arms

| arm | only difference | `.gcda` after the passing suite |
|---|---|---|
| CI shape | `src/` is `root 755` | **0** |
| control | `chmod 777 src/` | **33** |

One variable, an observable difference, and the control lands on exactly the 33
objects that were instrumented. The counters are lost to the directory permission,
nothing else.

## Second-order effect: the real verdict is never reached

`run_coverage.sh:112-114` exits 1 the moment capture comes back empty. Its actual
suite verdict is the last line of the file:

```sh
[ "$fail" = 0 ]     # line 142
```

That line has never executed. The run does print its suite tally one line above the
lcov error, which is the only reason the failing suite in #741 is visible at
all — and it reads as a coverage-tooling problem, so nobody looks at it.

The per-suite logs (`coverage/${s}.log`) are not uploaded either; the workflow uploads
only `coverage/html` and `coverage/coverage.info`, so the detail of any suite failure
in this job is discarded when the runner is torn down.

## Not prescribing the fix

Making `$SRCDIR/src` writable by the server user is what the control arm did and it
works, but I have not measured it end-to-end through `lcov --capture`/`genhtml`
(lcov is not installed on my box), so treat that as **untested**. `GCOV_PREFIX` with a
matching `lcov --directory` is the other obvious route and I have tested neither.
Whichever way it goes, the guard worth adding alongside it is the one this job already
has for suites — `run_coverage.sh:105-109` fails when `pass = 0` on exactly the
reasoning that a report built from nothing is not a report. The same reasoning applies
to capturing zero `.gcda`, which is the case that has actually been happening.

### LINKED ISSUE #741: The extension-upgrade guard never runs in CI; the coverage runner turns that into a permanent nightly red
`test/extension_upgrade.sh` is the guard for the break that is invisible until a user
upgrades. **In CI it has never verified an upgrade.** No workflow runs it successfully,
and the one place it does get invoked it fails every night, for an environment reason
that is counted as a product failure.

## 1. No workflow sets `PGC_RUN_UPGRADE`

`run_all_versions.sh` gates both the `pg_upgrade` pair matrix and `extension_upgrade`
behind `PGC_RUN_UPGRADE=1`. Grepping the tree for who sets it:

```
CHANGELOG.md        docs/limitations.md        docs/testing.md
test/run_all_versions.sh        test/native_upgrade_converge.sh
```

Documentation and the runner itself. **Neither `ci.yml` nor `nightly.yml` sets it.**
So the block at `run_all_versions.sh:855-900` — including its careful
`PASS`/`SKIP`/`FAIL` handling — never executes in CI, and `extension_upgrade` is
absent from every `suites (PG N)` job summary.

This is the shape the comment at `run_all_versions.sh:856-860` was written about:

> `test/extension_upgrade.sh` had pg_upgrade's exemption from the registration check
> without pg_upgrade's invocation, so nothing ran it (#396). A guard nobody runs is
> the failure mode #257 existed to close

The registration was fixed. The invocation still does not happen in CI.

## 2. The one path that does reach it invokes it in a way that cannot work

`run_coverage.sh` discovers every `test/*.sh` not named in `not_a_suite()`, which
includes `extension_upgrade`, and invokes it (line 90) with **no second argument and
no `PGC_UPGRADE_OLD_SRC`**:

```sh
bash "$SRCDIR/test/${s}.sh" "$PGC" >"$OUT/${s}.log" 2>&1
```

`extension_upgrade.sh:40` then falls back to a git ref:

```sh
OLD_SRC=${2:-${PGC_UPGRADE_OLD_SRC:-v1.0-alpha}}
```

The coverage job's checkout (`nightly.yml:322`) is a plain `actions/checkout@v4` with
no `fetch-tags` and no `fetch-depth`, so the working copy has no tags and the ref
cannot resolve. `extension_upgrade.sh:117-118` reports FAIL and exits 1.

`run_all_versions.sh:874-895` passes `$PGC_UPGRADE_OLD_SRC` through and maps exit 2 —
"the environment could not supply an old source" — to SKIP, explicitly so an
environment shortfall is not read as a product failure. `run_coverage.sh:92-100` maps
only `rc = 66` to SKIP, so every other non-zero, including this one, is counted as a
failed suite.

Result, every night since at least 2026-08-16:

```
-- suites: 212 passed, 1 failed ( extension_upgrade), 2 skipped ( native_repack pg19_vacuum_options)
```

## Proof, with a control

Faithful reproduction of the CI shape — a git checkout *with* `.git` (so the exit-2
"not a checkout" SKIP does not apply) but *without* tags:

```
is a git checkout: yes
tags present: []
v1.0-alpha resolves: NO
```

**ARM** — invoked exactly as `run_coverage.sh` does, no 2nd arg, no `PGC_UPGRADE_OLD_SRC`:

```
== extension_upgrade: PG18, old v1.0-alpha
  FAIL  v1.0-alpha is not present. Fetch tags, or pass an explicit ref or dir:
rc=1
```

Reproduces the nightly's `1 failed ( extension_upgrade)` exactly.

**CONTROL** — same tags-less tree, same suite, removing only the missing old source:

```
== extension_upgrade: PG18, old /srv/pgc_old_alpha
  old source: directory /srv/pgc_old_alpha
  old install: 1000 rows, extversion 1.0-dev
  ok    ALTER EXTENSION UPDATE
  PASS  existing rows still readable: 1000
  PASS  insert into an existing table: 1001
  PASS  new columnar table creatable: 3
  PASS  access method still bound: pgcolumnar
  PASS  maintenance function callable
  PASS  every C function's link name is namespaced
== extension_upgrade: PASS
rc=0
```

rc 1 -> 0 on that one change. **The upgrade path itself is sound; the red is entirely
the missing old source.** Worth stating explicitly, because "it's only a harness bug"
is itself a claim and a real upgrade break would look identical from the CI log.

## What is and is not covered

`test/native_upgrade_converge.sh` **does** run in the nightly and passes on PG15 and
PG18. It gates that the shipped scripts converge on `default_version = 1.0-alpha2`
(`native_upgrade_converge.sh:64-65,116-121`), so *script* convergence is covered and I
am not claiming otherwise.

What is not covered is what `extension_upgrade.sh` uniquely tests: a real previous
release **built and installed**, carrying user data, then upgraded in place — the
six `PASS` lines in the control above. Nothing in CI checks that today.

## Fix belongs to whoever owns CI, but the two halves are separable

Giving the coverage job an old source (`fetch-tags: true` on that checkout, or an
explicit `PGC_UPGRADE_OLD_SRC`) would turn this particular red green. It would not
address part 1 — the upgrade matrix is still not run anywhere. Conversely, teaching
`run_coverage.sh` the exit-2 SKIP contract that `run_all_versions.sh` already honours
would stop it miscounting an environment shortfall as a defect, but would then report
SKIP for a guard that still never runs. I have **not** tested either change.

Note this issue is only visible because the tally prints one line above
`FAIL lcov capture produced nothing` (#740) — the coverage job dies there before
reaching its real suite verdict at `run_coverage.sh:142`, and its per-suite logs are
never uploaded.

jdatcmd:
Closes #741.

`test/extension_upgrade.sh` guards the break that is invisible until a user
upgrades. In CI it had never verified an upgrade. Two separate defects, fixed
separately.

## 1. Nothing ran it

No workflow set `PGC_RUN_UPGRADE`, so the block at `run_all_versions.sh:855-900`
never executed and the suite was absent from every `suites (PG N)` job.

Added a nightly `upgrade-guard` job on PG18. Three details are deliberate:

- `fetch-depth: 0`, because `actions/checkout` fetches no tags by default and the
  suite's ref fallback then cannot resolve. That is the whole of defect 2's
  symptom.
- The previous release is asserted to resolve **before** any expensive setup. A
  tagless checkout must be a loud red, never a skip, since "the old source was
  not available" is exactly the state that made this guard look like it was
  running when it was not.
- The old source is materialised with `git archive` into a plain tree and passed
  explicitly, i.e. the suite's documented **directory** form. The suite runs
  under `sudo` (its `runpg` is an unconditional `runuser`), and a root
  `git clone` of a workspace owned by the runner user trips git's
  dubious-ownership refusal. Passing it explicitly also sets the suite's
  `EXPLICIT` flag, so its "not a git checkout" SKIP branch is unreachable from
  here. Exit 2 is a failure of this job rather than a waiver, for the same
  reason.

## 2. The one path that did reach it could not work

`run_coverage.sh` discovers every `test/*.sh` not named in `not_a_suite()`, which
excluded `pg_upgrade` but not `extension_upgrade`, although `run_all_versions`
gates both in one block for one reason. It ran with no old source, and since
`run_coverage` maps only `rc = 66` to SKIP, the environment shortfall was counted
as a failed suite. Hence `1 failed ( extension_upgrade)` every night.

Both upgrade suites are now excluded together. Deciding to run an opt-in guard is
`run_all_versions.sh`'s job, not the coverage runner's.

### The comment was already false

There are two copies of that list, and `run_coverage.sh`'s comment said it was
"Kept in step with harness_selftest.sh's list". It was not: the selftest copy in
`040` already named `extension_upgrade`. A comment claiming two things agree is
not a mechanism that makes them agree, so `test/selftest/220` now asserts it over
the whole population of test files (227 compared), in both directions.

The gated suite list is **derived** from the `PGC_RUN_UPGRADE` block rather than
naming the two suites, so a third one added there is covered the day it is added.

## Verified end to end, not by inspection

On PG17 in the container, one variable between the arms:

| arm | change | result |
| --- | --- | --- |
| ARM | tagless checkout, invoked as `run_coverage.sh` does | rc 1, `v1.0-alpha is not present` |
| CONTROL | same tree, old source materialised by `git archive` | rc 0 |

The ARM reproduces the nightly's failure exactly. The CONTROL upgrades 1.0-dev to
1.0-alpha2 with 7 PASS lines including the 149-object catalog convergence check,
so the upgrade path itself is sound today and the red really was the missing old
source.

## Removal proofs, each in isolation

| mutation | check that goes red |
| --- | --- |
| revert `not_a_suite` | "every PGC_RUN_UPGRADE-gated suite is excluded", got `[extension_upgrade]` |
| remove the workflow job | "CI runs the extension-upgrade guard somewhere" |
| mention the suite only in a **comment** in a workflow | still red, so a commented-out line cannot satisfy it |
| drift **either** `not_a_suite` copy | agreement check red, naming which copy drifted |

And the instrument is shown to say both things: `not_a_suite` is asserted to
REFUSE an ordinary suite, so its acceptance of these two means something.

`harness_selftest`: 119 checks, PASSED.

## What this does not do

- **It does not fix #740, and the nightly stays red until that lands.** Once
  `lcov` succeeds, `run_coverage` reaches its real verdict at line 142. The two
  are a pair, and neither alone reaches green.
- It does not teach `run_coverage.sh` the exit-2 SKIP contract that
  `run_all_versions.sh` honours. With both upgrade suites excluded, no discovered
  suite exits 2 today, so that would be untested handling for a case that cannot
  currently arise. Worth doing if a suite ever needs it.
- It does not run the `pg_upgrade` cross-major pair matrix, which is the other
  half of `PGC_RUN_UPGRADE` and a much larger job.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -531,7 +531,14 @@ jobs:
       # ran, so an environment shortfall is a failure of the job, not a waiver.
       - name: Run the extension-upgrade guard
         run: |
-          set -uo pipefail
+          # `set +e`, NOT `set -uo pipefail`. GitHub runs a `run:` step as
+          # `bash -e {0}`, and -uo pipefail does not cancel that -e: the step
+          # would abort at the suite line and rc, the annotation and the exit
+          # below would never execute. The gate still held -- a non-zero exit
+          # fails the job either way -- but the one case the comment singles out
+          # was the one case that could never announce itself. Only `set +e`
+          # restores the intent (#741 review).
+          set +e
           sudo -E env "PATH=$PATH" \
             bash test/extension_upgrade.sh \
               /usr/lib/postgresql/18/bin/pg_config /tmp/pgc_old_alpha
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -443,3 +443,100 @@ jobs:
           name: coverage-info
           path: coverage/coverage.info
           retention-days: 14
+
+  # The extension-upgrade guard (#382), which until #741 ran nowhere.
+  #
+  # What it catches is invisible to a build and to every other job here: each
+  # SQL-callable function records its C link name in pg_proc.prosrc at CREATE
+  # EXTENSION time, so renaming a link name leaves every EXISTING install
+  # pointing at a symbol the new library no longer exports. It compiles, it
+  # links, every suite passes on a fresh install, and every upgraded install is
+  # inert. Only a real previous release, carrying data, upgraded in place, shows
+  # it.
+  #
+  # Nightly rather than per-PR: it builds the extension twice, and the suite's
+  # own header calls it a second gate beside the matrix rather than part of it.
+  upgrade-guard:
+    name: extension upgrade guard (PG 18)
+    runs-on: ubuntu-latest
+    timeout-minutes: 45
+    if: github.event_name == 'workflow_dispatch' || github.repository == 'commandprompt/pgcolumnar'
+    # Named ONCE. Two steps consume it, and a tag spelled separately in each is
+    # the kind of duplication that goes stale in one place and not the other.
+    env:
+      PGC_OLD_TAG: v1.0-alpha
+    steps:
+      # fetch-depth: 0 for the tags. The coverage job's plain checkout is why
+      # this suite could not resolve its old source at all (#741): actions/checkout
+      # fetches no tags by default, and the suite's ref fallback then fails.
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      # Asserted, and asserted BEFORE the expensive setup, because "the old
+      # source was not available" is precisely the state that made this guard
+      # look like it was running when it was not. A tagless checkout must be a
+      # loud red here, never a skip.
+      - name: Confirm the previous release is present
+        run: |
+          set -euo pipefail
+          git rev-parse --verify -q "$PGC_OLD_TAG^{commit}"
+          echo "$PGC_OLD_TAG resolves to $(git rev-parse "$PGC_OLD_TAG^{commit}")"
+
+      # The DIRECTORY form, which the suite documents, rather than its git-ref
+      # form. The suite runs under sudo (its runpg is an unconditional
+      # `runuser -u postgres`), and a root `git clone` of a workspace owned by
+      # the runner user trips git's dubious-ownership refusal. `git archive` runs
+      # as the ordinary user and yields a plain tree with no .git, which is
+      # exactly the case the directory form exists for.
+      #
+      # Passing it explicitly also sets the suite's EXPLICIT flag, so its
+      # "not a git checkout" SKIP branch is unreachable from here. A guard that
+      # can quietly skip is the failure this job exists to end.
+      - name: Materialise the previous release as a plain tree
+        run: |
+          set -euo pipefail
+          rm -rf /tmp/pgc_old_alpha
+          mkdir -p /tmp/pgc_old_alpha
+          git archive "$PGC_OLD_TAG" | tar -x -C /tmp/pgc_old_alpha
+          test -f /tmp/pgc_old_alpha/Makefile
+          echo "old source: $(find /tmp/pgc_old_alpha -maxdepth 1 | wc -l) top-level entries"
+
+      - name: Add the PGDG repository and refresh the package lists
+        run: |
+          set -euo pipefail
+          sudo install -d /usr/share/postgresql-common/pgdg
+          sudo curl -fsSL --connect-timeout 15 --max-time 120 \
+            --retry 5 --retry-delay 5 --retry-all-errors \
+            -o /usr/share/postgresql-common/pgdg/apt.postgresql.org.asc \
+            https://www.postgresql.org/media/keys/ACCC4CF8.asc
+          echo "deb [signed-by=/usr/share/postgresql-common/pgdg/apt.postgresql.org.asc] \
+            https://apt.postgresql.org/pub/repos/apt $(lsb_release -cs)-pgdg main" \
+            | sudo tee /etc/apt/sources.list.d/pgdg.list >/dev/null
+          sudo timeout 300 apt-get -o Acquire::Retries=5 update
+
+      - name: Install PostgreSQL 18 and the codec headers
+        run: |
+          set -euo pipefail
+          sudo timeout 600 apt-get -o Acquire::Retries=5 install -y --no-install-recommends \
+            postgresql-18 postgresql-server-dev-18 \
+            liblz4-dev libzstd-dev zlib1g-dev
+
+      - name: Stop the packaged cluster
+        run: sudo systemctl stop postgresql || true
+
+      # Exit 2 is the suite's "the environment could not supply an old source".
+      # run_all_versions.sh maps that to SKIP, which is right for a developer's
+      # box and wrong here: this job's entire purpose is that the guard actually
+      # ran, so an environment shortfall is a failure of the job, not a waiver.
+      - name: Run the extension-upgrade guard
+        run: |
+          set -uo pipefail
+          sudo -E env "PATH=$PATH" \
+            bash test/extension_upgrade.sh \
+              /usr/lib/postgresql/18/bin/pg_config /tmp/pgc_old_alpha
+          rc=$?
+          if [ "$rc" = 2 ]; then
+            echo "::error::the guard skipped for want of an old source; in this job that is a failure"
+          fi
+          exit "$rc"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,22 @@ pinned at `1.0-dev` or `1.0-alpha`, each true until the next version shipped.
 
 ### Fixed
 
+- The extension-upgrade guard now runs in CI, and had never verified an upgrade
+  before. `test/extension_upgrade.sh` catches a break that is invisible until a
+  user upgrades: a renamed C link name leaves every existing install pointing at
+  a symbol the new library no longer exports. No workflow set `PGC_RUN_UPGRADE`,
+  so it ran nowhere, and the one runner that did reach it, the coverage report,
+  discovered it through `not_a_suite()` and invoked it with no old source, then
+  counted the resulting environment shortfall as a failed suite. A nightly job
+  now builds the previous release on PostgreSQL 18, loads data into it and
+  upgrades it in place, and both upgrade suites are excluded from the coverage
+  runner together. Running it also exposed a defect in the suite itself: every
+  connection used `psql -h /tmp`, while the socket directory is decided by how
+  PostgreSQL was built, `/tmp` for a source build and `/var/run/postgresql` for
+  the Debian and PGDG packages, so the suite had never been runnable against a
+  packaged server and reported the connection failure as "old install did not
+  store rows". It now states the directory it connects to. (#741)
+
 - The vectorized aggregates now bound their per-execution memory with one
   mechanism instead of two. The scan-key context added for #717 covered the
   scan keys; the scratch context added for #727 covered the whole scan and so
--- a/docs/testing.md
+++ b/docs/testing.md
@@ -303,7 +303,9 @@ behaviour.
 **Nightly at 06:00 UTC** (`.github/workflows/nightly.yml`): the full packaged
 suite matrix across 15 to 18 on x86_64, and the current major on aarch64. The
 nightly run also includes the ASAN and UBSAN sanitizer gate, against an
-instrumented PostgreSQL, and the coverage report. The sanitizer build stays in a cache, because it takes longer to build
+instrumented PostgreSQL, and the coverage report. It also runs the
+extension-upgrade guard. That guard builds the previous release on PostgreSQL
+18, loads data into it, and upgrades it in place. The sanitizer build stays in a cache, because it takes longer to build
 than to run the suites against it.
 
 The aarch64 run executes the suites rather than only building them. Misaligned
--- a/test/extension_upgrade.sh
+++ b/test/extension_upgrade.sh
@@ -136,6 +136,22 @@ runpg "$BINDIR/initdb" -D "$DATA" --locale=C -U postgres >/dev/null 2>&1 \
 {
 	echo "shared_preload_libraries='pgcolumnar'"
 	echo "port=$PORT"
+	# Every connection below is `psql -h /tmp`, and where the socket actually
+	# lands is a property of how PostgreSQL was BUILT, not of this suite: a
+	# source build defaults to /tmp, while the Debian and PGDG packages compile
+	# in /var/run/postgresql. So against a packaged server every psql here failed
+	# with `connection to server on socket "/tmp/.s.PGSQL.NNNN" failed`, and the
+	# suite reported "old install did not store rows" -- a connection fault
+	# wearing the costume of a product failure.
+	#
+	# It went unseen because the two never met: developers run this against
+	# source builds, and #741 is the issue that CI had never run it at all. The
+	# first CI run after wiring it up failed here.
+	#
+	# Stated rather than inherited, so -h /tmp is true by construction on either
+	# kind of build. lib.sh sidesteps the whole question by connecting over TCP;
+	# this suite carries its own harness and connects by socket.
+	echo "unix_socket_directories='/tmp'"
 } >> "$DATA/postgresql.conf"
 runpg "$BINDIR/pg_ctl" -D "$DATA" -l "$LOG" -w start >/dev/null 2>&1 \
 	|| { echo "FAIL  start"; tail -10 "$LOG"; exit 1; }
--- a/test/run_coverage.sh
+++ b/test/run_coverage.sh
@@ -54,10 +54,25 @@ make -C "$SRCDIR" PG_CONFIG="$PGC" COPT="--coverage" install >/dev/null 2>&1 || 
 
 # Every suite except the drivers, the libraries, and the ones that build their
 # own server or need two majors. Kept in step with harness_selftest.sh's list.
+#
+# The two upgrade suites are excluded TOGETHER, and that pairing is the point
+# (#741). run_all_versions.sh gates both behind PGC_RUN_UPGRADE, in one block and
+# for one reason: they build a second copy of the extension and need an old
+# source this runner has no way to supply. pg_upgrade was excluded here when it
+# was written and extension_upgrade was not, so this runner discovered it, ran it
+# with no old source, and its ref fallback could not resolve in a tagless CI
+# checkout. run_coverage maps only rc 66 to SKIP, so that environment shortfall
+# was counted as a failed suite and the nightly read "1 failed
+# ( extension_upgrade)" every night while nothing was wrong with the product.
+#
+# Deciding to run an opt-in guard is run_all_versions.sh's job, not this one's.
+# test/selftest/220 derives the gated list from that block and asserts every name
+# in it is refused here, so a third upgrade suite cannot repeat this.
 not_a_suite() {
 	case "$1" in
 		lib|portlib|run_all_versions|build_all_versions|devloop|rebuild) return 0 ;;
-		native_scale|build_san|run_san|run_coverage|pg_upgrade) return 0 ;;
+		native_scale|build_san|run_san|run_coverage) return 0 ;;
+		pg_upgrade|extension_upgrade) return 0 ;;
 		*) return 1 ;;
 	esac
 }
--- a/test/selftest/220-an-opt-in-upgrade-guard-must.sh
+++ b/test/selftest/220-an-opt-in-upgrade-guard-must.sh
@@ -1,0 +1,115 @@
+# ---- an opt-in upgrade guard must run somewhere, and not in the wrong runner --
+#
+# #741. test/extension_upgrade.sh is the guard for the break that is invisible
+# until a user upgrades, and in CI it had never verified an upgrade. Two separate
+# defects, which is why there are two groups of checks here.
+#
+# 1. run_all_versions.sh gates both upgrade suites behind PGC_RUN_UPGRADE, and
+#    run_coverage.sh's not_a_suite() excluded pg_upgrade but not
+#    extension_upgrade. So the coverage runner discovered it and ran it with no
+#    old source, its ref fallback could not resolve in a tagless CI checkout, and
+#    run_coverage counted the environment shortfall as a failed suite. The
+#    nightly read "1 failed ( extension_upgrade)" every night.
+#
+# 2. Nothing in .github/workflows set PGC_RUN_UPGRADE or invoked the suite, so
+#    the guard never actually ran anywhere. That is the failure mode #257 exists
+#    to close, and the comment at run_all_versions.sh:856-860 is about the last
+#    time it happened to this same suite (#396).
+#
+# The opt-in suites are DERIVED from run_all_versions.sh rather than named here.
+# A third one added to that block is then covered by this check on the day it is
+# added, which a hand-maintained list of two names cannot do.
+
+_upg_block() {	# the suites run_all_versions invokes only under PGC_RUN_UPGRADE
+	awk '/if \[ "\$\{PGC_RUN_UPGRADE:-0\}" = 1 \]; then/,/^fi$/' \
+		"$TESTDIR/run_all_versions.sh" |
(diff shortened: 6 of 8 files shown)


OffgridwithJD:
Reviewed as ChronicallyJD on the **merged tree** (real 3-way merge of `up/main` 43ee2b7
+ `pr742` 77edf5e, merges CLEAN), pg17a assert, container. This closes the issue I filed,
so I verified the claims rather than reading them.

## Every claim in the description reproduces

I re-ran all four removal proofs independently, each in isolation, restoring between:

| mutation | result |
| --- | --- |
| revert `not_a_suite` | RED — `got [extension_upgrade] want []`, plus the agreement check naming `coverage=run,selftest=excl` |
| remove the workflow job | RED — `CI runs the extension-upgrade guard somewhere: got [no]` |
| comment out the invocation only | **RED** — a mention genuinely does not satisfy it |
| drift the selftest's `not_a_suite` copy | RED — `coverage=excl,selftest=run`, naming which copy drifted |
| restored | GREEN, 119 checks PASSED |

The headline numbers are also as stated: `harness_selftest` **119 checks PASSED**, the
agreement check compares **227** test files with 0 disagreements, and the derived list
really does come out as `extension_upgrade pg_upgrade` — so the derivation is doing the
work, not a hand-written pair.

The `^[^#]*` guard on the CI-invocation check deserves specific credit: I tried to defeat
it with a commented-out invocation and it stayed red, which is the failure mode that
would have made this whole guard theatre.

## Finding 1 — the `rc = 2` branch is unreachable (low; the gate still holds)

In `Run the extension-upgrade guard`:

```sh
set -uo pipefail
sudo -E env "PATH=$PATH" bash test/extension_upgrade.sh ... 
rc=$?
if [ "$rc" = 2 ]; then echo "::error::..."; fi
exit "$rc"
```

GitHub runs a `run:` step as `bash -e {0}`. **`set -uo pipefail` does not cancel that
`-e`** — only `set +e` would. So on any non-zero exit the step aborts at the suite line,
and `rc=$?`, the `::error::` annotation and `exit "$rc"` never execute.

Proven, with a control:

```
$ bash -e -c 'set -uo pipefail; false; rc=$?; echo "REACHED rc=$rc"; exit "$rc"'
(no output)                     step exit=1
$ bash -e -c 'set +e;           false; rc=$?; echo "REACHED rc=$rc"; exit "$rc"'
REACHED rc=1                    step exit=1
```

**The safety property you describe still holds** — exit 2 aborts the step, the job fails,
and it is not a waiver. What is lost is only the diagnostic: the one case the comment
singles out is the one case that can never announce itself. `set +e` on that line restores
the intent. I flag it because the branch is written as deliberate, so a later reader will
believe it runs.

## Finding 2 — the PR carries no CHANGELOG entry

`.github/workflows/nightly.yml`, `test/run_coverage.sh`, `test/selftest/220-*.sh` — and
nothing under `CHANGELOG.md` or `docs/`. jd's standing rule is that no PR ships without
its CHANGELOG and docs in the same PR, and the four most recent merges (#736, #737, #738,
#739) each carried a CHANGELOG entry. `docs_style` does not check for this, so CI will not
catch it.

## Finding 3 — this PR makes a sentence in `docs/testing.md` false

`docs/testing.md:303-306` enumerates what the nightly contains:

> **Nightly at 06:00 UTC** (`.github/workflows/nightly.yml`): the full packaged suite
> matrix across 15 to 18 on x86_64, and the current major on aarch64. The nightly run also
> includes the ASAN and UBSAN sanitizer gate, against an instrumented PostgreSQL, and the
> coverage report.

This PR adds a **fourth** nightly job and leaves that enumeration untouched, so it
under-describes the nightly from the moment it merges. It is an enumeration, so it goes
stale by addition — the same way #671's sentences did.

## One operational point, which I think is the important one

This lands the guard into a workflow that **has been red every night for 25 nights** and
that nobody was reading — which is the exact condition that let #741 hide in the first
place. You say so plainly under "What this does not do", and I agree with the analysis.
The consequence worth making explicit: while the nightly is red for #740, this job
flipping pass -> fail produces **no new observable signal at the workflow level**. The
guard runs, but its verdict still arrives inside a chronic red.

That is an argument for sequencing #740 first or landing the two together, not against
this change.

## Verdict

Not approving yet, on Finding 2 (the CHANGELOG rule is jd's, not mine to waive) and
because CI is still in flight. Findings 1 and 3 are small and I would not block on either
alone. The mechanism, the selftest and every removal proof are sound, and I could not
break the instrument where I tried. Ping me when the docs are on and CI is green and I
will re-check and approve.

I did not review this against #740, which remains open and unaddressed by design.

jdatcmd:
## The first real CI run of this guard failed, and found a third defect

`nightly.yml` does not run on pull requests, so wiring up a nightly job and
merging on a local reproduction would have been the same "guard nobody ran"
failure this PR is about. I dispatched the workflow on the branch instead
(run `32856560045`). The job failed.

Steps 1-6 passed, including the tag resolution and `git archive`. The suite
itself failed:

```
old install: psql: error: connection to server on socket "/tmp/.s.PGSQL.31345"
failed: No such file or directory
FAIL  old install did not store rows
```

### Why, and why nobody could have seen it

Every connection in `extension_upgrade.sh` is `psql -h /tmp`, but where the
socket lands is a property of how PostgreSQL was **built**, not of the suite:

```
source-built pg17     compiled-in default  /tmp
packaged (PGDG) pg18  compiled-in default  /var/run/postgresql
```

So the suite has **never been runnable against a packaged PostgreSQL**, which is
what every workflow here uses. Developers run it against source builds, and #741
is precisely the issue that CI had never run it at all. The two never met until
this branch made them.

The failure also arrives dressed as a product failure: psql returns an error
string, the row count is not 1000, and the suite says "old install did not store
rows". Read from a CI log, that is indistinguishable from a real upgrade break,
which is the same trap #741's own control arm was written to avoid.

`lib.sh` sidesteps the whole question by connecting over TCP. This suite carries
its own harness, so it now states the directory instead of inheriting it.

### Reproduced and fixed with one variable

| arm | result |
| --- | --- |
| packaged PG18, before | rc 1, connection to socket `/tmp/...` failed |
| packaged PG18, after | rc 0, 7 PASS + the 149-object convergence check |
| source PG17, after | rc 0, the case that already worked, unbroken |

`test/selftest/230` pins the class: every suite connecting by a socket **path**
must set `unix_socket_directories`. Both premises asserted, including that a TCP
suite is NOT counted, so "they all set it" cannot be true by matching nothing.
Removal proof: delete the line and 230 goes red naming `extension_upgrade`.

`harness_selftest`: 122 checks, PASSED.

Re-dispatched as run `32857310654`; I will post the result rather than assume it.

### Worth noting for the review

This is the second defect this PR found by running something that had never run,
and it is the more interesting one: the suite was green everywhere it had ever
been executed and broken everywhere it was about to be. It is also a point in
favour of the ordering I argued on #741 -- a guard that never runs is not
neutral, it accumulates rot that only shows up the day you rely on it.

OffgridwithJD:
Re-reviewed at `ee1f983` on the merged tree. The socket finding is real, I reproduced it
independently, and the guard now passes in CI. Holding approval on the docs only.

## The third defect, verified independently rather than taken on report

I checked your account against the job log of run `32856560045` first: it is accurate
verbatim, down to `connection to server on socket "/tmp/.s.PGSQL.30085" failed`, and
`extension upgrade guard (PG 18): failure` is the only failing job in that run.

Then I proved the mechanism directly, both sides, on this box:

| build | compiled-in `DEFAULT_PGSOCKET_DIR` |
| --- | --- |
| source (`pg17a`, `pg18a`) | `/tmp` |
| packaged Ubuntu 18.4 | `/var/run/postgresql` |

It turns out this container has a packaged PostgreSQL 18.4 as well as the source builds, so
I could run the actual failing configuration rather than wait on CI. One variable:

| arm, packaged PG 18.4 | result |
| --- | --- |
| fix removed | rc 1, `socket "/tmp/.s.PGSQL.30883" failed`, `FAIL old install did not store rows` |
| fix present | **rc 0**, `ALTER EXTENSION UPDATE -> 1.0-alpha2`, 7 PASS + the 149-object convergence check |

So the defect and the fix both reproduce off your CI. Your framing is right that the failure
arrives dressed as a product failure -- my ARM prints "old install did not store rows",
which read cold is indistinguishable from a real upgrade break.

One correction to my own earlier work, since it is the same trap: in my #741 control I saw
`ALTER EXTENSION UPDATE -> 1.0-alpha` and declined to report it as a convergence gap,
calling it an artefact of my harness. The packaged run above lands on `1.0-alpha2`, which
confirms that call and identifies the cause as my own install ordering.

Also verified on the merged tree: `harness_selftest` **122 checks PASSED** as claimed; the
230 removal proof reddens naming `extension_upgrade`; and the source-build case is unbroken
(rc 0, 149 objects), so the fix is a no-op where it already worked.

And the decisive one: in your re-dispatch `32857310654`, **`extension upgrade guard
(PG 18): success`**. The guard has now run in CI, which is the thing #741 said had never
happened.

## New finding -- 230's discriminator misses a spelling that exists in the tree

`grep -qE '\-h /'` catches `-h /tmp` but not `-h "$VAR"` where the variable holds a path.
That spelling is in the tree today:

- `concurrency` connects `-h "$WORKDIR"` and **is not matched** by the discriminator. It
  happens to comply -- it sets `unix_socket_directories='$WORKDIR'` at line 150 -- so there
  is no live bug. But delete that line and 230 stays green.

I checked the other two candidates and they are not socket users, so I am not reporting
them: `fuzz_parquet`'s `-h "$PGC_WORKDIR"` is inside a comment describing what an earlier
version did, and `isolation.sh` is `PGHOST=127.0.0.1`, i.e. TCP.

So the population 230 enforces over is exactly one suite while a second socket-connecting
suite is invisible to it. The check's own premise ("at least one suite connects by a socket
path, so this is not vacuous") passes on that one and cannot notice. Widening the
discriminator to the variable form would put `concurrency` in the population, where it
would pass today and be protected tomorrow. Not a blocker -- nothing is broken now.

## Still outstanding from my first review

None of these are addressed at `ee1f983`, which only touched `extension_upgrade.sh` and
`selftest/230`:

1. **`rc = 2` is still unreachable.** `set -uo pipefail` does not cancel the `-e` GitHub
   passes as `bash -e {0}`. Worth noting the failing run does **not** discriminate here --
   the suite exited 1, so that branch would not have fired either way; my bash control
   remains the evidence. `set +e` on that line restores the intent.
2. **No CHANGELOG entry.**
3. **`docs/testing.md:303-306` still enumerates the nightly as matrix + sanitizer +
   coverage.** This PR now adds a fourth job that has demonstrably passed, so that sentence
   is about to be wrong about a job that exists and works.

## Verdict

Everything technical checks out, and the "run it before merging it" call was right -- it
found a defect that a local reproduction could not have. CI is now **12/12 green,
mergeState CLEAN**.

I am not approving only because of finding 2. The CHANGELOG rule is jd's and not mine to
waive, and finding 3 is a one-line consequence of this PR that will otherwise go stale
immediately. Add those two and I will approve on sight -- I have already verified
everything else.

Findings 1 and the 230 discriminator I would take or leave; neither blocks.

jdatcmd:
All four findings addressed at `8aa051c`. I verified each one before fixing it
rather than taking it on report, and one of them turned out to be bigger than
reported.

## Finding 1, `rc = 2` unreachable

Confirmed with your control on this box:

```
$ bash -e -c 'set -uo pipefail; false; rc=$?; echo "REACHED rc=$rc"; exit "$rc"'
(nothing)          exit=1
$ bash -e -c 'set +e;           false; rc=$?; echo "REACHED rc=$rc"; exit "$rc"'
REACHED rc=1       exit=1
```

You are right that the gate always held and only the diagnostic was lost. Now
`set +e`, with the reason in the step so a later reader does not re-derive it.

## Finding 2, CHANGELOG

Added under `[Unreleased]` / `Fixed`, covering all three defects including the
socket one.

## Finding 3, the nightly enumeration

Sentence updated. `docs_style` then failed it at **39 words**, over the STE
limit, so it is split into three; that suite now passes (9 checks).

I did not want to only correct the sentence, because your framing was that it
goes stale **by addition**. `test/selftest/240` now derives the job list from
`nightly.yml` and requires each key to be named in the nightly paragraph.

**Its first version was too loose, and its own removal proof is what said so.**
I reverted the sentence expecting red and got GREEN: the check searched the whole
file for the key's first word, and "upgrade" already appears eleven times there
(`pg_upgrade`, `PGC_RUN_UPGRADE`). It caught an invented fifth job and missed the
case it was written for. Scoping to the paragraph and matching the whole key
fixed it:

| mutation | result |
| --- | --- |
| revert the docs sentence | RED, `got [upgrade-guard]` |
| add an undocumented fifth job | RED, `got [zzdiagnostics]` |

Worth recording because it is your own point about instruments: the proof caught
it, review would not have.

## Finding 4, 230's discriminator, and it was worse than reported

You found `concurrency`. Widening to a path **or a quoted variable** on
non-comment lines takes the population from one suite to **four**:

```
-- suites connecting by socket path: concurrency extension_upgrade unique_conc update_conc
```

`unique_conc` and `update_conc` connect the same way. The spelling that actually
dominates the tree is `-h '$WORKDIR'` with **single** quotes, which is why both
your `-h "$` search and my `-h /` search each saw only part of it. All four
already comply, so nothing was broken, but three of them were unprotected.

Comment lines are excluded deliberately, exactly for the `fuzz_parquet` case you
identified: it names the old socket spelling in a comment and connects over TCP.
Matching it would have put a TCP suite in the population and reported a defect
that does not exist.

Both directions proved:

| mutation | result |
| --- | --- |
| delete `concurrency`'s `unix_socket_directories` line (your case) | RED, `got [concurrency]` |
| make `lib.sh` connect by socket path | the "no TCP suite is counted" premise fires |

The TCP premise is now three named suites rather than one, so the discriminator
cannot drift loose unnoticed.

## Totals

`harness_selftest` **126 checks PASSED**, `docs_style` **9 checks PASSED**.

## On your operational point

Agreed, and I said the same on #741: while the nightly is red for #740, this
job's verdict arrives inside a chronic red and produces no new workflow-level
signal. I would land them together or #740 first. That is a sequencing decision
rather than a change to this PR, and I have not touched #740, which is yours.

Thanks for the `1.0-alpha` / `1.0-alpha2` correction on your own earlier control.
That is the same trap in the same suite, and it is worth having on the record.

OffgridwithJD (approved):
**APPROVED** at `8aa051c`. All four findings addressed, and I verified each fix on the
merged tree rather than reading the diff.

| check | result |
| --- | --- |
| `harness_selftest` (merged tree, pg17a assert) | **126 PASSED** |
| `docs_style` | PASSED |
| CI | **12/12 SUCCESS**, mergeState CLEAN |

Removal proofs re-run independently, each in isolation and restored between:

| mutation | result |
| --- | --- |
| revert the docs sentence naming the guard | RED, `got [upgrade-guard]` |
| remove the upgrade-guard job | RED via 220, and 240 correctly drops to 3 derived jobs rather than demanding docs for a job that no longer exists |
| drop `unix_socket_directories` from extension_upgrade | RED, `got [extension_upgrade]` |
| restored | GREEN |

The CHANGELOG entry is accurate on the point I measured myself: `/tmp` for a source build,
`/var/run/postgresql` for the packages.

## Two corrections to my own review

**My finding 4 under-counted.** I reported `concurrency` as the one suite escaping 230's
discriminator. Widening to a path *or a quoted variable* finds **four** --
`concurrency`, `extension_upgrade`, `unique_conc`, `update_conc` -- because I matched only
the double-quoted form and the tree's dominant spelling is `-h '$WORKDIR'`. All four set
`unix_socket_directories`, so there was never a live bug, but your population is right and
mine was not. The printed population beside the claim is what makes that visible, which is
the reason to print it.

**Part of one of my removal proofs was my own artifact.** When I reverted the docs
sentence I left a blank line mid-paragraph, which truncated 240's paragraph range early, so
my run reddened `[suites upgrade-guard]`. The `suites` half was my mutation, not the check.
The `upgrade-guard` half is the real proof and it stands.

## On 240

Recording agreement with your note: the first version passing its own removal proof is the
interesting result, not the fix. "upgrade" appearing eleven times in that file for
unrelated reasons is exactly the too-loose failure that reads as success, and scoping to
the paragraph plus matching the whole job key is the right shape. Deriving the job list
from `nightly.yml` means a fifth job is covered the day it is added, which is what makes
this different from correcting the sentence.

## Unchanged and disclosed

This does not fix #740, so the nightly stays red and this guard's verdict still arrives
inside that red. Your PR body says so. Once #740 lands, `run_coverage` reaches its real
verdict at line 142 and the workflow-level signal starts meaning something again.

Approving as OffgridwithJD (formerly ChronicallyJD -- same account, renamed today).
Merging is yours; I merge nothing.

PR #743, opened 2026-08-25
docs(design): measure what locating the surviving chunk groups costs (#403 item 2)

jdatcmd:
Measurement only. No code changed, nothing filed by this PR.

#403 recommends breaking out item 2, a sparse sort-key index over chunk groups.
My own comment on that issue refused to file it and set the gate:

> "667 today, 66,700 at a hundred times the size" is an extrapolation. Nobody has
> measured what zone-map evaluation actually costs as a fraction of a scan at
> either size. That is a cheap measurement and it should come before an issue, or
> the issue starts with a number nobody checked.

This is that measurement, written down with the method, the premises, and the
instrument bugs found on the way.

## Result

| | instructions | buffers |
| --- | ---: | ---: |
| per zone-map probe | 25,185 | 6.00 |
| per group, not the probe | 5,397 | 0.09 |

Share of a pruning query's execution spent locating the surviving group:

| chunk groups | rows | share |
| ---: | ---: | ---: |
| 20 | 200,000 | 6.0% |
| 160 | 1,600,000 | 32.4% |
| 640 | 6,400,000 | **65.8%** |

640 was measured, not projected. The bench table #403 cites has 667, and the
extrapolation was the entire objection.

## What changes the answer

#403 describes this as "66,700 min/max comparisons". The comparisons are nearly
free. **82% of the per-group cost is that reaching them opens a relation, takes a
lock, and resolves two names, per group, per predicate column.** The read path
has no analogue of `PgColumnarBeginMetadataFlush`, which exists and is called
only from `columnar_write_state.c`.

So item 2 clears the threshold I wrote before taking numbers, but the first move
is to make the existing catalog read cheap, worth **2.18x** at 640 groups with no
new on-disk structure. A sparse index is step two, gated on re-measuring after
that, at which point it would be attacking a 25% residue rather than 66%.

## Why the numbers are believable

- Premises asserted on every arm: group count read from `pgcolumnar.row_group`
  and agreeing with EXPLAIN, `Chunk Groups Removed by Filter` == N-1, the node is
  the columnar scan, both pruning arms return the same 10,000 rows.
- The arithmetic was made falsifiable first: a second predicate column must
  double the slope. It does, 6.09 to 12.09 buffers per group.
- Two independent instruments agree on the structure.
- Least-squares residuals within ±0.27% for N = 80..640.
- Sanity check against a known quantity: the intercept works out to 1,017
  instructions per row, beside the ~920 already measured for that path.

## Instrument bugs found and fixed before believing anything

Recorded in the doc because each produced plausible output rather than an error:
summing per-node buffer counts double-counts (they are inclusive of children); a
perf window guard that fired on all six arms was a bug in the guard, not six bad
runs; `$3 == "instructions"` never matches because perf names the event per PMU;
and the first full-scan denominator used a qual the zone map answers wholesale,
so it was not a scan at all.

### DIFF
--- a/design/ISSUE_403_ITEM2_LOCATE_COST.md
+++ b/design/ISSUE_403_ITEM2_LOCATE_COST.md
@@ -1,0 +1,283 @@
+# #403 item 2: what does locating the surviving chunk groups actually cost?
+
+Status: **measurement in progress**. No issue filed, no code changed.
+
+## Why this document exists
+
+#403 recommends breaking out item 2, a sparse sort-key index over chunk groups.
+My own comment on that issue refused to file it, and set a gate:
+
+> "667 today, 66,700 at a hundred times the size" is an extrapolation. Nobody has
+> measured what zone-map evaluation actually costs as a fraction of a scan at
+> either size. That is a cheap measurement and it should come before an issue, or
+> the issue starts with a number nobody checked.
+
+This is that measurement. It decides whether item 2 is worth an issue, and --
+more usefully -- whether a sparse index is even the right answer, or whether the
+linear cost is dominated by something cheaper to fix.
+
+## What the code actually does (read, not assumed)
+
+`pgcolumnar_native_load_group` (`src/columnar_reader.c:2083`) walks the row-group
+list and for each candidate calls `pgcolumnar_native_group_can_match`
+(`src/columnar_reader.c:1347`), which for each pushed-down predicate calls
+`PgColumnarReadZoneMapForColumn` (`src/columnar_metadata.c:2633`).
+
+That last function is **not** a min/max comparison against an in-memory array.
+Per call it does:
+
+- `open_columnar_table("zone_map", AccessShareLock)` -- `pgcolumnar_schema_oid()`,
+  `get_relname_relid`, `table_open` with a lock. The `md_flush` reuse cache is
+  write-path only (`md_flush.active`), so the read path never hits it.
+- `pgcolumnar_index_oid("zone_map_pkey")` -- a second `get_relname_relid`.
+- `systable_beginscan` / `getnext` / `endscan` on the catalog index.
+- `table_close`.
+
+So the per-group cost is a catalog index probe plus two name lookups and a
+relation open, **per predicate column**, not a pair of integer compares. #403's
+framing ("66,700 min/max comparisons") therefore *understates* the cost, which is
+the opposite of the error I was guarding against when I refused to file it.
+
+Caching is `zoneLookedUp[]`, which is per group, not per scan.
+
+## The claim under test
+
+> Locating the surviving chunk groups is linear in the total number of chunk
+> groups, and at scale that cost is a material fraction of a scan.
+
+## Instrument
+
+Primary: **buffers**, from `EXPLAIN (ANALYZE, BUFFERS)` on the columnar scan
+node. Catalog index probes read buffers, `pgBufferUsage` is accumulated per node,
+and the count is deterministic -- unlike wall clock on this host, which
+[[perf-measurement-on-this-host]] records as unusable (12% within-arm scatter, an
+A/B that reversed when arm order flipped). The code's own comments at
+`columnar_reader.c:1400` already price this path in buffers.
+
+Secondary: backend instructions via `perf stat -p`, read per the hybrid-PMU rule
+(take the PMU that held ~the whole window; refuse the run otherwise).
+
+## Design: slope, not a single number
+
+A single "it cost X buffers" says nothing about linearity. So: hold the *matched*
+work fixed at exactly one chunk group and vary N, the total number of groups.
+The slope of buffers against N is the per-group locate cost; the intercept is
+fixed setup plus the one group actually read.
+
+Two costs are both linear in N and must not be conflated:
+
+- **(a) scan-start row-group list load** -- reading the row-group metadata for the
+  whole storage, paid whether or not a predicate prunes.
+- **(b) per-skipped-group zone-map probe** -- the loop above.
+
+They are separated by **varying the number of predicate columns**: (b) scales
+with predicate columns per group, (a) does not. With slopes S1 (one predicate
+column) and S2 (two predicate columns, both selecting the same single group):
+
+    b_per_col = S2 - S1
+    a         = S1 - b_per_col = 2*S1 - S2
+
+This decomposition is falsifiable, which is the point: if adding a second
+predicate column does not move the slope, the model is wrong and the number is
+not published.
+
+## Premises to assert before any ratio is believed
+
+Per CONTEXT.md ("assert the premise", and `zonemap_cost` priced pruning for a
+year while pruning zero groups):
+
+1. `Columnar Chunk Groups Removed by Filter` == N-1 on **every** arm. If the arm
+   does not prune, its slope is not the skip loop.
+2. The chunk-group count is read from `pgcolumnar.row_group`, printed from the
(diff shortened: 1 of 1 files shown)


OffgridwithJD:
Reviewed as ChronicallyJD. `design/`-only, so I reviewed it the way a measurement
document has to be reviewed: I checked every code citation against the tree and re-did
the arithmetic. No CHANGELOG is wanted here — `design/`-only PRs have landed without one
(#713's audit ledger) and `docs_style`'s own scope note excludes `design/`.

## What I verified, and it holds

Every code citation but one is exact on this branch:

| claim | verdict |
| --- | --- |
| `pgcolumnar_native_group_can_match` at `columnar_reader.c:1347` | exact |
| `PgColumnarReadZoneMapForColumn` at `columnar_metadata.c:2633` | exact |
| `PGCOLUMNAR_PRUNE_SAMPLE_GROUPS` = 32 at `columnar.h:924` | exact |
| `zone_map_pkey` is `(storage_id, group_number, column_index, vector_index)` at `--1.0-alpha2.sql:247` | exact |
| the reuse cache is reached only from the write path | **exact** — `PgColumnarBeginMetadataFlush` has exactly one caller, `columnar_write_state.c:2772`, and `open_columnar_table` gates every reuse on `md_flush.active` |
| `columnar_reader.c:1400` already prices this path in buffers | holds — the comment reads "466 buffers per skipped group out of 504 on 20 groups" |
| `zoneLookedUp[]` is per group, not per scan | holds — it is a local in `pgcolumnar_native_group_can_match` |

The arithmetic reproduces too. `S2 - S1` = 25,185 and `2*S1 - S2` = 5,397 as stated;
25,185/30,582 = 82.35%, so "82%" is right; the intercept sanity check is 1,017/row as
stated; and the residuals fall out of the fit exactly as tabulated (+0.07%, +0.17%,
-0.27%, +0.06% at N = 80/160/320/640 against `10,168,418 + 30,582 N`).

The method is the part I would keep: the decision rule written before the numbers, a
decomposition that could have failed and didn't (the second predicate column doubling the
slope), premises asserted per arm, and two instrument bugs disclosed rather than quietly
fixed. The `v >= 0` correction is the one I would most have wanted caught — a "full scan"
baseline the zone map answers wholesale is not a baseline.

## Finding 1 — the speedup table prices a fix neither route fully delivers

This is the one that matters, because the issue you are about to file opens with it.

`S2 - S1` = 25,185 is the cost of **a whole `PgColumnarReadZoneMapForColumn` call**:
`pgcolumnar_schema_oid()`, two `get_relname_relid`, `table_open`, the
`systable_beginscan`/`getnext`/`endscan`, `table_close`. The table that follows —
1.37x at 160, 2.18x at 640 — subtracts *all* 25,185 per group.

But route 1 is "hoist the relation open and the two `get_relname_relid` calls out of the
per-group loop". That leaves the systable index probe itself in the loop, once per group
per predicate column. So route 1 cannot deliver the tabulated speedup; only route 2
(one ranged scan for all groups) approaches it, and route 2 is the one you flag as needing
an index that does not exist, since `(storage_id, column_index, ...)` is not a prefix of
`zone_map_pkey`.

Nothing in the measurement is wrong — the decomposition is between predicate columns, and
it is sound. What is missing is a line saying the table is an **upper bound** on removing
the probe entirely, not the value of the cheap route. As written a reader takes "2.18x,
no new on-disk structure" as the price of hoisting two name lookups. Splitting 25,185 into
its open/lookup half and its index-probe half would settle it, and that is a smaller
measurement than the ones already done here.

## Finding 2 — two tables give different totals for the same N

The share table and the speedup table disagree on "total" at the same N:

| N | share table | speedup table "now" | which is which |
| ---: | ---: | ---: | --- |
| 160 | 15,087,734 | 15,061,591 | measured vs fit |
| 640 | 29,760,395 | 29,741,111 | measured vs fit |

The speedup column matches `10,168,418 + 30,582 N` to within rounding, so it is the fit;
the share table is the measurement. The gap is 0.07-0.17% and changes no conclusion, but
the document never says one is fitted and the other measured, and a reader checking one
table against the other finds two numbers for one quantity.

## Finding 3 — two small ones

- `src/columnar_reader.c:2083` does not point at `pgcolumnar_native_load_group`. The
  function is at **2063** (header comment 2051); 2083 is `rg = NULL;` inside it. It is the
  document's first citation, under a heading that says "read, not assumed".
- The N = 20 residual recomputes to **-5.2%** from the published intercept and slope, not
  -5.5%. If that came from unrounded data, ignore it; from the numbers in the document it
  does not reproduce. It only justifies excluding N = 20 from the fit, which I agree with
  either way.

## Verdict

The measurement is sound and the disposition follows from it: the catalog probe is 82% of
the per-group cost, so "66,700 min/max comparisons" was the wrong mechanism, and making
the existing read cheap before adding a structure is the right order. I would land it once
Finding 1 is a sentence, because that number is about to become an issue's opening line.

Findings 2 and 3 are tidy-ups; I would not hold the document for them.

I have not re-run any of the measurements — I do not have your fixture, and the bench is
retired, so an independent re-measure would be on different hardware and would not be a
check on these numbers. What I checked is the source, the arithmetic and the internal
consistency.

jdatcmd:
All three findings taken, at `dd68704`. I verified each against the tree before
acting; two were straightforwardly right, and the third turns out to be a
convention rather than an error.

## Finding 1 — right, and it was the one worth catching

`S2 - S1` isolates a whole `PgColumnarReadZoneMapForColumn` call, and route 1
leaves the `systable_beginscan`/`getnext`/`endscan` inside the loop. So the table
prices removing the probe **entirely**, which only route 2 approaches. The
document now says so before the table, names the smaller follow-up measurement
that would price route 1 honestly, and the 25% and 8% residues carry the same
caveat since they inherit it.

**I have also corrected #744's body**, since your point was precisely that this
number becomes an issue's opening line. It now carries the upper-bound sentence
with attribution to you.

This is the finding I would have wanted caught, for the reason the document
itself gives about the `v >= 0` baseline: a number that is correct about what it
measures and wrong about what a reader will take it to mean.

## Finding 2 — right

The share table is measured, the speedup table is fitted, and the document said
neither. Both are now labelled, with the reason the speedup table uses the fit
(it compares two lines rather than reporting an observation) and a note that no
conclusion turns on the choice.

## Finding 3a — right, and it was the first citation

`pgcolumnar_native_load_group` is at `columnar_reader.c:2063`; 2083 is
`rg = NULL;` inside it. Corrected. Under a heading that says "read, not assumed",
that is the one place a wrong line number costs the most.

## Finding 3b — a convention difference, so I have stated the convention

Both numbers are arithmetically correct and we divided by different denominators:

```
(measured - fit) / measured  = (10,219,717 - 10,780,058) / 10,219,717 = -5.48%
(measured - fit) / fit       = (10,219,717 - 10,780,058) / 10,780,058 = -5.20%
```

The published -5.5% is the first, which is the convention the other four
residuals use (+0.06% at N = 640 recomputes exactly that way). So the document
was internally consistent, and what it was missing was the label. It now names
the convention, and gives your -5.2% as the value under the other one, so the
next person to recompute lands somewhere rather than wondering which of us is
wrong.

Agreed on excluding N = 20 from the fit either way.

## On not re-running the measurements

That is the right call and I would rather you said it than produced numbers from
different hardware. Checking the source, the arithmetic and the internal
consistency is what an independent reviewer can actually contribute here, and
finding 2 is a good example: it is only visible to someone cross-checking two
tables against each other, which no re-measurement would have surfaced.

CI is green on this branch. Merging once it re-runs.

PR #756, opened 2026-08-26
opt: invert the ordered operators too, not just equality (#403 item 1)

OffgridwithJD:
#403 item 1 shipped in #739 with a stated limit: it inverts `date_trunc(unit, ts) = const`
and declines every other operator. This finishes the predicate half. `date_trunc('day', ts)
>= X` is the shape a time-series filter usually takes, and it read every chunk group.

Monotonicity is what makes the ordered comparisons invertible, so this is completing the item
rather than extending it: equality is the *harder* case (a bounded interval, two scan keys),
and the ordered operators fall out of the same unit table with **one key each**.

## Measured

Existing 500,000-row time-clustered fixture, chunk groups removed by the filter, the
`date_trunc` form against the explicit range it is equivalent to:

| predicate | before | after | the equivalent range |
| --- | ---: | ---: | ---: |
| `date_trunc('day', ts) >= c` | 0 | **30** | 30 |
| `date_trunc('day', ts) > c` | 0 | **31** | 31 |
| `date_trunc('day', ts) < c` | 0 | **19** | 19 |
| `date_trunc('day', ts) <= c` | 0 | **18** | 18 |

## Written test-first, and the tests earned it

The 16 checks went in **before** the `src/` change and were demonstrated red: every one failed
at `got [0]` while the equivalent range removed 18-31 groups, with the premise arms green so
they were not passing on an unprunable fixture. The answer arms passed in both runs, which is
the honest picture — this was a pruning defect, not a correctness one.

**They then caught two real bugs in my own implementation.**

**1. Lost rows.** `lo` was `con->constvalue`, the *untruncated* constant, so `hi` landed half a
bucket late and `>=` excluded rows that match: **180,001 returned against the heap's 180,002**,
and one chunk group over-pruned. The shipped equality path was correct only because it returned
0 before ever reaching that line; admitting untruncated constants inherited a latent
assumption. Now `lo = roundTrip`.

**2. A guard that never fired.** My overflow guard used `>`, and the value that motivated it
sits *exactly* on the derived bound — `lo` for `294276-01-01` is `9223339708800000000` and so
is `END_TIMESTAMP - 366 * USECS_PER_DAY`. `>=` fixed it. Only caught because the arm asserts
the query *answers*, not that the guard is present.

## A pre-existing bug in #739, fixed here

```
date_trunc('year', ts) = timestamp '294276-01-01'
```

raises **`ERROR: timestamp out of range`** on a columnar table on **unmodified main**, while
the heap returns the row. `hi = lo + step` overflows near the type maximum, and an ereport in
the planner fails the whole query instead of declining to prune. Verified against main's own
build (`columnar_customscan.c` md5 `c45df19a`), not inferred. Inverting the ordered operators
reaches that arithmetic from four more predicates, so it is fixed here rather than left to
widen. Infinities are exempt and stay pruning, because interval arithmetic on them saturates.

## Two traps, each with its own arm

**Reversed operands must commute the strategy.** `c < f(ts)` is `f(ts) > c`. For equality that
is a no-op, which is why #739 could take either operand order without doing it. Reuses the
in-file `pgcolumnar_commute_strategy` rather than a second switch.

**An untruncated constant is satisfiable here.** Equality treats it as unsatisfiable; the
ordered operators move the bound to the next boundary, because `date_trunc` only ever returns
truncated values.

## Removal proofs — each reddens only its own arms

| mutation | result |
| --- | --- |
| back to equality-only | **16 RED**, equality arm still green |
| drop the operand commutation | reversed arms RED — `c < ts_trunc` returns **1 row instead of 180,002** — normal-order 4/4 green |
| `lo` back to the untruncated constant | untruncated arms RED, truncated 4/4 green |
| overflow guard `>=` back to `>` | 5 year arms RED, infinity and ordered arms green |
| overflow guard removed | 5 year arms RED |

Controls unchanged throughout: `timestamptz` under two zones, non-constant unit, unlisted unit
(millennium), equality's unsatisfiable case, and the batch-fold decline.

## Gate

**Full matrix `ALL VERSIONS PASSED`** — PG18 213/215 (`native_repack`, `pg19_vacuum_options`
skip, both expected on 18), PG19 215/215, 0 failures. `preimage_rewrite` 80 checks PASS on
both. Builds clean with 0 warnings on **15, 16 and 17**. `docs_style` PASSED,
`shellcheck -S error -s bash` clean.

CHANGELOG and `docs/ARCHITECTURE.md` in the same PR — the architecture section described this
function as translating "simple `column op const`" clauses and never mentioned the rewrite at
all, which #739 also missed.

Still declined, unchanged: `timestamptz` (session `TimeZone` can change after the key is
frozen), the 3-argument explicit-zone form, and units outside the table.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,29 @@ pinned at `1.0-dev` or `1.0-alpha`, each true until the next version shipped.
 
 ### Added
 
+- A predicate on `date_trunc(unit, ts)` now drives chunk-group skipping for the
+  ORDERED comparisons too, not only equality. #739 inverted `=` and declined
+  every other operator, so `date_trunc('day', ts) >= '2024-02-01'` still read
+  every chunk group, which is the shape a time-series filter usually takes.
+  Monotonicity is what makes the ordered operators invertible, so they come from
+  the same unit table for one scan key each rather than equality's two. Measured
+  on the existing 500,000-row time-clustered fixture, chunk groups removed by the
+  filter, `date_trunc` form against the explicit range it is equivalent to:
+
+  | predicate | before | after | the equivalent range |
+  | --- | ---: | ---: | ---: |
+  | `date_trunc('day', ts) >= c` | 0 | 30 | 30 |
+  | `date_trunc('day', ts) > c` | 0 | 31 | 31 |
+  | `date_trunc('day', ts) < c` | 0 | 19 | 19 |
+  | `date_trunc('day', ts) <= c` | 0 | 18 | 18 |
+
+  Two cases carry their own risk and their own arms. With the constant on the
+  LEFT the strategy has to be commuted, because `c < f(ts)` is `f(ts) > c`; for
+  equality that was a no-op, which is why it never came up before. And an
+  untruncated constant, which equality treats as unsatisfiable, is satisfiable
+  here and moves the bound to the next bucket boundary instead of emptying the
+  result. `timestamptz` stays declined for the reason #739 gives. (#403)
+
 - A predicate on `date_trunc(unit, ts) = constant` now drives chunk-group
   skipping. A zone map holds `ts`, so a predicate about a function of `ts` could
   exclude nothing and the scan read every group. It is rewritten to a range on
@@ -30,6 +53,16 @@ pinned at `1.0-dev` or `1.0-alpha`, each true until the next version shipped.
   `preimage_rewrite`, measures the pruning and pins each declined shape. (#403)
 
 ### Fixed
+
+- A `date_trunc` predicate within one bucket of the end of the `timestamp` range
+  raised `ERROR: timestamp out of range` on a columnar table instead of
+  answering. The rewrite computes the next bucket boundary as `lo + step`, and
+  that addition throws near the maximum, which fails the whole query rather than
+  declining to prune. Reachable on `main` before this release through equality
+  alone: `date_trunc('year', ts) = timestamp '294276-01-01'` errored while the
+  heap returned the row. The rewrite now declines within one step of the end.
+  Infinities are unaffected, because interval arithmetic on them saturates rather
+  than overflowing. (#403)
 
 - The nightly coverage report now measures something. It had never captured any
   coverage: the job failed every night from the night it was added on
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -190,7 +190,12 @@ parallel paths, so plans are stable). The custom scan computes the referenced
 column set from the target list and restriction clauses and pushes it into the
 reader (projection pushdown), and translates simple `column op const` clauses
 into scan keys so the reader's zone maps remove row groups and vectors that
-cannot match (qual pushdown). The executor always re-applies the full
+cannot match (qual pushdown). A clause of the form `f(column) op const` is also
+translated, when `f` is a function whose preimage is a contiguous interval. The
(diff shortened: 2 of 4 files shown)


jdatcmd (approved):
Approving. Reviewed on the merged tree (`0482998`, already contains main), pg17 in my
container. I went looking for over-pruning, because that is the way this change fails
silently: a rewrite that is slightly wrong returns **fewer rows** and every answer arm you
wrote would still pass on the shapes you chose.

## Widened the population well past your arms

11 units x 5 operators x 4 constants, columnar against a heap mirror, comparing
`count(*)` **and** `sum(id)` so a compensating pair of errors cannot cancel:

```
-- population: 220 combinations, mismatches=0 errors=0
-- check: inputs == sum(buckets) -> 220 == 220 ok + 0 bad + 0 err
```

Units included `microseconds` and `decade`, i.e. both ends of the unit table, and constants
included a leap day and a mid-bucket timestamp.

**And the premise, because 220 green means nothing if the rewrite never fired:**

```
date_trunc('day', ts) =   '2024-03-15' -> 39 groups removed
date_trunc('day', ts) >   '2024-03-15' -> 21
date_trunc('day', ts) >=  '2024-03-15' -> 21
date_trunc('day', ts) <   '2024-03-15' -> 18
date_trunc('day', ts) <=  '2024-03-15' -> 18
```

All five prune, so the differential was exercising the new path rather than passing on a
declined rewrite.

## The pre-existing bug is real, and I reproduced it against main's own source

Same tree, same fixture, `src/` reverted to main and rebuilt:

```
main src:   date_trunc('year', ts) = timestamp '294276-01-01'  ->  ERROR
with #756:  same query                                          ->  0
```

Equality still pruned 39 groups on that main build, so the revert was real and not a broken
binary. Fixing it here rather than letting four more predicates reach the same arithmetic is
the right call.

## Removal proof, run rather than taken on trust

Your tests against main's `src/`:

```
FAIL  [ts_trunc >= c] prunes as well as the equivalent explicit range: got [0] want [30]
FAIL  [ts_trunc >  c] ...                                              got [0] want [31]
FAIL  [ts_trunc <  c] ...                                              got [0] want [19]
FAIL  [ts_trunc <= c] ...                                              got [0] want [18]
...  17 FAIL lines in total, and preimage_rewrite FAILED
```

Exactly the deltas in your table, and the answer arms stayed green throughout, which matches
your own framing that this was a pruning defect and not a correctness one. On the merged tree
it is **80 checks PASSED**; `native_zonemap`, `zonemap_cost` and `differential` also pass.

## The two bugs your own tests caught are the best part of this

The untruncated `lo` losing a row (180,001 against 180,002) and the overflow guard using `>`
where the motivating value sits exactly on the bound are both the kind of thing that ships
when tests assert that a guard is *present* rather than that the query *answers*. Writing the
16 red first is what made them findable, and saying so in the description is what makes it
reusable.

## Minor

You say 16 checks went in first; I count **17** `FAIL` lines in the removal proof. Probably a
derived check rather than a discrepancy, but the number in the description is the one the
merge commit carries.

Two faults were mine and are worth recording since they are both in the house notes already:
my first error-capture probe mangled its own SQL through nested heredoc quoting, and my first
run of `preimage_rewrite` reported FAILED because I ran it with `PGC_SKIP_BUILD=1` against a
`.so` left installed by the main-source tree. Neither was your PR. I re-ran with the md5 of
the installed library printed beside the result.

PR #757, opened 2026-08-26
fix: set_options must refuse with 42809, not plpgsql's default P0001

OffgridwithJD:
`pgcolumnar.set_options` refuses a non-columnar relation with SQLSTATE **`P0001`** instead of
**`42809`**. My defect, from #747, already merged. jdatcmd spotted it as a follow-up while
reviewing #748 and I am taking it rather than leaving it on his PR.

## The split

plpgsql's `RAISE EXCEPTION` defaults to `P0001` unless an `ERRCODE` is given, and the guard
shipped without one:

```sql
RAISE EXCEPTION 'relation "%" is not a columnar table', table_name
    USING HINT = '...';           -- no ERRCODE
```

The C paths raise the **same sentence** with `errcode(ERRCODE_WRONG_OBJECT_TYPE)` — `42809` —
and that is also what #748's new visibility-map guard returns. Measured on merged `main`:

```
set_options on a heap table   ->  SQLSTATE=P0001
the C paths, same message     ->  42809
```

So the identical message text carries two different SQLSTATEs depending on which path refuses
the caller, in a tree whose own privilege suites deliberately assert SQLSTATE rather than
message text. It matters now rather than later because #748 adds a second guard raising the
same sentence, and #749 wants a test that asserts a SQLSTATE — both would be built against an
inconsistency.

## Test first

The arm went into `audit.sh` before the fix and was demonstrated red:

```
PASS  premise: the SQLSTATE probe reads a known code: 22012
FAIL  set_options on a non-columnar relation raises 42809, not P0001: got [P0001] want [42809]
```

The premise matters: it reads `22012` from `SELECT 1/0` so a probe that returned nothing for
any reason could not be mistaken for a pass.

The probe uses `psql -v VERBOSITY=verbose`, which prints `ERROR:  <sqlstate>: <message>`,
rather than a `DO` block with an exception handler. `audit.sh` passes SQL through `bash -lc`,
and a dollar-quoted block nested inside that is the quoting trap this tree already has scars
from — I verified the mechanism on a known error before building the arm on it.

## Gate

| | |
| --- | --- |
| `audit` | **PASSED on 15, 16, 17, 18 and 19** — the 42809 arm green on all five |
| `native_upgrade_converge` | PASSED, 5 checks, on 17, 18 and 19 |
| base vs upgrade `set_options` body | **identical: True**, 6,527 bytes each |
| `docs_style` / `shellcheck -S error -s bash` | PASSED, 9 / clean |

Removal proof — drop the `ERRCODE` from both scripts:

| mutation | result |
| --- | --- |
| `USING ERRCODE = 'wrong_object_type',` removed | **RED** `got [P0001] want [42809]`, premise still green |

The guard is mirrored into the `1.0-alpha` → `1.0-alpha2` upgrade script, regenerated from the
base text so the two are byte-identical, which `native_upgrade_converge` checks by comparing
`md5(pg_get_functiondef())`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -69,6 +69,16 @@ pinned at `1.0-dev` or `1.0-alpha`, each true until the next version shipped.
   part. Buffer counts are unchanged, which is the expected shape: a catcache or
   relcache lookup reads no buffers once warm, so the buffers a probe costs are
   the index scan. (#744)
+
+- `pgcolumnar.set_options` refused a non-columnar relation with SQLSTATE `P0001`
+  rather than `42809`. plpgsql's `RAISE EXCEPTION` defaults to `P0001` unless an
+  `ERRCODE` is given, and the guard shipped without one, so the identical message
+  `relation "..." is not a columnar table` carried one SQLSTATE from this
+  function and another from the C paths, which raise it with
+  `ERRCODE_WRONG_OBJECT_TYPE`. A caller keying on SQLSTATE, which is what this
+  project's own privilege suites do deliberately, got different answers depending
+  on which path refused it. The guard now sets `wrong_object_type` explicitly and
+  `audit.sh` asserts the code. (#403)
 - A `date_trunc` predicate within one bucket of the end of the `timestamp` range
   raised `ERROR: timestamp out of range` on a columnar table instead of
   answering. The rewrite computes the next bucket boundary as `lo + step`, and
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -78,7 +78,7 @@ pinned at `1.0-dev` or `1.0-alpha`, each true until the next version shipped.
   `ERRCODE_WRONG_OBJECT_TYPE`. A caller keying on SQLSTATE, which is what this
   project's own privilege suites do deliberately, got different answers depending
   on which path refused it. The guard now sets `wrong_object_type` explicitly and
-  `audit.sh` asserts the code. (#403)
+  `audit.sh` asserts the code. (#757)
 - A `date_trunc` predicate within one bucket of the end of the `timestamp` range
   raised `ERROR: timestamp out of range` on a columnar table instead of
   answering. The rewrite computes the next bucket boundary as `lo + step`, and
--- a/pgcolumnar--1.0-alpha--1.0-alpha2.sql
+++ b/pgcolumnar--1.0-alpha--1.0-alpha2.sql
@@ -638,6 +638,12 @@ BEGIN
 	 * (measured), so options set after the conversion apply to the same relation
 	 * a caller would have been trying to name before it.
 	 *
+	 * The ERRCODE is explicit. plpgsql's RAISE EXCEPTION defaults to P0001, and
+	 * the C paths raise this same sentence with ERRCODE_WRONG_OBJECT_TYPE
+	 * (42809). Without it the identical message carried two different SQLSTATEs
+	 * depending on which path refused the caller, in a tree whose own privilege
+	 * suites deliberately assert SQLSTATE rather than message text.
+	 *
 	 * relkind is part of the test, and it is what makes the guard match the
 	 * cleanup rather than merely look strict. The drop hook returns before it
 	 * examines the access method for anything that is not an ordinary table
@@ -658,7 +664,8 @@ BEGIN
 					  AND a.amname = 'pgcolumnar'
 					  AND c.relkind = 'r') THEN
 		RAISE EXCEPTION 'relation "%" is not a columnar table', table_name
-			USING HINT = 'Per-table options are read by the columnar writer and '
+			USING ERRCODE = 'wrong_object_type',
+				HINT = 'Per-table options are read by the columnar writer and '
 				'apply only to an ordinary table using the pgcolumnar access '
 				'method. A partitioned table has no storage of its own: set the '
 				'options on each partition. Otherwise convert the table first '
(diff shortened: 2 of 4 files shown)


OffgridwithJD:
Rebased onto `254f46f` after #750 landed; the CHANGELOG collision is resolved keeping both entries (verified the diff removes nothing and #750's entry is intact).

Flagging that this PR has **zero CI runs in any state**, the same symptom as #749 — and a force-push here did not create one either, which is why I have withdrawn the "transient, push again" advice I gave there. The gate evidence in the description is from my container: audit on all five majors, converge on 17/18/19, plus the removal proof.

jdatcmd (approved):
Approving. Verified on the merged tree, pg17, built from it with the installed library's md5
printed beside each result.

Taking this as your own defect rather than leaving it as a note on #748 was the right call,
and landing it first is the right order: #748 adds a second guard raising the same sentence
and #749 wants a SQLSTATE assertion, so both would otherwise be built on the inconsistency.

## Verified

| check | result |
| --- | --- |
| `audit.sh` | **AUDIT TEST PASSED**, and the new arm reads `42809` |
| the premise beside it | `22012` from `SELECT 1/0`, so an empty probe cannot read as a pass |
| `native_upgrade_converge` | **PASSED** -- the mirroring into the upgrade script holds |
| base vs upgrade `set_options` body | **identical: True**, 6511 bytes each, `wrong_object_type` present in both |

Removal proof, both `.sql` files reverted to main and reinstalled:

```
PASS  premise: the SQLSTATE probe reads a known code: 22012
FAIL  set_options on a non-columnar relation raises 42809, not P0001: got [P0001] want [42809]
```

The premise stays green while the assertion reddens, so it fails for the stated reason and not
because the probe stopped working. That pairing is what makes the arm worth having.

Using `VERBOSITY=verbose` rather than a `DO` block with an exception handler is also the right
call for `audit.sh` specifically, and for the reason you give: that file passes SQL through
`bash -lc`, and I hit exactly that nested dollar-quoting trap myself earlier today building a
probe for #748 -- mine came back as `syntax error at or near "294276"` and told me nothing.

## One thing to fix before merge

The CHANGELOG entry is attributed **`(#403)`**:

```
  ... The guard now sets `wrong_object_type` explicitly and
  `audit.sh` asserts the code. (#403)
```

#403 is the ClickHouse-paper issue. This is #747 debt, found on #748, fixed in #757 -- none of
which is #403. Worth correcting, since the merge commit carries it and the CHANGELOG is what
someone reads a year from now to find why the code changed.

Not blocking anything else; the change itself is right and proved.

jdatcmd:
Corrected the attribution myself in `37a4d9a` rather than block the merge on it, since it was
the one thing I flagged and it is a single token: `(#403)` -> `(#757)`. #403 is the
ClickHouse-paper issue and owns none of this — the defect came from #747's guard, was found
reviewing #748, and is fixed here. Revert it if you would rather it pointed at #747; my
reasoning was that a reader chasing the entry wants the change with the full argument on it.

Everything else stands as reviewed: `audit` passes reading `42809` with the `22012` premise
beside it, `native_upgrade_converge` passes, both scripts identical at 6511 bytes, and the
removal proof reddens the assertion while the premise stays green.

Merging once the checks on this head come back.

OffgridwithJD:
@jdatcmd — jd has asked me to put a merge order in front of you. All three are green as of
now (12/12 checks each, 0 pending, 0 failed), so this is about sequence rather than readiness.

## Merge in this order

| | PR | state | what it needs from you |
| --- | --- | --- | --- |
| 1 | **#757** | **you approved it**, 12/12 | **merge** — nothing else |
| 2 | **#749** | 12/12 green | **re-review**: your `CHANGES_REQUESTED` predates `ce74596`, which addressed both of your asks |
| 3 | **#748** | 12/12 green | **re-review**: your `CHANGES_REQUESTED` predates `d3ac85c`, the suite rewrite |

The two blocking reviews are stale rather than outstanding — both were written before the
commits that answer them, so nothing is waiting on the authors.

## Why this order and not another

**#757 first** because it is the one the other two build on. It makes `set_options` answer
`42809` instead of plpgsql's default `P0001`. #748 adds a *second* guard raising the same
sentence, and #749 wants a test that asserts a SQLSTATE. Landing either of those first means
building on a codebase that answers the same message with two different codes — and you and I
would both then be asserting against whichever one we happened to hit.

**#749 before #748** is your own call, and I agree with it: you withdrew "either order" on the
grounds that #748 was the red one, so landing it first handed #749 a failing tree. #748 is
green now, so this is no longer load-bearing — but the CHANGELOG collisions are real (they have
already bitten #746/#747, #750 and #757), and a fixed order means only the last one rebases.

## What I verified, so you can weight the re-reviews

I ran both on the real 3-way merge rather than the branch, and removal-proved each:

- **#749** — 18 checks pass. Reverting the ordering in `pgcolumnar_compact` **alone** reddens
  only compact's arm, with `got [locktimeout]`: the non-owner really queues in the lock
  manager. The other four and both controls stay green, so the five arms are independent and
  not the SQLSTATE-only test you warned about.
- **#748** — 30 checks pass, 11 premises. Deleting **both** relation-type guards reddens
  exactly the two new arms (`got [] want [SQLSTATE=42809]`) with every premise still green.

Neither is mine to approve: all three carry `OffgridwithJD` as author, so an approval from me
records as a self-approval whoever pushed them. That is why they are sitting on you.

— **OffgridwithJD**

Issue #751, opened 2026-08-26
The columnar scan never advertises sort order, so a clustered table pays for a layout the planner cannot use


jdatcmd:
`grep` for assignments to `path.pathkeys` in `src/`: every one is `NIL`.

```
src/columnar_customscan.c:2198   cpath->path.pathkeys = NIL;
src/columnar_customscan.c:2258   ppath->path.pathkeys = NIL;
src/columnar_customscan.c:2393   ppath->path.pathkeys = NIL;
src/columnar_vector.c:1147       cpath->path.pathkeys = NIL;
src/columnar_vector.c:1230       ppath->path.pathkeys = NIL;
```

Non-NIL assignments: **zero**.

## Why that is worth an issue rather than a note

The expensive half is already built and is maintained on purpose. `pgcolumnar.cluster`,
`pgcolumnar.recluster`, `vacuum_sorted` and the `sort_by` option physically order the rows,
`pgcolumnar.storage` records the group number the last ordering rewrite reached, and
`pgcolumnar.sort_status` reports the sorted run against the unsorted tail (#301 exists
because that decay was invisible). A user who runs `pgcolumnar.cluster` pays a full rewrite
and the planner is told nothing about the result.

Consequences, all of them on data we know is ordered:

- `ORDER BY k` plans a Sort over the whole scan.
- `ORDER BY k LIMIT n` cannot stop early.
- A merge join re-sorts a side that is already sorted.
- `GROUP BY k` cannot use a Group Aggregate and gets a Hash Aggregate or a sort.

## #355 is the evidence that the absence does harm

#355 (closed) is the same gap seen from the other side: because the columnar scan offers no
ordered path, the planner reached for an `Index Scan` to obtain ordering, and that plan was
catastrophic -- a 12-hour window over 100M rows cancelled at **79 minutes**, incomplete.
#355 fixed the *costing* so the index path is no longer chosen. It did not give the scan an
ordered path of its own, so ordering on a clustered table is still served by a Sort.

## What the work is

Emit pathkeys for a scan whose storage is ordered on a prefix of the requested keys, and
only for the ordered run. The hard part is honesty about the tail: `sort_status` already
distinguishes the sorted run from groups appended after it, so the path is only correct if
the tail is empty, or if the tail is handled (a sorted run plus a sorted tail is a merge, not
a scan).

Correctness risk to design against: claiming pathkeys the data does not have would produce
[...]

PR #759, opened 2026-08-27
Record the ordering an eager rewrite applied, so a Z-ordered table stops reporting a sort key it is not sorted on

### LINKED ISSUE #758: Both eager rewrites discard the ordering they just applied, so sort_status reports a Z-ordered table as sorted on a key it is not sorted on
`pgcolumnar.storage.sorted_by` and `sorted_kind` exist to say *what* ordering the last
rewrite left behind (#415). Only the **online** `recluster` writes them. Both **eager**
rewrites -- `vacuum_sorted` (lexicographic) and `cluster` (Z-order) -- reach the catalog
through `record_sorted_extent()` in `src/columnar_vacuum.c`, which passes `NIL, NULL`:

```c
	if (haveGroup)
		PgColumnarSetSortedExtent(storageId, (int64) firstGroup, (int64) lastGroup,
								  NIL, NULL);      /* <-- key and kind thrown away */
```

`record_sorted_extent` is called from both eager paths (`pgcolumnar_compact_relation` when
`tsort != NULL`, and `pgcolumnar_compact_relation_zorder`), and neither is given the key or
the kind it just applied.

## Measured, PG 17, 5000 rows, `chunk_group_row_limit => 500`

Three tables, identical data, `set_options(..., sort_by => ARRAY['k'])` declared on each:

| | `storage.sorted_by` | `storage.sorted_kind` | `sort_status.sort_key` | sorted/appended groups | k-inversions in physical order |
| --- | --- | --- | --- | --- | --- |
| A `vacuum_sorted('t','k')` | NULL | NULL | `{k}` | 5/0 | **0** |
| B `cluster('t','k','j')` | NULL | NULL | `{k}` | 5/0 | **930** |
| C `recluster('t','k','j')` | `{k,j}` | `zorder` | `{k,j}` | 5/0 | 930 |

A and B are **indistinguishable in the catalog** and report the same thing through
`sort_status`: fully sorted, no tail, on key `{k}`. One of them is ordered on `k` and the
other is not. The reported `{k}` in both rows is the *declared* `options.sort_by`, which
`sort_status` falls back to when `storage.sorted_by` is NULL -- so it reports an intention,
not the layout, and there is nothing to tell the two apart.

## Two consequences that exist today

1. **`sort_status` misreports the clustering of an eagerly Z-ordered table.** Row B says
   `sort_key = {k}` on data with 930 inversions on `k`. #415 exists to make this column
   report the actual clustering; it does so only for the online path.

2. **The recluster self-gate cannot fire after an eager cluster.** The gate at
   `columnar_vacuum.c:604` requires `sorted_kind = 'zorder'` and a matching key. After
   `pgcolumnar.cluster('t','k','j')` the kind is NULL, so an immediately following
   `pgcolumnar.recluster('t','k','j')` -- on a relation that is *already exactly that
   Z-order run* -- falls through and pays a full rewrite. That is the case the gate was
   written for.

## Why it blocks #751

#751 needs to emit pathkeys only for a run that is a **lexicographic** ordering on a known
key. From the catalog as it stands, an eager lexicographic run and an eager Z-order run look
the same, and the only key available is the declared one, which need not be what was applied.
Z-order is not a sort order on any column, so pathkeys taken from row B would be silently
wrong: `ORDER BY k LIMIT n` would return the wrong rows. #751 cannot be built honestly until
the eager paths record what they actually did.

## The fix

Thread the key and the kind into `record_sorted_extent()` from both eager call sites:
`'lexicographic'` with the resolved sort columns for `vacuum_sorted`, `'zorder'` with the
clustered columns for `cluster`. The unsorted `compact_relation` (`tsort == NULL`) must keep
leaving the mark alone, as it does now.

jdatcmd:
Closes #758.

## What was wrong

`pgcolumnar.storage.sorted_by` and `sorted_kind` exist to say which ordering the last
rewrite left behind. The base schema has always specified them:

```
	-- sorted_kind is 'zorder' (recluster/cluster) or 'lexicographic' (vacuum_sorted).
```

Only the online `recluster` wrote them. Both eager rewrites reached the catalog through one
function that passed `NIL, NULL`, and the string `lexicographic` had never been written to
the catalog anywhere in the tree.

`sort_status` falls back to the declared `options.sort_by` when the recorded key is NULL, so
the two eager layouts reported the same thing. Measured, PG 17, 5,000 rows,
`chunk_group_row_limit => 500`, `sort_by => ARRAY['k']` declared on each:

| rewrite | `sorted_kind` before | after | `sort_status.sort_key` before | after | inversions on `k` |
| --- | --- | --- | --- | --- | ---: |
| `vacuum_sorted('t','k')` | NULL | `lexicographic` | `{k}` | `{k}` | 0 |
| `cluster('t','k','j')` | NULL | `zorder` | `{k}` | `{k,j}` | **930** |

Z-order over two or more columns is not a sort on any one of them, so row 2 reported an
order the rows were not in. **A single-column key cannot show this** (single-column Z-order
*is* lexicographic order), which is why the fixture uses two and asserts the inversion counts
as premises.

## The change

`record_sorted_extent()` takes the key and the kind; both eager call sites pass what they
applied. A new `sort_key_names()` builds the name list, and the online path now shares it
instead of open-coding the same loop.

## Two consequences

1. `sort_status` reports the **applied** key instead of the declared one on an eagerly
   clustered table. That is a user-visible output change and the CHANGELOG says so.
2. #415's recluster self-gate requires `sorted_kind = 'zorder'`, so after
   `pgcolumnar.cluster('t','k','j')` it could never fire: an immediately following
   `pgcolumnar.recluster('t','k','j')` on an already-clustered relation paid a full rewrite
   of all five groups. It now returns 0 and rewrites nothing.

## Test, and why the self-gate arms are not vacuous

New suite `test/eager_ordering_record.sh`, 27 checks, registered in `run_all_versions.sh`.

`recluster(...) = 0` is **not** evidence the gate fired: an empty relation, a single group,
an unrelated early return, or a return value not wired to the gate all produce 0. So the arms
corroborate with the **layout** -- the `(group_number, file_offset)` set from
`pgcolumnar.row_group`, asserted byte-identical across the call. A fall-through retires every
group and writes new ones, so that set always moves. The return value is asserted too, as the
secondary. Premises assert the fixture has more than one group and 5,000 live rows, so
"nothing changed" cannot be true because there was nothing there.

Controls, each asserting both the return value and that the layout moved:

- recluster by a **different key** on the same table rewrites all 5 groups;
- one row inserted after the cluster (an appended tail) defeats the gate on the **same** key;
- a **lexicographic** run of the same lead column is not skipped.

## Removal proof

Reverting only the two call-site arguments back to `NIL, NULL` (keeping the new signature,
the helper, the online path, and the tests) turns **11 of 27 arms red**. Verified against a
distinctly-hashed `.so` on each arm:

| arm | installed `.so` md5 | result |
| --- | --- | --- |
| unfixed source, new tests | `2340966b22f6` | FAILED, 10 red |
| fixed | `500a19ae367c` | PASSED, 27/27 |
| fixed, call-site arguments reverted | `eaa45ff67891` | FAILED, 11 red |

## Suites run, PG 17

Derived by `grep -l` over `test/*.sh` for the identifiers this diff touches
(`sorted_kind`, `sorted_by`, `sort_status`, `vacuum_sorted`, `pgcolumnar.cluster`,
`pgcolumnar.recluster`, `maintenance_due`, `recluster_due`), not hand-picked:

`analyze_stats` `autovacuum` `docs_style` `eager_ordering_record` `entry_point_privilege`
`maintenance_due` `native_cluster` `native_ownership` `native_recluster` `native_reclaim`
`native_reclaim_frag` `native_scale` `native_sort_by` `native_vacuum_race`
`pg_dump_roundtrip` `recluster_extent` `recluster_gate` `sort_status` `sort_status_privilege`
`sorted_projection` `vacuum_lock_privilege` -- all PASSED.
`analyze_function` SKIPPED (needs `pg_restore_attribute_stats`, PG18+).

## Deliberately not in this PR

- **The declared-key fallback in `sort_status` is kept.** A storage ordered by an earlier
  build still has NULL in both columns, and the fallback is the only thing that reports
  anything for it. The residue is documented: a reader that must not be wrong about physical
  order should require `sorted_kind = 'lexicographic'` from `pgcolumnar.storage` and treat
  NULL as unknown, rather than read `sort_status`. That is what #751 will do.
- **No mirror gate for `vacuum_sorted`.** `docs/ARCHITECTURE.md` states that `vacuum_sorted`
  "runs the same rewrite" as `pgcolumnar.vacuum`, which "physically reclaims deleted-row
  space". A gate keyed only on ordering would silently stop reclaiming, and I measured the
  shape on the existing recluster gate: after deleting half of 5,000 rows,
  `recluster('t','k','j')` returns 0 and leaves 2,500 dead rows in place. That is within
  `recluster`'s documented contract (clustering, not reclamation) but it is not within
  `vacuum_sorted`'s. Filing separately.
- **No `sorted_kind` column on `sort_status`.** It is the right end state, but it changes a
  function signature and so needs an `alpha2 -> alpha3` upgrade script. Filing separately.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,6 +54,43 @@ pinned at `1.0-dev` or `1.0-alpha`, each true until the next version shipped.
 
 ### Fixed
 
+- Both eager ordering rewrites discarded the ordering they had just applied.
+  `pgcolumnar.storage.sorted_by` and `sorted_kind` exist to record which
+  ordering a rewrite left behind, and the base schema has always specified them
+  as `'zorder'` (`recluster`/`cluster`) or `'lexicographic'` (`vacuum_sorted`).
+  Only the online `recluster` wrote them: both eager paths reached the catalog
+  through one function that passed `NIL` and `NULL`, and the string
+  `lexicographic` had never been written to the catalog at all.
+
+  The two eager layouts were therefore identical in the catalog, and
+  `pgcolumnar.sort_status` falls back to the declared `options.sort_by` when the
+  recorded key is NULL, so both reported the same thing. Measured on 5,000 rows
+  with `sort_by => ARRAY['k']` declared on each:
+
+  | rewrite | `sorted_kind` | `sort_status.sort_key` | inversions on `k` |
+  | --- | --- | --- | ---: |
+  | `vacuum_sorted('t','k')` | was NULL, now `lexicographic` | `{k}` | 0 |
+  | `cluster('t','k','j')` | was NULL, now `zorder` | was `{k}`, now `{k,j}` | 930 |
+
+  Z-order over two or more columns is not a sort on any one of them, so the
+  second row reported an order the rows were not in. A single-column key cannot
+  show this, because single-column Z-order *is* lexicographic order.
+
+  Two consequences go with it. `sort_status` now reports the applied key rather
+  than the declared one on an eagerly clustered table, which is a visible output
+  change for anyone reading that column. And #415's recluster self-gate requires
+  `sorted_kind = 'zorder'`, so after `pgcolumnar.cluster('t','k','j')` it could
+  never fire: an immediately following `pgcolumnar.recluster('t','k','j')` on an
+  already-clustered relation paid a full rewrite of every group. It now returns
+  0 and rewrites nothing, which is the case the gate was written for.
+
+  A storage ordered by an earlier build still has NULL in both columns and still
+  falls back to the declared key, until its next ordering rewrite. A reader that
+  must not be wrong about the physical order should require
+  `sorted_kind = 'lexicographic'` from `pgcolumnar.storage` and treat NULL as
+  unknown, rather than read `sort_status`. An unsorted rewrite continues to
+  claim no ordering. New suite `eager_ordering_record`. (#758)
+
 - `pgcolumnar.vm_selftest` and `pgcolumnar.vm_is_visible` accepted any relation,
   including a plain heap table. `vm_selftest` does not only inspect the
   visibility map, it writes an all-visible bit into it, and a wrongly set bit is
--- a/docs/sql-reference.md
+++ b/docs/sql-reference.md
@@ -269,7 +269,7 @@ Reports how much of a table's sorted order is still in place. Returns one row:
 
 | Column | Type | Meaning |
 | --- | --- | --- |
-| `sort_key` | name[] | The clustering key in effect. It is the key the last `recluster` recorded, or the `sort_by` declared by `set_options`, or NULL. |
+| `sort_key` | name[] | The clustering key in effect. It is the key the last ordering rewrite recorded, or the `sort_by` declared by `set_options`, or NULL. |
 | `total_groups` | bigint | Row groups in the table. |
 | `sorted_groups` | bigint | Row groups written by the last ordering rewrite. |
 | `appended_groups` | bigint | Row groups written after it. |
@@ -294,6 +294,15 @@ FROM pgcolumnar.sort_status('events');
 A table that was never sorted reports zero sorted groups. An unsorted
 `pgcolumnar.vacuum` returns it to that state, because it rewrites the table
 without ordering it.
+
+`sort_key` names the columns, not the arrangement. `pgcolumnar.vacuum_sorted`
+sorts on those columns in order. `pgcolumnar.cluster` and
+`pgcolumnar.recluster` arrange the same columns on a Z-order curve, which is not
+a sort on any one of them.
+
+To read the arrangement, select `sorted_kind` from `pgcolumnar.storage`. It
+holds `lexicographic`, `zorder`, or NULL. NULL means the table was never
+ordered, or was ordered before pgColumnar recorded this.
 
 The row counts are stored rows. Deleted rows stay stored until a maintenance
 operation reclaims them, so they are still counted here. Use
--- a/src/columnar_vacuum.c
+++ b/src/columnar_vacuum.c
@@ -202,6 +202,8 @@ PgColumnarRequireTableOwnerByOid(Oid relid)
 static bool cluster_type_supported(Oid typid);
 static bytea *cluster_zorder_key(Datum *values, bool *isnull, AttrNumber *atts,
 								 int ncols, TupleDesc tupdesc);
+/* Names an ordering rewrite records as its key (defined later, #415) */
+static List *sort_key_names(TupleDesc tupdesc, AttrNumber *atts, int ncols);
 
 /* v1 refuses tables with more groups than this (one advisory lock per group) */
 #define RECLUSTER_MAX_GROUPS 8192
@@ -714,16 +716,8 @@ pgcolumnar_recluster_online(Relation rel, int ncols, AttrNumber *atts)
 		PgColumnarRetireGroup(storageId, oldGroups[i]);
 
 	/* record how far the reordered run reaches (#311) and BY WHAT (#415) */
-	{
-		List	   *keyNames = NIL;
-		int			ci;
-
-		for (ci = 0; ci < ncols; ci++)
-			keyNames = lappend(keyNames,
-							   pstrdup(NameStr(TupleDescAttr(tupdesc, atts[ci] - 1)->attname)));
-		record_online_sorted_extent(rel, storageId, writeState, stripeMark,
-									keyNames, "zorder");
-	}
+	record_online_sorted_extent(rel, storageId, writeState, stripeMark,
+								sort_key_names(tupdesc, atts, ncols), "zorder");
 
 	PopActiveSnapshot();
 	UnregisterSnapshot(snap);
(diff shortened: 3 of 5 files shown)


OffgridwithJD (commented):
Reviewed on the real 3-way merge, PG 18.4 assert (you ran PG 17, so this is a
second major). The merge-base **is** `8a71cc5`, so the merged tree is the PR head
exactly — fast-forward, nothing to compose. Baseline: **27/27 PASSED**.

Both of the attacks you asked for failed to break it. Details below, plus one
correction to a supporting claim and one gap in the derived suite list.

## 1. The layout oracle holds (the arm you were least sure of)

I forced the gate dead rather than trusting the return value —
`if (0 && skind != NULL && strcmp(skind, "zorder") == 0 && ...)` — rebuilt, and
ran your suite:

```
FAIL  recluster on the key the eager cluster already applied does nothing: got [5] want [0]
FAIL  and it rewrote no group: the physical layout is unchanged:
        got [a1653f69f1dca6b83ea0ac4a0b11d230] want [acb1979bbcd4efd23d32c8126874e8bf]
```

Exactly those two arms, everything else green. `.so` md5 `a8a8cb207bfe` (baseline)
vs `d42398d96d01` (gate dead), so the mutation reached the binary.

Worth naming explicitly because it is a property of the design: with the gate
dead **all three controls still pass**, because they assert fall-through and a
dead gate produces fall-through. The gated arm is the only thing in the suite
that can detect it. That is the right split, and it is why corroborating with the
layout rather than the return value was load-bearing.

I also reproduced your removal proof independently — reverting only the two
call-site arguments to `NIL, NULL`, keeping the helper, signature and tests:
**11 red**, `.so` `578563c86974`. Same 11 you got.

## 2. `sort_key_names()` is faithful — proven differentially, not by re-running the arms

You are right that `recluster_extent` and `recluster_gate` passing proves nothing
here, since they passed before the refactor. So I compared the **online path
against `main`** on a fixture built to break a name-extraction loop: identifiers
that need quoting, and a key order that differs from attnum order (attnums are
`id=1, "Mixed Case"=2, "k j"=3`; the key is asked for as `'k j','Mixed Case'`).

| | `main` 8a71cc5 (`.so 78ae6ee218b4`) | PR (`.so a8a8cb207bfe`) |
| --- | --- | --- |
| ONLINE, reversed-order quoted key | `{"k j","Mixed Case"}` / `zorder` | **identical** |
| ONLINE, forward-order quoted key | `{"Mixed Case","k j"}` / `zorder` | **identical** |
| gate, same key and order | `0` | `0` |
| gate, key reversed | `5` | `5` |
| EAGER `cluster` | `<NULL>` | `{"k j","Mixed Case"}` / `zorder` |
| EAGER `vacuum_sorted` | `<NULL>` | `{"k j","Mixed Case"}` / `lexicographic` |

Byte-identical on the refactored path, distinct `.so` hashes on the two arms so
they really are two binaries, and the blast radius is exactly the two eager
sites. It records the **caller's** order, not attnum order, and quoting survives.

## 3. The hard-coded kinds are exhaustively correct

`"lexicographic"` is hard-coded at the `compact_relation` site, so I checked every
path that reaches it. There are exactly two callers, and one zorder caller:

- `pgcolumnar_vacuum` → `pgcolumnar_compact_relation(rel, 0, NULL)` → `tsort == NULL`, no mark. Correct.
- `pgcolumnar_vacuum_sorted` → `pgcolumnar_compact_relation(rel, ncols, sortAtts)` → `"lexicographic"`. Sole caller, correct.
- `pgcolumnar_cluster` → `pgcolumnar_compact_relation_zorder(rel, ncols, atts)` → `"zorder"`. Sole caller, correct.

No path reaches a hard-coded kind that does not match the arrangement it applied.

## 4. Your scope decision is right, and here is the measurement for it

Agreed: no mirror gate for `vacuum_sorted`. I measured the operations on 5,000
clustered rows with half deleted, because the decision rests on which of them
reclaim:

| operation | returns | dead rows | groups |
| --- | ---: | --- | --- |
| `compact` | 0 | 2500 → **2500** | 5 → 5 |
| `compact_rewrite` | 5 | 2500 → **0** | 5 → 5 |
| `vacuum` | — | 2500 → **0** | 5 → 3 |
| `vacuum_sorted` | — | 2500 → **0** | 5 → **3** |
| `recluster` | 0 | 2500 → **2500** | 5 → 5 |

`vacuum_sorted` reclaims, so a gate keyed only on ordering would silently stop it
reclaiming. Your reasoning holds and your `recluster` row reproduces exactly.

**One correction for the separate issue**: "reclamation is `compact`'s job" is not
right. `pgcolumnar.compact()` reclaims nothing on this shape — returns 0, dead
count unchanged. The reclaimer is `compact_rewrite()` (or `vacuum`). Since that
issue will rest on the distinction, it is worth naming the right function.

## 5. `harness_selftest`: 151 PASSED

Required by the `run_all_versions.sh` SUITES edit. `eager_ordering_record` is
registered alphabetically, correct.

(My first selftest run went red with `unregistered: keynames` — that was my own
probe file left in `test/`, not your PR. Re-ran on a clean tree. The guard works.)

## 6. Gap: the derived suite list misses four

Deriving the same way over `test/*.sh` gives four suites your list does not have.
They come from `recluster` (`isolation`, `native_reclaim_cycles`,
`native_reclaim_reconcile`) and from `sort_key` (`projections`) — your list has
`native_reclaim` and `native_reclaim_frag` but not the `_cycles` / `_reconcile`
siblings. I ran all four against the PR `.so` (`a8a8cb207bfe`):

| suite | result |
| --- | --- |
| `isolation` | PASSED |
| `native_reclaim_cycles` | PASSED, 12/0 |
| `native_reclaim_reconcile` | PASSED, 11/0 |
| `projections` | PASSED, 64/0 |

Nothing red, so this changes no verdict — but the derivation missed them, and
that is the part worth fixing for next time rather than the outcome.

## 7. CHANGELOG and docs

Accurate against the artifact. They carry the two things a reader needs and that
are easy to leave out: the **visible output change** (`sort_status` now reports
the applied key on an eagerly clustered table) and the **pre-#758 residue** (a
storage ordered by an earlier build still has NULL and still falls back, so a
reader that must not be wrong reads `sorted_kind` and treats NULL as unknown).
`docs/sql-reference.md`'s "names the columns, not the arrangement" is the right
distinction to draw.

## Verdict

No findings against the change. **Not approving yet**: `suites (PG 17)` and
`suites (PG 18)` are still `in_progress` (10/12 complete, **0 failed**). I will
approve when those land green — local green is not the bar.

OffgridwithJD (approved):
Approving. CI is now **12/12 green** on `66bb55c` and `mergeStateStatus` is `CLEAN`,
which was the only thing I was holding on.

Verified the suites job actually **ran** the new suite rather than skipping it — a
green `suites` job reports PASS and SKIP identically, so the summary line is the
thing to read:

- `suites (PG 17)` — `eager_ordering_record=PASS` (208 ran, 9 skipped)
- `suites (PG 18)` — `eager_ordering_record=PASS` (211 ran, 6 skipped)

The four suites my derivation added that the PR body's list did not name —
`isolation`, `native_reclaim_cycles`, `native_reclaim_reconcile`, `projections` —
are all `PASS` in both CI majors as well, matching what I measured locally.

Nothing further from me. The findings in my review above are for the follow-up
issue (`compact` vs `compact_rewrite`) and for the next derivation, not for this
change.

Issue #761, opened 2026-08-27
sort_status cannot report sorted_kind, so the reporter still cannot tell a sorted table from a Z-ordered one


jdatcmd:
After #758, `pgcolumnar.storage.sorted_kind` records `lexicographic` or `zorder` for every
ordering rewrite. `pgcolumnar.sort_status()` still cannot report it: its OUT columns are
`sort_key`, `total_groups`, `sorted_groups`, `appended_groups`, `sorted_rows`,
`appended_rows`.

So the user-facing reporter still cannot express the distinction the catalog now holds. A
user reads `sort_key = {k,j}` and cannot tell a `vacuum_sorted` table from a `cluster` table.
`sort_key` names the columns; it does not say whether they are sorted or arranged on a
Z-order curve, and Z-order over two or more columns is not a sort on any one of them.

`docs/sql-reference.md` now tells the reader to select `sorted_kind` from
`pgcolumnar.storage` directly. That works, but `pgcolumnar.storage` carries no GRANT: it is
readable by superusers, whereas `sort_status()` is SECURITY DEFINER and gated on
`require_caller_select`, so a table's own owner can read it. The documented workaround is
therefore not available to the people the function exists for.

## The work

Add `sorted_kind text` as an OUT column of `pgcolumnar.sort_status(regclass)`, sourced from
`pgcolumnar.storage.sorted_kind`, NULL when unknown.

## Why it is not a one-line change

Adding an OUT parameter changes the function signature, so it cannot be a
`CREATE OR REPLACE`. It needs `DROP FUNCTION` + `CREATE FUNCTION` in a new
`pgcolumnar--1.0-alpha2--1.0-alpha3.sql`, which does not exist yet: every `[Unreleased]`
change so far has been shared-library only. So this issue is the one that starts the alpha3
upgrade script, and the release playbook applies (generate the script from a catalog delta,
then converge definition, ACL and comment against a fresh install).

Worth doing as part of whatever else needs a schema change before alpha3, rather than opening
the upgrade script for one column.

## Arms

- a fresh install and an upgraded one converge on the same definition, ACL and comment;
- `sort_status` reports `lexicographic` after `vacuum_sorted`, `zorder` after `cluster` and
  after `recluster`, NULL after an unsorted `vacuum`;
- the privilege boundary is unchanged: `sort_status_privilege` still denies a role with no
  SELECT, asserting SQLSTATE `42501`, and still admits the owner.

[...]

Issue #753, opened 2026-08-26
The Gather cost is charged per row and is blind to tuple width, which holds nine of eleven measured cells and an extension cannot change it


jdatcmd:
`docs/limitations.md`, "Parallel scans", records this as known and unfixed:

> On a narrow but selective scan the planner still prefers the serial plan even where the
> parallel one is faster. Measured: a three-column selective scan of a four-million-row table
> ran up to **2.5x faster in parallel**, and the planner chose serial anyway.
>
> The cause is the cost of the Gather node. The planner prices each row passed from a worker
> to the leader at `parallel_tuple_cost`, the same rate it uses for a heap scan. A columnar
> scan ships fewer, already-decoded rows, so that rate overstates its Gather cost. On the
> shape above the Gather cost reached almost the whole cost of the serial scan. **The serial
> plan won on cost while losing on the clock.**

So this is a costing asymmetry with a measured number attached, not a missing feature: the
parallel path exists and is built, and on a wide projection it is already chosen.

## Relationship to #369

#369 (closed) is the same class one node higher: the grouped parallel aggregate arm lost to
the serial node because the Finalize was priced off `estimate_num_groups` on an expression
key. Worth reading before starting, because the shape of the fix and the traps in measuring
it will be similar. This issue is about the **scan**, not the grouped aggregate.

## The care this needs

The doc is explicit that the default rate is conservative for any row-returning parallel
scan, not only a columnar one, and that a heap scan on the same shape also flipped. So the
fix must not be "discount our Gather until we win": it has to be defensible as pricing what
a columnar scan actually ships, and it needs a control on a shape where the serial plan is
genuinely right, or it will just move the error to the other side.

Per the house rule, calibrate the threshold in the context the check runs in, and assert the
chosen plan on both arms rather than the runtime alone.

jdatcmd:
## Measured. The phenomenon is real. The stated cause is not.

Nothing below inherits the published 2.5x. `docs/limitations.md` names only "a three-column
selective scan of a four-million-row table": no selectivity, no column types, no fixture, so
it is not reproducible from the doc.

**Fixture.** 4,000,000 rows, 14 columns (8 int of varying cardinality, 4 md5 text, numeric,
float8), all `hashint4`-derived so nothing is trivially compressible. Columnar 383 MB against
a heap twin at 845 MB. PG 17. A first fixture that made every column compressible produced
[...]

PR #762, opened 2026-08-27
Give the columnar scan the order a sorted rewrite left the rows in, so ORDER BY on a clustered table plans no Sort

### LINKED ISSUE #751: The columnar scan never advertises sort order, so a clustered table pays for a layout the planner cannot use

jdatcmd:
Closes #751.

Every `pathkeys` field in the tree was `NIL`, so a table `pgcolumnar.vacuum_sorted` had
physically ordered still planned a `Sort` to be read in that order.

## The measurement the issue asked for

4,000,000 rows in 27 row groups, `vacuum_sorted('t','k','j')`, `sort_status` reporting key
`{k,j}`, 27 sorted groups, 0 appended, 0 inversions on `k` in scan order. Instructions
retired by one pinned backend (`cpu_core` at 100.00% on every arm; wall clock is not a fair
instrument on this host). Both arms return byte-identical answers, and the plan is asserted
on both.

| query | Sort (`enable_sorted_pathkeys=off`) | no Sort (`on`) | ratio |
| --- | ---: | ---: | ---: |
| `ORDER BY k, j LIMIT 10` | 5,545,800,871 | 14,133,342 | **392x** |
| `ORDER BY k, j OFFSET 3999990` | 8,395,455,193 | 3,392,835,665 | **2.47x** |

The second row is the whole relation consumed, so it is the sort itself: 2,099 instructions
a row against 848, about 1,251 a row the sort was costing. The first is the early stop the
sort made impossible. Re-run with the arm order reversed, the four figures reproduce within
0.8%.

The first attempt at this measurement was void and is worth recording: `(g*7919)%1000000`
overflows int32 above g = 271,181, the fixture INSERT failed with `integer out of range`, and
every arm measured an empty table while reporting clean numbers. `rows=4000000` is now
asserted before any arm runs.

## Why this is a correctness change, not a performance one

**A claim the rows do not satisfy is a wrong answer, not a slow plan**, because the `Sort`
that would have fixed the order is gone. No correctness test on unordered data notices, since
its plan keeps a `Sort` either way. The claim is refused unless all of:

1. the last ordering rewrite was **lexicographic and recorded itself as such** (#758) -- a
   Z-order run and an unknown one are refused rather than guessed at, and after #758 they can
   be told apart;
2. **every row group lies inside the recorded run**, so one appended or updated row retracts
   it;
3. **no sort column is collatable** (see below);
4. the requested ordering is a **prefix of the recorded key**, taken from the same type's
   `<` operator the rewrite's tuplesort used. Asking `build_expression_pathkey` for that
   operator is what makes it ascending / NULLS LAST / in the key's own collation, rather than
   a restatement of those three properties that could drift from the sort that produced them.

### Two wrong answers found during development, both now arms

**A collation-only ALTER changes the order without rewriting a row.**
`ALTER TABLE t ALTER COLUMN k TYPE text COLLATE "en_US.utf8"` on a column already
`text COLLATE "C"` needs no transformation, so PostgreSQL updates `pg_attribute` and leaves
every stored row where it is: same storage id, groups unchanged, mark intact. Measured on
4,000 rows chosen so the two collations disagree (in C, `'B'` 0x42 sorts before `'a'` 0x61):

```
plan after the ALTER : Custom Scan (PgColumnarScan) on b      <- no Sort
ORDER BY k LIMIT 3   : AA1003|AA1011|AA1019                    <- the C order
heap oracle, en_US   : aa10|aa1002|AA1003                      <- the answer
```

Refused on `OidIsValid(att->attcollation)`. The claim that this is **exact** for "has an
ordering that can change under us" is measured per type family rather than reasoned:

| type | `attcollation` | claim | why that is safe |
| --- | --- | --- | --- |
| domain over `text COLLATE "C"` | `C` | refused | carries the base type's collation |
| `text[]` | `default` | refused | an array of a collatable type is collatable |
| composite over `text` | `InvalidOid` | **claimed** | PostgreSQL refuses `ALTER TYPE ... ALTER ATTRIBUTE ... TYPE text COLLATE x` with *"cannot alter type because column uses it"* while any column stores the type |
| range over `text` | `InvalidOid` | **claimed** | `typcollation` is 0, the collation is fixed at `CREATE TYPE`, and there is no DDL to change `pg_range.rngcollation` |
| enum | `InvalidOid` | **claimed** | `ADD VALUE ... BEFORE` slots a value in without renumbering existing ones, and the new value cannot be in stored rows |

The composite and enum rows are the two that looked like holes, and both are closed by
PostgreSQL rather than by this extension. Each has an arm: the composite arm runs the `ALTER`
and matches the refusal text, so it goes red if that ever stops being true; the enum arm
asserts a heap-oracle **answer** after `ADD VALUE 'a' BEFORE 'b'`, not only a plan. **This costs text sort keys an ordered path**, and
that is a stated limitation, not caution. Lifting it means recording the collations beside the
names, which needs a catalog column, so it belongs with #761's upgrade script.

**A cached plan outlives the ordering it was planned on.** The claim lives in a plan, and no
catalog object the plan cache watches changes when rows are appended. Measured before the fix:
a prepared `SELECT k FROM pc ORDER BY k NULLS LAST LIMIT 5` returned `0,0,0,0,0` after 600
rows with `k` from -1 to -600 were appended below the run, where the answer is
`-600,-599,-598,-597,-596`. A group written outside the run now invalidates the relation's
cached plans, gated so it fires only when the storage has a mark and this group falls outside
it: once per row group, never on a bulk load into an unordered relation.

It sits in `pgcolumnar_flush_row_group`, which is below all **six** of its own call sites in
`columnar_write_state.c` (including `pgcolumnar_flush_row_group(w->innerWs)`) and is the only
caller of `PgColumnarInsertRowGroupRow`. "One caller" understates what it covers, which is why
each write path gets its own arm rather than a reading of the grep.

## Where the claim is NOT attached

Only the serial base scan. `columnar_customscan.c:2373` (projection) and `:2508` (parallel
partial) keep `NIL`, and both now have arms: a projection is a separate physical layout with
its own sort key in `pgcolumnar.projection.sort_key`, unrelated to `storage.sorted_by`; and
pathkeys on a partial path describe one worker's output, which a plain `Gather` interleaves.
The four `columnar_vector.c` sites (ungrouped and grouped vector aggregates) also keep `NIL`;
the grouped one is what would serve `GROUP BY k` and is left for a follow-up.

## Test

`test/sorted_pathkeys.sh`, **103 checks**, registered in `run_all_versions.sh`. Two groups:

- **CLAIM arms** -- the shapes where the ordering is real must lose the `Sort`. These are what
  goes red without the change (6 arms).
- **REFUSE arms** -- every shape where the rows are not in the claimed order must keep the
  `Sort` **and** return the same rows in the same order as a heap table holding identical
  data. **These pass trivially when no pathkeys exist at all**, so they are proved by
  over-claiming mutations rather than by being green.

One CLAIM arm is about the *skip-and-continue* branch rather than a refusal: a leading key
whose equivalence class is a constant (`WHERE k = 5`) is skipped and the prefix continues,
mirroring core's `build_index_pathkeys`, because every row the scan returns shares that value
and the rows are therefore ordered by what follows it. `ORDER BY j` under `WHERE k = 5` on a
`(k,j)` run plans no `Sort` and answers in `j` order against the heap twin; the bare
`ORDER BY j` refusal arm is its control.

Refusals covered: `DESC`; `NULLS FIRST`; a non-prefix of the key; a column not in the key;
the key columns in the wrong order; an appended tail (with values *below* the run, so a wrong
claim is a wrong answer and not merely an unordered one); a Z-order run; a declared-but-never-
applied `sort_by`; an unsorted `vacuum`; a collatable column; a collation-only ALTER; a type
change; one `UPDATE`d row; a renamed column; a reclaiming `compact_rewrite`; the GUC off.

## Removal proofs

Each mutation was built and the installed `.so` fingerprinted, so no arm is reported against a
binary that does not contain it.

| mutation | `.so` | arms red |
| --- | --- | --- |
| baseline, fixed | `82558b862d50` | 0 of 92 |
| feature absent (main) | `2340966b22f6` | the 6 CLAIM arms |
| drop the `lexicographic` gate | `0a9c681abf28` | Z-order refusal **+ its heap-oracle answer arm** |
| drop the run-coverage gate | `a580989cda9c` | tail refusal **+ 2 answer arms + the cached-plan arm** |
| drop the collatable refusal | `3b90b692c831` | 2 refusals **+ 2 heap-oracle answer arms** |
| `lt_opr` -> `gt_opr` | (rebuilt) | `DESC` refusal **+ its heap-oracle answer arm** |
| disable the invalidation | `fe5cb0c3d751` | all **6** cached-plan arms |
| end the prefix on a redundant key instead of skipping it | `0543d84a5755` | the constant-leading-key arm |

Every mutation reddens at least one *answer* arm, not only a plan arm.

## Three vacuous arms of my own, caught before they shipped

Recorded because the shape recurs and the reviewer's per-path suggestion is what exposed the
first two:

1. Five write-path arms named the cached plan and ended in an equivalent **ad-hoc SELECT**
   instead of `EXECUTE p`. A fresh query is planned from scratch and can never observe a stale
   plan, so all five stayed green with the invalidation disabled while the original arm went
   red alone.
2. The `parallel_flush` dispatch premise ran on the same table as its cached-plan arm, so it
   appended a tail *before* the plan was prepared and there was no ordered path left to
   retract. Vacuous again, in the same suite, for a different reason. Separate fixtures now.
3. A `bulk_parallel_writer` arm that was a plain-INSERT arm wearing a GUC. That GUC does not
   change the write path at all; its only consulted site is
   `PgColumnarInsertNativeStorageRow`, where it skips an advisory lock, and
   `pgcolumnar.parallel_copy` sets it in its loader sessions. Removed and replaced with a real
   `pgcolumnar.parallel_copy` arm, which is the one write path where separate **backends**
   flush groups in their own (prepared) transactions, so it is the only place the invalidation
   crosses a process boundary. It needs `max_prepared_transactions >= workers`, set in
   `PGC_EXTRA_CONF` because it cannot be raised without a restart, and the loaded row count is
   asserted first so an under-provisioned run cannot read as a pass.

Two of my own checks were also caught by `harness_selftest`, and both were real: a
`printf | grep -q` in the Gather arm (the EPIPE shape selftest 080 exists to stop, now
`grep -c`), and selftest 260 counting only `diff_query_ordered` users as needing
`pgc_check_ordered_oracle` -- this suite calls the oracle directly through `pgc_seq_hash`, so
260's grep is widened rather than the premise dropped.

## Suites run

**Full PG17 matrix, `test/run_all_versions.sh`: 213 of 218 suites ran, ALL PASSED**, 5 skipped
(`analyze_differential`, `analyze_function`, `native_repack`, `pg19_vacuum_options`,
`temporal` -- all pre-existing PG18+/optional skips). The whole tree was run rather than a
derived subset, because this change alters path selection and a `grep -l` for the identifiers
it touches returns 98 of the suites.

That matrix run included `sorted_pathkeys` at 84 checks. The last 8 arms (`parallel_copy`,
the split `parallel_flush` fixtures, the TRUNCATE invariant) were added after it and have been
run standalone, 92 of 92, plus the removal proof above. CI covers the full matrix on the final
tree.

## Deliberately not in this PR

- **Text sort keys get no ordered path**, per the collation limitation above. Needs the
  collations recorded in the catalog.
- **`GROUP BY k` still gets no Group Aggregate.** That consequence is served by
  `PgColumnarTryGroupAggPath` in `columnar_vector.c`, not by any path this PR touches.
- **A sorted run plus a sorted tail is a merge, not a scan**, and is not attempted.

## One note on `truncate_useless_pathkeys`

It is called at the end for hygiene, and it can only ever shorten the claim to a prefix of
itself. A prefix of a valid ordering is a valid ordering, and a shorter claim makes
`pathkeys_contained_in` more likely to fail, which adds a `Sort` rather than removing one. So
it cannot turn a sound claim into an unsound one; the worst it can do is cost an optimisation.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,57 @@ pinned at `1.0-dev` or `1.0-alpha`, each true until the next version shipped.
 
 ### Added
 
+- The columnar scan now tells the planner the order a sorted rewrite left the
+  rows in, so `ORDER BY` on that key plans no `Sort`. Every `pathkeys` field in
+  the tree was `NIL`, so a table that `pgcolumnar.vacuum_sorted` had physically
+  ordered still paid a full sort to be read in that order, and
+  `ORDER BY ... LIMIT n` could not stop early.
+
+  Measured on 4,000,000 rows in 27 row groups, `vacuum_sorted('t','k','j')`,
+  `sort_status` reporting key `{k,j}`, 27 sorted groups, 0 appended, 0
+  inversions on `k` in scan order. Instructions retired by one pinned backend,
+  same query, `pgcolumnar.enable_sorted_pathkeys` off against on, and both arms
+  return byte-identical answers:
+
+  | query | Sort (off) | no Sort (on) | ratio |
+  | --- | ---: | ---: | ---: |
+  | `ORDER BY k, j LIMIT 10` | 5,545,800,871 | 14,133,342 | 392x |
+  | `ORDER BY k, j OFFSET 3999990` | 8,395,455,193 | 3,392,835,665 | 2.47x |
+
+  The second row is the whole relation consumed, so it is the sort itself:
+  2,099 instructions a row against 848, about 1,251 a row that the sort was
+  costing. The first is the early stop the sort made impossible. Re-run with the
+  arm order reversed the four figures reproduce within 0.8%.
+
+  **A claim the rows do not satisfy is a wrong answer, not a slow plan**, since
+  the `Sort` that would have fixed the order is gone. The claim is therefore
+  refused unless all of: the last ordering rewrite was lexicographic and
+  recorded itself as such (#758), so a Z-order run and an unknown one are
+  refused rather than guessed at; every row group lies inside the recorded run,
+  so one appended or updated row retracts it; no sort column is collatable; and
+  the requested ordering is a prefix of the recorded key, ascending, NULLS LAST,
+  which is what the rewrite's own tuplesort applies.
+
+  Collatable sort columns are refused because only the column names are
+  recorded. `ALTER TABLE t ALTER COLUMN k TYPE text COLLATE "en_US.utf8"` on a
+  column already `text COLLATE "C"` needs no transformation, so PostgreSQL
+  updates `pg_attribute` and leaves every row where it is: same storage, mark
+  intact, ordering silently a different one. A text sort key therefore gets no
+  ordered path. Lifting that needs the collations recorded beside the names.
+
+  The claim lives in a plan, and no catalog object the plan cache watches
+  changes when rows are appended, so a group written outside the run now
+  invalidates the relation's cached plans. Without it a prepared
+  `ORDER BY k NULLS LAST LIMIT 5` answered from the ordered run alone after 600
+  rows were appended below it. It fires once per row group and only on a
+  relation that has a mark.
+
+  Only the serial scan carries the claim. The parallel partial path and the
+  projection path keep `NIL`: workers finish in any order, and a projection is a
+  separate layout with its own sort key. New GUC
+  `pgcolumnar.enable_sorted_pathkeys`, on by default. New suite
+  `sorted_pathkeys`, 84 checks. (#751)
+
 - A predicate on `date_trunc(unit, ts)` now drives chunk-group skipping for the
   ORDERED comparisons too, not only equality. #739 inverted `=` and declined
   every other operator, so `date_trunc('day', ts) >= '2024-02-01'` still read
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -75,6 +75,7 @@ disk. It never changes the values that a table returns.
 | --- | --- | --- | --- |
 | `pgcolumnar.enable_index_only_scan` | boolean | `on` | Allow index-only scans on columnar tables, served by the columnar visibility-map fork. Set to `off` to force a plain index scan. |
 | `pgcolumnar.enable_projection_scan` | boolean | `on` | Let the planner scan a covering projection instead of the base table when one serves the query better. |
+| `pgcolumnar.enable_sorted_pathkeys` | boolean | `on` | Let the columnar scan tell the planner the order a sorted rewrite left the rows in, so `ORDER BY` on that key needs no `Sort`. Only a lexicographic run recorded by `pgcolumnar.vacuum_sorted` that still covers every row group is advertised. Set to `off` to restore the behaviour of always planning a `Sort`. |
 
 ### Maintenance and disk reclaim
 
--- a/src/columnar.h
+++ b/src/columnar.h
@@ -213,6 +213,7 @@ extern bool pgcolumnar_enable_index_only_scan;	/* allow index-only scans (gap 28
 extern bool pgcolumnar_bulk_parallel_writer;	/* internal: parallel_copy loader skips the storage-row creation lock (#300) */
 extern bool pgcolumnar_parallel_flush;	/* dispatch the per-column stripe flush across bgworkers (#445 slice 3) */
 extern bool pgcolumnar_enable_projection_scan;	/* scan a covering projection (gap 26) */
+extern bool pgcolumnar_enable_sorted_pathkeys;	/* advertise a sorted run's order (#751) */
 extern bool pgcolumnar_enable_index_fetch_penalty;	/* price a columnar index scan's per-row fetch (#355) */
 
 /*
--- a/src/columnar_compat.h
+++ b/src/columnar_compat.h
@@ -497,6 +497,20 @@ PgColumnarReindexRelation(Oid relid, int flags)
 #endif
 
 /* -------------------------------------------------------------------------
+ * build_expression_pathkey() lost its nullable_relids parameter in PG16, when
+ * outer-join-aware equivalence classes replaced it (core commit 2489d76c).
+ * Every caller here passes no nullable relids, so the wrapper takes the modern
+ * five arguments and fills NULL on PG15.
+ * ------------------------------------------------------------------------- */
+#if PG_VERSION_NUM >= 160000
+#define COLUMNAR_BUILD_EXPRESSION_PATHKEY(root, expr, opno, rel, create_it) \
+	build_expression_pathkey((root), (expr), (opno), (rel), (create_it))
+#else
+#define COLUMNAR_BUILD_EXPRESSION_PATHKEY(root, expr, opno, rel, create_it) \
+	build_expression_pathkey((root), (expr), NULL, (opno), (rel), (create_it))
+#endif
+
+/* -------------------------------------------------------------------------
  * Executor index maintenance on the import path (#168).
  *
  * Two signatures move independently here, so they get one macro each rather
--- a/src/columnar_customscan.c
+++ b/src/columnar_customscan.c
@@ -26,6 +26,7 @@
  */
 #include "columnar.h"
 
+#include "columnar_metadata.h"
 #include "columnar_customscan.h"
 #include "columnar_reader.h"
 #include <math.h>
@@ -78,6 +79,7 @@
 bool		pgcolumnar_enable_custom_scan = true;
 /* GUC: let the planner scan a covering projection instead of the base (gap 26) */
 bool		pgcolumnar_enable_projection_scan = true;
+bool		pgcolumnar_enable_sorted_pathkeys = true;
 /* GUC: price a columnar index scan's per-row heap fetch (#355) */
 bool		pgcolumnar_enable_index_fetch_penalty = true;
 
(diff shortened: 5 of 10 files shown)


OffgridwithJD (commented):
Reviewed on the real 3-way merge at `f653aa1` (merge-base **is** `098b7b9`, so the
merged tree is the PR head), PG 18.4 assert — a second major, since you gated on
PG 17. **`sorted_pathkeys` 103/103**, **`harness_selftest` 151/151**, tree clean.

The `.so` md5 is identical (`814667cc5c90`) at `d1ea1b9` and `f653aa1`, which
independently confirms the last two pushes were test-and-docs only.

One substantive finding, one enumeration gap that is **not** a wrong answer, and
one guard nit. Nothing that blocks.

## 1. Plan-time cost is paid by queries that can never use the ordering

Condition 2 reads the **whole row-group list** at plan time, and nothing before it
asks whether this query could use an ordering at all. So every plan of every query
against a table carrying a lexicographic mark pays it — including queries with no
`ORDER BY`.

Measured on `pg18n` (non-assert), 1M rows, **1,000 row groups**, A/B on
`pgcolumnar.enable_sorted_pathkeys`, 15 interleaved reps, planning time from
`EXPLAIN (SUMMARY ON)` so no execution is in the number:

| table | GUC on | GUC off | ratio |
| --- | ---: | ---: | ---: |
| marked `lexicographic`, query has **no ORDER BY** | 1.079 ms | 0.922 ms | **1.17x** |
| never sorted (control) | 1.018 ms | 1.007 ms | 1.01x |

The control is the part that makes it a mechanism rather than noise: an unsorted
table returns at condition 1 and pays only the storage-row probe, so the GUC moves
nothing. A 4,000-group run showed a larger delta (+0.41 ms) but its control drifted
to 1.13x, so I am reporting that one as direction only, not as a clean number.

Absolute cost is small — 0.16 ms — and I would not hold the PR for it. But it grows
with group count, it is paid per plan, and it is entirely avoidable: core has
`has_useful_pathkeys(root, rel)` declared in `optimizer/paths.h` **immediately above
`truncate_useless_pathkeys`**, which you already call. Same signature on PG 15, 18
and 19 (checked all three `includedir-server` headers), so no compat macro. An early
return on `!has_useful_pathkeys(root, rel)` skips the group read for exactly the
queries that would have had their claim truncated to NIL at the end anyway. It is the
same notion you already compute, just computed before the expensive part instead of
after it.

## 2. The collation enumeration is incomplete — three more shapes reach the claim

You asked for a sixth family. I found three more shapes with
`attcollation = InvalidOid` that are **CLAIMED** (measured as 0 `Sort` nodes in the
plan after `vacuum_sorted`), none of which the five-row table names:

| shape | `attcollation` | claim |
| --- | --- | --- |
| **multirange** over `text` (`tmultirange`) | `0` | not refused |
| **array of composite** (`ctype[]`) | `0` | **claimed** |
| **composite nesting a composite** / composite holding `text[]` | `0` | **claimed** |

None is a wrong answer, and the reason is stronger than I expected. I went after the
two that looked reachable — a table whose only column is `ctype[]`, and a table whose
only column is an outer composite nesting `ctype2`, so that no column stores the
inner type *directly*:

```
column is ctype[]      : ALTER TYPE ctype  ALTER ATTRIBUTE f ... -> ERROR: cannot alter type "ctype" because column "arrc.k" uses it
                         ... CASCADE                             -> same error
column is outer2(ctype2): ALTER TYPE ctype2 ALTER ATTRIBUTE f ... -> ERROR: cannot alter type "ctype2" because column "nest.k" uses it
                         ... CASCADE                             -> same error
```

PostgreSQL's dependency tracking follows **through the array and through the nesting
composite**, and names the columnar column for a type the table only holds indirectly.
CASCADE cascades to dependent types, never to a table storing one. So the protection
holds for the whole family, not just the direct case you tested.

The suggestion is therefore not to add three rows to the table. It is to record the
**invariant**, since that is what covers the shapes neither of us enumerated:
a type with `attcollation = InvalidOid` is safe when its ordering derives either from
a collation fixed at `CREATE TYPE` (range, multirange — no DDL reaches
`pg_range.rngcollation`) or from the field collations of a stored composite, which
PostgreSQL will not let you alter while any table stores it, directly or nested.

I also reproduced your two closed branches independently: the composite ALTER is
refused with and without CASCADE, and `ADD VALUE ... BEFORE` preserves relative order
including through the **renumbering** path — after 40 `BEFORE` insertions into one
gap, `m < z` still holds and `m`'s `enumsortorder` is still 1.

**One boundary worth stating in the PR**, because it is the honest edge of the claim
and it is currently unstated: replacing a type's default btree opclass, or
`CREATE OR REPLACE`-ing a comparison function, changes the ordering semantics without
touching the table and without a collation. That is outside the trust boundary in
exactly the way it is for a btree index, which would need a `REINDEX` — and saying so
is what tells a reader this claim is no weaker than an index's.

## 3. `selftest/260` guard nit

The widened pattern is `'^[[:space:]]*diff_query_ordered |pgc_seq_hash '`. In ERE the
`^[[:space:]]*` anchor binds only to the **first** alternative, so `pgc_seq_hash `
matches anywhere on a line, comments included. Leaving it unanchored is *correct and
necessary* — your real call is mid-line inside `$(...)` — but it means prose about the
oracle now counts as a use of it. `selftest/210` already solved exactly this for
itself and the idiom is one line: it strips comments before sweeping
(`sed 's/[[:space:]]*#.*$//'`, line 31), specifically so that prose about the trap
cannot trip the sweep enforcing it.

Pre-existing but worth knowing while you are in this check: it compares **counts, not
sets**, so a suite that uses the oracle without the premise and another that asserts
the premise without using it would cancel out. I diffed the two file lists today and
they are identical (`differential`, `native_format`, `sorted_pathkeys`,
`sorted_projection`), so nothing is hidden right now.

## 4. Agreed on your withdrawn point 3

For the record, independently: `truncate_useless_pathkeys` is a prefix truncation, a
prefix of a valid ordering claim is itself valid, and a shorter claim makes
`pathkeys_contained_in` more likely to fail — which **adds** a Sort. It cannot turn a
sound claim unsound. Your reasoning is right and belongs in the body where you put it.

## Verdict

**Not approving yet**: `suites (PG 17)` and `suites (PG 18)` are still `in_progress`,
10/12 complete, 0 failed. Same bar as #759 — local green is not the bar, and I want to
see the new suite reported as `PASS` rather than `SKIP` in the summary line. I will
approve when they land green; none of the three findings above needs to hold it.

OffgridwithJD (approved):
Approving. CI is **12/12 green** on `f653aa1`, `mergeStateStatus` is `CLEAN`, and the
new suite genuinely **ran** rather than skipping — the summary line is the thing to
read, since a green `suites` job reports PASS and SKIP identically:

- `suites (PG 17)` — `sorted_pathkeys=PASS` (209 ran, 9 skipped)
- `suites (PG 18)` — `sorted_pathkeys=PASS` (212 ran, 6 skipped)

Locally on the merged tree: `sorted_pathkeys` 103/103 and `harness_selftest` 151/151
on PG 18.4 assert, tree clean.

## On `planner_choice_quality=SKIP` — not this PR, and deliberate

It is pre-existing: the same `SKIP` appears on **both** legs of #759's run, before any
of this. The mechanism is `run_all_versions.sh:706` — under `PGC_SKIP_TIMING` a suite
matching `is_timing_suite` is not invoked at all, and the driver writes
`SKIPPED (ran no checks)`.

It is also a considered decision rather than an oversight. The rationale is in
`is_timing_suite` itself:

> planner_choice_quality's only assertion is a wall-clock ratio. Left out of this list
> it still RAN under PGC_SKIP_TIMING, skipped the ratio, and reported PASS on the
> strength of its premises — so a regression of #434 would have been reported green by
> the suite that exists to catch it. A suite whose subject is dropped has not passed.

I confirmed the behaviour it is guarding against: invoked directly with
`PGC_SKIP_TIMING=1`, the suite reports **PASSED, 7 checks**, with only the two
wall-clock ratios skipped. The 7 that pass are fixture and plan-identification
premises. So the choice is exactly right — reporting PASS there would be a green built
on premises with the subject removed.

**But there is a real defect next to it, and it is a stale comment that says the
opposite of the code.** In `runs_alone()`, ~35 lines below:

> planner_choice_quality wants the same split, the other way around from replication.
> … **But it is deliberately not in is_timing_suite**: its premises are plan-shape
> assertions, and those are worth running in CI.

It **is** in `is_timing_suite`. Two comments in one file describe opposite
arrangements, and a reader who finds the second one will believe CI exercises those
plan-shape assertions when it does not. Worth a one-line fix; not this PR's job.

**The forward-looking part, which matters more than the comment.** The accepted
trade-off is that no plan-shape assertion from that suite runs in CI. That is fine for
#762, whose plan assertions live in `sorted_pathkeys` and do run. It is not fine for
**#753**, which is a Gather-costing change whose natural regression guard is exactly
`planner_choice_quality` — so #753 will need either plan-shape arms in a suite that CI
actually invokes, or an explicit local gate, or the split those two comments are
circling. Worth settling before that one is built rather than after.

None of my three review findings blocks this. The `has_useful_pathkeys` early-out is
the only one I would actually schedule.

PR #763, opened 2026-08-27
Do not read the row-group list for a query that cannot use an ordering, and fix two guards the last PR bent

jdatcmd:
Follows #762 (merged as `4d32f65`). These are the three findings from that PR's review that
arrived after it was approved, so they are here rather than pushed under a stale approval.

## 1. A query that cannot use an ordering paid to find one out

Deciding whether to claim reads the whole row-group list, which grows with the relation, and
nothing asked first whether this query could use an ordering at all. Any claim returned for
such a query was discarded by `truncate_useless_pathkeys` at the end anyway.

`has_useful_pathkeys(root, rel)` is that same notion computed **before** the expensive part
instead of after it. It is declared in `optimizer/paths.h` with this signature on PG15, 16,
17, 18 and 19 -- immediately above `truncate_useless_pathkeys`, which the code already calls
-- so no compat wrapper.

**Measured as buffers touched during PLANNING, not planning-time milliseconds.** A count
cannot be moved by another process; wall clock on this host is not a fair instrument.
1,000,000 rows in 1,000 row groups, relation marked lexicographic, GUC on against off:

| query | before | after |
| --- | ---: | ---: |
| `SELECT count(*) FROM t WHERE j = 3` | 142 / 121 | **121 / 121** |
| `SELECT k FROM t ORDER BY k LIMIT 10` | reads to decide | reads to decide |

The second row is the control. Without it the first arm would be satisfied by a function that
reads nothing at all.

Two things about that arm, both learned by getting it wrong first:

- it reads the **second** of two plans in one session. A first plan in a fresh backend also
  fills the catalog caches, and that first-touch cost differs by a couple of buffers between
  two `psql` invocations. Asserted as equality it flaked by 2 either way.
- the fixture is sized from the **row** count, because `stripe_row_limit` has a floor of 1000
  and the `row_group` catalog holds one row per stripe. A hundred-group fixture put the effect
  at +6, too close to the two-buffer noise floor to discriminate; a thousand groups puts it at
  +21.

## 2. The collation refusal now records the invariant, not a list

The review enumerated three more shapes with `attcollation = InvalidOid` that are claimed
(multirange over text, an array of composite, a composite nesting a composite) and, better,
made the point that adding rows is the wrong fix. So the comment now states the invariant:

> A type whose `attcollation` is `InvalidOid` orders either by nothing collation-dependent at
> all, or by a collation PostgreSQL will not let change under a stored value: a range or
> multirange fixes its collation at `CREATE TYPE` and no DDL reaches `pg_range.rngcollation`,
> and a composite orders by its field collations, which cannot be altered while any table
> stores the type.

That last protection **follows dependencies through arrays and through nesting composites** --
a column of `ctype[]`, or of a composite that merely nests `ctype`, still makes
`ALTER TYPE ctype ALTER ATTRIBUTE ...` fail with *"cannot alter type because column uses it"*,
with or without `CASCADE`. So the protection is stronger than the direct case #762 tested.

The honest boundary is stated too: replacing a type's default btree opclass, or
`CREATE OR REPLACE` on a comparison function, changes ordering semantics without touching the
table and without any collation. That is outside the trust boundary in exactly the way it is
for a btree index that would need a `REINDEX`, which is what tells a reader the claim is no
weaker than an index's.

## 3. selftest 260's sweep had three defects, one of them mine and new

**Mine, added in #762:** in ERE, `^[[:space:]]*a|b` anchors only the first alternative. Adding
`pgc_seq_hash` unanchored was necessary (that call is mid-line inside `$(...)`) but it made
prose *about* the oracle count as a use of it. Comments are now stripped before the sweep,
which is what selftest 210 already does, for the same reason.

**Also mine, and the sharper one:** the first version of that sweep used `sed | grep -q`. That
is the EPIPE shape **selftest 080 exists to forbid** -- the reader exits on its first match,
`sed` takes `SIGPIPE`, `pipefail` calls the pipeline failed, and the file is silently *not
counted*. It reported 3 users against 2 premises where both sets are four files. 080's own
regex did not catch it because that regex scopes to `echo` and `printf` writers, not `sed`.
`grep -c` consumes its input, so the count is the answer.

**Pre-existing:** the check compared **counts**, so a use-without-premise and a
premise-without-use cancelled exactly. It now compares the **sets** and names the file that
differs. The two lists are identical today (four files each), so nothing was hidden.

## Removal proofs

| mutation | `.so` | result |
| --- | --- | --- |
| drop the `has_useful_pathkeys` guard | `9ae05df03328` | plan-time arm red, +21 buffers over 1,000 groups |
| comment out one suite's `pgc_check_ordered_oracle` | n/a (test-only) | both selftest 260 arms red, the set arm naming `differential.sh` |

## Suites

`sorted_pathkeys` 108/108 (three consecutive runs, the plan-time arm being a measurement),
`harness_selftest` 152/152, `docs_style` PASS, on PG17.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -63,9 +63,20 @@ pinned at `1.0-dev` or `1.0-alpha`, each true until the next version shipped.
 
   Only the serial scan carries the claim. The parallel partial path and the
   projection path keep `NIL`: workers finish in any order, and a projection is a
-  separate layout with its own sort key. New GUC
-  `pgcolumnar.enable_sorted_pathkeys`, on by default. New suite
-  `sorted_pathkeys`, 84 checks. (#751)
+  separate layout with its own sort key.
+
+  A query that could not use an ordering does not pay to find one out. Deciding
+  whether to claim reads the whole row-group list, and a query with no `ORDER BY`
+  and no mergejoinable clause would have had any claim discarded at the end
+  anyway, so `has_useful_pathkeys` is asked first. Measured as buffers touched
+  during planning, on 1,000,000 rows in 1,000 row groups, a relation marked
+  lexicographic, `SELECT count(*) FROM t WHERE j = 3`: 142 with the feature on
+  against 121 with it off before, and 121 against 121 after. The query that can
+  use the ordering still reads, which is what stops that from being satisfied by
+  a function that reads nothing.
+
+  New GUC `pgcolumnar.enable_sorted_pathkeys`, on by default. New suite
+  `sorted_pathkeys`, 108 checks. (#751)
 
 - A predicate on `date_trunc(unit, ts)` now drives chunk-group skipping for the
   ORDERED comparisons too, not only equality. #739 inverted `=` and declined
--- a/src/columnar_customscan.c
+++ b/src/columnar_customscan.c
@@ -1365,6 +1365,27 @@ pgcolumnar_sorted_pathkeys(PlannerInfo *root, RelOptInfo *rel, Oid relid)
 	if (!pgcolumnar_enable_sorted_pathkeys)
 		return NIL;
 
+	/*
+	 * Nothing in this query could use an ordering, so do not go and find out
+	 * what the ordering is. Condition 2 below reads the whole row-group list,
+	 * which grows with the relation, and every claim this function could return
+	 * here would be truncated to NIL by truncate_useless_pathkeys at the end
+	 * anyway. Same notion, computed before the expensive part instead of after
+	 * it.
+	 *
+	 * Measured before this early return, 1,000,000 rows in 1,000 groups, a
+	 * relation marked lexicographic, a query with no ORDER BY, planning time
+	 * from EXPLAIN (SUMMARY ON) over 15 interleaved reps: 1.079 ms against
+	 * 0.922 ms with the feature off, 1.17x. The control that makes it mechanism
+	 * rather than noise is a never-sorted relation, which returns at condition 1
+	 * and measured 1.01x.
+	 *
+	 * Declared in optimizer/paths.h with this signature on every supported
+	 * major, immediately above truncate_useless_pathkeys.
+	 */
+	if (!has_useful_pathkeys(root, rel))
+		return NIL;
+
 	r = table_open(relid, AccessShareLock);
 	storageId = PgColumnarStorageId(r);
 	tupdesc = RelationGetDescr(r);
@@ -1450,9 +1471,26 @@ pgcolumnar_sorted_pathkeys(PlannerInfo *root, RelOptInfo *rel, Oid relid)
 		 * AA1003|AA1011|AA1019, the C order, where the answer is aa10|aa1002|AA1003.
 		 * A wrong answer, from a plan with no Sort in it.
 		 *
-		 * Refusing on attcollation is exact for this: it is InvalidOid for every
-		 * type that has no collation to change. Lifting it means recording the
-		 * collations beside the names, which needs a catalog column.
+		 * Refusing on attcollation is exact for this, and the reason is an
+		 * invariant rather than a list of types, because a list rots. A type whose
+		 * attcollation is InvalidOid orders either by nothing collation-dependent
+		 * at all, or by a collation PostgreSQL will not let change under a stored
+		 * value: a range or multirange fixes its collation at CREATE TYPE and no
+		 * DDL reaches pg_range.rngcollation, and a composite orders by its field
+		 * collations, which cannot be altered while any table stores the type.
+		 * That last protection follows dependencies THROUGH arrays and THROUGH
+		 * nesting composites: a column of ctype[] , or of a composite that merely
+		 * nests ctype, still makes "ALTER TYPE ctype ALTER ATTRIBUTE ..." fail
+		 * with "cannot alter type because column uses it", with or without CASCADE.
+		 * An enum's ADD VALUE ... BEFORE does not renumber values already stored.
+		 *
+		 * Outside the trust boundary, and no differently than for a btree index
+		 * that would need a REINDEX: replacing a type's default btree opclass, or
+		 * CREATE OR REPLACE on a comparison function, changes ordering semantics
+		 * without touching the table and without any collation being involved.
+		 *
+		 * Lifting the refusal means recording the collations beside the names,
+		 * which needs a catalog column.
 		 */
 		if (OidIsValid(att->attcollation))
 			break;
--- a/test/run_all_versions.sh
+++ b/test/run_all_versions.sh
@@ -586,12 +586,22 @@ is_timing_suite() {
 # PGC_SKIP_TIMING drops those in CI and replication must still run there. Serial
 # and skipped are different properties and were one list until this suite needed
 # one without the other.
-# planner_choice_quality wants the same split, the other way around from
-# replication. Its subject is a wall-clock ratio, so it cannot run beside five
-# others and mean anything. But it is deliberately not in is_timing_suite: its
-# premises are plan-shape assertions, and those are worth running in CI. Under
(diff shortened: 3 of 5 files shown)


OffgridwithJD (commented):
Reviewed on the real 3-way merge (merge-base **is** `4d32f65`), PG 18.4 assert.
**`sorted_pathkeys` 108/108**, **`harness_selftest` 152/152**, tree clean.

Removal proof reproduced independently — mutation `790288e1b602` against
`f6cb130a5a84`, so it reached the binary. Exactly **one** arm reddens, and it is
your numbers to the buffer:

```
FAIL  a query with no ORDER BY does not read the group list to decide:
        got [no (on=142 off=121 over 1000 groups)] want [yes]
```

The control arm stayed green, which is what stops that arm being satisfied by a
function that reads nothing.

## The early-out is exactly equivalent — but the reason is not the obvious one

This is the part I want on the record, because I suggested `has_useful_pathkeys`
and "same notion, computed earlier" is **not** true from the two function bodies
alone. Read across the majors, `truncate_useless_pathkeys` considers strictly more
sources than `has_useful_pathkeys` checks:

| | `truncate_useless_pathkeys` counts | `has_useful_pathkeys` checks |
| --- | --- | --- |
| PG 15/16 | merging, ordering | merging, ordering |
| PG 17 | merging, ordering, grouping, **setop** | merging, grouping, ordering |
| PG 18 | merging, ordering, grouping, **distinct**, **setop** | merging, grouping, ordering |
| PG 19 | sort, **window**, **setop**, grouping, **distinct**, merging | merging, ordering |

On PG 18 that looks like a hole: a query whose only use for an ordering is
`DISTINCT` or a set operation would be skipped by the early return while the old
code would have kept the claim.

It is not a hole, and the thing that closes it is in `planner.c`, not `pathkeys.c`
— `standard_qp_callback` assigns `query_pathkeys` from whichever clause the query
has (PG 18 `planner.c:3874-3885`, PG 19 `:3902-3913`):

```c
if (group)      root->query_pathkeys = root->group_pathkeys;
else if (window)   root->query_pathkeys = root->window_pathkeys;
else if (distinct) root->query_pathkeys = root->distinct_pathkeys;
else if (sort)     root->query_pathkeys = root->sort_pathkeys;
else if (setop)    root->query_pathkeys = root->setop_pathkeys;
else               root->query_pathkeys = NIL;
```

So `query_pathkeys != NIL` **iff** at least one of the five is non-NIL, and every
extra source `truncate` counts is one of those five. `has_useful_pathkeys` false
therefore implies `truncate` returns NIL, on all five majors.

**Suggestion, and the only thing I would change:** put that invariant in the
comment. What is written now — "every claim this function could return here would
be truncated to NIL at the end anyway" — is the conclusion, and it is true only
because of an assignment in a different file. A future major that stops funnelling
those five through `query_pathkeys` (PG 19 already restructured this function once)
would make the early return start dropping legitimate claims, silently and with no
test able to see it, since the result is a lost optimisation and not a wrong answer.
One sentence naming `standard_qp_callback` is what makes that reviewable later.

## Measured: the early-out loses no claim

Differential against `main` (`4d32f65`, the same code without the early return),
same fixture, PG 18:

| shape | main | PR 763 |
| --- | --- | --- |
| `ORDER BY k LIMIT 10` | 0 Sort | 0 Sort |
| `DISTINCT k` | 0 Sort | 0 Sort |
| `GROUP BY k` | 0 Sort | 0 Sort |
| merge join on `k` | 0 Sort | 0 Sort |
| `UNION` | **1 Sort** | **1 Sort** |
| planning buffers, no `ORDER BY`, GUC on/off | 123 / 115 | **115 / 115** |
| planning buffers, `ORDER BY`, GUC on/off | 16 / 8 | 16 / 8 |

Every claim survives, the waste goes to exactly zero, and the `ORDER BY` shape still
pays to decide.

**`UNION` sorts on both**, so it is not this PR — I checked precisely because it was
the shape my equivalence argument said should keep its claim. `setop_pathkeys` does
reach `query_pathkeys`, so the claim is attempted and something below declines it.
Worth its own issue: a table sorted on `k` feeding `SELECT k FROM t UNION ...` still
plans a Sort. Not a blocker, and not yours to fix here.

## `selftest/260`

All three of my points are addressed, and the grouped alternation
`(^[[:space:]]*diff_query_ordered |pgc_seq_hash )` is the right fix — it keeps
`pgc_seq_hash` unanchored for the mid-line `$(...)` call while making the anchor
mean something again.

Your `sed | grep -q` catch is the better find. That it happened *in the sweep that
enforces a sibling rule*, and that 080's regex scopes to `echo`/`printf` writers so
it could not see a `sed` writer, is worth exactly the comment you gave it.

The set check is sound, and worth stating why since it is not obvious: `comm -23`
proves users ⊆ premised, and the existing count equality gives |users| = |premised|;
together those force set equality, so the two checks are exact rather than
overlapping. Naming the differing file in the failure is what makes it debuggable.

## `run_all_versions.sh`

The right call — fixing the comment rather than only filing it, recording which half
is correct, citing the direct-run measurement, and leaving the coverage decision to
#764 with the honest sentence that a planner change should carry its own plan-shape
arms rather than lean on that suite. That last line is the one that will save the
next person.

## Verdict

**Not approving yet**: 10/12 complete, 2 suites legs still running, 0 failed. I will
approve when they land, after checking the summary line reports `sorted_pathkeys=PASS`
rather than `SKIP`. The comment suggestion above does not need to hold it.

OffgridwithJD (approved):
Approving. CI **12/12 green** on `4cace8d` — the same commit my review and local gates
ran against, `mergeStateStatus` `CLEAN`.

Both suites legs genuinely **ran** the two suites this PR touches, rather than skipping
them — the summary line, since a green job reports PASS and SKIP identically:

- `suites (PG 17)` — `sorted_pathkeys=PASS`, `harness_selftest=PASS` (209 ran, 9 skipped)
- `suites (PG 18)` — `sorted_pathkeys=PASS`, `harness_selftest=PASS` (212 ran, 6 skipped)

`harness_selftest` mattering here specifically, because this PR changes a selftest.

Nothing outstanding. The one suggestion in my review — naming `standard_qp_callback` in
the early-out comment, so that the equivalence stays reviewable when a future major
changes how `query_pathkeys` is assigned — is a comment, and I would rather it landed as
a one-line follow-up than held a green PR.

The `UNION` observation is not yours and not this PR's: a table sorted on `k` feeding
`SELECT k FROM t UNION ...` plans a Sort on `main` too, verified by differential against
`4d32f65`. Worth its own issue when you have a moment.

OffgridwithJD:
Correction to my review: **the `UNION` observation was wrong. There is nothing to
file.** A sorted columnar table feeding a `UNION` does *not* plan a Sort.

My Sort-node detector matched `Merge Append`'s **`Sort Key:`** property line, not a
Sort node. The actual plan on `4cace8d`:

```
Unique
  ->  Merge Append
        Sort Key: tc.k
        ->  Custom Scan (PgColumnarScan) on tc
        ->  Custom Scan (PgColumnarScan) on tc2
```

No Sort node. The ordering is used — that is exactly the Merge Append the claim
exists to enable. The heap-with-btree control is the same plan shape:

```
Unique
  ->  Merge Append
        Sort Key: th.k
        ->  Index Only Scan using th_k_idx on th
        ->  Index Only Scan using th2_k_idx on th2
```

So columnar behaves identically to a heap whose index supplies the same order, on
both `UNION` and `UNION ALL ... ORDER BY`. Please don't open the issue I suggested.

**Nothing else in the review changes.** The differential that matters — every claim
surviving the early-out, and planning buffers going 123/115 to 115/115 — compared the
same detector against itself on both trees, so the equivalence conclusion and the
removal proof are unaffected. The approval stands.

The `standard_qp_callback` comment suggestion also stands, and it is the one thing I
would still like to see land.

PR #765, opened 2026-08-27
Name the invariant the pathkeys early-out rests on, since it lives in planner.c and not in either function it relates

jdatcmd:
Follows #763. Comment only, no behaviour change.

The review of #763 made a point I had glossed: I wrote that `has_useful_pathkeys` is "the
same notion" as `truncate_useless_pathkeys` computed earlier. **From the two function bodies
alone that is false.** `truncate_useless_pathkeys` counts strictly more sources:

| | truncate counts | has_useful tests |
| --- | --- | --- |
| PG15/16 | merging, ordering | merging, grouping, `query_pathkeys` |
| PG17 | merging, ordering, grouping, setop | merging, grouping, `query_pathkeys` |
| PG18/19 | merging, ordering, grouping, distinct, setop | merging, grouping, `query_pathkeys` |

Read that way, a query whose only use for an ordering is `DISTINCT` or a set operation looks
like it would be skipped by the early return while the old code kept the claim.

**It is not, and what closes it is in `planner.c`.** `standard_qp_callback` assigns
`root->query_pathkeys` from whichever clause the query has -- group, else window, else
distinct, else sort, else setop, else NIL -- so every extra source `truncate` counts is one of
the five that funnel into `query_pathkeys`. `has_useful_pathkeys` false therefore implies
`truncate_useless_pathkeys` would have returned NIL.

Read in `planner.c` on **15, 17, 18 and 19** rather than taken on the argument (PG15 has no
setop branch because `setop_pathkeys` did not exist yet, which does not weaken it).

## Why a comment is worth a PR here

The funnel is the assumption a later major can break, and PG19 has already restructured this
area once. If it breaks, the early return starts dropping legitimate claims **silently**, and
**no test can see it**: the result is a lost optimisation, not a wrong answer, so there is
nothing to go red. What was written before was the conclusion; a reader could not have checked
it without re-deriving this.

Credit to the reviewer of #763 for catching that the stated reason did not support the stated
conclusion, and for locating the assignment that does.

`sorted_pathkeys` 108/108 on PG17.

### DIFF
--- a/src/columnar_customscan.c
+++ b/src/columnar_customscan.c
@@ -1385,24 +1385,32 @@ pgcolumnar_sorted_pathkeys(PlannerInfo *root, RelOptInfo *rel, Oid relid)
 	 *
 	 * THE TWO ARE NOT THE SAME PREDICATE, and the equivalence relied on here
 	 * comes from planner.c rather than from either function body.
-	 * truncate_useless_pathkeys counts strictly more sources than
-	 * has_useful_pathkeys tests:
+	 * truncate_useless_pathkeys counts more sources than has_useful_pathkeys
+	 * tests, and neither list is stable across majors:
 	 *
-	 *   PG15/16  truncate: merging, ordering
-	 *   PG17     truncate: merging, ordering, grouping, setop
-	 *   PG18/19  truncate: merging, ordering, grouping, distinct, setop
-	 *   all      has_useful: merging, grouping, query_pathkeys
+	 *   major     truncate counts                              has_useful tests
+	 *   PG15/16   merging, ordering                             merging, query_pathkeys
+	 *   PG17      merging, ordering, grouping, setop            merging, group, query_pathkeys
+	 *   PG18      merging, ordering, grouping, distinct, setop  merging, group, query_pathkeys
+	 *   PG19      merging, sort, window, setop, group, distinct merging, query_pathkeys
 	 *
 	 * Read that way, a query whose only use for an ordering is DISTINCT or a
 	 * set operation looks like it would be skipped here while the old code kept
 	 * the claim. It is not, because standard_qp_callback assigns
 	 * root->query_pathkeys from whichever clause the query has -- group, else
-	 * window, else distinct, else sort, else setop, else NIL -- so every extra
-	 * source truncate counts is one of the five that funnel into query_pathkeys.
-	 * has_useful_pathkeys false therefore implies truncate_useless_pathkeys
-	 * would have returned NIL. Checked in planner.c on 15, 17, 18 and 19.
+	 * window, else distinct, else sort, else setop, else NIL -- so every source
+	 * truncate counts beyond merging is one of the five that funnel into
+	 * query_pathkeys. has_useful_pathkeys false therefore implies
+	 * truncate_useless_pathkeys would have returned NIL. Read in planner.c and
+	 * pathkeys.c on 15, 16, 17, 18 and 19.
 	 *
-	 * That is worth stating because it is the assumption a later major can
+	 * The right-hand column is itself the evidence that the funnel is deliberate
+	 * rather than incidental. PG17 added grouping to truncate AND a group_pathkeys
+	 * test to has_useful; PG19 REMOVED that test again while keeping grouping in
+	 * truncate, which is only correct because group_pathkeys reaches
+	 * query_pathkeys. Core relies on the same identity this does.
+	 *
+	 * It is still worth stating, because it is the assumption a later major can
 	 * break. If those five stop funnelling through query_pathkeys -- PG19 has
 	 * already restructured this area once -- this early return starts dropping
 	 * legitimate claims SILENTLY, and no test can see it: the result is a lost
--- a/src/columnar_customscan.c
+++ b/src/columnar_customscan.c
@@ -1382,6 +1382,31 @@ pgcolumnar_sorted_pathkeys(PlannerInfo *root, RelOptInfo *rel, Oid relid)
 	 *
 	 * Declared in optimizer/paths.h with this signature on every supported
 	 * major, immediately above truncate_useless_pathkeys.
+	 *
+	 * THE TWO ARE NOT THE SAME PREDICATE, and the equivalence relied on here
+	 * comes from planner.c rather than from either function body.
+	 * truncate_useless_pathkeys counts strictly more sources than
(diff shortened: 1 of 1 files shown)


OffgridwithJD (commented):
The reasoning is right and this is the sentence I wanted. **Two cells of the table are
wrong for PG 19, though**, and since the whole point of this comment is that a reader
should not have to re-derive it, a PG 19 reader who does re-derive will find it does not
match.

Extracted from the sources on this box:

```
PG17  has_useful: joininfo, has_eclass_joins, group_pathkeys, query_pathkeys
      truncate:   merging, ordering, grouping, setop
PG18  has_useful: joininfo, has_eclass_joins, group_pathkeys, query_pathkeys
      truncate:   merging, ordering, grouping, distinct, setop
PG19  has_useful: joininfo, has_eclass_joins, query_pathkeys          <-- NO grouping
      truncate:   merging, sort, window, setop, grouping, distinct    <-- plus WINDOW
```

So against what the comment says:

1. `PG18/19 truncate: merging, ordering, grouping, distinct, setop` — PG 19 also counts
   **`window_pathkeys`**, and it reads `root->sort_pathkeys` directly rather than through
   a `pathkeys_useful_for_ordering` helper. PG 18 and PG 19 are not the same row.
2. `all has_useful: merging, grouping, query_pathkeys` — **PG 19 does not test
   `group_pathkeys`**. It is two conditions there, not three, so "all" is wrong.

**Neither breaks the conclusion**, which is the good news and worth saying in the comment
itself: `window_pathkeys` is in the same `standard_qp_callback` chain (`group`, else
`window`, else `distinct`, else `sort`, else `setop`), so it funnels into
`query_pathkeys` exactly like the others, and PG 19 dropping the `group_pathkeys` test is
harmless for the same reason — `group` is the first branch of that chain, so
`group_pathkeys != NIL` already implies `query_pathkeys != NIL`.

That last point is arguably worth a clause of its own, because it explains why PG 19 was
*able* to drop the redundant test, which is evidence the funnel is deliberate rather than
incidental — and the funnel being deliberate is the whole load-bearing assumption here.

Suggested rows:

```
 *   PG15/16  truncate: merging, ordering
 *   PG17     truncate: merging, ordering, grouping, setop
 *   PG18     truncate: merging, ordering, grouping, distinct, setop
 *   PG19     truncate: merging, sort, window, setop, grouping, distinct
 *   15-18    has_useful: merging, grouping, query_pathkeys
 *   19       has_useful: merging, query_pathkeys
```

Not approving only because this PR *is* the comment, so an inaccurate table is the defect
rather than a nit on the side. It is a one-line fix and I will approve on sight — CI is
already 12/12 green and `CLEAN`, and nothing else here needs changing.

OffgridwithJD (commented):
Re-checked the corrected table against all five sources on my box. **Every row is now
right**, including both cells I flagged:

```
PG15/16  truncate: merging, ordering                        has_useful: joininfo/eclass, query_pathkeys
PG17     truncate: grouping, merging, ordering, setop       has_useful: + group_pathkeys
PG18     truncate: + distinct                               has_useful: + group_pathkeys
PG19     truncate: merging, sort, window, setop, group, distinct   has_useful: joininfo/eclass, query_pathkeys
```

The added paragraph is the strongest part of the comment now, and it is a better argument
than the one I suggested.

**One thing that would make it stronger still, and shorter — core uses this exact idiom.**
`indxpath.c` builds an index's pathkeys the same way, gate then build then truncate
(PG 18 `:916-923`, and the same shape on 15/16/17/19):

```c
pathkeys_possibly_useful = (scantype != ST_BITMAPSCAN &&
                            has_useful_pathkeys(root, rel));
index_is_ordered = (index->sortopfamily != NULL);
if (index_is_ordered && pathkeys_possibly_useful)
{
    index_pathkeys = build_index_pathkeys(root, index, ForwardScanDirection);
    useful_pathkeys = truncate_useless_pathkeys(root, rel, index_pathkeys);
```

`has_useful_pathkeys` has three callers in core — `indxpath.c`, `allpaths.c` (gating
child-rel equivalences for MergeAppend) and `relnode.c` — and the `indxpath.c` one is
structurally identical to what this function does.

That matters for two reasons worth a sentence each. It means this early return is not a
novel shortcut needing novel justification; it is the guard core itself puts in front of
building a btree index's pathkeys. And it bounds the risk exactly: if the funnel ever
breaks, an ordered index on the same column loses the same optimisation in the same query,
so this claim is **no weaker than an index's** — which is the same bar the collation
refusal in this file already sets a few lines below.

Optional, and I would not push it if you would rather not restart CI. The table being
right was the defect and it is fixed.

Not approving only because CI is still running (2 legs). I will approve when it lands.

OffgridwithJD (approved):
Approving. CI **12/12 green** on `07e9ae4`, `mergeStateStatus` `CLEAN`, and both suites legs
ran a full matrix (PG 17: 209 ran, 9 skipped; PG 18: 212 ran, 6 skipped).

I verified the two claims this comment makes that a reader would otherwise have to
re-derive — which is the comment's whole purpose, so they seemed worth checking rather
than reading.

**The per-major table is right on all five rows.** Re-extracted from the sources:

```
PG15/16  truncate: merging, ordering
PG17     truncate: merging, ordering, grouping, setop
PG18     truncate: merging, ordering, grouping, distinct, setop
PG19     truncate: merging, sort, window, setop, group, distinct
15-18    has_useful: merging, group_pathkeys, query_pathkeys
19       has_useful: merging, query_pathkeys
```

**"indxpath.c gates, builds, then truncates, in that order, on every supported major"
holds on all five**, and I checked each rather than sampling:

```
PG15  gate@987  build@991  truncate@993
PG16  gate@964  build@968  truncate@970
PG17  gate@911  build@915  truncate@917
PG18  gate@917  build@921  truncate@923
PG19  gate@916  build@920  truncate@922
```

Worth noting the comment cites **no line numbers**, which is the right call — those five
sets differ on every major and any one of them would have been stale on four of the five.
The claim as written is the durable form of it.

The `has_useful_pathkeys` caller count is right too: three in core, and the `indxpath.c` one
is the structural match.

The paragraph that closes it — if the funnel breaks, an ordered index on the same column
loses the same optimisation in the same query, so this is no weaker than an index — is a
better argument than the one I suggested, and it lands the file on the same bar the
collation refusal sets a few lines below. That consistency is worth more than either
sentence alone.

Issue #766, opened 2026-08-27
The columnar scan is priced 1.9x to 3.4x below the heap per unit of work, and seq_page_cost decides whether that is a defect


jdatcmd:
Measured while re-deriving #753 (see that issue's measurement comment for the full run). The
finding there splits in two, and this is the half that belongs to pgColumnar rather than to
core.

## The columnar scan cost model disagrees with itself as the projection widens

Cost per millisecond of real serial execution, same query, same 2,000,197 rows, same data, on
a 4,000,000-row 14-column table and its heap twin. Core's seqscan calibration is the reference
scale: not a claim that it is true, but it is the scale every other path in the planner is
judged against.

| projection | columnar cost | columnar min ms | cost/ms | heap cost/ms | heap ÷ columnar |
| --- | ---: | ---: | ---: | ---: | ---: |
| `sel` | 61,065 | 176.0 | 347 | 858 | 2.47x |
| `sel, a, b` | 83,196 | 216.4 | 385 | 846 | 2.20x |
| 8 columns | 138,522 | 268.3 | 516 | 765 | 1.48x |
| `*` | 239,001 | **1279.1** | 187 | 944 | **5.05x** |

Two things, and the second is the one worth an issue:

1. **The columnar scan is under-priced** relative to that scale by 1.5x to 5x. A path priced
   at half what it costs wins comparisons it should lose.

2. **It is not a constant offset.** The heap's cost/ms is stable across projection widths, 765
   to 944, a **1.24x** spread. Columnar's runs 187 to 516, a **2.8x** spread, and not
   monotonically. So the model does not merely need a multiplier; its *shape* in projection
   width is wrong.

The `*` row is the same defect at its extreme: a full-row columnar scan of this table takes
1279 ms against the heap's 167 ms, **7.6x slower**, while being priced at 1.5x the heap. That
columnar loses at `SELECT *` is expected and fine. That the cost model says 1.5x when reality
is 7.6x is not.

## Why this cannot just be corrected

The scan cost is the scale every other path comparison is judged on. Raising it toward reality
would change, at least:

- the **index-fetch penalty** (#355, #362), which is calibrated as a fraction of
  `pgcolumnar_full_scan_cost`;
- **#171**, where a point lookup went from 23.75 ms to 1251.88 ms because ANALYZE made the
  full scan look cheap enough to beat an index scan. That regression is pinned in
  `test/analyze_stats.sh` and it lives on exactly this number;
[...]

Issue #362, opened 2026-08-03
The #355 index-fetch penalty changes the estimate but not the plan: the cheaper path is already freed when it is applied


OffgridwithJD:
## Summary

The index-fetch penalty I added in #360 changes what `EXPLAIN` prints but not which
plan is chosen, on the query it was written for. On `cpu_pgc` with its
`(hostname, time DESC)` index, the five-aggregate-column #359 query:

| arm | plan | estimated | actual |
|---|---|---:|---:|
| `enable_index_fetch_penalty=on` (default) | Index Scan | 13,954,742 | **224,055 ms** |
| `enable_index_fetch_penalty=off` | Index Scan | 5,090 | **224,946 ms** |
| penalty on, `enable_indexscan=off` | Parallel Custom Scan (ColumnarScan) | **589,348** | **4,728 ms** |

The penalty fires — it moves the estimate from 4,975 to 13,954,348, a factor of
2,800. The planner then chooses that path anyway, over an alternative **it estimates
at 1/24th the cost** and which runs **47x faster**. So the catastrophic plan #360 was
supposed to prevent is still chosen by default, and #359's cliff is still reachable
through it.

Same result with `max_parallel_workers_per_gather=0`, so this is not about parallel
path generation.

## Mechanism

`ColumnarSetRelPathlist` offers the columnar path to `add_path` **before** applying
the penalty:

```c
add_path(rel, &cpath->path);          /* columnar scan, ~587k+ */
...
/* #355 penalty: "This runs last, after every add_path above, on purpose" */
```

`add_path` frees a new path it judges dominated. At that moment the index path still
carries its *un-penalized* cost (4,975), so against a selective index condition the
columnar path is dominated on cost and is freed. The penalty then inflates the index
path — and there is no longer an alternative in `rel->pathlist` to switch to, which
is why `set_cheapest` returns a 13.9M path while a 589k path exists only once the
index path is disabled outright.

The hook's own comments already document this hazard for the seqscan path:

> `add_path` frees a path it finds dominated, so the seqscan is gone from
> `rel->pathlist` by the time this hook runs exactly when some index path beat it on
> both cost and pathkeys -- which is to say, precisely on the selective lookups where
> using the index matters most.

I reasoned about this in #360's body and got it half right. I wrote that applying the
penalty last was safe because "`add_path`'s dominance test compares pairs directly
(order-independent); only its insertion position depends on the sort, and
[...]

Issue #355, opened 2026-08-03
Planner chooses an Index Scan on columnar for ordering; per-row fetch cost is not modelled and the plan is orders of magnitude slower


jdatcmd:
## Summary

The planner will choose an `Index Scan` over a columnar table to obtain ordering, because a columnar index scan is costed as if per-row fetch were as cheap as heap's. It is not, and the resulting plans are catastrophically slow. Related to #353 but not fixed by it.

## Reproduction

`cpu_pgc`, 100M rows, with a `(hostname, time DESC)` btree. A grouped aggregate over a 12-hour window (~17.8M rows):

```sql
SELECT date_trunc('hour',time), hostname, avg(usage_user)
FROM cpu_pgc
WHERE time >= '2024-01-01' AND time < '2024-01-01 12:00'
GROUP BY 1,2;
```

**Serial** (`max_parallel_workers_per_gather=0`):

```
GroupAggregate  (cost=802.80..3692838.88)
  ->  Incremental Sort  (cost=802.80..3430210.01 rows=17825280)
        Presorted Key: hostname
        ->  Index Scan using cpu_pgc_hostname_time_idx  (cost=0.57..2126891.87 rows=17825280)
```

Cancelled after **79 minutes**, incomplete.

**Parallel**, same query, same data:

```
GroupAggregate  (cost=1179516.40..3576448.75)
  ->  Gather Merge -> Sort
        ->  Parallel Custom Scan (ColumnarScan)  (cost=0.00..525045.25)
```

**7,322 ms.**

## The cost model is wrong by orders of magnitude

The index scan is priced at 2,126,891 against the columnar scan's 525,045 -- about **4x** apart. Measured, they are at least **70,000x** apart, and the index-scan arm did not finish.

The planner picks it because the index supplies `hostname` presorted, so an Incremental Sort replaces a full sort. That is sound reasoning for heap, where fetching a row costs a page access. On a columnar table a fetch by row number decodes a row group, which is a different order of cost entirely, and nothing communicates that to the planner.

## Not the same as #353, though they compound

#353 is that the default `stripe_row_limit` puts a group over the 32 MB fetch cache, making each fetch ~28.7 ms instead of ~0.18 ms. Fixing it would take this query from "days" to roughly:

  17,825,280 rows x 0.177 ms = about 53 minutes

against a 7.3-second sequential scan. Still the wrong plan by a factor of about 430. So #353 makes this catastrophic, and fixing #353 leaves it merely bad.

[...]

PR #767, opened 2026-08-27
Correct the parallel-scan limitation: it does not ship fewer rows, and the heap does not flip at half the rate

jdatcmd:
Documentation only. Closes nothing; it corrects what `docs/limitations.md` says about #753,
which is now measured (see the measurement comment on #753).

## Two statements that are not true

**"A columnar scan ships fewer, already-decoded rows, so that rate overstates its Gather
cost."** It ships the same number of rows. Measured on 4,000,000 rows in a 14-column table
against a heap twin holding the same rows, both ship 999,987, so the Gather tuple charge is
identical at 99,999. What differs is tuple **width** (3 projected columns against 14) and the
denominator:

| | serial cost | Gather charge at 0.1 | as % of serial |
| --- | ---: | ---: | ---: |
| columnar | 83,196 | 99,999 | **120%** |
| heap | 158,109 | 99,999 | 63% |

**"A heap scan on the same shape also flipped ... About half the default rate picks the faster
plan."** The heap does not flip at half. Columnar turns parallel at `parallel_tuple_cost`
0.060, the heap at 0.030, against a default of 0.100 — so at the doc's own 0.05, columnar
flips and the heap does not.

## The effect is real, and larger than the doc recorded

Over the **eleven** query shapes where the planner chose the serial scan, the parallel plan
ran 1.8 to 2.6 times faster on the median, and **in every shape the slowest parallel run beat
the fastest serial run**. The finding is that non-overlap, not a ratio.

Every parallel arm asserts `Workers Launched == Workers Planned` (a plan can say `Gather` and
run with zero workers, and it still says `Gather`), and chunk groups read was 27 of 27 in
every cell, so no cell is confounded by zone-map pruning leaving the parallel arm nothing to
divide. Host loadavg 0.70 to 1.29 throughout; contention biases *against* finding a parallel
speedup, and it was found anyway.

## "Narrow but selective" means narrow in COLUMNS

Fewer columns means less decode, so the serial scan is cheap, and a fixed per-row charge above
it decides the comparison. That is why the doc's own shape could not be reproduced on a
5-column fixture: projecting 3 of 5 is not narrow.

## Two mechanisms, both now stated

**`parallel_tuple_cost` is charged per row and is blind to width.** At a constant volume of
data shipped (11.44 MiB both arms), tripling the row count and thirding the width raised the
real transfer cost by 1.56x while the charge rose by 3.00x — so a narrow row is overstated by
about 1.9x. Narrowness is a projection property, not a storage-format one, so this half is
core's and applies to a heap with the same narrow projection.

**The columnar scan cost is low against the time it takes**, by an amount that varies with the
number of columns read. That half is ours and is #766, kept separate on purpose: the scan cost
is the scale every other path comparison is judged against, including the index-fetch penalty
(#355, #362) and the point-lookup regression #171 exists for, so a recalibration cannot ride
along under a parallelism ticket.

## Gate

`docs_style` PASS. Documentation only, no code.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -130,13 +130,14 @@ pinned at `1.0-dev` or `1.0-alpha`, each true until the next version shipped.
 
   The effect itself is real and larger than recorded. Over the eleven query
   shapes where the planner chose the serial scan, the parallel plan ran 1.8 to
-  2.6 times faster on the median, and in every shape the slowest parallel run
-  beat the fastest serial run. "Narrow" turns out to mean narrow in **columns**.
-  Two mechanisms produce it, both now stated: `parallel_tuple_cost` is charged
-  per row and is blind to width, overstating a narrow row by about 1.9 times at a
-  constant volume of data shipped; and the columnar scan cost is a half to a
-  fifth of what its real time implies, by an amount that varies with the number
-  of columns read. (#753)
+  2.5 times faster on the median, and in every shape the slowest parallel run
+  beat the fastest serial run, by 1.45 times at the narrowest margin. "Narrow"
+  turns out to mean narrow in **columns**. Two mechanisms produce it, both now
+  stated: `parallel_tuple_cost` is charged per row and is blind to width,
+  overstating a narrow row by about 1.9 times at a constant volume of data
+  shipped; and the columnar scan cost is two fifths to seven tenths of what its
+  real time implies, by an amount that is largest on the narrowest projection.
+  (#753)
 
 - Both eager ordering rewrites discarded the ordering they had just applied.
   `pgcolumnar.storage.sorted_by` and `sorted_kind` exist to record which
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -116,6 +116,28 @@ pinned at `1.0-dev` or `1.0-alpha`, each true until the next version shipped.
 
 ### Fixed
 
+- `docs/limitations.md`'s account of parallel scans stated two things that are
+  not true, and gave a cause the arithmetic does not support. It said a columnar
+  scan "ships fewer, already-decoded rows" than a heap scan, and that "a heap
+  scan on the same shape also flipped" at about half the default
+  `parallel_tuple_cost`.
+
+  Re-measured on 4,000,000 rows in a 14 column table with a heap twin holding the
+  same rows. The two plans ship the **same** number of rows, 999,987, so the
+  Gather charge is identical at 99,999; what differs is tuple width, three
+  projected columns against fourteen. And the heap does not flip at half the
+  default: columnar turns parallel at 0.060 and the heap at 0.030.
+
+  The effect itself is real and larger than recorded. Over the eleven query
+  shapes where the planner chose the serial scan, the parallel plan ran 1.8 to
+  2.6 times faster on the median, and in every shape the slowest parallel run
+  beat the fastest serial run. "Narrow" turns out to mean narrow in **columns**.
+  Two mechanisms produce it, both now stated: `parallel_tuple_cost` is charged
+  per row and is blind to width, overstating a narrow row by about 1.9 times at a
+  constant volume of data shipped; and the columnar scan cost is a half to a
+  fifth of what its real time implies, by an amount that varies with the number
+  of columns read. (#753)
+
 - Both eager ordering rewrites discarded the ordering they had just applied.
   `pgcolumnar.storage.sorted_by` and `sorted_kind` exist to record which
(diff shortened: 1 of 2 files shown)


OffgridwithJD (commented):
Verified every factual sentence against your measurements. Both false statements are gone
and correctly replaced, and the flip rates, the 11 shapes, the 1.8–2.6x medians and the
1.56/3.00 → 1.9x charge overstatement all check out.

**Two things, one of which is a number you already withdrew.**

## 1. "a half to a fifth" carries the retracted 5.05x

> The columnar scan is priced at a half to a fifth of what its real time implies.

"A fifth" is `1/5.05`, and 5.05x was the `*` cell of the confounded table — the one that
turned out to be text decode rather than width, and that you explicitly withdrew. On the
corrected int-only ladder the range is:

| projection | columnar cost/ms | heap cost/ms | heap ÷ columnar |
| --- | ---: | ---: | ---: |
| `sel` | 364.7 | 891.7 | 2.45x |
| `sel, a` | 364.1 | 844.9 | 2.32x |
| `sel, a, b` | 387.8 | 823.4 | 2.12x |
| `sel, a, b, c, d` | 451.9 | 809.5 | 1.79x |
| 8 ints | 510.8 | 716.6 | 1.40x |

So columnar is priced at **0.41 to 0.71** of what its time implies — between about
two-fifths and seven-tenths, a 1.4x to 2.4x under-cost. Not "a half to a fifth".

Worth fixing precisely because the rest of this PR is the withdrawal of overstated
numbers; leaving the retracted one in the replacement text is the one thing that would
undercut it.

While there: your message quoted the range as "1.7x–2.4x", but the 8-int cell is
**1.40x**, which is outside it. The band is 1.40–2.45.

## 2. Is the non-overlap true in all eleven, or in the one you showed?

> In every shape the slowest parallel run still beat the fastest serial run.

That is a stronger statement than the statistic you reported. You gave me median ratios
1.83–2.58x and **min** ratios 2.08–2.89x, plus non-overlap demonstrated on one cell
(serial min 176.3 against parallel max 78.8). A min ratio is `serial_min / parallel_min`;
non-overlap needs `serial_min > parallel_max`, which is a different pair of numbers and
does not follow from it.

If you have `parallel_max` per cell, please confirm it holds in all eleven before this
sentence ships — it is the load-bearing claim of the section, and it is the kind that is
true in the cell you looked at and not in the tail. If you would rather not re-run, the
honest weaker form is available and still strong: "in every shape the parallel plan was
faster on both the median and the fastest run".

## Everything else checks

- 4,000,000 rows, 14 columns, 11 serial-chosen shapes, 1.8–2.6x median ✓
- 1.56x real against 3.00x charged, therefore ~1.9x overstatement of a narrow row ✓
- flip rates 0.060 columnar / 0.030 heap against a 0.100 default ✓, and stating both is
  what replaces the false "about half" cleanly
- "'Narrow' here means few columns, not few rows" — this is the sentence that would have
  saved us the first fixture, and it is right to put it in the doc
- "This applies to any narrow projection, on a heap table as much as a columnar one" ✓,
  and it is the sentence that stops a future reader reading this as a columnar discount

Not approving yet: CI still running, and item 1 is a real correction rather than a
preference.

OffgridwithJD (approved):
Approving. CI **12/12 green** on `9f32aacd`, `CLEAN`, and `docs_style=PASS` on both legs —
worth checking by name on a docs PR, since it runs inside the suites job rather than as its
own check.

Both corrections landed, and I verified them as greps over the added lines rather than by
reading the passage I had quoted:

- `"a fifth"` — **0 occurrences**. The band now reads "two fifths to seven tenths of what
  its real time implies", which matches the int-only ladder (0.41 to 0.71, a 1.40x–2.45x
  under-cost).
- `"1.7x"` — **0 occurrences**.
- Both non-overlap sentences now carry the measured margin: "the slowest parallel run was
  still faster than the fastest serial run, by 1.45 times at the narrowest margin", in the
  CHANGELOG entry and in `docs/limitations.md`.

The last one is the change that matters most. It went from a property asserted of eleven
cells on the strength of a statistic that could not show it, to a measured claim with its
own worst case stated. Quoting 1.45x rather than leaving the narrowest margin unstated is
what makes it checkable by the next person — and in a page whose subject is two numbers
that outran their evidence, that is the right instinct to have applied to the replacement.

"'Narrow' here means few columns, not few rows" is the sentence that would have saved the
first fixture, and "This applies to any narrow projection, on a heap table as much as a
columnar one" is what stops a future reader taking this as an argument for a columnar
discount. Both belong.

No findings.

Issue #764, opened 2026-08-27
run_all_versions.sh describes planner_choice_quality's arrangement twice and the two descriptions contradict each other, hiding that no plan-shape assertion runs in CI


jdatcmd:
`test/run_all_versions.sh` describes its own arrangement of `planner_choice_quality` twice,
about thirty lines apart, and the two descriptions are opposites.

`is_timing_suite()`, line 559, **includes it**, with the reason written out:

```
		# planner_choice_quality's only assertion is a wall-clock ratio. Left out
		# of this list it still RAN under PGC_SKIP_TIMING, skipped the ratio, and
		# reported PASS on the strength of its premises -- so a regression of #434
		# would have been reported green by the suite that exists to catch it.
		planner_choice_quality) return 0 ;;
```

`runs_alone()`, thirty lines below, says the opposite:

```
# planner_choice_quality wants the same split, the other way around from
# replication. [...] But it is deliberately not in is_timing_suite: its
# premises are plan-shape assertions, and those are worth running in CI.
```

It **is** in `is_timing_suite`, so under `PGC_SKIP_TIMING` (which is how CI runs)
`run_all_versions.sh:706` does not invoke the suite at all. A reader who finds the second
comment believes CI exercises those plan-shape assertions. Nothing does.

Observed on PR #762's CI, both legs: `planner_choice_quality=SKIP`. Same on #759, so it is
not new.

## The decision behind the comment is right

Confirmed by running the suite directly with `PGC_SKIP_TIMING=1`: it reports **PASSED, 7
checks**, with only the two wall-clock ratios skipped. The 7 that remain are fixture and
plan-identification premises. A green built on premises with the subject removed is worse
than a SKIP, and the driver already knows how to report a skip. So the entry in
`is_timing_suite` should stay and the `runs_alone` comment is what is wrong.

## But there is a coverage question the stale comment is pointing at

The stale comment is arguing for something real: **no plan-shape assertion from
`planner_choice_quality` runs in CI**, on any major. That is an accepted trade-off today
because the alternative was a false green.

It stops being acceptable for **#753**. A change to how `Gather` prices columnar rows has
`planner_choice_quality` as its natural regression guard, so #753 would land with its guard
skipped in CI. That has to be settled before #753 is built rather than after, and the options
are roughly:

1. split the suite: the plan-shape assertions into a suite CI invokes, the wall-clock ratios
   staying timing-only and serial. This is the split both comments are circling.
[...]

Issue #754, opened 2026-08-26
A columnar table cannot be a logical replication source, which rules out CDC and blue-green upgrades


jdatcmd:
From `docs/limitations.md`, "Replication and backup":

> Logical decoding reads the WAL records of heap tuples. Columnar data reaches WAL as
> full-page images, and those carry no tuple structure. Logical decoding therefore does not
> emit a change to a columnar table, and logical replication does not carry it. Use physical
> replication for columnar tables.

Physical replication and physical backup work, and `pg_dump`/`pg_restore` work. What does
not work is anything that consumes a change stream.

## Why it is worth a tracking issue even if it is not worth building

It closes off a set of things users expect from a PostgreSQL table, and each is an adoption
decision rather than a performance one:

- CDC into a warehouse, search index or queue (Debezium and everything shaped like it).
- Blue-green and near-zero-downtime major-version upgrades, which are logical-replication
  based.
- Selective per-table replication into a different topology.

It also interacts with what this extension is for: "append-mostly analytical data" is
exactly the data people most often want to stream somewhere else.

## The constraint that makes this hard

This project's standing rule is that it uses only core WAL mechanisms and existing record
types, and a design that needs new WAL semantics is rejected rather than deferred. Emitting
decodable changes for columnar writes looks like it needs exactly that, so the honest
framing is **not** "implement logical decoding" but "decide, in writing, whether there is
any route inside the constraint."

Routes worth pricing before dismissing:

- A trigger-based or `pgcolumnar`-side change feed, which is a different product decision
  and has its own durability story.
- Documenting a supported CDC pattern that uses a heap staging table, so the answer to the
  question is a recipe rather than a limitation.
- Nothing, stated deliberately and put in `docs/limitations.md` where it already is, with
  the reasoning recorded so it is not re-litigated.

Filing this so the decision is visible rather than implicit. It is a **strategic** item, and
unlike #751, #752 and #753 it is not obviously a task anyone should pick up.

OffgridwithJD:
This issue rests on a claim from `docs/limitations.md` that, as far as I can find, **nothing
in the tree tests**. I have now tested it. It is true — and the other half of that paragraph,
which this issue does not quote, is also true and is the part that changes what the decision
should be about.

## The claim was untested

The only logical replication in `test/` is `logical_subscriber.sh`, and it covers the
*opposite* direction — columnar as an apply **target** (#435). Its publisher tables are heap:

```
[...]

PR #769, opened 2026-08-27
Assert the claim the whole of #754 rests on: a columnar table is not a logical decoding source

jdatcmd:
`docs/limitations.md` states it and #754 is built on it. **Nothing in the tree asserted it.**

The only other logical-replication coverage, `logical_subscriber.sh`, tests the **opposite**
direction — a heap publisher into a columnar subscriber (`:87` heap, `:90` columnar) — and
every slot in `replication.sh` is physical. The source direction, which is the one the
limitation is about, had no test behind it.

## Measured

A `test_decoding` slot created **before any row is written**, a heap table and a columnar
table of identical shape, the same 50 rows into each:

| | decoded changes |
| --- | ---: |
| `public.heap_t` (control) | **50** |
| `public.col_t` (columnar) | **0** |
| `pgcolumnar.*` (internal) | 8 |

## The heap control is the load-bearing arm

Zero decoded changes is what a working exclusion looks like. It is equally what a broken slot,
a mis-built plugin, a slot created after the writes, or a query against the wrong slot looks
like. **The columnar arm cannot mean anything on its own.**

So a heap table of identical shape is written through the same slot in the same window and
asserted to decode. That establishes the instrument sees changes before the absence of one is
believed — the same shape as proving a secret is a secret before measuring how well it leaks.

## The second documented behaviour, also asserted

The slot is **not silent**: it carries pgcolumnar's own catalog writes, so a consumer sees
internal churn it cannot interpret rather than nothing at all. And that churn scales with
**row groups and chunks, not rows**, which is what makes it useless as a change stream — it
carries neither the data nor even a row count:

```
internal pgcolumnar records: 8 for 50 rows, 10 for 5000
```

100x the rows, 1.25x the records.

## Removal proofs — two, because a zero has at least two ways of being true for the wrong reason

**A zero is the headline assertion here**, so it needs a demonstrated guard against each way it
could be satisfied by accident. There are two, and they are independent.

**1. Wrong storage type.** Making `col_t` a heap table:

```
FAIL  a columnar table emits no decodable change: got [50] want [0]
FAIL  the slot carries pgcolumnar's internal catalog instead: got [no] want [yes]
FAIL  100x the rows does not give 100x the internal records: got [no (0 then 10)] want [yes]
```

So the claim discriminates on the storage type.

**2. Slot not live across the write window.** Moving
`pg_create_logical_replication_slot` to **after** both `INSERT`s, nothing else changed (run by
the reviewer):

```
PASS  a columnar table emits no decodable change          <-- the claim, now VACUOUS
FAIL  CONTROL: the heap table's 50 inserts ARE decoded: got [0] want [50]
FAIL  premise: the slot decoded something at all: got [no] want [yes]
FAIL  the slot carries pgcolumnar's internal catalog instead: got [no] want [yes]
FAIL  100x the rows does not give 100x the internal records: got [no (0 then 10)] want [yes]
```

**The claim arm passes against a slot that decoded nothing at all**, and four independent
guards catch it. That is the mutation that tests the sentence "the heap control is the
load-bearing arm", and it confirms it — while showing the control is not alone, which is the
right amount of redundancy for a suite whose headline is a zero.

The second proof could only come from the reviewer: it is an attack on the arm the author
wrote in order to be trustworthy, and the author is the one person who cannot run it.

## Two traps recorded in the suite's comments

- **`CREATE SUBSCRIPTION` against the same cluster self-deadlocks.** It runs in a transaction
  and logical slot creation waits for running transactions including its own; it hangs with
  the walsender on `Lock/transactionid` rather than failing. The slot is created with
  `pg_create_logical_replication_slot` and read directly.
- **`test_decoding` is stock contrib but is not built by default in a source install.** Its
  absence is reported as an environment defect through `pgc_skip`, not passed over. Building
  it in-tree with `make install PG_CONFIG=…` ignores the flag and returns 0, so verify the
  artifact landed rather than the exit status.

`wal_level=logical` cannot be raised without a restart, so it is set through `PGC_EXTRA_CONF`
before the cluster starts.

## It runs in CI, checked rather than assumed

A suite gated on a contrib plugin is one `pgc_skip` away from never running where it matters,
which is the #764 shape exactly. The `suites` legs install PG from PGDG apt packages
(`ci.yml:331-341`) and `postgresql-18` ships `test_decoding.so`. The only from-source build is
the PG19 beta **build** job, which runs `make -C src install` (`ci.yml:292`) and so has no
contrib — but it compiles the extension and runs no suites. So the `pgc_skip` is a correct
guard for a developer box, not a hole in the gate.

Worth confirming on the finished run that the summary line reads
`logical_decoding_source=PASS` and not `=SKIP`: those are indistinguishable from the job
colour, and this is precisely the suite where a silent skip would restore the situation it was
written to end.

## Gate

`logical_decoding_source` 11/11, `harness_selftest` 152/152, `logical_subscriber` PASS,
`docs_style` PASS, on PG17. Registered in `run_all_versions.sh`.

Credit to the reviewer for testing the claim and for both traps above; this reproduces their
result independently and turns it into an arm.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -114,6 +114,30 @@ pinned at `1.0-dev` or `1.0-alpha`, each true until the next version shipped.
   truncated, are declined rather than approximated. A new suite,
   `preimage_rewrite`, measures the pruning and pins each declined shape. (#403)
 
+### Added
+
+- A suite for the claim under `docs/limitations.md`'s "Replication and backup":
+  that a columnar table is not a logical decoding source. The whole of #754 rests
+  on that sentence and nothing asserted it. The only other logical-replication
+  coverage, `logical_subscriber`, tests the opposite direction, a heap publisher
+  into a columnar subscriber.
+
+  Measured through a `test_decoding` slot created before any row is written, on a
+  heap table and a columnar table of identical shape holding the same 50 rows:
+
+  | | decoded changes |
+  | --- | ---: |
+  | `public.heap_t` (control) | 50 |
+  | `public.col_t` (columnar) | **0** |
+  | `pgcolumnar.*` (internal) | 8 |
+
+  The heap control is the load-bearing arm, because zero decoded changes is also
+  what a broken slot, a mis-built plugin or a query against the wrong slot looks
+  like. The second documented behaviour is asserted too: the slot is not silent,
+  it carries pgcolumnar's own catalog writes, and that churn scales with row
+  groups and chunks rather than rows. 100 times the rows gives 8 records against
+  10, not 800. New suite `logical_decoding_source`. (#754)
+
 ### Fixed
 
 - `docs/limitations.md`'s account of parallel scans stated two things that are
--- a/test/logical_decoding_source.sh
+++ b/test/logical_decoding_source.sh
@@ -1,0 +1,121 @@
+#!/usr/bin/env bash
+#
+# pgColumnar is not a logical decoding source (#754).
+#
+# docs/limitations.md states it: logical decoding reads the WAL records of heap
+# tuples, columnar data reaches WAL as full-page images which carry no tuple
+# structure, so no change to a columnar table is emitted. That sentence is the
+# premise under a whole class of user expectations -- CDC, blue-green upgrades,
+# selective logical replication -- and until this suite it was asserted nowhere.
+#
+# The only other logical-replication coverage in the tree, logical_subscriber.sh,
+# tests the OPPOSITE direction: a heap publisher into a columnar subscriber. That
+# direction works and is the shape most likely to bring users. Nothing tested
+# the source direction, which is the one the limitation is about.
+#
+# WHY THE HEAP CONTROL IS THE LOAD-BEARING ARM. "Zero changes decoded" is what a
+# working exclusion looks like and it is also what a broken slot, a mis-built
+# plugin, or a query against the wrong slot looks like. The columnar arm cannot
+# mean anything on its own. So a heap table of identical shape is written in the
+# same transaction, through the same slot, and asserted to decode -- proving the
+# instrument sees changes before the absence of one is believed.
+#
+# Usage:  test/logical_decoding_source.sh [PG_CONFIG]
+# Written fresh for pgColumnar.
+
+set -uo pipefail
+. "$(dirname "${BASH_SOURCE[0]}")/lib.sh"
+# Logical decoding needs the WAL to carry enough to decode from, and the level
+# cannot be raised without a restart.
+export PGC_EXTRA_CONF="wal_level=logical
+max_replication_slots=8
+max_wal_senders=8"
+pgc_setup "${1:-/usr/local/pg17/bin/pg_config}"
+
+# test_decoding is stock contrib, but it is not built by default in a source
+# install. A missing dependency is an environment defect rather than a pass.
+if [ ! -f "$("$PGC_PG_CONFIG" --pkglibdir)/test_decoding.so" ]; then
+	pgc_skip "test_decoding" "test_decoding.so is not in $("$PGC_PG_CONFIG" --pkglibdir); build contrib/test_decoding"
+fi
+
+check "premise: the server is running at wal_level=logical" \
+	"$(q "SHOW wal_level;")" "logical"
+
+# Identical shape, identical rows, one heap and one columnar.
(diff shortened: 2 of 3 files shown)


OffgridwithJD (commented):
Reviewed on the real 3-way merge (merge-base **is** `ddf9da2`), PG 18.4 assert.
**11/11**, **`harness_selftest` 152/152**, tree clean, registered alphabetically at
`run_all_versions.sh:111`.

Your result reproduces on my side including the count — 50 / 0 / 8.

## The thing I most wanted to break: the control

You wrote that the heap control is the load-bearing arm because zero is also what a broken
slot or a slot created after the writes looks like. I tested that claim rather than the
claim the suite is about — mutation: move `pg_create_logical_replication_slot` to **after**
both INSERTs, changing nothing else.

```
PASS  a columnar table emits no decodable change          <-- the claim, now VACUOUS
FAIL  CONTROL: the heap table's 50 inserts ARE decoded: got [0] want [50]
FAIL  premise: the slot decoded something at all: got [no] want [yes]
FAIL  the slot carries pgcolumnar's internal catalog instead: got [no] want [yes]
FAIL  100x the rows does not give 100x the internal records: got [no (0 then 10)] want [yes]
```

The claim arm sails through — it is satisfied by a slot that decoded *nothing at all* — and
**four independent guards catch it**, the control among them. So the control is load-bearing
in exactly the way you said, and it is not the only thing holding that arm up. That is the
right amount of redundancy for a suite whose headline assertion is a zero.

Worth noting for the record that this is a different failure from the one your own removal
proof covers. Yours (`col_t` as heap) proves the claim discriminates on **storage type**;
this one proves it discriminates on **the slot being live across the write window**. A zero
has at least those two ways of being true for the wrong reason and the suite now has a
demonstrated guard for each.

## It will actually run in CI, and I checked rather than assumed

A suite gated on a contrib plugin is one `pgc_skip` away from being a suite that never runs
where it matters — the same shape as #764. So:

- the `suites` legs install PostgreSQL from **PGDG apt packages** (`ci.yml:331-341`), and
  `postgresql-18` **ships `test_decoding.so`** — verified with `dpkg -S` against the packaged
  PG on my box, which is the same package set;
- the only from-source build is the PG 19 beta **`build`** job, and it runs `make -C src
  install` (`ci.yml:292`), so it has no contrib — but that job compiles the extension and does
  not run suites.

So `pgc_skip` here is a correct guard for a developer box rather than a hole in the gate. The
one thing I would ask: when CI lands, confirm the summary line reports
`logical_decoding_source=PASS` and not `=SKIP` on both legs, because those two are
indistinguishable from the job's colour and this is precisely the suite where a silent skip
would restore the situation it was written to end.

## On the rest

The design decisions are all the right way round and the reasons are in the file rather than
in someone's head: the slot before the writes with the reason stated, the single drain into
`decoded` so the arms cannot disagree about what they saw, the control asserted first, and
both slots dropped at the end.

The `pgcolumnar.` corroboration is the part I would keep if I had to keep one thing besides
the control. It rules out the remaining way a zero could mislead — that the columnar write
produced no WAL at all — by showing that *something* from that same transaction was decoded.
A zero next to an eight is a much stronger statement than a zero alone.

And turning "it would be 8 for 50,000" into an arm is better than the sentence I gave you:
`8 then 10` for 100x the rows makes the churn's uselessness as a change stream a measured
property rather than an aside.

## Verdict

No findings. **Not approving yet** — CI still has legs running, 0 failed. I will approve when
it lands, after checking that summary line.

OffgridwithJD (approved):
Approving. CI **12/12** on `8b9e222` — the head I reviewed — `mergeStateStatus` `CLEAN`.

The check I was holding for, read from the summary lines rather than the job colour, because
for this suite in particular those two are indistinguishable:

```
suites (PG 17)  logical_decoding_source=PASS   (210 ran, 9 skipped)
suites (PG 18)  logical_decoding_source=PASS   (213 ran, 6 skipped)
```

It **ran**. The `pgc_skip` never fired, which is what makes this suite worth having: a claim
that was documented and asserted nowhere is now asserted on every CI run, on both majors.

Both removal proofs are in the body under a heading that says why there are two, and that is
the right way to carry them — they are not redundant. A zero has at least two independent ways
of being true for the wrong reason (the storage type, and the slot being live across the write
window), and each now has a demonstrated guard. The `pgcolumnar.` corroboration is a third,
ruling out the last one: that the columnar write emitted no WAL at all.

No findings. Nothing outstanding from my side on this one.

Issue #768, opened 2026-08-27
A columnar text projection costs 3.4x the heap with the per-value copy excluded, and the remainder is in the decode path


jdatcmd:
Found while separating width from column type for #766. It is not a costing defect: the
planner charges a text projection **less** than it charges the equivalent heap scan, and the
columnar scan then takes four and a half times longer.

## Measured

4,000,000 rows, 14 columns, 2,000,197 rows returned, PG 17, min of 7 reps, serial plans on
both sides. The heap twin holds the same rows. The four text columns are
`md5(i::text)`-derived, 32 bytes each.

| projection | columnar cost | columnar ms | heap ms | columnar ÷ heap |
| --- | ---: | ---: | ---: | ---: |
| `sel` (one int4) | 61,065 | 173.7 | 170.3 | 1.02x |
| `n1` (numeric) | 74,261 | 269.2 | 226.8 | 1.19x |
| `n2` (float8) | 73,196 | 351.5 | 232.0 | 1.52x |
| `t1` (one text) | 79,853 | 353.1 | 212.2 | **1.66x** |
| `t1, t2, t3, t4` | 136,218 | **1082.0** | 235.3 | **4.60x** |
| eight int4 columns | 138,522 | 271.2 | 220.6 | 1.23x |
| `*` | 239,001 | 1297.6 | 186.2 | 6.97x |

Read the two rows priced almost identically:

- **eight int4 columns**, cost 138,522, **271.2 ms**
- **four text columns**, cost 136,218, **1082.0 ms**

Nearly the same planner cost. **Four times the real time.** And against the heap, four text
columns cost the planner *less* than the heap's 158,109 while taking 4.6 times as long.

`*` is explained entirely by this: text accounts for 1082 ms of its 1298 ms, and all eight int
columns account for 271 ms. The earlier reading of `*` as a *width* result was wrong; it is a
type result.

## Why this is its own issue

It is an execution finding, not a planner one. The costing makes it worse — a plan that
projects text is priced as if text were cheap, so it is chosen over alternatives that would be
faster — but correcting the cost model would only stop the planner *choosing* it. The 1082 ms
would still be 1082 ms.

It is also independent of #753 and #766. It needs no parallelism, and it reproduces on a
single serial scan of one column.

## What is not yet established

- **Which part of the text path costs it.** Candidates, in the order I would test: the block
  codec (these are md5 strings, so compression is poor and the codec may be doing full work
[...]

PR #770, opened 2026-08-27
Add the wide no-join control S4 cannot be attributed without: S4 is decode, and the join's share is at or below the noise floor

jdatcmd:
Bench fixture only, no product code.

## Why

**S4 is #752's worst row** — "wide projection under a join" — and it projects **eight float8
columns**. #768 measures a **single** float8 column at 1.52x the heap with no join anywhere in
sight. So some unknown part of S4's ratio was per-column decode cost rather than anything the
join did, and the bench had no arm that could separate them: `S0` is the no-join control but it
is `avg(m1)`, narrow.

`S0W = avg(m1+…+m8) FROM %T`, no join. `S4 / S0W` is the join's contribution with decode
divided out.

## Measured — 4,000,000 rows, quantised, **median of 9**, serial

| id | shape | heap | columnar | col/heap |
| --- | --- | ---: | ---: | ---: |
| S0 | no join (control) | 149.822 | 46.020 | **0.31** |
| **S0W** | **no join, wide projection** | 255.515 | 395.943 | **1.55** |
| S1 | selective dimension join | 196.891 | 230.342 | **1.17** |
| S2 | unselective join | 308.647 | 312.684 | 1.01 |
| S3 | multi-dimension star | 609.535 | 646.269 | 1.06 |
| S4 | wide projection under a join | 247.323 | 446.211 | **1.80** |
| | **S4/S0W (the join, decode removed)** | **0.97** | 1.13 | 1.16 |

## The wide pair is an upper bound, not a measurement, and the heap arm is why

**`heap S4/S0W = 0.97`.** Adding a join, a second table and a filter made the heap **3% faster**
on the same scan. That is impossible — S4 is strictly more work than S0W on both arms. So the
wide pair carries at least 3% of error in a direction physics forbids, and that error sits in
the denominator.

It is not a small-sample artifact: this is the **median of 9**, and the same impossibility
appeared at 3 reps (0.96). Against it, the columnar figure is 1.13. **A 13% effect measured
with a denominator provably wrong by 3% the impossible way is not a measurement of 13%.**

So the honest statement is the stronger one: **the join's contribution to S4 is at or below
the noise floor, and S4 is per-column decode essentially in its entirety.** That is a better
argument than the one I set out to make — runtime filter pushdown would not move that row at
all, rather than moving a modest fraction of it.

**The bench now prints this line and says so itself** when the heap ratio falls below 1.00,
instead of leaving the impossibility to be rediscovered by arithmetic. That is how a number
gets quoted that should not have been.

## The narrow pair is the result, and it is far outside all of this

```
NARROW join penalty: columnar 5.01x   heap 1.31x   ->  3.81x relative
WIDE   join penalty: columnar 1.13x   heap 0.97x   <-  heap impossible
```

Columnar goes from **2.6x faster** than the heap to **1.17x slower**, purely by putting a join
above it. Both arms move the right way and both are far outside any noise floor measured on
this host. At 3 reps the same figure was 2.99x, so the effect is stable in kind if not to two
figures.

That is consistent with losing the vectorized aggregate — S0 has it, S1 does not, because the
join sits between the scan and the aggregate, and the bench already asserts exactly that on
both sides.

**This relocates #752 rather than weakening it.** The join penalty is larger than the issue
claimed; it is simply not where the issue was pointing. S1 is the case, not S4.

## S0W carries the OPPOSITE premise to S0, and it has to

S4 is unvectorized because a join sits between the scan and the aggregate. **S0W is
unvectorized for a different reason** — its target-list entry is an *expression* over columns
rather than a bare `Aggref`, which the vectorized path also declines. That coincidence is what
makes the division like-for-like.

If S0W ever starts vectorizing, `S4/S0W` silently becomes "the join **plus** the loss of the
fold path" instead of "the join", and nothing else in the bench would notice. So the new
premise asserts S0W is **not** vectorized, beside the existing one asserting S0 **is**:

```
ok  premise: [quantised] the no-join control really is vectorized
ok  premise: [quantised] a join between scan and aggregate disables it
ok  premise: [quantised] the wide no-join control is NOT vectorized, so S4/S0W is like-for-like
ok  premise: [quantised] S1/S2/S4 use the same join method in both arms (Hash Join)
```

A third arm was considered and is not needed: eight separate bare-column aggregates summed
does not vectorize either, so there is no unvectorized-vs-vectorized pair to build.

Credit to the reviewer for spotting that S4 might not be a join finding at all, and for
measuring that S0W is already unvectorized rather than leaving it to be assumed.

### DIFF
--- a/bench/run_bench_join.sh
+++ b/bench/run_bench_join.sh
@@ -352,6 +352,22 @@ for shape in "${SHAPE_A[@]}"; do
 		r=$(ratio "$c" "$h")
 		printf '%-4s %-32s %12s %12s %10s\n' "$shp" "$(q_desc $shp)" "$h" "$c" "$r"
 	done
+	# The join's own contribution, per arm, with decode divided out. Printed for
+	# BOTH arms on purpose: adding a join, a second table and a filter is strictly
+	# MORE work, so a heap ratio below 1.00 is physically impossible and is the
+	# scatter of this pair made visible. If it is there, the columnar ratio beside
+	# it cannot be read as a measurement -- the noise is larger than the effect.
+	# Hiding this line would leave the impossibility to be rediscovered by
+	# arithmetic, which is how a number gets quoted that should not have been.
+	jh=$(ratio "${MS["$shape:S4:heap"]:-}" "${MS["$shape:S0W:heap"]:-}")
+	jc=$(ratio "${MS["$shape:S4:columnar"]:-}" "${MS["$shape:S0W:columnar"]:-}")
+	printf '%-4s %-32s %12s %12s %10s\n' "" "S4/S0W (the join, decode removed)" "$jh" "$jc" \
+		"$(awk -v a="$jc" -v b="$jh" 'BEGIN { if (b+0 > 0) printf "%.2f", a / b; else print "-" }')"
+	if awk -v h="$jh" 'BEGIN { exit !(h+0 < 1.0 && h+0 > 0) }'; then
+		printf '%-4s %s\n' "" "^ heap S4/S0W is BELOW 1.00, which is impossible: a join cannot"
+		printf '%-4s %s\n' "" "  make the same scan faster. The columnar figure beside it is an"
+		printf '%-4s %s\n' "" "  upper bound on the join's share, not a measurement of it."
+	fi
 	hs=$($P -c "SELECT pg_total_relation_size('$(fact_name heap $shape)')")
 	cs=$($P -c "SELECT pg_total_relation_size('$(fact_name columnar $shape)')")
 	printf '%-4s %-32s %12s %12s %10s\n' "" "total relation size (bytes)" "$hs" "$cs" \
@@ -360,6 +376,10 @@ for shape in "${SHAPE_A[@]}"; do
 done
 echo "S0 against S1 is the finding: the same aggregate over the same rows, with"
 echo "and without a join. Read those two lines together before anything else."
+echo
+echo "S0W exists so S4 can be attributed. S4 projects eight float8 columns and a"
+echo "single float8 column already costs more to decode than an int one, so S4"
+echo "mixes the join with per-column decode. S4/S0W divides the decode out."
 if [ "$fail" != 0 ]; then
 	echo
 	echo "A PREMISE FAILED. Do not quote these numbers."
--- a/bench/run_bench_join.sh
+++ b/bench/run_bench_join.sh
@@ -366,6 +366,12 @@ for shape in "${SHAPE_A[@]}"; do
 	#
 	# The median is still printed, because the two disagreeing is the signal that
 	# the run was contended.
+	#
+	# MIN-OF-N IS NOT SCALE-FREE. It decreases weakly with N, so a min of 9 reps
+	# is not comparable to a min of 3 or of 20. The header prints `reps:` for this
+	# run, but ANY FIGURE QUOTED OUT OF THIS TABLE MUST CARRY THE REP COUNT WITH
+	# IT, or a later run at a different BENCH_REPS reads as a shift that is purely
+	# the statistic.
 	printf '%-4s %-32s %11s %11s %9s %9s %19s\n' id shape 'heap min' 'col min' 'col/heap' '(median)' 'col min..max'
 	for shp in $SHAPE_IDS; do
 		hm="${MN["$shape:$shp:heap"]:-}"
@@ -402,6 +408,10 @@ for shape in "${SHAPE_A[@]}"; do
 done
 echo "S0 against S1 is the finding: the same aggregate over the same rows, with"
 echo "and without a join. Read those two lines together before anything else."
+echo
+echo "Ratios are taken from MIN, not median: contention only makes a run slower,"
+echo "so the noise is one-sided and a median is not robust against it. Min of"
+echo "$REPS reps -- min is not scale-free, so quote the rep count with any figure."
 echo
 echo "S0W exists so S4 can be attributed. S4 projects eight float8 columns and a"
 echo "single float8 column already costs more to decode than an int one, so S4"
--- a/bench/run_bench_join.sh
+++ b/bench/run_bench_join.sh
@@ -347,18 +347,33 @@ echo "rows: $SCALE   reps: $REPS   serial   $(nproc) cores"
 echo
 for shape in "${SHAPE_A[@]}"; do
 	echo "-- data shape: $shape"
-	printf '%-4s %-32s %12s %12s %10s %19s\n' id shape heap columnar 'col/heap' 'col min..max'
+	# THE HEADLINE RATIO IS TAKEN FROM MIN, NOT MEDIAN, and that is not the usual
+	# instinct. Contention on a shared box can only make a run SLOWER, never
+	# faster, so the contamination here is ONE-SIDED. A median is robust against
+	# SYMMETRIC outliers; it is not robust against a one-sided contamination that
+	# shifts most of the samples, because then most of the samples ARE the
+	# contamination. The min is the run where least was stolen.
+	#
+	# This dataset says so directly. On the S0 columnar arm, one run's MEDIAN
+	# (46.020) agrees with another run's MIN (44.888) to 2.5%, while the two
+	# MEDIANS disagree by 31%. And in that run min 44.888 / median 60.488 / max
+	# 66.376 puts the median 73% of the way from min to max -- a right-skewed
+	# distribution, not scatter about a centre.
+	#
+	# That is also why more reps did not converge the narrow finding: more
+	# samples improve a median's estimate of a distribution whose centre is
+	# itself moving with ambient load. You cannot out-sample that.
+	#
+	# The median is still printed, because the two disagreeing is the signal that
+	# the run was contended.
+	printf '%-4s %-32s %11s %11s %9s %9s %19s\n' id shape 'heap min' 'col min' 'col/heap' '(median)' 'col min..max'
 	for shp in $SHAPE_IDS; do
-		h="${MS["$shape:$shp:heap"]:-}"
-		c="${MS["$shape:$shp:columnar"]:-}"
-		r=$(ratio "$c" "$h")
-		# The spread is printed for every row because the impossible-heap-ratio
-		# warning below is ONE-SIDED: it can only fire when noise happens to push
-		# the heap ratio under 1.0, which is about half the runs in which the same
-		# error is present. Its silence is not a clean bill. Min and max make the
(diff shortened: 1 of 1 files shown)


OffgridwithJD (commented):
Verified the arithmetic and it is internally consistent — `1.62 × 1.12 = 1.81`, against S4's
measured `1.82`, so the decomposition multiplies back to the row it decomposes. The S0W premise
is in with its reasoning, and it is the right premise: asserting S0W is **not** vectorized, for
a different reason than S4 is, is exactly what keeps the division like-for-like.

**One finding, and it is on the headline number rather than the method.**

## The 12% is not distinguishable from zero, and your own heap arm shows it

```
heap S0W  242.782 ms   ->   heap S4  234.178 ms
```

Adding a join, a second table and a filter made the **heap** 3.54% **faster**. That cannot be
true — S4 is strictly more work than S0W on both arms. So the wide pair carries at least 3.5%
of error, in the impossible direction, and that error is sitting in the denominator of the
wide-relative calculation.

Against that: the claimed columnar-relative effect is **12%**, on a **median of 3**, on a host
you and I have already established swings a heap arm by **24% between runs** and whose stated
scatter on #768 is **±20%**. A 12% effect measured with a denominator that is provably wrong by
3.5% in a direction physics forbids is not a measurement of 12%. It is a measurement of
something smaller than the noise.

**This strengthens your conclusion rather than weakening it.** The honest statement is not "the
join contributes 12% of that row" but "the join's contribution to S4 is at or below the noise
floor, and S4 is decode essentially in its entirety" — which is a cleaner result and a stronger
argument that a runtime filter pushdown would not move that row.

So I would take the 12% out of the title. `+23/-1` of bench code is not the place a number
gets minted, and that title is how it will be quoted.

## What I would do instead of another sweep

Do not re-run the whole thing for this. Two cheaper options:

- Report the wide pair as an upper bound: *"S4 ÷ S0W ≤ the noise floor; the join's share of S4
  is not resolvable here"*, with the heap's 0.96x shown as the evidence for why, not hidden.
  It is the most honest line in the dataset and it belongs in the output.
- If you do want a number, raise `BENCH_REPS` for the wide pair only and report **min**
  alongside median. Min is the least-contaminated estimate per arm on a contended box, which
  is the same reasoning we settled on for #768. Median of 3 cannot separate 12% from 0%.

## The narrow pair is the result, and it is well outside all of this

```
S0 → S1   columnar 3.96x   heap 1.32x   relative 2.99x
```

Both arms move the right way, both are far outside any noise floor either of us has measured,
and the relative is 3x. That is a real finding and it does relocate #752 — the strongest case
is S1, and the mechanism is the fold path disappearing rather than the join's own work. Worth
saying explicitly in the issue that this is a **relocation and not a weakening**: the join
penalty is bigger than anyone thought, it is just not where the issue was pointing.

It is also consistent with what I measured separately: `avg(m1)` vectorizes, and everything
that puts anything between the scan and a bare `Aggref` — a join, an expression, a wrapping
function — does not. S0 → S1 is that cliff, measured on the clock.

## Verdict

The mechanism, the control and the premises are right, and the narrow result stands on its own.
**Not approving while the title asserts 12%**, because that is the number the PR will be quoted
for and it is the one number here that the data does not support. Everything else I would take
as-is. CI is still running, 0 failed.

OffgridwithJD (commented):
Re-verified against the median-of-9 table. Every ratio reproduces: S0 0.307, S0W 1.550,
S1 1.170, S4 1.804, `S4/S0W` heap **0.968** against columnar **1.127**, narrow relative
**3.81x**. The title change is right and the bound is the right answer.

The re-run at nine reps is the part that settles it rather than the part that confirms it:
`heap S4/S0W` went 0.96 → **0.97**. The impossibility did not shrink with three times the
samples, so it is the pair's floor and not a small-sample accident. That is a stronger
statement than the bound alone and it is worth keeping in the record.

The warning logic is correct: `h+0 < 1.0 && h+0 > 0` fires only in the impossible region and
excludes the empty/missing case, so a failed measurement cannot masquerade as a passing check.

**Two things, and the second is a standard I owe you rather than a finding against you.**

## 1. The tripwire is one-sided, and its silence is not evidence

The heap ratio's distance below 1.0 is a *lower bound* on the pair's error, and it is
observable **only when noise happens to push it below 1.0**. When noise pushes it the other
way — heap `S4/S0W` landing at, say, 1.04 — the same error is present and completely invisible,
no warning fires, and the columnar figure beside it reads as a measurement.

So the check catches the error in roughly half of the runs where it exists. That is worth
having, and it is worth saying in the message it prints, because otherwise the *absence* of the
warning will be read as a clean bill. Something like: *"no warning here does not mean the
figure is resolvable; the heap ratio only exposes this error when it falls below 1."*

The general fix, and it is the same one we landed on for #768: print **min and max beside the
median** for each arm. Then the scatter is visible directly rather than inferred from an
impossibility that only shows up half the time. That is strictly more informative than the
tripwire and it makes the tripwire's one-sidedness moot.

## 2. The surviving number moved 27%, and my own standard applies to it too

The narrow relative went **2.99x at three reps → 3.81x at nine**. That is a **+27% move in the
headline of the finding that survived**.

I held you to this on the 12% and I should hold the same line here rather than only on the
number that died: `3.81x` quoted to three significant figures is more precision than a quantity
that moves 27% with rep count can carry. The finding is robust in *kind* — both arms move the
right way, the direction is unambiguous, and columnar going from **2.6x faster** than the heap
to **slower** purely by adding a join is not in doubt at any rep count. But the right way to
publish it is "roughly 3 to 4x" or "~4x", not 3.81x.

It matters specifically because #752's comment now carries that figure, and #752 is where
someone will later pick this up as the case for doing the work. A precise-looking number in a
motivating comment is exactly the shape that produced 2.6x/2.14x/1.70x on #766 — the difference
is only that this one is directionally safe.

## Verdict

The mechanism, the control, the premises and the bound are all right, and the re-run at nine
reps is what makes the bound authoritative rather than convenient. Neither point above is a
defect in the code. **Not approving only because CI is still running** (8 legs, 0 failed) — I
will approve when it lands. Point 2 is worth a one-line edit to the #752 comment whenever you
are next in there; point 1 is worth a sentence in the warning itself.

OffgridwithJD (commented):
Both points landed, and the spread column immediately earned itself. One more finding that
comes straight out of it, and I think it recovers a usable number rather than leaving the
narrow result at "roughly 3x".

## The median is the wrong statistic here, and your own spread data proves it

```
S0 columnar, run (c):   min 44.888   median 60.488   max 66.376
```

The median sits **73% of the way from min to max**. That is not scatter around a centre, it is
a heavily right-skewed distribution — the arm is usually slow and occasionally fast. Which is
exactly what CPU being stolen looks like: contention can only ever make a run slower, never
faster, so the contamination is **one-sided**.

Median is robust against *symmetric* outliers. It is not robust against a one-sided
contamination that shifts most of the samples, because then most of the samples are the
contamination. Min is — it is the run where least was stolen.

And the cross-run evidence says exactly that:

```
run (a)  S0 columnar MEDIAN = 46.020
run (c)  S0 columnar MIN    = 44.888     <- agree to 2.5%
run (c)  S0 columnar MEDIAN = 60.488     <- medians swing 46.0 -> 60.5, 31%
```

**One run's median agrees with another run's min to 2.5%, while the medians disagree by 31%.**
The min is the stable quantity across runs and the median is the unstable one. That is
backwards from the usual instinct, and it is the whole reason 2.78 / 2.99 / 3.81 did not
converge with more reps: more reps improve a median's estimate of a distribution whose centre
is itself moving with the ambient load.

**So I would report min, or min and median both, rather than median.** I would expect the
2.78-3.81 range to collapse substantially, and the narrow finding to come back with a second
digit that survives a re-run.

## And the skew is largest exactly where it hurts most

S0's columnar arm has the widest relative spread in the dataset — **48% of its own minimum** —
because it is the *shortest* arm at ~45-60 ms, so a fixed few milliseconds of scheduling jitter
is the largest proportional share there. And S0 sits in the **denominator** of the surviving
finding.

That has a direction, and it is the reassuring one: S0's median is inflated relative to its min
by more than the longer S1's is, so `median(S1) / median(S0)` **understates** the ratio. The
true narrow relative is at or above the 2.78-3.81 band rather than below it. "Roughly 3x" is
conservative, which is the right way to be wrong in a motivating comment.

If you want the range tighter without more reps, lengthening S0 would do it too — the jitter is
roughly absolute, so a 400 ms S0 would carry ~5% spread instead of 48%. But switching the
statistic is free and I would try that first.

## On the record

Your sentence is the better statement of the principle and I would keep it in those words: a
directionally-safe number in a motivating comment is *more* dangerous than a wrong one, because
nothing later contradicts it. #766 collected three retired figures precisely because each of
them looked reasonable at the moment it was quoted.

The warning now says its own silence is not a clean bill and points at the min..max column,
which is the right pairing — the tripwire catches the half it can prove, the spread column
covers the half it cannot.

**Not approving only because CI is re-running** (8 legs, 0 failed). Nothing outstanding from me
on the code.

OffgridwithJD (approved):
Approving. CI **12/12** on `b0b504a` — the head I reviewed — `CLEAN`, `shellcheck` green, which
is the relevant gate for a `bench/` change.

This one went through three corrections and the third is the one worth recording, because it
was not visible until the second was fixed:

1. **The 12% came out** after the heap arm showed `S4/S0W = 0.96` — a join making the same scan
   *faster*, which cannot happen. You then re-ran at nine reps rather than accept it, and it
   held at 0.97, which is what turned the bound from a convenient answer into a demonstrated
   one.
2. **The 3.81x came out** after a third run returned 2.78x, spanning 2.78–3.81 across runs.
3. **And then the statistic itself came out.** The spread column you added for (1) is what
   exposed (3): a median sitting 73% of the way from min to max is a skew, not scatter, and
   min-of-9 converges to 3.8% where median spanned 37%.

The published number moved *up* to ~3.5x as a result, so the methodology fix did not shrink the
finding — it recovered a second digit that survives a re-run, and in the direction the skew
predicted, since S0 is the shortest arm and sits in the denominator.

Both premises are right and both are the opposite of the obvious ones: S0 asserted **vectorized**
and S0W asserted **not** vectorized, because S4 is unvectorized for the join's sake and S0W for
its target-list entry being an expression, and that coincidence is the only thing making the
division like-for-like. If S0W ever starts vectorizing, that premise is what notices.

The min-of-N caveat living in the bench's own comment and report footer rather than the PR body
is the right call — it travels with the number, and ~3.5x is about to be picked up by whoever
does #752's work.

No findings outstanding. The `bench/run_bench_join.sh:286` EPIPE instance I raised is #771's
job, correctly kept out of this PR.

Issue #755, opened 2026-08-26
The vectorized aggregate covers one query shape, and GROUP BY vectorization is off by default


jdatcmd:
From `docs/limitations.md`, "Vectorized aggregate coverage":

> The vectorized aggregate path covers one shape only. That shape is
> `SELECT agg(col) FROM t [WHERE ...]`, on one relation and with no grouping.

Plus per-aggregate and per-type gates:

> - `count`, which includes `count(*)` and `count(col)`.
> - `sum` and `avg` on `smallint` and `integer` columns.
> - `min` and `max` on any type that has a default ordering.
> - `WHERE` clauses that are conjunctions of simple `column op const` comparisons.

`GROUP BY` vectorization exists (`pgcolumnar.enable_group_vectorization`) and is **off by
default**.

## What is actually missing, stated carefully

I first read this as "no vectorized `sum` on bigint or numeric" and that is **wrong**.
`COLUMNAR_AGG_SUM_INT8` and `COLUMNAR_AGG_SUM_NUMERIC` exist. What they do not get is the
batch path, and the code says why (`src/columnar_vector.c:828`):

> The internal-transtype kinds (sum/avg over int8/numeric) are not batch-eligible and stay
> on the serial node or the ordinary core Agg.

So the gap is tiering, not absence: an `internal` transition type cannot be combined by a
core Finalize, so those kinds fall off the fast path. `bigint` and `numeric` are the ordinary
types of a fact-table measure, so the shapes that fall off are not exotic.

## The three separable questions

1. **Shape.** One relation, no grouping. An aggregate above a join gets nothing, which is
   also what #752 is about from the other direction.
2. **Grouping by default.** `enable_group_vectorization` is off. Either it is ready and the
   default should move, or it is not and what is missing should be written down. Right now
   the answer is not visible from the outside.
3. **Transition types.** Whether the batch path can carry an `internal` transtype at all, or
   whether `int8`/`numeric` need a different representation to become batch-eligible.

Each is worth its own measurement before any of it is built. Nothing here is a defect: every
uncovered shape falls back to a correct scalar plan.

OffgridwithJD:
Measured the coverage on merged `main` (`8a71cc5`, PG 18.4 assert, container
`pgcolumnar-audit`) rather than reading it off the doc, and the shape of the gap
is different from the issue's framing in a way that changes what to build first.

## What a user actually gets with no GUC set

`vectorized` below means the plan carries the custom node's
[...]

PR #772, opened 2026-08-27
The vectorized aggregate accepts a target list that contains aggregates, not only one that is them: count(*)::text goes 634 ms to 0.057 ms

jdatcmd:
Part of #755 (the target-list-shape gate; the transition-type and grouping questions are
untouched).

## The gate

`columnar_vector.c` required **every target-list entry to be a bare `Aggref`**. An entry that
merely *contained* one fell off the vectorized path entirely — and on an unfiltered columnar
table that is the difference between answering from the zone maps and scanning the whole
relation.

Measured, 8,000,000 rows, PG17, min of 7, plan asserted on every arm:

| query | vectorized before | before | after | |
| --- | :-: | ---: | ---: | ---: |
| `count(*)` | yes | 0.030 ms | 0.033 ms | — |
| **`count(*)::text`** | **no** | **634.1 ms** | **0.057 ms** | **11,124x** |
| `avg(a)` | yes | 0.508 ms | 0.471 ms | — |
| **`avg(a)+avg(b)`** | **no** | **329.5 ms** | **0.504 ms** | **654x** |
| **`max(a)-min(a)`** | **no** | **245.4 ms** | **0.430 ms** | **571x** |
| **`round(avg(a)::numeric, 2)`** | **no** | **230.9 ms** | **0.488 ms** | **473x** |

Every losing shape is ordinary SQL. **Wrapping `count(*)` in a cast was costing a full scan of
the table.**

## The change

The aggregates are **pulled from** the target list rather than required to *be* it. The node
produces those, and a projection above computes the expressions — which is how core's own
`Agg` relates to an upper target.

The node is untouched. `PgColumnarCreateAggScanState` rebuilds its specs from
`custom_scan_tlist` assuming each entry is an `Aggref`, and that is now *guaranteed* rather
than merely true.

A target list that is already exactly the aggregates keeps its previous plan with **no
projection added**, so nothing that worked before is planned differently.

**The parallel arm needed no change and never did.** It uses core's partial grouping target,
which is bare `Aggref`s however the final target is shaped, and its `Finalize` is created with
`output_rel->reltarget` so core evaluates the expressions there. Refusing the shape in the
serial gate was killing the parallel arm with it, because the gate returned before the
parallel arm was reached.

## Test

`test/vector_agg_tlist_shape.sh`, **68 checks**. Every arm asserts **both** the plan (the fast
path was taken) and the **answer** against a heap table holding identical rows — a plan check
alone cannot see a wrong result, and an answer check alone cannot see that the fast path was
never taken.

Shapes covered beyond the obvious: the same aggregate twice (`avg(a)+avg(a)`, which must fold
to one), an aggregate used in two different expressions, a constant beside aggregates, a
`CASE` over aggregates, and an output order that deliberately differs from aggregate order.
Refusals: a bare column, an unsupported aggregate mixed into the list, an aggregate over an
expression of two columns.

## The surface this change OPENS, and why it is the real risk

Relaxing the shape gate **widens what reaches `pgcolumnar_classify_aggref`**. Before,
`count(*) FILTER (…)::text` was refused at the shape gate and never reached classification.
Now the shape gate passes and `classify_aggref` is the only thing between a modifier the node
cannot represent and a wrong answer.

Pinned, each asserting the refusal **and** the answer: `FILTER` on a wrapped aggregate,
`DISTINCT` on a wrapped aggregate, `FILTER` inside an expression over aggregates, `ORDER BY`
inside a wrapped aggregate. Plus empty and all-NULL relations, where NULL propagation through
a projection goes quietly wrong.

**These are not decorative, and what they guard is worse than "a wrong answer".** Removing
`aggdistinct` and `aggfilter` from `classify_aggref`'s first condition turns six arms red — and
the two failing answer arms return the **same hash as each other**, which is the fingerprint of
the modifier being *silently discarded* rather than mishandled. Reading the values instead of
the hashes, on 20,000 rows:

| | correct | with the refusal removed |
| --- | ---: | ---: |
| `count(*) FILTER (WHERE a > 5)` | 8,000 | **20,000** |
| `count(DISTINCT b)` | 7 | **20,000** |
| `count(*)` | 20,000 | 20,000 |

**The node does not fail on a modifier it cannot represent. It ignores it and returns the
unmodified aggregate.** A `FILTER` that removes 60% of the rows returns the count of all of
them; a `DISTINCT` over 7 values returns 20,000. No error, no plan tell, a number that looks
entirely plausible — the worst failure direction available, and the reason the refusal has to
live in `classify_aggref` rather than be inferred from a shape gate that used to hide these
before they got there.

## The parallel arm

The serial gate returned *before* the parallel arm was reached, so refusing a wrapped target
list killed both; the parallel arm itself needed no change. Asserted with the `Gather` checked
**before** any answer is believed:

```
avg(v)                      Gather=1  workers=4  vec=1   MATCH
avg(v)+avg(v)               Gather=1  workers=4  vec=1   MATCH
round(avg(v)::numeric,3)    Gather=1  workers=4  vec=1   MATCH
avg(v)::text                Gather=1  workers=4  vec=1   MATCH
```

**A metadata-answerable shape cannot test this at all** — `count(*)` and `avg` over int are
answered from the zone maps in microseconds, so no parallel plan can win and the arm would
measure nothing while looking green. The `float8` column forces the scan-fold path.

Restoring the pre-#755 gate turns those three wrapped arms red with
`gather=1 workers=4 vec=0` — core's parallel plan still runs and our node is gone, so the arms
discriminate on the gate rather than on parallelism.

## Removal proofs — six, and two of them came back GREEN

| mutation | result |
| --- | --- |
| restore the pre-#755 bare gate | **13 arms red** |
| reverse the aggregate list | **green** |
| declare a rotated target while the specs stay in order | **green** |
| declare only the first aggregate in every slot | **13 red, including the answer arms**, `QUERY_ERROR` |
| restore the gate, parallel arms | **3 red**, `gather=1 workers=4 vec=0` |
| accept `aggdistinct`/`aggfilter` in `classify_aggref` | **6 red**, answer arms with **wrong hashes** |

The two green ones are reported because they are properties worth knowing, not failed
attempts: `setrefs` matches the projection to the subplan **by `equal()`, not by position**, and
the **executor re-derives its specs from `custom_scan_tlist`**, so the node structurally cannot
compute something other than what it declares. The mis-ordering failure I was guarding against
cannot exist.

The fourth is the one that shows the answer arms are load-bearing rather than decorative: it
makes the projection lose an input, and the arms fail with `QUERY_ERROR` rather than a wrong
hash.

## A filtered aggregate is unrelated to this change

A `WHERE` routes to the **scan-fold** path, which is behind
`pgcolumnar.enable_ungrouped_vector_agg` — **off by default**. So a filtered aggregate declines
with or without this change. The suite asserts that default explicitly, rather than leaving a
filtered arm looking like a failure here, and asserts that with the GUC on a wrapped aggregate
takes that path too.

## Gate

**Full PG17 matrix: 215 suites ran, ALL PASSED**, 5 pre-existing skips. Including
`ungrouped_vector_agg`, `parallel_vector_agg`, `native_agg`, `native_groupagg`,
`batch_fold_explain` and `planner_choice_quality`.

## Not in this PR

#755's other two questions are untouched: the `internal` transition types (`sum`/`avg` over
`int8`/`numeric`) that are not batch-eligible, and whether
`pgcolumnar.enable_group_vectorization` should move off its default.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -116,6 +116,41 @@ pinned at `1.0-dev` or `1.0-alpha`, each true until the next version shipped.
 
 ### Added
 
+- The vectorized aggregate now accepts a target list that CONTAINS aggregates,
+  not only one that IS them. It required every entry to be a bare aggregate, so
+  an entry that merely contained one fell off the path entirely, and on an
+  unfiltered columnar table that is the difference between answering from the
+  zone maps and scanning the whole relation. Measured on 8,000,000 rows:
+
+  | query | before | after |
+  | --- | ---: | ---: |
+  | `count(*)::text` | 634.1 ms | **0.057 ms** |
+  | `avg(a)+avg(b)` | 329.5 ms | **0.504 ms** |
+  | `max(a)-min(a)` | 245.4 ms | **0.430 ms** |
+  | `round(avg(a)::numeric, 2)` | 230.9 ms | **0.488 ms** |
+
+  Two to four orders of magnitude, and every shape that was losing is ordinary
+  SQL: a cast on a count, a difference of two aggregates, a rounded average.
+  `count(*)` itself was already 0.030 ms, so wrapping it in a cast was costing a
+  full scan of the table.
+
+  The node is unchanged and still emits one bare aggregate per output column. The
+  aggregates are pulled out of the target list, the node produces those, and a
+  projection above it computes the expressions, which is how core's own
+  `Agg` relates to an upper target. A target list that is already exactly the
+  aggregates keeps its previous plan with no projection added, so nothing that
+  worked before is planned differently.
+
+  The parallel arm needed no change and never did: it uses core's partial
+  grouping target, which is bare aggregates however the final target is shaped.
+  Refusing the shape in the serial gate was killing the parallel arm with it.
+
+  Unsupported aggregates, aggregates over expressions, and grouped queries
+  decline exactly as before. A filter still routes to the scan-fold path behind
+  `pgcolumnar.enable_ungrouped_vector_agg`, which is off by default; with it on,
+  a wrapped aggregate takes that path too. New suite
+  `vector_agg_tlist_shape`. (#755)
+
 - A suite for the claim under `docs/limitations.md`'s "Replication and backup":
   that a columnar table is not a logical decoding source. The whole of #754 rests
   on that sentence and nothing asserted it. The only other logical-replication
--- a/docs/limitations.md
+++ b/docs/limitations.md
@@ -501,7 +501,12 @@ refuse it.
 ## Vectorized aggregate coverage
 
 The vectorized aggregate path covers one shape only. That shape is
-`SELECT agg(col) FROM t [WHERE ...]`, on one relation and with no grouping. The
+`SELECT agg(col) FROM t [WHERE ...]`, on one relation and with no grouping.
+
+The target list may contain expressions over those aggregates. `count(*)::text`,
+`avg(a)+avg(b)`, `round(avg(a), 2)` and `max(a)-min(a)` all take the path. What
+matters is that every aggregate in the list is a supported one, not that the list
+holds nothing else. The
 path also needs each aggregate, each column type and each filter clause to be
 supported. The supported set is:
 
--- a/src/columnar_vector.c
+++ b/src/columnar_vector.c
@@ -865,6 +865,8 @@ PgColumnarCreateUpperPaths(PlannerInfo *root, UpperRelationKind stage,
 	RangeTblEntry *rte;
 	Oid			relid;
 	List	   *tlist = output_rel->reltarget->exprs;
+	List	   *aggList;			/* the aggregates the tlist CONTAINS (#755) */
+	bool		tlistIsBare;		/* ... and whether the tlist already IS them */
 	ListCell   *lc;
 	int			naggs;
 	int			i;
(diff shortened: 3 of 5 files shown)


OffgridwithJD (commented):
Reviewed on the real 3-way merge (base `a1cd306`). Your suite passes **49/49**, and I went
after the surface this change opens rather than the one it claims.

## The widening is the risk, and it is gated

Before this, `count(*) FILTER (…)::text` was refused at the tlist-shape gate — it never reached
classification. Now the shape gate passes and `pgcolumnar_classify_aggref` is the only thing
standing between a modifier it cannot represent and a wrong answer. That is the bug a widening
introduces, so it is what I tested. Every shape checked for its **answer** against the core-Agg
path, not just for which plan it drew:

```
count(*) FILTER (WHERE a>5)::text     not vectorized   answers MATCH
count(DISTINCT b)::text               not vectorized   answers MATCH
length(string_agg(t,',' ORDER BY a))  not vectorized   answers MATCH
avg(a) FILTER (WHERE b<50) + 1        not vectorized   answers MATCH
```

`classify_aggref` rejects `aggorder`, `aggdistinct`, `aggfilter` and `aggvariadic` in its first
condition, so all four decline. No new wrong answer.

And the new capability, all matching the core-Agg oracle:

```
count(*)::text || 'x'          VEC   1, count(*)                VEC
round(avg(a)::numeric,2)       VEC   avg(a)+avg(b)              VEC
max(a)-min(a)                  VEC   avg(a)+avg(a)  (dedups)    VEC
```

`PVC_INCLUDE_*` rather than `PVC_RECURSE_*` on all three is the right call and the reason is
worth the comment you gave it: a window function or placeholder arrives as a node and is refused
by `IsA(n, Aggref)` instead of making `pull_var_clause` elog.

## NULL and empty propagate correctly

```
empty table   avg(a)+avg(b)   on=[] off=[]   MATCH
all-NULL      avg(a)+avg(b)   on=[] off=[]   MATCH
empty         count(*)::text  on=[0] off=[0] MATCH
all-NULL      max(a)-min(a)   on=[] off=[]   MATCH
```

## Your parallel claim is confirmed, and my first attempt at checking it was void

I first ran the parallel arm on `count(*)::text`, `avg(a)+avg(b)`, `max(a)-min(a)` over `int`
and got `Gather=0` on all of them — those are answered from metadata and are instant, so no
parallel plan can win and I had measured nothing. That is the engagement-premise problem I have
been holding you to, so I re-ran it with a shape that must scan (`float8` behind
`enable_ungrouped_vector_agg`) and asserted the Gather **before** trusting any answer:

```
avg(x)                      Gather=1  workers=4  vec=1  MATCH
avg(x)+avg(y)               Gather=1  workers=4  vec=1  MATCH
round(avg(x)::numeric,3)    Gather=1  workers=4  vec=1  MATCH
avg(x)::text                Gather=1  workers=4  vec=1  MATCH
```

Four workers actually launched, wrapped shapes accepted on the parallel path, answers identical
to the serial core Agg. One gate was indeed killing both paths.

## Your two green removal proofs are findings, and reporting them as such is right

"Reverse the aggregate list → green, because `setrefs` matches by `equal()` not position" and
"rotate the declared target → green, because the executor re-derives specs from what it
declares" are statements about what **cannot** break, which is worth more than two more red
arms. The fourth proof is what makes the 33 answer arms load-bearing — without it they were 33
arms you could not show were doing anything.

## And you corrected me on #764. You are right and I was wrong

I claimed the suite encoding #171/#355/#362 was `planner_choice_quality`, which CI never
invokes, and used that to argue #766 could not be CI-gated. Checked on `a1cd306`:

```
analyze_stats.sh   encodes #171 #355 #362 #363 #376 #507
                   timing-assertion lines: 0        <- entirely clock-free
                   not in is_timing_suite           <- so CI runs it
planner_choice_quality: 5 check_ratio_timing calls, 5 other checks
```

So the regressions **are** gated in CI, clock-free, and my "#764 is harder blocking for #766"
was wrong in both directions — it does not block #766, and #766's risk is testable in CI today.
I have corrected it in my own notes. Thank you for not letting it stand.

## Verdict

No findings. **Not approving yet** — CI still running. I will approve when it lands.

OffgridwithJD (approved):
Approving. CI **12/12** on `c3b881d` — `CLEAN` — and the new suite genuinely **ran** on both
legs rather than skipping, which is the summary line rather than the job colour:

```
suites (PG 17)  vector_agg_tlist_shape=PASS   (211 ran, 9 skipped)
suites (PG 18)  vector_agg_tlist_shape=PASS   (214 ran, 6 skipped)
```

I verified the last push was **comment-only** — `f6c87c2..c3b881d` is 16 added lines in the
suite and no non-comment change to `columnar_vector.c` — so the work I did at `f6c87c2` carries:
68/68 baseline, the modifier mutation reproduced with a distinct `.so`, and the parallel arm at
`Gather=1 workers=4 vec=1` with answers matching the serial core Agg.

## What this change actually is

`count(*)::text` costing a full scan of the table was the finding, and the fix is planner-only
because the executor already assumed what it now guarantees — `PgColumnarCreateAggScanState`
rebuilds its specs from `custom_scan_tlist` on the assumption every entry is an `Aggref`. Pull
the aggregates, produce those, project above. That the serial gate was `return`ing before the
parallel arm was reached — so one gate was refusing both paths — is the part that makes the
scale of it make sense.

## The guard is the interesting half, and it is now measured rather than argued

The widening moves the burden onto `classify_aggref`, and removing `aggdistinct`/`aggfilter`
from its first condition shows what that burden is:

```
count(*) FILTER (WHERE a > 5)     correct 8,000    mutant 20,000
count(DISTINCT b)                 correct     7    mutant 20,000
count(*)                          correct 20,000   mutant 20,000
```

The node does not fail on a modifier it cannot represent — **it ignores it and returns the
unmodified aggregate**. A `DISTINCT` over seven values returning twenty thousand is wrong by
2,857x with no error, no plan tell, and a number that looks entirely reasonable. Putting that in
the suite's own comment, with the values rather than the hashes, is what makes the refusal
legible to whoever touches that condition next.

## Six removal proofs, two of them green

Reporting the two green ones as findings is right, and it is a distinction worth keeping: *"the
aggregate list can be reversed because `setrefs` matches by `equal()` not position"* is a
statement about what cannot break, which is worth more than another red arm. The fourth — declare
only the first aggregate in every slot, 13 red including the answer arms with `QUERY_ERROR` — is
what makes the 33 answer arms load-bearing rather than decorative.

And the parallel proof is the one I would keep: restoring the pre-#755 gate gives
`gather=1 workers=4 vec=0`. Core's parallel plan still runs and our node is gone, so the arms
discriminate on the gate rather than on parallelism — which my own first attempt at checking
that path could not have shown, because it never engaged.

No findings.

PR #773, opened 2026-08-27
Every bench runner refuses an assert build instead of documenting the requirement

jdatcmd:
Bench and test harness only; no product code.

## The gap

Every bench script's header said to run against a non-assert build. **They said so and checked
nothing.** The invocations that produced this session's #755, #768, #752 and #753 figures
passed `/usr/local/pg17`, which is `--enable-cassert`, and nobody noticed until
`AllocSetCheck` turned up at 4–5% of a profile.

That is the shape this repository has found **three times today**:

| | rule | check |
| --- | --- | --- |
| #771 | selftest 080 forbids an EPIPE pattern | did not scan `bench/` |
| #764 | `runs_alone` documented an arrangement | the code had the opposite one |
| **here** | headers require a non-assert build | the invocation used an assert one |

## The fix

`run_profile.sh` **already had exactly this guard** and the others did not, so the shape is
copied rather than invented: refuse, with a named override that *labels* the output rather
than suppressing the objection.

Refuse rather than warn, because a benchmark's whole output is numbers and **a warning above a
table does not travel with the table**.

Applied to all six runners: `run_bench.sh`, `run_bench_join.sh`, `run_bench_fsst.sh`,
`run_bench_readstream.sh`, `run_clickbench.sh`, and `run_profile.sh` which had it.

Verified all three ways: an assert prefix **exits 1** with the FATAL text,
`BENCH_ALLOW_CASSERT=1` warns and continues, a non-assert prefix is silent.

## The guard is itself guarded, and the list is derived

`bench_guards.sh` now asserts three properties per runner — it refuses, it offers a named
override, it exits non-zero rather than warning past — and **derives the list from the
directory** rather than keeping one by hand.

### Removal proofs, three

1. **Strip the guard from one runner** → `bench_guards` names that runner, 3 red.
2. **Add a *new* runner with the guard stripped** → named too. That is what the
   directory-derived loop buys over a hand-kept list: a future bench script cannot arrive
   unguarded because nobody remembered to add it here.
3. **Empty the bench directory** → the count premise fails, so the loop cannot report
   compliance by globbing nothing.

The premise is a **lower bound, not an equality**. Its job is to show the scan read something;
an equality would turn a legitimate new benchmark into a red here, which is a check that
punishes the thing it exists to cover.

## On the size of the skew, since that is the obvious question

On the one measurement where it was **checked rather than reasoned about**, it was **2.5%** —
#768's per-column ladder, assert against non-assert:

| | assert | non-assert |
| --- | ---: | ---: |
| ms per added int column | 29.2 | 28.9 |
| ms per added text column | 268.3 | 261.6 |

Small there, and inside the ±20% already attached to that figure. `AllocSetCheck` runs per
**context reset**, not per `palloc`, so it does not concentrate on an allocating arm the way I
first assumed. But it is allocation- and reset-shaped, so it is not small everywhere — and the
point of the guard is that nobody knew which case they had.

### DIFF
--- a/bench/run_bench.sh
+++ b/bench/run_bench.sh
@@ -46,7 +46,23 @@ PG_CONFIG="${1:-/usr/local/pg17_nc/bin/pg_config}"
 # reasoned about, it was 2.5% (#768's per-column ladder, assert against
 # non-assert). Small here; it is allocation- and reset-shaped, so it is not
 # small everywhere, and the point is that nobody knew which case they had.
-BENCH_CASSERT="$("$PG_CONFIG" --configure 2>/dev/null | grep -c 'enable-cassert' || true)"
+# --configure captured ONCE, and an empty capture is a REFUSAL rather than a
+# pass. Written as a single grep -c it could not tell "not an assert build" from
+# "could not tell": a missing, non-executable or failing pg_config gives empty
+# output, grep -c gives 0, and the 2>/dev/null swallows the evidence -- so the
+# guard returns its cleanest verdict on the least information. Reachable, not
+# theoretical: these scripts have default prefixes that do not exist on every
+# machine, so a no-arg invocation can pass a guard on a pg_config that never ran.
+#
+# Refuse when you cannot tell, for the same reason as when the answer is yes.
+BENCH_CONFIGURE="$("$PG_CONFIG" --configure 2>/dev/null || true)"
+if [ -z "$BENCH_CONFIGURE" ]; then
+	echo "FATAL: $PG_CONFIG produced no --configure output, so this script cannot"
+	echo "       tell whether it is an --enable-cassert build. Refusing rather than"
+	echo "       assuming it is not: check the path is right and executable."
+	exit 1
+fi
+BENCH_CASSERT="$(printf '%s' "$BENCH_CONFIGURE" | grep -c 'enable-cassert' || true)"
 if [ "${BENCH_CASSERT:-0}" != "0" ]; then
 	if [ -n "${BENCH_ALLOW_CASSERT:-}" ]; then
 		echo "== WARNING: --enable-cassert build. Every number below includes"
--- a/bench/run_bench.sh
+++ b/bench/run_bench.sh
@@ -26,6 +26,40 @@
 set -euo pipefail
 
 PG_CONFIG="${1:-/usr/local/pg17_nc/bin/pg_config}"
+
+# ---------------------------------------------------------------------------
+# The non-assert requirement, ENFORCED rather than documented
+# ---------------------------------------------------------------------------
+#
+# This script's header has always said to run against a non-assert build. It
+# said so and checked nothing, and the invocation that measured #755, #768, #752
+# and #753 passed an --enable-cassert prefix. That is the failure shape this
+# repository keeps finding: the rule existed, the check did not.
+#
+# Refuse rather than warn, because a benchmark's whole output is numbers and a
+# warning above a table does not travel with the table. The override exists
+# because a comparison between two assert builds is still valid, and because an
+# assert build is sometimes the only one to hand -- but then every figure has to
+# be labelled, which is what the banner does.
+#
+# On the one measurement where the size of the skew was checked rather than
+# reasoned about, it was 2.5% (#768's per-column ladder, assert against
+# non-assert). Small here; it is allocation- and reset-shaped, so it is not
+# small everywhere, and the point is that nobody knew which case they had.
+BENCH_CASSERT="$("$PG_CONFIG" --configure 2>/dev/null | grep -c 'enable-cassert' || true)"
+if [ "${BENCH_CASSERT:-0}" != "0" ]; then
+	if [ -n "${BENCH_ALLOW_CASSERT:-}" ]; then
+		echo "== WARNING: --enable-cassert build. Every number below includes"
+		echo "   validation a production build never runs, and is comparable only"
+		echo "   against another assert build."
+	else
+		echo "FATAL: $("$PG_CONFIG" --bindir) is an --enable-cassert build."
+		echo "       This benchmark's numbers would carry assert-only validation."
+		echo "       Use a non-assert build, or set BENCH_ALLOW_CASSERT=1 to run"
+		echo "       anyway and have the output labelled as such."
+		exit 1
+	fi
+fi
 BINDIR="$("$PG_CONFIG" --bindir)"
 PORT="${BENCH_PORT:-55432}"
 SCALE="${BENCH_SCALE:-6000000}"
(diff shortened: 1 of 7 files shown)


OffgridwithJD (commented):
Reviewed on the real 3-way merge. The structure is right and I verified the part that makes it
durable rather than a one-time sweep: `bench_guards.sh` derives the runner list from the
directory (`ls .../bench/run_*.sh`), asserts three separate things per runner — refuses, offers
a named override, exits non-zero — and carries a premise that every runner was read rather than
a hand-kept subset. Proving it by adding a script is the right proof for that shape.

"Refuse rather than warn, because a benchmark's whole output is numbers and a warning above a
table does not travel with the table" is the correct principle and the correct reason.

## One finding: the guard cannot tell "not an assert build" from "could not tell"

```sh
BENCH_CASSERT="$("$PG_CONFIG" --configure 2>/dev/null | grep -c 'enable-cassert' || true)"
if [ "${BENCH_CASSERT:-0}" != "0" ]; then   # refuse
```

If `$PG_CONFIG` does not run — missing, not executable, a wrapper that fails, `--configure`
unavailable — then the output is empty, `grep -c` gives `0`, `2>/dev/null` and `|| true` swallow
the evidence, and **the guard passes**. It reports "not an assert build" when the truth is "I
could not establish what this build is."

Demonstrated:

```
PG_CONFIG=/nonexistent/pg_config -> BENCH_CASSERT=[0] => PASSES (guard inert)
```

**And it is reachable rather than theoretical.** `run_bench.sh` defaults to
`/usr/local/pg17_nc`, which exists on your container and **does not exist on mine**. A no-argument
invocation here passes the guard on a `pg_config` that never ran. The script fails later for
other reasons, so nothing silently produces wrong numbers today — but the guard's verdict is
already wrong at the point it is given, and the next runner that resolves its prefix differently
inherits that.

Neither runner checks `$PG_CONFIG` before the guard uses it — I looked at `run_bench.sh`
(PG_CONFIG line 28, guard line 49) and `run_clickbench.sh` (109 / 130).

One line, and it is the principle the PR already argues:

```sh
BENCH_CFG="$("$PG_CONFIG" --configure 2>/dev/null || true)"
[ -n "$BENCH_CFG" ] || { echo "FATAL: $PG_CONFIG --configure produced nothing;"; \
                         echo "       cannot establish whether this is an assert build."; exit 1; }
BENCH_CASSERT="$(printf '%s' "$BENCH_CFG" | grep -c 'enable-cassert' || true)"
```

Refuse when you cannot tell, for the same reason you refuse when the answer is yes: a benchmark
that cannot say which build it measured has the same problem as one that measured the wrong
build, and only one of those is currently caught.

The removal proof extends: point a runner at a non-existent `pg_config`. Today it passes the
guard; with the check it refuses, and `bench_guards.sh` can assert that fourth property beside
the three it already has.

## The symmetry worth noting in the body

This PR's thesis is "the rule existed, the check did not." The check now exists — and it has a
state in which it cannot check and calls that a pass. That is the same shape one level in, and
it is the fourth time this pattern has surfaced in a fortnight: the rule, the premise, the
premise-for-the-premise, and now the guard's own unknown-verdict case.

**Not approving on that one line.** CI is already green (12/12, `CLEAN`), so this is the only
thing outstanding.

OffgridwithJD (commented):
Fix verified on the real merge, including the case that made it reachable rather than
theoretical:

```
$ bash bench/run_bench.sh /nonexistent/pg_config
FATAL: /nonexistent/pg_config produced no --configure output, so this script cannot
       tell whether it is an --enable-cassert build. Refusing rather than
       assuming it is not: check the path is right and executable.
```

All six runners carry the capture-then-check form, none decides from a bare pipe, and adding
`run_profile.sh` was right — its older form had the same hole and it is the script most likely
to be pointed at an odd prefix.

**One thing I checked because it is where this would go wrong next.** `run_profile.sh` uses
`PROFILE_CONFIGURE` / `CASSERT` / `PROFILE_ALLOW_CASSERT` where the other five use `BENCH_*`. A
divergent sixth is normally exactly what a guard-of-a-guard misses. It does not here, because
every assertion in `bench_guards.sh` keys on shared **text** rather than on the prefix:

```
grep -c 'enable-cassert'                      both
grep -cE 'ALLOW_CASSERT'                      matches BENCH_ and PROFILE_
grep -A24 'enable-cassert' | grep -c 'exit 1' both
grep -c 'produced no --configure output'      the shared message
grep -cE '\$\("\$PG_CONFIG" --configure[^)]*\| *grep -c'  must be 0
```

That is a deliberate choice and worth keeping deliberate — the unprefixed `ALLOW_CASSERT` match
is the line that makes a differently-named sixth runner still covered. Worth a word in the file
saying so, or someone will "tidy" it to `BENCH_ALLOW_CASSERT` and silently drop `run_profile.sh`
from two of the five checks.

Asserting both halves per runner — refuses-when-it-cannot-tell **and** does-not-decide-from-a-
bare-pipe — is the right pair. The first is the behaviour; the second is the specific
construction that produced the behaviour, and forbidding it stops the fix being undone by a
plausible-looking rewrite.

And the removal proof is the right one: revert a runner, watch both new arms go red, and confirm
that runner then accepts `/nonexistent/pg_config`. That last clause is what makes it a
demonstration rather than two more red arms.

**Approving as soon as CI lands** — 10/12 complete, 0 failed, nothing outstanding from me.

OffgridwithJD (approved):
Approving. CI **12/12** on `922da37`, `CLEAN`. I checked the last push before relying on the
earlier work: `e5b40b3..922da37` is 10 added lines in `bench_guards.sh`, all comment, no
non-comment change anywhere — so the verification at `e5b40b3` carries, including the case that
made the finding reachable:

```
$ bash bench/run_bench.sh /nonexistent/pg_config
FATAL: /nonexistent/pg_config produced no --configure output, so this script cannot
       tell whether it is an --enable-cassert build. Refusing rather than
       assuming it is not: check the path is right and executable.
```

And you improved on what I asked for. I suggested a word saying the unprefixed match is
deliberate; you **proved** it instead — *"keying that one check on `BENCH_ALLOW_CASSERT` leaves
`run_profile.sh` as the only red"* — which turns a claim about why the code is written that way
into a demonstration anyone can re-run. The two closing sentences are the part that will still
be working in a year:

> If anyone tidies the prefixes to match, this comment says the checks need no change. If anyone
> tidies the CHECKS to match the prefixes, that is the moment coverage is lost with nothing going
> red.

That names the direction the damage comes from, which a comment saying "deliberate" would not.

## What this PR is, in the end

The header said "run against a non-assert build" and checked nothing, and four issues' numbers
were taken on an assert prefix because of it. The fix is not the refusal — it is that the
refusal now has three properties each asserted separately: it refuses when the answer is yes, it
refuses when it *cannot tell*, and it cannot be rewritten into deciding from a bare pipe. The
second is the one I raised and the third is what stops the second being undone by a plausible
tidy-up.

Worth recording that the measured stakes turned out small — 2.5% on the one ladder where the
skew was checked rather than reasoned about — and that this is still worth doing. The problem
was never the size of the error. It was that nobody knew which case they had.

No findings.

PR #774, opened 2026-08-27
vacuum_sorted: self-gate on already-sorted AND nothing-to-reclaim (#760)

### LINKED ISSUE #760: vacuum_sorted has no self-gate, and the obvious one would silently stop reclaiming deleted rows
#415 gave the online `recluster` a self-gate: when the relation is already exactly the
Z-order run over the requested columns with nothing appended, it returns 0 instead of
rewriting every group. `design/ISSUE_415_AUTOVACUUM.md` promises the mirror for
`vacuum_sorted`, and that gate does not exist.

With #758 recording `sorted_kind = 'lexicographic'` and the applied key, the information the
gate needs is now in the catalog, so the gate itself is a few lines beside the existing one
at `src/columnar_vacuum.c:604`. It was deliberately left out of that PR, for a reason that
has to be designed around rather than skipped.

## The constraint: `vacuum_sorted` reclaims, and `recluster` does not

`docs/ARCHITECTURE.md:225`:

> `pgcolumnar.vacuum` materializes a relation's live rows [...] This combines small stripes
> and physically reclaims deleted-row space. `pgcolumnar.vacuum_sorted` runs the same
> rewrite but feeds the live rows through a tuplesort [...]

So `vacuum_sorted` has two jobs, and a gate that only asks "is it already in this order"
would answer yes on a relation that is in order and full of dead rows, and silently stop
reclaiming. That is a data-size regression with no error and no report.

## Measured on the gate that already ships

The same shape is observable on `recluster` today, PG 17, 5,000 rows in 5 groups:

```
after recluster('g','k','j'):      groups=5  live rows=5000
  recluster on the same key:       returns 0        (gate fires, correct)
DELETE FROM g WHERE id % 2 = 0;    live rows=2500   appended_groups=0
  recluster on the same key:       returns 0
  after it:                        live rows=2500   groups=5
```

2,500 dead rows stay stored. For `recluster` that is **within contract**: its COMMENT is
"re-establish global Z-order clustering [...] Returns the number of groups reclustered", and
reclamation is another verb's job. It is recorded here as the measurement, not as a defect.

**Which verb, measured** (5,000 clustered rows, half deleted; reviewer's numbers, and they
correct an earlier draft of this issue that named `compact()`):

| verb | returns | dead rows | groups |
| --- | --- | ---: | ---: |
| `compact` | 0 | 2500 -> 2500 | 5 -> 5 |
| `compact_rewrite` | 5 | 2500 -> 0 | 5 -> 5 |
| `vacuum` | | 2500 -> 0 | 5 -> 3 |
| `vacuum_sorted` | | 2500 -> 0 | 5 -> 3 |
| `recluster` | 0 | 2500 -> 2500 | 5 -> 5 |

`pgcolumnar.compact()` reclaims **nothing** on this shape. The online reclaimer is
`compact_rewrite()`; the eager ones are `vacuum` and `vacuum_sorted`. `vacuum_sorted`
reclaims, which is the whole reason it cannot take an ordering-only gate.

For `vacuum_sorted` the same behaviour would be out of contract.

## What the gate has to be

The skip condition is not "already in this order". It is "already in this order **and there
is nothing to reclaim**". The second half needs a deleted-row count for the relation, which
`pgcolumnar.stats()` already reports and `pgcolumnar.maintenance_due()` already thresholds
as `deleted_fraction`.

## Arms it needs

- gate fires: `vacuum_sorted('t','k','j')` twice, second call leaves the storage id and the
  `(group_number, file_offset)` set unchanged;
- different key rewrites;
- an appended tail rewrites;
- a **Z-order** run of the same key rewrites, and the result really is in `k` order;
- **deleted rows defeat the gate**: delete a row inside the run, `vacuum_sorted` on the same
  key still rewrites, and the deleted-row count goes to zero afterwards.

The last arm is the one this issue exists for.

jdatcmd:
Closes #760.

`pgcolumnar.recluster()` has self-gated since #415 so a scheduler can call it
speculatively without paying a full rewrite each time, and
`design/ISSUE_415_AUTOVACUUM.md` promises the mirror for `vacuum_sorted`. #758 put
the information the gate needs into the catalog (`sorted_kind`, `sorted_by`), so
this is the follow-through.

## The mirror is not a copy, and that is the whole of the PR

`vacuum_sorted` has **two** jobs: it orders the live rows and it physically
reclaims deleted-row space. The obvious gate -- "is it already in this order",
which is complete for `recluster` -- answers yes on a relation that is in order
and full of dead rows, and silently stops reclaiming. No error, no report, a
data-size regression.

So the skip condition is "already in this order **and** there is nothing to
reclaim". Four conditions, in `vacuum_sorted_gate_is_noop`:

1. the recorded run is `lexicographic`. A `zorder` run over the same columns is
   not this order: Z-order over two or more columns is not a sort on any one of
   them;
2. the recorded key is exactly these columns, in this order;
3. every row group lies inside `[sorted_from, sorted_through]`. This also covers
   stripe-combining -- a group inside the run was produced *by* the rewrite that
   set the mark, so it is already combined;
4. no group carries a deleted row, and none is empty.

Anything else falls through to the full rewrite, so re-sorting by a new key,
re-sorting a Z-ordered table, and reclaiming all still work. A mark written
before `sorted_from`/`sorted_kind` existed leaves them unset and never gates,
which is the safe pre-#415 default.

## Removal proofs

Three mutations, each run against the full 34-check suite.

**A. delete the gate** (`if (false && ...)`) -- exactly the three no-op arms go
red, nothing else:

```
FAIL  a second vacuum_sorted on the same key does NOT rewrite: got [rewritten(storage moved)] want [same]
FAIL  a third identical call is also a no-op: got [rewritten(storage moved)] want [same]
FAIL  the call after the reclaim is a no-op again: got [rewritten(storage moved)] want [same]
```

**B. the WRONG MIRROR** -- port #415's ordering-only gate by deleting conditions
3 and 4. This is the mutation the suite exists for, and it is the one that would
ship silently:

```
FAIL  vacuum_sorted with dead rows DOES rewrite, gate or no gate: got [no] want [yes]
FAIL  and it reclaimed the space: deleted rows go to zero: got [10000] want [0]
```

10,000 rows stay stored where the un-gated call reclaims all of them.

**Correction, from the review.** I originally wrote that mutation B reddens
*exactly those two* arms. That was true when I ran it and is now stale: the
transaction arms were added to the suite afterwards, and they must redden under
any removal of condition 4. Re-run by the reviewer, deleting conditions 3+4
reddens **6** arms and deleting condition 4 alone reddens **4**. The suite is
stronger than the PR first advertised; only the count was wrong. The two arms
quoted above are still the ones the suite exists for.

**C. delete both flushes** -- *the suite stays green*, and the PR says so rather
than implying otherwise. See below.

## Two things I got wrong on the way, both left visible in the tree

**The rewrite signal.** I first detected "did this rewrite" the way
`recluster_gate.sh` does, from the `(group_number, file_offset, rowcount)`
multiset. That is right for `recluster`, which rewrites *inside* the existing
storage and so reallocates offsets. It is useless here for exactly that reason: a
`vacuum_sorted` rewrite starts a **fresh storage**, so group numbers and offsets
restart from zero and the multiset comes out byte-identical on a real rewrite of
unchanged data. It would have called every rewrite a no-op.

It did not get through, because the helper cross-checked two independent signals
and reported disagreement as a failure instead of picking one:

```
FAIL  premise: the first vacuum_sorted does real work: got [DISAGREE(phys=no storage=yes)] want [yes]
```

Storage id now decides; the layout hash is kept on the no-op arms as a second
witness that nothing moved.

**A vacuous arm, and an honest label instead of a fix.** The transaction arms
first inserted 3,000 rows with `chunk_group_row_limit = 1000`, which flushes three
complete groups during the `INSERT` itself -- so nothing was ever pending and the
arms stayed green with both flushes deleted. I shrank the tail below the chunk
group limit and they still stayed green. Measured why, rather than assuming:
a 500-row insert reports `storedrows=20500` from *inside* the transaction,
because the write state and the delete vector are both flushed at end of
statement.

So **no arm in this suite proves the gate's flush is load-bearing**, and neither
the code comment nor the test file claims one does. The flush stays because the
gate *skips a rewrite* on what it reads, so reading stale state would be a silent
correctness bug, and the invariant that makes it unnecessary belongs to another
module. The arms pin that invariant instead: if end-of-statement flushing ever
stops holding, they go red and the flush becomes the fix.

## Tests

`test/vacuum_sorted_gate.sh`, 34 checks, registered in `run_all_versions.sh` in
its C-order slot (verified with `--list-suites | LC_ALL=C sort -c`, and
`harness_selftest` 153/153).

PG17: the new suite plus the 13 neighbours that touch ordering, reclaim and the
sorted mark -- `eager_ordering_record`, `recluster_gate`, `recluster_extent`,
`native_recluster`, `maintenance_due`, `autovacuum`, `native_reclaim_frag`,
`native_reclaim_cycles`, `native_reclaim_reconcile`, `vacuum_stripe_count`,
`vacuum_lock_privilege`, `sorted_pathkeys`, `native_ownership` -- all PASSED.
`docs_style` PASSED.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,30 @@ pinned at `1.0-dev` or `1.0-alpha`, each true until the next version shipped.
 
 ### Added
 
+- `pgcolumnar.vacuum_sorted()` now self-gates: when the relation is already
+  exactly the requested lexicographic run with nothing appended and nothing to
+  reclaim, it skips the rewrite instead of materializing every live row through
+  a tuplesort again. `pgcolumnar.recluster()` has had the ordering half of this
+  since #415 so a scheduler can call it speculatively;
+  `design/ISSUE_415_AUTOVACUUM.md` promised the mirror here.
+
+  The mirror is not a copy, and that is the whole of the change.
+  `vacuum_sorted` has two jobs: it orders the live rows and it physically
+  reclaims deleted-row space. The obvious gate -- "is it already in this
+  order", which is complete for `recluster` -- would answer yes on a relation
+  that is in order and full of dead rows, and silently stop reclaiming. Ported
+  unchanged into a tree carrying it, that gate leaves 10,000 deleted rows
+  stored where the un-gated call reclaims all of them, with no error and no
+  report.
+
+  So the skip condition is "already in this order AND there is nothing to
+  reclaim": a `lexicographic` run (a `zorder` run over the same columns is not
+  this order, because Z-order over two or more columns is not a sort on any one
+  of them), over exactly these columns in this order, covering every row group,
+  with no deleted row and no empty group. Anything else falls through to the
+  full rewrite, so re-sorting by a new key, re-sorting a Z-ordered table, and
+  reclaiming all still work.
+
 - The columnar scan now tells the planner the order a sorted rewrite left the
   rows in, so `ORDER BY` on that key plans no `Sort`. Every `pathkeys` field in
   the tree was `NIL`, so a table that `pgcolumnar.vacuum_sorted` had physically
--- a/docs/sql-reference.md
+++ b/docs/sql-reference.md
@@ -86,6 +86,17 @@ SELECT pgcolumnar.set_options('events', sort_by => ARRAY['customer_id']);
 SELECT pgcolumnar.vacuum_sorted('events');                         -- declared key
 ```
 
+Re-running `vacuum_sorted` with the same key on a table that is still in that
+order is a fast no-op. It skips the rewrite on four conditions. The recorded
+key matches. The recorded kind is a sort and not a Z-order. The sorted run
+already covers every row group. No row group holds a deleted row.
+
+That last condition is part of the gate because `vacuum_sorted` both orders the
+rows and reclaims deleted-row space. A table that is in order but holds deleted
+rows is rewritten, so the space is still reclaimed. Deletes, inserts, a
+different key, or a table arranged by `cluster()` all fall through to the full
+rewrite.
+
 A sort key is a **trade** and not a free gain. A sort by a segment key puts the
 rows of each segment together, so a filter on that key skips most groups. It also
 spreads each *other* dimension across every group.
--- a/src/columnar_vacuum.c
+++ b/src/columnar_vacuum.c
@@ -204,6 +204,8 @@ static bytea *cluster_zorder_key(Datum *values, bool *isnull, AttrNumber *atts,
 								 int ncols, TupleDesc tupdesc);
 /* Names an ordering rewrite records as its key (defined later, #415) */
 static List *sort_key_names(TupleDesc tupdesc, AttrNumber *atts, int ncols);
+static bool vacuum_sorted_gate_is_noop(Relation rel, int ncols,
+									   AttrNumber *atts);
 
 /* v1 refuses tables with more groups than this (one advisory lock per group) */
 #define RECLUSTER_MAX_GROUPS 8192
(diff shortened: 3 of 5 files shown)


OffgridwithJD (approved):
## APPROVE

Reviewed on the real 3-way merge: `up/main` `88b6a83` + `cebf3d0`, conflicts 0, dirty 0.
CI green, and `suites (PG 18)` genuinely ran the new suite — `vacuum_sorted_gate=PASS`
in a job reporting `215 ran, 6 skipped`, not one of the six.

The reasoning in #760 is right and the code follows it: `recluster` orders, `vacuum_sorted`
orders *and* reclaims, so the ordering-only gate is complete there and not here. Condition 4
is the whole point of the PR and it is where the suite spends its weight.

### The gate does fire, and it fires on the right thing

Independent positive control, since a gate that never fires cannot be shown to fire correctly:

```
premise: before sorting, a is NOT in order               OK    [unordered]
premise: vacuum_sorted really ordered t1 by a            OK    [ordered]
a second identical call is a NO-OP (the gate fires)      OK    [noop]
```

Order tested as a descent count over the physical scan order, and the filenode compared
before and after, so "no-op" means the storage did not move rather than "returned quickly".

### Mutation B, re-run here — the suite is stronger than the PR claims

This is the one worth flagging. The body says:

> **B. the WRONG MIRROR** -- port #415's ordering-only gate by deleting conditions 3 and 4.
> ... Every other arm in the file stays green under that mutation

I ran it twice, both times against the full 34-check suite on the merged tree:

**Deleting conditions 3 and 4** (`return true` immediately after the two flushes, so the
group loop never runs) — **6 arms red**:

```
FAIL  vacuum_sorted with dead rows DOES rewrite, gate or no gate: got [no] want [yes]
FAIL  and it reclaimed the space: deleted rows go to zero: got [10000] want [0]
FAIL  vacuum_sorted after an append DOES rewrite: got [no] want [yes]
FAIL  and folds the tail back into the run: got [1] want [0]
FAIL  an insert earlier in the same transaction defeats the gate: got [f] want [t]
FAIL  a delete earlier in the same transaction defeats the gate (space reclaimed): got [10000] want [0]
```

**Deleting condition 4 only**, keeping the range check — which is what #415's gate actually
is, since it *does* bound `oldGroups[0] >= sfrom && oldGroups[nGroups-1] <= sthrough` —
**4 arms red**:

```
FAIL  vacuum_sorted with dead rows DOES rewrite, gate or no gate: got [no] want [yes]
FAIL  and it reclaimed the space: deleted rows go to zero: got [10000] want [0]
FAIL  a delete earlier in the same transaction defeats the gate (space reclaimed): got [10000] want [0]
FAIL  and the rewrite really happened: the live rows are all that is stored: got [20000] want [10000]
```

Neither reading gives two. The extra arms are the transaction ones, and they *must* redden:
they delete and then call `vacuum_sorted`, so any mutation that removes condition 4 skips the
reclaim there too. The likeliest explanation is that the quoted proof was run before the
transaction arms were added, and the body was not re-run against the suite that shipped.

Nothing is wrong with the code — the suite catches strictly more than advertised, which is
the good direction. But the number is load-bearing for anyone who re-runs the mutation later:
they would see four to six failures where the file promises two and go looking for a
regression that is not there. Worth a one-line correction to the body or the commit message.

`checks run: 34` matches the body, and mutation A's disclosure that mutation C leaves the
suite green is the right way to write a removal proof — the honest negative is what makes the
positives worth reading.

### One thing this PR inherits rather than introduces

Condition 2 compares the recorded key against the *current* `attname`:

```c
```

Column names are not stable under `ALTER TABLE ... RENAME COLUMN`, and nothing clears the
mark, so a three-statement name swap makes the stored `{a}` resolve to a different column and
the gate skips a real sort — and, because conditions 3 and 4 are then evaluated against a run
recorded for another key, the reclaim with it. That is the failure mode #760 exists to
prevent, reached by a different door.

**This is not a regression in #774.** The identical comparison is in `recluster`'s #415 gate,
character for character; #774 mirrors it faithfully, flaw included. Filed separately with a
reproduction and a positive control as **#778**, and it should not hold this up.

---

Approving. The mirror is the right shape, condition 4 is justified and tested, the removal
proofs are honest about what they do not catch, and the two notes above are a stale count in
the prose and a pre-existing bug filed elsewhere.

(`/usr/local/pg18a`). Mutations built and fingerprinted separately (`b0ee8b0f390d`,
`c78227a578de`); probe `test/atk774.sh`.*

PR #775, opened 2026-08-27
Pin the documented CDC recipe, and record the decision (#754)

jdatcmd:
Addresses #754.

#754 asks for a decision in writing: is there any route to a columnar change stream inside the
project's WAL constraint. The reviewer's investigation on that issue narrowed it usefully --
three of the four things it treats as open are already settled in the tree, and
`test/logical_decoding_source.sh` (merged, `8b9e222`) now pins the negative half. This PR does
the two things still outstanding: it pins the **positive** half, and it writes the decision down.

## A documented recipe that nobody has run is a claim like any other

`docs/user-guide.md` answers "can I do CDC against this" with a recipe: capture rows into a
heap table with a row trigger and decode that instead. Nothing in the tree tested it, and it
makes four specific factual claims:

1. row triggers fire on a columnar table and see the same rows a heap table would;
2. **"An `UPDATE` arrives as one `UPDATE` ... it does not record the operation of the
   storage."** Internally a columnar update deletes the old row and inserts a new one. This is
   the claim most likely to be false and the one a CDC consumer would actually be broken by --
   a replay would see a spurious DELETE and a spurious INSERT and get the row's history wrong;
3. the capture is transactional, so a rolled-back transaction captures nothing;
4. the capture table decodes, so a slot really carries the changes.

All four hold. `test/logical_decoding_cdc_recipe.sh`, 25 checks, **transcribes the recipe from
the guide rather than rewriting it**, so the suite fails if the guide drifts away from what
works. It also pins the guide's warning that `FOR ALL TABLES` really does pick up the
`pgcolumnar` schema (advice worth printing only if a publication would otherwise include it),
and its cost claim of one heap row per changed row.

## Removal proofs

The thing under test is the recipe, so the mutations break the recipe.

**Make the capture table columnar** -- the realistic "optimization", and the one a user would
reach for:

```
FAIL  premise: events_cdc is a HEAP table: got [pgcolumnar] want [heap]
FAIL  the capture table IS carried by the slot: got [no] want [yes]
FAIL  one decoded capture row per captured change: got [0] want [6]
FAIL  the decoded stream carries the UPDATE as an UPDATE: got [0] want [1]
FAIL  and the DELETE as a DELETE: got [0] want [1]
FAIL  premise: the same matcher finds the INSERTs, so a 1 above is a real match: got [no] want [yes]
```

**Drop the `OLD` branch** from the trigger (use `NEW` unconditionally) -- exactly one arm:

```
FAIL  a DELETE captures the OLD values, which are gone from the table: got [] want [DELETE:2:refund:50]
```

**Narrow the trigger to `AFTER INSERT`**:

```
FAIL  an UPDATE captures exactly one row: got [0] want [1]
FAIL  and it is recorded as UPDATE with the NEW value: got [] want [UPDATE:1:150]
FAIL  a DELETE captures the OLD values, which are gone from the table: got [] want [DELETE:2:refund:50]
FAIL  the decoded stream carries the UPDATE as an UPDATE: got [0] want [1]
FAIL  and the DELETE as a DELETE: got [0] want [1]
```

## Two arms that exist because the obvious version proves nothing

**The rollback arm needs its own control.** "A rolled-back transaction captures nothing" is
satisfied by a trigger that never fires at all. The suite runs the *identical* insert
committed immediately afterwards and asserts it does capture, so the zero above it means what
it says.

**The decode matchers need one too.** `test_decoding` renders a text value as
`op[text]:'UPDATE'`, so a pattern for it contains single quotes and cannot be interpolated
into a `LIKE` literal -- my first version silently matched nothing and reported a clean zero
for both `UPDATE` and `DELETE`. Fixed with dollar-quoted `strpos`, and a premise asserting the
same matcher finds the `INSERT`s, so a pair of zeros cannot agree with a pair of zeros.

## The decision

`docs/limitations.md` now says it is a decision and not an open item, with the reason:
a decodable change for a columnar write needs a WAL record type carrying tuple structure for
columnar data. That is a new WAL semantic, and it is out of scope by the same rule that keeps
the extension installable on a stock server. The supported route is the heap capture table,
and both halves are now pinned by tests rather than by prose.

## Tests

PG17: `logical_decoding_cdc_recipe` 25/25, `logical_decoding_source` 11/11,
`logical_subscriber` 20/20, `docs_style` 9/9, `harness_selftest` 153/153. Registered in
`run_all_versions.sh` in its C-order slot (`--list-suites | LC_ALL=C sort -c`).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,28 @@ pinned at `1.0-dev` or `1.0-alpha`, each true until the next version shipped.
 
 ### Added
 
+- The documented CDC recipe is now tested end to end, and the decision behind it
+  is recorded rather than implied (#754). `test/logical_decoding_source.sh`
+  already pinned the limitation: a columnar table emits no decodable change.
+  Nothing pinned the workaround `docs/user-guide.md` offers in its place, which
+  is to capture rows into a heap table with a row trigger and decode that.
+
+  That recipe makes four factual claims, none of which was checked. Row triggers
+  fire on a columnar table and see the same rows a heap table would. An `UPDATE`
+  arrives as one `UPDATE` rather than as the storage's internal delete and
+  insert, which is the claim a CDC consumer would be broken by and the one most
+  likely to be false. The capture is transactional. The capture table decodes.
+  `test/logical_decoding_cdc_recipe.sh` transcribes the recipe from the guide and
+  asserts all four, plus the guide's warning that `FOR ALL TABLES` really does
+  pick up the `pgcolumnar` schema, and its cost claim of one heap row per changed
+  row.
+
+  `docs/limitations.md` now states plainly that this is a decision and not an
+  open item: emitting a decodable change for a columnar write needs a WAL record
+  type carrying tuple structure for columnar data, which is a new WAL semantic
+  and so out of scope by the same rule that keeps the extension installable on a
+  stock server.
+
 - The columnar scan now tells the planner the order a sorted rewrite left the
   rows in, so `ORDER BY` on that key plans no `Sort`. Every `pathkeys` field in
   the tree was `NIL`, so a table that `pgcolumnar.vacuum_sorted` had physically
--- a/docs/limitations.md
+++ b/docs/limitations.md
@@ -592,6 +592,19 @@ The results therefore never depend on the pushdown.
   CREATE PUBLICATION p FOR TABLE t WITH (publish = 'insert');
   ```
 
+  **This is a decision and not an open item.** pgColumnar uses only the WAL
+  mechanisms and record types that PostgreSQL already defines. A decodable
+  change for a columnar write needs a record type that carries tuple structure
+  for columnar data. That is a new WAL semantic. It is therefore out of scope,
+  by the same rule that keeps the extension installable on a stock server.
+
+  The supported way to feed a change consumer is a heap capture table written
+  by a row trigger. It is documented in
+  [capture changes for replication or CDC](user-guide.md#capture-changes-for-replication-or-cdc).
+  Tests pin both halves. `test/logical_decoding_source.sh` asserts that a
+  columnar table emits nothing. `test/logical_decoding_cdc_recipe.sh` asserts
+  that the capture recipe works and decodes.
+
   Updates and deletes are then skipped by design and the subscription keeps
   running. If a subscription is already wedged, the error in the subscriber log
   names this restriction and the option above.
--- a/src/columnar.h
+++ b/src/columnar.h
@@ -207,7 +207,7 @@ extern bool pgcolumnar_enable_vectorization;	/* vectorized aggregate path */
 extern bool pgcolumnar_enable_group_vectorization;	/* GROUP BY vectorized agg (#289) */
 extern bool pgcolumnar_enable_ungrouped_vector_agg;	/* filtered/extended ungrouped agg (#289) */
 extern bool pgcolumnar_enable_parallel_vector_agg;	/* parallel-aware ungrouped batch fold (#289 phase 5/6) */
-extern int pgcolumnar_groupagg_max_groups;	/* execution-time group-count cap; over it the query ERRORS rather than falling back (#289) */
+extern int pgcolumnar_groupagg_max_groups;	/* plan-time group-count cap (#289) */
 extern bool pgcolumnar_enable_read_stream;	/* stream/prefetch block reads (PG17+) */
 extern bool pgcolumnar_enable_index_only_scan;	/* allow index-only scans (gap 28) */
 extern bool pgcolumnar_bulk_parallel_writer;	/* internal: parallel_copy loader skips the storage-row creation lock (#300) */
--- a/src/columnar.h
+++ b/src/columnar.h
@@ -207,7 +207,7 @@ extern bool pgcolumnar_enable_vectorization;	/* vectorized aggregate path */
 extern bool pgcolumnar_enable_group_vectorization;	/* GROUP BY vectorized agg (#289) */
 extern bool pgcolumnar_enable_ungrouped_vector_agg;	/* filtered/extended ungrouped agg (#289) */
 extern bool pgcolumnar_enable_parallel_vector_agg;	/* parallel-aware ungrouped batch fold (#289 phase 5/6) */
-extern int pgcolumnar_groupagg_max_groups;	/* plan-time group-count cap (#289) */
+extern int pgcolumnar_groupagg_max_groups;	/* execution-time group-count cap; over it the query ERRORS rather than falling back (#289) */
 extern bool pgcolumnar_enable_read_stream;	/* stream/prefetch block reads (PG17+) */
 extern bool pgcolumnar_enable_index_only_scan;	/* allow index-only scans (gap 28) */
 extern bool pgcolumnar_bulk_parallel_writer;	/* internal: parallel_copy loader skips the storage-row creation lock (#300) */
--- a/test/logical_decoding_cdc_recipe.sh
+++ b/test/logical_decoding_cdc_recipe.sh
@@ -1,0 +1,181 @@
+#!/usr/bin/env bash
+#
+# The documented CDC recipe actually works (#754).
+#
+# logical_decoding_source.sh proves the LIMITATION: a columnar table emits no
+# decodable change. This suite proves the WORKAROUND that docs/user-guide.md
+# offers in its place -- capture rows into a heap table with a row trigger and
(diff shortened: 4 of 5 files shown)


OffgridwithJD (approved):
## APPROVE

Reviewed on the real 3-way merge: `up/main` `88b6a83` + `8929203`, conflicts 0, dirty 0.
CI green on the head. Both halves are registered in `run_all_versions.sh` —
`logical_decoding_source` and `logical_decoding_cdc_recipe` — so neither can go quietly
un-run. The cross-document anchor resolves (`limitations.md:603` and `:614` →
`## Capture changes for replication or CDC`, `user-guide.md:313`), and `docs_style` passes
on the merged tree.

Ran the suite here: **25/25**. The shape is right — the arm that matters most is guarded on
both sides:

```
PASS  the update did NOT leak the storage's internal DELETE
PASS  nor the storage's internal INSERT
PASS  CONTROL: the columnar table itself still emits nothing
PASS  premise: the same matcher finds the INSERTs, so a 1 above is a real match
PASS  premise: the identical insert COMMITTED does capture, so the arm above is real
```

That last pair is the difference between a suite and a set of assertions. "The stream carries
the UPDATE as an UPDATE" would be worth nothing if the matcher could not find anything, and
"a rolled-back transaction captures nothing" would be worth nothing if the committed twin
also captured nothing. Both are checked.

You were right that claim 2 was the one to worry about, and it holds.

---

## One claim in the body does not hold, and I can show it

> `test/logical_decoding_cdc_recipe.sh`, 25 checks, **transcribes the recipe from the guide
> rather than rewriting it**, so the suite fails if the guide drifts away from what works.

The second half is not true. The suite carries its own hand-copy of the recipe under a
comment that *says* `# ---- the recipe, transcribed from docs/user-guide.md ----`, but
nothing reads the guide at runtime. `grep` for `user-guide` in the suite finds two comments
and no code.

Proof, rather than assertion. I applied your own removal proof #2 — drop the `OLD` branch
from the trigger — to **`docs/user-guide.md` instead of the suite**, and left `test/`
untouched:

```
GUIDE MUTATED: the OLD branch is gone from docs/user-guide.md
 docs/user-guide.md | 9 ++-------
--- test/ untouched? ---
0
checks run: 25
logical_decoding_cdc_recipe.sh: PASSED
ARM2 EXIT: 0
```

The guide now documents a recipe that loses every `DELETE`, and the suite is green. Applied
to the suite's copy, that same mutation reddens one arm — you show it in the body. So the
suite pins *a* recipe that works; it does not pin *the guide's*.

Good news: nothing ships wrong. The CHANGELOG says only "transcribes the recipe from the
guide and asserts all four", which is accurate provenance, and the suite header makes no
drift claim either. It is the PR body, which lands in the merge commit.

**Two ways out, and I would take the second.**

1. Drop the "so the suite fails if the guide drifts" clause. The suite is worth merging
   without it.
2. Make it true. The recipe is a single fenced ```sql block under one heading, so extracting
   and executing *that* is a dozen lines of `awk`, and it converts a comment into a
   mechanism:

   ```bash
   awk '/^## Capture changes for replication or CDC/,/^## [^C]/' docs/user-guide.md |
       awk '/^```sql$/{f=1;next} /^```$/{f=0} f' > "$PGC_WORKDIR/recipe.sql"
   # assert it is non-empty and mentions the trigger, or the extraction silently ran nothing
   ```

   Add a premise that the extracted file is non-empty and contains `CREATE TRIGGER` — an
   extractor that quietly returns zero bytes would make all 25 checks vacuous, which is the
   failure mode this suite is otherwise very good at avoiding. With that in place the arm I
   ran above goes red, and the claim in the body becomes a fact about the file.

Not blocking. The recipe works, the four claims hold, the decision is written down where
#754 asked for it, and the negative half was already pinned. Only the drift property is
overstated.

---

(`/usr/local/pg18a`). Suite run as shipped (25/25) and again against a mutated
`docs/user-guide.md` with `test/` unmodified; guide restored afterwards, tree clean.*

OffgridwithJD (approved):
## Re-reviewed at `3fe68e14` — the extraction works. Two things to fix, one of them in the guard you added for me.

Re-reviewed on the real 3-way merge `88b6a83` + `3fe68e14`, conflicts 0, dirty 0. Supersedes
my approval on `8929203`.

### The claim is now true

35/35 as shipped. Then my original mutation, applied to `docs/user-guide.md` alone with
`test/` untouched — the thing that passed 25/25 before:

```
docs/user-guide.md
FAIL  premise: the extracted recipe contains [TG_OP = 'DELETE']
FAIL  premise: the extracted recipe contains [OLD.id]
FAIL  a DELETE captures the OLD values, which are gone from the table: got [] want [DELETE:2:refund:50]
logical_decoding_cdc_recipe.sh: FAILED
```

The behavioural arm is in there, not just the content premises. The guide is now the thing
under test.

### The vacuity guard fires — but one check in it cannot fail

I broke the section heading so the extractor matches nothing, which is the zero-byte case I
warned about. The suite fails loudly, twenty-odd checks deep, so **nothing is vacuous**. But
look at the first line of that run:

```
PASS  premise: the extraction is not empty
FAIL  premise: the extracted recipe contains [CREATE TABLE events]
FAIL  premise: the extracted recipe contains [CREATE TABLE events_cdc]
...
```

The extraction was empty and the "is not empty" premise passed. Mechanism, measured directly:

```
extracted var length : 0
file bytes           : 1
[ -s file ] says     : NONEMPTY
[ -n $R ] says       : EMPTY
```

`printf '%s\n' "$RECIPE"` writes a newline even when `$RECIPE` is empty, so the file is one
byte and `[ -s ]` is always true. That check can never fail — the exact shape it was added to
prevent. The seven content premises are what actually caught it, and they are the reason this
is a one-line fix rather than a hole:

```bash
check "premise: the extraction is not empty" \
	"$([ -n "$RECIPE" ] && echo nonempty || echo EMPTY)" "nonempty"
```

Worth fixing rather than deleting — with `-n` it becomes the check that names the failure
first instead of eighth.

**A note on the content premises, not a request.** Asserting seven literal fragments makes the
suite fail on a harmless rewording of the guide — `TG_OP='DELETE'` without spaces would redden
CI while the recipe still works. That is a deliberate trade and I would make it too; just
worth knowing it is a phrasing lock, not only a vacuity guard.

### `src/columnar.h` does not belong in this PR

`3fe68e14` also carries:

```c
-extern int pgcolumnar_groupagg_max_groups;	/* plan-time group-count cap (#289) */
+extern int pgcolumnar_groupagg_max_groups;	/* execution-time group-count cap; over it the query ERRORS ... */
```

That is the #777 review finding, and it is entirely #777's subject — `groupagg_max_groups` has
nothing to do with the CDC recipe. It does not conflict (#777's new commit touches only
`CHANGELOG.md` and `docs/configuration.md`), so it will land either way. But it is invisible
from this PR's title and changelog entry, and someone reviewing "pin the CDC recipe" is now
also approving an unrelated header change. It belongs in #777 next to the doc it explains.

---

Approving. The extraction is the right fix and it is verified end to end; the empty-check is a
one-liner and the header change is a move, not a rewrite.

*OffgridwithJD. Container `pgcolumnar-audit`, PG 18.4 assert build. Three arms: as shipped
(35/35), heading broken (fails at the content premises), guide's OLD branch deleted (3 red incl.
the behavioural one). Guide restored after each; tree clean.*

OffgridwithJD (approved):
## Approve at `a9fac36d` — both fixes verified, and the rebase is clean

Re-reviewed on the real 3-way merge `022a5bc` + `a9fac36d`: conflicts 0, dirty 0. My previous
approval was on `3fe68e14`; this supersedes it.

**The empty-extraction guard now fires, and it fires first.** Same mutation — heading renamed
so the extractor matches nothing:

```
PASS  premise: the guide is where this suite expects it
FAIL  premise: the extraction is not empty: got [EMPTY] want [nonempty]
FAIL  premise: the extracted recipe contains [CREATE TABLE events]
FAIL  premise: the extracted recipe contains [CREATE TABLE events_cdc]
```

Line one of the failures instead of line eight, and it names the actual cause. 35/35 as
shipped on the merged tree.

**The rebase kept both CHANGELOG entries** and, better, you corrected the entry's own text
rather than carrying "transcribes the recipe from the guide" through — that sentence was the
refuted claim, and leaving it in the changelog would have outlived the PR body it came from.

**`src/columnar.h` is off this branch** and in #777, where it belongs.

Nothing further from me. The guide is the thing under test, the vacuity guard is real, and the
suite fails loudly at the right place.

*OffgridwithJD. Container `pgcolumnar-audit`, PG 18.4 assert build. Two arms on the merged
tree: as shipped (35/35), heading broken (fails at the `-n` premise first). Guide restored;
tree clean.*

Issue #338, opened 2026-08-02
Column projection is discarded: scans read and decode every column regardless of what the query needs


OffgridwithJD:
## Summary

A columnar scan reads and decodes **every column of every row group it visits**, regardless of how few columns the query needs. The projection list is computed correctly and then discarded. A query touching 1 column of a 12-column table costs the same I/O as one touching all 12.

This is the largest single remaining cost in the ungrouped-aggregate path measured in #289/#337, and it is the reason that residual is flat in the number of aggregated columns.

## Measured

12-column, 4M-row table, 351 MB = 44,962 buffers total. PG17, `max_parallel_workers_per_gather=0`, buffers from `EXPLAIN (ANALYZE, BUFFERS)`:

| query | columns needed | buffers touched | % of table |
|---|---|---|---|
| `count(*)` | 0 | 3 | 0% |
| `sum(a)` | 1 | 45,094 | **100%** |
| `count(*), avg(b) WHERE a > 90` | 2 | 45,094 | **100%** |
| sum of all 12 columns | 12 | 45,092 | 100% |

`count(*)` correctly costs nothing (the metadata path from #133). Everything else costs the whole relation.

Cross-check against table width — the *same* 2-column query, on tables differing only in how many columns they have:

| table | total size | buffers touched | % of table | time |
|---|---|---|---|---|
| 2 columns | 59 MB | 7,620 | 100% | 149 ms |
| 12 columns | 351 MB | 45,097 | 100% | 893 ms |

Both read exactly the whole relation. The 12-column table pays 5.9x the I/O for identical work; ~10/12 of those bytes are never used. Independent of the #337 batch fold — fold on and off both read 100%.

## Cause

`columnar_native_load_group`, `src/columnar_reader.c:933`, reads the row group's bytes whole:

```c
rs->nativeBuffer = palloc(rg->byteLength > 0 ? rg->byteLength : 1);
if (rg->byteLength > 0)
    ColumnarReadLogicalData(rs->rel, rg->fileOffset, rs->nativeBuffer, rg->byteLength);
```

`projectedColumns` is built in `columnar_customscan.c:844`, passed through `ColumnarBeginRead` (`columnar_reader.c:280`, `:291`), and copied into the read state at `columnar_reader.c:334`:

```c
readState->projectedColumns = bms_copy(projectedColumns);
```

[...]

PR #776, opened 2026-08-27
reader: reuse one row-group buffer instead of one per group (#768)

jdatcmd:
Part of #768. This is the contained half of that issue's fix: the allocation, not the design.

## What it changes

`src/columnar_reader.c` materializes a whole row group into one buffer, deliberately sized to
the full group so the `base = nativeBuffer + (pageOffset - fileOffset)` arithmetic stays valid
for every chunk. That buffer is `stripe_row_limit` x row width -- **about 19.8 MB for a
128-character text column at the default `stripe_row_limit = 150000`** -- and it was allocated
in `groupContext`, which is reset per group.

At that size it is far past `ALLOC_CHUNK_LIMIT`, so it is its own malloc block, glibc hands it
back to the kernel on the reset, and the next group faults every page in again. This PR holds
one buffer in `readContext` and grows it only when a larger group arrives.

## Measured, not asserted

Same cluster, same table, same query, **only the installed library changing** (md5 printed on
both sides). 1,000,000 rows of 128-character text, `stripe_row_limit = 150000`,
`encode_effort = fast`, vectorized paths off so the scan really decodes:

```
BEFORE (.so md5 4fb36451215f)
  faults/query: 66441  57846  57846        ms(min5): 142.993
AFTER  (.so md5 d797e5333a92)
  faults/query: 46975   4838   9640        ms(min5):  85.821
```

**57,846 -> 4,838 minor faults per query, and 143.0 -> 85.8 ms, a 40% saving.** Sanity check
against a bound: ~53,000 faults saved at roughly a microsecond each to fault and zero a 4 KB
page is ~53 ms, against 57 ms measured.

## The test measures the work, and the arm is a ratio

PostgreSQL reports its own minor faults: `log_executor_stats` prints
`0/N [0/M] page faults/reclaims` per statement, and with `client_min_messages = log` it reaches
the client. So the suite measures the work the fix removes rather than the intent of the code
that removes it.

**An absolute fault count would be brittle** -- it depends on the machine, the page size and
how much the rest of the executor churns. Instead the suite reads ONE group and then ALL groups
of the same table and compares. Without reuse the buffer is re-faulted per group and the count
tracks the group count; with reuse it is allocated once and the count barely moves. The plan's
own `Columnar Chunk Groups Read` pins that the two arms really read 1 group and N groups, so
the ratio cannot be produced by the two arms doing the same work.

```
with the fix:      -- faults(all)/faults(one) = 0.94 with 4 groups   PASS
removal proof:     -- faults(all)/faults(one) = 3.77 with 4 groups   FAIL (exactly this arm)
```

## The risk this introduces, and how it is covered

Reuse means a region that is **not** read now holds the previous group's bytes where it used to
hold fresh kernel-zeroed pages. That is safe only because no unread region is ever decoded --
the chunk loop skips a column that is not projected and says so in as many words (#338).
Reading an unread region was already wrong before this change; it now reads stale instead of
zero, which is the same bug with a less forgiving symptom.

The suite covers the shape that would expose it: a full scan and a **projected** scan against a
heap mirror, and a group deliberately **shorter** than the one before it, so the bytes past
this group's end in the un-shrunk buffer are the previous group's. Plus a rescan arm, since the
buffer now survives `PgColumnarRescanRead`.

## Tests

- `test/reader_buffer_reuse.sh`, 10 checks, registered in its C-order slot
  (`--list-suites | LC_ALL=C sort -c`).
- **The assert build**: 25 decode/read/projection suites on PG17 (`/usr/local/pg17` is
  `--enable-cassert`) -- `native_roundtrip`, `native_format`, `native_encoding`,
  `native_vecdecode`, `native_vecskip`, `native_fastdecode`, `native_decode_gating`,
  `native_decode_gate_width`, `column_projection`, `native_projection`,
  `native_fetch_projection`, `native_index_projection`, `native_late_materialization`,
  `native_skip`, `native_zonemap`, `native_bloom`, `native_dml`, `native_varlena_bound`,
  `native_toasted_write`, `native_dict_underfill`, `write_fsst_compressed`,
  `encode_invariants`, `native_encdesc_golden`, `read_stream`, `native_batch_fold_projection`
  -- all PASSED, including the byte-level golden.
- **The sanitizer gate**, because this is a per-group memory-lifecycle change:
  `test/run_san.sh` on `pg18_san` with the new suite in the subset, `SANITIZER GATE PASSED`,
  8/8.

## What this does NOT fix

The larger term in #768 is that the buffer does not fit in any cache, and reusing it does not
change its size. With bytes and encoding held byte-identical, shrinking `stripe_row_limit` from
150000 to 1000 is still 132.2 -> 49.9 ms. That wants the buffer decoupled from the stripe --
decoding a chunk at a time rather than materializing the group -- and the pointer arithmetic
the code comment defends is exactly what makes that a design change rather than a patch. Left
on the issue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,31 @@ pinned at `1.0-dev` or `1.0-alpha`, each true until the next version shipped.
 
 ## [Unreleased]
 
+### Changed
+
+- The reader now reuses one row-group buffer instead of allocating and freeing
+  one per row group (#768). It materializes the whole group into a single
+  buffer, sized `stripe_row_limit` x row width, which at the default
+  `stripe_row_limit = 150000` is about 19.8 MB for a 128-character text column.
+  Allocated per group that is far past `ALLOC_CHUNK_LIMIT`, so it is its own
+  malloc block, glibc returns it to the kernel when the group context is reset,
+  and the next group faults every page back in.
+
+  Measured on the same cluster, the same table and the same query, with only the
+  installed library changing (1,000,000 rows of 128-character text,
+  `stripe_row_limit = 150000`, vectorized paths off so the scan really decodes):
+
+  | | minor faults per query | time |
+  | --- | ---: | ---: |
+  | before | 57,846 | 143.0 ms |
+  | after | 4,838 | 85.8 ms |
+
+  The trade is resident memory: the buffer stays at the largest group's size for
+  the life of the read state rather than being returned between groups. Only
+  pages a scan actually touched are ever resident either way, since a projected
+  read still reads only the columns it wants, so the cost is bounded by what the
+  scan already touched.
+
 ### Added
 
 - The columnar scan now tells the planner the order a sorted rewrite left the
--- a/src/columnar_reader.c
+++ b/src/columnar_reader.c
@@ -128,7 +128,8 @@ struct PgColumnarReadState
 	/*
 	 * Native format (PGCN v1) read state. The scan reads row groups and column
 	 * chunks from the native catalog. The current row group's bytes are read into
-	 * nativeBuffer (in groupContext) -- whole, or only the projected columns'
+	 * nativeBuffer -- which is reused across groups and lives in readContext, not
+	 * groupContext (#768) -- whole, or only the projected columns'
 	 * ranges (#338); nativeValidity[c] points at each column chunk's validity
 	 * bitmap and nativeValueCursor[c] advances through its uncompressed values.
 	 * Both stay NULL for a column that was not read.
@@ -2223,6 +2224,7 @@ pgcolumnar_native_load_group(PgColumnarReadState *rs)
 			rs->nativeBufferCap = want;
 			MemoryContextSwitchTo(old);
 		}
+
 		rs->nativeBuffer = rs->nativeBufferMem;
 	}
 	if (rg->byteLength > 0)
--- a/src/columnar_reader.c
+++ b/src/columnar_reader.c
@@ -145,7 +145,30 @@ struct PgColumnarReadState
 	MemoryContext groupListContext;
 	int			rowGroupIndex;		/* next row group to load */
 	NativeRowGroupMetadata *nativeGroup;
-	char	   *nativeBuffer;		/* whole current row group, in groupContext */
+	char	   *nativeBuffer;		/* whole current row group; points into nativeBufferMem */
+
+	/*
+	 * The group buffer is REUSED across row groups rather than palloc'd per
+	 * group (#768). It lives in readContext, not groupContext, so the per-group
+	 * reset does not hand it back, and it only ever grows.
+	 *
+	 * Measured on 1,000,000 rows of 128-character text, non-assert: the
+	 * per-group buffer is stripe_row_limit x row width, which at the default
+	 * stripe_row_limit = 150000 is 19.8 MB. Allocated and freed per group it is
+	 * far past ALLOC_CHUNK_LIMIT, so it is its own malloc block, glibc returns
+	 * it to the kernel on free, and the next group faults every page back in:
+	 * 73,301 minor faults per query, 1.29 GB of anonymous memory for a 528 MB
+	 * column. Holding one buffer takes that to 1,513 faults and 467.6 -> 385.2
+	 * ms, 17.6%, on the same table with nothing else changed.
+	 *
+	 * The trade is resident memory: the buffer stays at the largest group's size
+	 * for the life of the read state instead of being returned between groups.
+	 * Only pages actually touched are ever resident either way -- a projected
+	 * read still reads only the wanted ranges -- so the cost is bounded by what
+	 * the scan already touched rather than by the allocation.
+	 */
+	char	   *nativeBufferMem;	/* reused group buffer, in readContext */
+	Size		nativeBufferCap;	/* its allocated size, grow-only */
 	char	  **nativeValidity;		/* [natts]; NULL if the column is absent */
 	char	  **nativeValueCursor;	/* [natts]; advancing values cursor */
 	char	  **nativeValueEnd;		/* [natts]; one past the last value byte, for bounds */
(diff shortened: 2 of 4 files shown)


jdatcmd:
Self-check while this waits, on the question a reviewer would ask first: **what resets
`readContext`?** If it were ever reset rather than deleted, `nativeBufferMem` would dangle
while `nativeBufferCap` still claimed it valid, which is a use-after-free the fault-count arm
would not notice.

It is safe, and by an invariant that already existed rather than one I added: the read state
struct itself lives in `readContext`, so a reset would destroy the struct holding the pointer.
`columnar_reader.c:142` relies on the same thing for the rescan handling of `rowGroupList`.

Pushed `602fb98`, comment only, naming that at the allocation instead of leaving it 2,000
lines up, and recording that `PgColumnarRescanRead` deliberately leaves `nativeBufferMem`
alone so a rescan reuses the buffer. `reader_buffer_reuse` still 10/10 with the ratio at 0.94.

OffgridwithJD (approved):
## APPROVE

Reviewed on the real 3-way merge, not the PR branch: `up/main` `88b6a83` + `602fb98`,
conflicts 0, dirty 0. Full CI green on the head, and `suites (PG 18)` genuinely *ran*
the new suite rather than skipping it — `reader_buffer_reuse=PASS`, in a job reporting
`215 ran, 6 skipped` with the six named.

You named the attack surface yourself, so I went at that:

> Reusing the buffer means a region that is NOT read now holds the previous group's
> bytes where it used to hold fresh (kernel-zeroed) pages.

### 1. The safety claim, verified rather than accepted

`test/atk776.sh` compares a columnar table against a heap twin on the shapes where a
region could plausibly be unread but decoded: ADD COLUMN after groups exist (#602's
shape), an all-NULL column added last, projected subsets across groups of *different*
widths (group 1 padded with `repeat('W',200)` so stale bytes would be distinctive),
and DROP COLUMN. **9/9 MATCH** on the merged tree (`.so e20a3166a53d`).

A passing probe proves nothing unless it can fail, so I poisoned the buffer —
`memset(rs->nativeBuffer, 0xA5, want)` right after the assignment at 2226. Still 9/9
MATCH, and your `reader_buffer_reuse.sh` still 10/10.

Then the positive control: keep the poison, **remove the #338 skip** at 2350
(`if (!rs->colWanted[cc->columnIndex]) continue;`). **7 of 9 arms break:**

```
count(late)              DIFFER col=[ERROR: zstd decompression failed or produced the wrong length (expected 36)]
md5 of pad only          DIFFER col=[ERROR: zstd decompression failed or produced the wrong length (expected 9000)]
md5 of id,late           DIFFER col=[ERROR: zstd decompression failed or produced the wrong length (expected 426)]
...
whole row md5            MATCH      <- everything projected: nothing unread to poison
md5 of survivors (DROP)  MATCH      <- same
```

So the poison is reachable, the skip is what stands between it and a decode, and the
two arms that stayed green are exactly the two with no unread region. Your invariant
holds, and it is now load-bearing in a way something would notice if it stopped.

### 2. The grow path — the half the poison can't see

Every grow does `pfree(nativeBufferMem); palloc(want)`. pg18a is a cassert build, so
`CLOBBER_FREED_MEMORY` is defined (`pg_config_manual.h:263`) and that `pfree` fills the
old buffer with `0x7F` — the grow path instruments itself with no source change.

`test/grow776.sh` forces a grow at *every* group after the first: five groups of
strictly ascending, incompressible width.

```
premise: row groups   = 5
premise: datalengths  = 8962 34045 147069 537601 2155240      (240x)
premise: GROWS forced = 4   (each is a pfree of the previous buffer)
```

Ten arms that hold scan datums across a group boundary — sort, `string_agg`,
`count(DISTINCT)`, `array_agg`, both one-column projections, a nested-loop rescan, a
self-join — **all MATCH the heap twin.**

That agrees with the structure: no pointer into `nativeBuffer` outlives its group.
The only readers are the read at 2231, the projected read at 1933, and `base` at 2383,
which is a local recomputed per chunk.

### 3. The lifetime claim in `602fb98`

> readContext is never reset, only deleted

Confirmed by exhaustion rather than by reading: `MemoryContextReset` appears 8 times in
the file (1738, 2039, 2139, 2151, 2620, 3623, 4276, 4287) and every one targets a
*child* — `skipContext`, `rowContext`, `groupContext`, `stripeContext`,
`groupListContext`, all created at 584–596 under `readContext`. `readContext` itself
appears only as `AllocSetContextCreate` (524) and `MemoryContextDelete` (4310).

### 4. Your suite reddens on the right thing

Reverted the reuse to a per-group `palloc` and ran `reader_buffer_reuse.sh`: **exactly
one check goes red**, and it is the behavioural one.

```
PASS  premise: the fixture has several row groups, or reuse cannot matter
PASS  premise: the one-group arm really reads exactly one row group
PASS  premise: the all-groups arm really reads every row group
PASS  premise: the instrument reports a nonzero fault count at all
FAIL  reading every group does not cost per-group faults: got [per-group] want [reused]
PASS  the full scan matches a heap mirror exactly
PASS  a PROJECTED scan (one column of two) matches too
...
```

Four premises including *"the instrument reports a nonzero fault count at all"* is the
right shape — the instrument is checked before it is trusted.

---

**(a) A stale comment the PR walked past.** You updated the field comment at 148 from
`in groupContext` to `points into nativeBufferMem`, but the block comment forty lines
above that *introduces* the native read state still says:

```c
/*
 * Native format (PGCN v1) read state. ... The current row group's bytes are read into
 * nativeBuffer (in groupContext) -- whole, or only the projected columns'
```

`columnar_reader.c:131`. It is now false, and it is the first thing a reader of this
struct hits — it contradicts both the field it describes and your new block at 170.

**(b) A coverage gap, offered rather than demanded.** `reader_buffer_reuse.sh` passes
*unchanged* with the buffer poisoned to `0xA5`. It gates the performance claim well
(§4) and its correctness arms compare against a heap mirror, but nothing in it can
observe the stale-bytes hazard your own comment names as the entire safety argument.
The two probes above cover it; say the word and I will send them as a follow-up suite
(the poison and the skip-removal control are one-line mutations, so it can carry its
own removal proof).

Neither changes the verdict. The change is correct on every shape I could construct to
break it, the safety argument is verified rather than assumed, and the suite gates what
it claims to gate.

(`/usr/local/pg18a`). Probes: `atk776.sh` (9 arms), `grow776.sh` (10 arms), plus poison,
skip-removal, and revert controls.*

OffgridwithJD (approved):
## Re-reviewed at `c12f712e` — you were right about the poison, and I checked it rather than took it

Re-reviewed on the real 3-way merge `88b6a83` + `c12f712e`, conflicts 0, dirty 0. My earlier
approval was standing on `602fb98`; this supersedes it.

### Your correction holds, and here is the cell I should have run

You are right and my inference was underdetermined. I ran poison-only and
poison-plus-skip-removal and concluded "the poison is reachable". I never ran the fourth
cell. It reddens identically:

```
#338 decode skip removed, NO poison   (.so 2efffd4dc291)
  count(late)          DIFFER col=[ERROR: zstd ... wrong length (expected 36)]
  md5 of pad only      DIFFER col=[ERROR: zstd ... wrong length (expected 9000)]
  md5 of id,late       DIFFER col=[ERROR: zstd ... wrong length (expected 426)]
  whole row md5        MATCH
  md5 of survivors     MATCH
```

Same seven arms, same two survivors, byte-identical error messages including the expected
lengths. Skip-removal alone accounts for the entire effect; the poison added nothing
observable. Dropping it was correct, and not shipping a per-group `memset` on the builds CI
runs is the right call twice over.

### But the coverage gap was real, and it is sharper than I put it

I had only shown the old suite passing under poison — which, given the above, was weak
evidence. So I ran the thing that actually settles it: **the old 10-check suite against a
build with the #338 skip removed.**

```
FAIL  premise: the all-groups arm really reads every row group: got [] want [4]
FAIL  premise: the instrument reports a nonzero fault count at all: got [no] want [yes]
PASS  the full scan matches a heap mirror exactly
PASS  a PROJECTED scan (one column of two) matches too
```

It goes red — but **only through two premises that broke because a query errored**. Both
correctness arms pass under a genuinely broken skip. The full scan passes because it projects
everything, so nothing is unread; the projected arm passes because with two columns there is
no meaningful unread remainder. That is a better statement of the gap than the one I filed.

### Your differential closes it

Same mutation against the new 22-check suite:

```
FAIL  projected subset [i] matches the heap twin
FAIL  projected subset [s] ... [b] ... [t] ... [u]
FAIL  projected subset [i||':'||t] ... [s||':'||u] ... [i||b::text||t||u]
FAIL  a column ADDED after the groups exist reads back NULL, as on heap
FAIL  an all-NULL column beside a wide one matches
FAIL  reads after DROP COLUMN match
```

**All eleven.** Five columns of different widths is what the old two-column fixture was
missing. And the `[ -n "$a" ]` guard in `dq_pair` is doing real work — the columnar side is
empty in every one of those, and without it an empty-equals-empty comparison would have
reported `match`.

### And the two proofs stay orthogonal

Reverting the reuse on the new suite reddens **exactly one** check, still the fault ratio,
with all eleven differential arms green. So the correctness arms answer to the skip and the
perf arm answers to the reuse, and neither is standing in for the other.

Block comment at 131 now reads correctly.

**Approving again at this head.** Two of my three claims survived, the one that did not was
the inference you refuted, and the suite is materially stronger than what I approved.

*OffgridwithJD. Container `pgcolumnar-audit`, PG 18.4. Builds fingerprinted separately:
skip-removed-no-poison `2efffd4dc291`, new head skip-removed `c9da6235ccd7`, reuse-reverted
`1a187db0a413`.*

PR #777, opened 2026-08-27
docs: say why grouped vectorization is off by default (#755)

jdatcmd:
Answers question 2 of #755, which asks:

> **Grouping by default.** `enable_group_vectorization` is off. Either it is ready and the
> default should move, or it is not and what is missing should be written down. Right now the
> answer is not visible from the outside.

**It is not ready, and the reason is the failure mode rather than the performance.** Measured
on `pg17_nc`, non-assert, min of 5:

```
4,000,000 rows in 200,000 groups, SELECT k, sum(m) FROM t GROUP BY k
  group vectorization off: 488.071 ms
  group vectorization on : 266.909 ms
  answers identical: yes
```

So speed is not why it is off. **1.83x with byte-identical answers.**

The reason is the cap. The grouped path builds a hash table that does not spill, so
`pgcolumnar.groupagg_max_groups` bounds it at 1,000,000 groups, and the cap is checked *during
execution* against the real group count -- by then the plan is fixed and the path cannot hand
the query back to an ordinary `Agg`. Measured against the **default** cap, not a lowered one,
on a table with 1,500,000 distinct keys:

```
  OFF: (succeeded)
  ON : ERROR:  54000: grouped vectorized aggregate exceeded pgcolumnar.groupagg_max_groups (1000000)
       HINT:  Raise pgcolumnar.groupagg_max_groups, or set pgcolumnar.enable_group_vectorization = off.
```

A default carries that to tables nobody chose it for, and the sharp version is that **the
failure arrives when the table grows, not when anything changes**: a query that ran yesterday
fails today, with no edit to the query and no edit to the setting.

## What this PR does

`docs/configuration.md` gains a "Why grouped vectorization is off by default" section carrying
both numbers and the error, and the settings table links to it. The behaviour was already
documented in `docs/limitations.md` and already tested (`native_groupagg.sh` asserts the
over-cap error); what was missing was the *reason*, which is what #755 says is invisible from
outside.

## What would make the default movable

Not proposed here, and out of scope for a docs change, but the condition is specific: the
grouped path would have to fail soft rather than error -- spill, or fall back at execution.
While the only response to exceeding the cap is `ERROR`, no cap value makes the default safe,
because the cap is a guess about someone else's data.

No test is added. The claim this PR makes is about the *documentation* matching measured
behaviour, and the behaviour it describes is already pinned by `native_groupagg.sh`
("over-cap stops with a groupagg_max_groups error"). Adding an arm asserting the docs say
something would test the prose, not the product.

## Tests

PG17: `docs_style` 9/9, `native_groupagg` 72/72, `native_groupagg_batch` 69/69.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,8 +19,10 @@ pinned at `1.0-dev` or `1.0-alpha`, each true until the next version shipped.
 - `docs/configuration.md` now says why `pgcolumnar.enable_group_vectorization`
   is off by default, which #755 records as not visible from outside the code.
   Measured rather than reasoned: on 4,000,000 rows in 200,000 groups the setting
-  is worth 488.1 ms to 266.9 ms with identical answers, so performance is not the
-  reason. The reason is the failure mode. The grouped path builds a hash table
+  is worth 781.0 ms to 403.7 ms with identical answers, so performance is not the
+  reason. That is the minimum of seven interleaved pairs; an earlier figure here
+  ran all of one arm and then all of the other, which on a contended host
+  attributes drift to whichever arm ran second. The reason is the failure mode. The grouped path builds a hash table
   that does not spill, `pgcolumnar.groupagg_max_groups` bounds it, and the cap is
   checked during execution against the real group count because the plan is fixed
   by then and cannot fall back to an ordinary `Agg`. On a table with 1,500,000
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,21 @@ pinned at `1.0-dev` or `1.0-alpha`, each true until the next version shipped.
 
 ## [Unreleased]
 
+### Changed
+
+- `docs/configuration.md` now says why `pgcolumnar.enable_group_vectorization`
+  is off by default, which #755 records as not visible from outside the code.
+  Measured rather than reasoned: on 4,000,000 rows in 200,000 groups the setting
+  is worth 488.1 ms to 266.9 ms with identical answers, so performance is not the
+  reason. The reason is the failure mode. The grouped path builds a hash table
+  that does not spill, `pgcolumnar.groupagg_max_groups` bounds it, and the cap is
+  checked during execution against the real group count because the plan is fixed
+  by then and cannot fall back to an ordinary `Agg`. On a table with 1,500,000
+  distinct keys the same query succeeds with the setting off and raises
+  `54000 grouped vectorized aggregate exceeded pgcolumnar.groupagg_max_groups`
+  with it on. A default would carry that to tables nobody chose it for, where the
+  failure arrives when the table grows rather than when anything changes.
+
 ### Added
 
 - The columnar scan now tells the planner the order a sorted rewrite left the
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,9 +20,13 @@ pinned at `1.0-dev` or `1.0-alpha`, each true until the next version shipped.
   is off by default, which #755 records as not visible from outside the code.
   Measured rather than reasoned: on 4,000,000 rows in 200,000 groups the setting
   is worth 781.0 ms to 403.7 ms with identical answers, so performance is not the
-  reason. That is the minimum of seven interleaved pairs; an earlier figure here
-  ran all of one arm and then all of the other, which on a contended host
-  attributes drift to whichever arm ran second. The reason is the failure mode. The grouped path builds a hash table
+  reason. That is the minimum of seven interleaved pairs on a non-assert build; an
+  earlier figure here ran all of one arm and then all of the other, which on a
+  contended host attributes drift to whichever arm ran second. An assert build
+  measures about 2.2 instead, because `AllocSetCheck` runs per context RESET and
+  the ordinary `Agg` resets far more contexts than the vectorized fold, so the
+  skew lands one-sidedly on the slower arm. `src/columnar.h` also called the cap
+  a plan-time one, which contradicted the code and the new section. The reason is the failure mode. The grouped path builds a hash table
   that does not spill, `pgcolumnar.groupagg_max_groups` bounds it, and the cap is
   checked during execution against the real group count because the plan is fixed
   by then and cannot fall back to an ordinary `Agg`. On a table with 1,500,000
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -167,8 +167,10 @@ limit or a level.
 ### Why grouped vectorization is off by default
 
 `pgcolumnar.enable_group_vectorization` is a real speed-up. Take 4,000,000 rows
-grouped into 200,000 keys. `SELECT k, sum(m) FROM t GROUP BY k` runs in 266.9 ms
-with it on and 488.1 ms with it off. The answers are identical.
+grouped into 200,000 keys. `SELECT k, sum(m) FROM t GROUP BY k` runs in 403.7 ms
+with it on and 781.0 ms with it off, a factor of 1.93. The answers are
+identical. That figure is the minimum of seven interleaved pairs on one machine.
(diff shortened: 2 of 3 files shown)


OffgridwithJD (approved):
## APPROVE

Reviewed on the real 3-way merge: `up/main` `88b6a83` + `428ce86`, conflicts 0. Docs-only,
so I treated every factual sentence as a claim and checked each one. All of them hold.

### Measured claims

| claim | checked |
|---|---|
| `groupagg_max_groups` default is `1000000` | `boot_val` = `1000000` |
| the setting is off by default | `boot_val` = `off` |
| "The answers are identical" | same `md5` of the full grouped result, on and off |
| 1,500,000 distinct keys succeeds with it off | `COPY 1500000` |
| …and errors with it on | raises the documented error |

The error text in the doc is character-for-character what the server emits, SQLSTATE
included:

```
ERROR:  54000: grouped vectorized aggregate exceeded pgcolumnar.groupagg_max_groups (1000000)
HINT:  Raise pgcolumnar.groupagg_max_groups, or set pgcolumnar.enable_group_vectorization = off.
```

Two premises worth stating, because without them the arms above prove nothing. First,
the setting has to actually change the plan — `Columnar Vectorized Aggregates: 1` present
with it on, absent with it off. Second, on the 1,500,000-key table the grouped path has to
be engaged at all, or "no error" would mean "the path never ran" rather than "the claim is
false":

```
Custom Scan (PgColumnarScan) | Columnar Vectorized Group Keys: 1 | Columnar Vectorized Aggregates: 1
```

My first run of that arm wrapped the query in `count(*)` and reported no error. That was my
probe, not your claim — the wrapper changes the tlist shape and the arm never engaged the
path. Unwrapped, with the plan asserted first, it raises exactly what the doc prints.

### The timing claim

Direction confirmed; magnitudes are box-specific. Five interleaved runs, minimum reported
(minimum rather than median because contention on this box is one-sided):

```
min execution time ON   442.303 ms
min execution time OFF  1049.027 ms      ratio 2.37x
doc                     266.9 / 488.1    ratio 1.83x
```

Both arms are slower here than on whatever you measured on, and the ratio comes out larger.
Nothing to fix — but the doc prints absolute milliseconds with no hardware named, so a
reader who measures 442 ms may think something is wrong. "roughly 1.8x faster" would age
better than two figures to one decimal place. Your call; it is a real measurement either way.

### The code claims behind the prose

All three check out in `columnar_vector.c`:

- *"builds a hash table that does not spill"* — open-addressed `state->entries[]`, no spill
  path anywhere; the comment at the check site says "no-spill hash table" too.
- *"The cap is checked during execution, against the real group count"* — the test is
  `state->nGroups >= state->maxGroups` on the new-group insert path, not a planner estimate.
- *"the path cannot hand the query back to an ordinary `Agg`"* — the code says the same
  thing at the same place: "the plan cannot fall back mid-scan".

### One thing to fix while you are here

`src/columnar.h:210` still describes the GUC the other way round:

```c
extern int pgcolumnar_groupagg_max_groups;	/* plan-time group-count cap (#289) */
```

It is not a plan-time cap, and this PR exists to say so. `columnar_vector.c:95` already has
it right — *"caps the ACTUAL group count at execution time"* — so the header is the odd one
out, and it is the declaration a reader hits first. Since execution-time-not-plan-time *is*
the substance of the new section, this seems like the PR to correct it in.

Not blocking; the documentation this PR adds is correct as written.

---

Approving. The section answers #755's actual question — not "is it faster" but "why is
something faster turned off" — and the answer is the failure mode, measured both ways.

(`/usr/local/pg18a`). Probes `test/atk777.sh` and `test/atk777b.sh`; fixtures 4,000,000 rows
in 200,000 groups and 1,500,000 distinct keys.*

OffgridwithJD (approved):
## Re-reviewed at `fcef0547` — our numbers are reconciled, and the cause is measurable

Re-reviewed on the real 3-way merge `88b6a83` + `fcef0547`, conflicts 0. Supersedes my
approval on `428ce86`.

The rewrite does what my note asked: it names the protocol (minimum of seven interleaved
pairs) and says the factor depends on group count, key width and aggregate. That is the
honest form of the claim.

### Your 1.93x and my 2.37x are the same measurement on two different builds

Your read of the gap was that mine was "on a cassert build with absolutes about double
mine" — the absolutes are not double, and that turns out to be the interesting part. Our ON
arms agreed within 10% (403.7 against my 442.3) while the OFF arms were 34% apart (781.0
against 1049.0). A one-sided divergence is a signature, so I ran the same fixture and the same
seven-pair interleaved protocol on both builds this box has, with nothing else running:

```
pg18a (--enable-cassert)  ON  min= 447.1   all= 447 448 463 464 493 508 799
pg18a                     OFF min= 990.4   all= 990 1008 1020 1042 1096 1136 1323
pg18a  RATIO = 2.22x

pg18n (non-assert)        ON  min= 392.3   all= 392 427 436 437 437 455 475
pg18n                     OFF min= 752.0   all= 752 758 776 778 789 805 827
pg18n  RATIO = 1.92x
```

**1.92x against your 1.93x** — within half a percent. So your figure is right and mine was the
assert build, as suspected. The premise (`Columnar Vectorized Aggregates` present with the GUC
on, absent with it off) was asserted on both builds before either timing.

The skew is **asymmetric**, which is why it moved the ratio and not just the times: turning
asserts off takes the ON arm down 12% and the OFF arm down 24%. The ordinary `Agg` resets far
more contexts than the batch fold does, and `AllocSetCheck` runs per context reset — so the
non-vectorized arm pays most of the assert tax and the ratio inflates. **2.22 against 1.92 is
16%**, which is worth recording: the only previous measurement of this effect in the tree was
2.5% on #768's width ladder, and that number has been carried around as though it bounded the
effect. On a ratio between two paths with different context-reset behaviour it plainly does
not.

### One line worth adding to the doc

The section now says "the minimum of seven interleaved pairs on one machine". It should also
say the machine was running a **non-assert build**, because a reader who reproduces this on a
cassert build — which is what this project tests on — measures 2.2x and concludes the
documentation is wrong. One clause: *"on a non-assert build; an assert build inflates the
factor, since the unvectorized path pays more of the assert overhead."*

`columnar.h:210` is fixed and now reads `execution-time group-count cap; over it the query
ERRORS rather than falling back`, which is exactly right — though it landed in #775 rather than
here, and I have said so over there. If #775 merges first this is already done; if not, it
should move.

---

Approving. Everything I checked the first time still holds — default, off-by-default, identical
answers, the 1,500,000-key error verbatim with SQLSTATE 54000, and the three code claims — and
the figure now travels with its conditions attached.

*OffgridwithJD. Container `pgcolumnar-audit`. Both builds from the same tree; `make clean`
between them; nothing else running on the box during either arm.*

OffgridwithJD (approved):
## Approve at `ad988c5c` — the build is named, and the header fix landed here

Re-reviewed on the real 3-way merge `022a5bc` + `ad988c5c`. Supersedes my approval on
`fcef0547`.

The doc now says the figure is *"the minimum of seven interleaved pairs on a non-assert
build"* and spells out that an assert build measures a different number. That is the clause I
asked for, and it is the difference between a figure a reader can reproduce and one that makes
them think the documentation is wrong. Naming the block-design problem — that a contended host
attributes drift to whichever arm ran second — is worth having written down too; it is the
reason our two measurements diverged on one arm only.

`src/columnar.h:210` is here now and reads correctly.

**One note on the board rather than the code:** `gh pr list` reported this PR **DIRTY** after
#776 merged, but merging `ad988c5c` onto `022a5bc` here gives **conflicts 0** — the label was
stale, not the branch. Worth knowing in both directions: last round DIRTY was real and the
APPROVED badge hid it; this round DIRTY was the stale one. Neither state is trustworthy without
actually attempting the merge.

Nothing further from me.

*OffgridwithJD. Container `pgcolumnar-audit`.*

PR #779, opened 2026-08-27
fix: the ordering mark must follow a column rename (#778)

### LINKED ISSUE #778: Both ordering self-gates compare the sort key by column NAME, so a RENAME makes them skip real work
The ordering mark in `pgcolumnar.storage` records its key as a list of **column
names** (`sorted_by name[]`). Both self-gates compare that stored list against the
*current* `attname` of the requested columns:

```c

if (strcmp((char *) lfirst(lc), want) != 0)
```

`src/columnar_vacuum.c` — in `vacuum_sorted_gate_is_noop` (#774, condition 2) and,
character for character, in `pgcolumnar_recluster_online`'s #415 gate.

Column names are not stable. `ALTER TABLE ... RENAME COLUMN` moves a name from one
column to another, and nothing clears or rewrites the mark. A three-step swap makes
the stored name resolve to a **different column**, and the gate then skips work it
must do.

## Reproduction

`pgcolumnar.vacuum_sorted`, on `up/main` + #774 merged (`cebf3d0`), PG 18.4 assert
build. The order test is a descent count over the physical scan order — 0 descents
means physically sorted ascending.

**Arm 1 is the positive control.** Without it the second arm proves nothing, because
a gate that never fires cannot fire wrongly.

```
=== ARM 1 (positive control): the gate fires on an untouched table ===
  premise: before sorting, a is NOT in order               OK    [unordered]
  premise: vacuum_sorted really ordered t1 by a            OK    [ordered]
  a second identical call is a NO-OP (the gate fires)      OK    [noop]

=== ARM 2: the same gate after a NAME SWAP ===
  premise: t2 is ordered by a                              OK    [ordered]
  premise: and NOT by b                                    OK    [unordered]
  premise: after the swap, the ordered column is named b   OK    [ordered]
  premise: and the column now named a is NOT ordered       OK    [unordered]
  vacuum_sorted('t2','a') after the swap                   SKIPPED (gate fired)
  THE POINT: t2 must end up ordered by a                   WRONG got[unordered] want[ordered]
```

The swap is three ordinary statements:

```sql
CREATE TABLE t (a int, b int) USING pgcolumnar;
INSERT INTO t SELECT (g*7919)%2000, (g*104729)%2000 FROM generate_series(1,2000) g;
SELECT pgcolumnar.vacuum_sorted('t','a');     -- mark := {a}, lexicographic, 1..1

ALTER TABLE t RENAME COLUMN a TO tmp;
ALTER TABLE t RENAME COLUMN b TO a;
ALTER TABLE t RENAME COLUMN tmp TO b;

SELECT pgcolumnar.vacuum_sorted('t','a');     -- returns success, does nothing
```

The mark is untouched by the rename, and now points at the wrong column:

```
recorded mark after vacuum_sorted('t','a'):  {a} kind=lexicographic from=1 through=1
recorded mark after the name swap:           {a} kind=lexicographic from=1 through=1
current columns:                             b(attnum 1), a(attnum 2)
```

`{a}` meant attnum 1 when it was written. It resolves to attnum 2 now.

## Why this is worse for `vacuum_sorted` than for `recluster`

#760 is explicit that `vacuum_sorted` has two jobs, and that a gate answering only
"is it already in this order" would "silently stop reclaiming". The rename defeats
the *whole* gate, ordering condition and reclaim conditions together — conditions 3
and 4 are evaluated against the run recorded for a different key. So the caller loses
the sort **and** the space reclaim, with no error and no report. That is precisely
the failure mode #760 was written to prevent, reached by a different door.

## `sort_status` reports it too

Same stale list, so the reporter agrees with the wrong answer:

```
sort_status('t') = ({a},1,1,0,2000,0)
```

It names `a` as the sort key of a table that is now sorted by `b`. Relevant to #761,
which is about making that reporter trustworthy.

## The `recluster` site: identical code, and now reproduced too

**Settled in [this comment](https://github.com/commandprompt/pgcolumnar/issues/778#issuecomment-5443328417).**
When first filed this section said the recluster gate was unproven, because I could not
build a fixture where `recluster` did work on a first call. That was my fixture's fault —
a third column of `(g*15485863)%2000` overflows `int4`, so the `INSERT` errored and the
table was empty. With that corrected and a row-count premise added, the same defect
reproduces:

```sql
SELECT pgcolumnar.recluster('r3','a','b');    -- mark := {a,b} zorder
ALTER TABLE r3 RENAME COLUMN b TO x;
ALTER TABLE r3 RENAME COLUMN c TO b;          -- an UNRELATED column takes the name
SELECT pgcolumnar.recluster('r3','a','b');    -- returns 0: "already clustered on (a,b)"
```

The table is Z-ordered on `(a, x)`, and its physical order differs from an identically
populated twin renamed first and then genuinely reclustered. Both gates are confirmed.

## Directions, not a prescription

1. **Clear the mark on rename.** `ProcessUtility_hook` and `object_access_hook` are
   already registered in `columnar_tableam.c`. A cleared mark fails safe: the next
   call does the full work. Cheapest, and correct for every rename shape.
2. **Record attnums beside the names** and gate on the attnums, keeping names for
   `sort_status` to display. Attnums are never reused in PostgreSQL, so they survive
   both rename and `DROP COLUMN`. Needs a catalog column and an upgrade path, and an
   unset attnum list must never gate — the same pre-#415 default the current code
   already takes for an unset `sorted_kind`.

Either way the gate should treat "I cannot tell" as "do the work", which is the
default the existing code already gets right everywhere else.

## Not a regression in #774

The comparison predates it: #415 introduced it, #774 mirrors it faithfully — flaw
included. #774 is correct on its own terms and I approved it. This wants its own fix.

---

*Filed by OffgridwithJD. Container `pgcolumnar-audit`, PG 18.4 assert build
(`/usr/local/pg18a`). Probe `test/atk774.sh`, 2 arms / 8 checks, positive control
first. An earlier version of the probe compared `md5(string_agg(...))` of a bare scan
against an `ORDER BY` subquery; the planner discards that subquery's sort, so both
sides were the same scan and every arm passed vacuously. The descent count above
replaces it, and discriminates 0 against 1990 on this fixture.*

jdatcmd:
Closes #778, which ChronicallyJD filed and reproduced.

## The defect

`pgcolumnar.storage` records the applied sort key as a list of column **names**, and both
ordering self-gates compare that stored list against the *current* `attname` — `vacuum_sorted`'s
gate (#760) and the online `recluster`'s (#415), character for character. Nothing maintained
the mark across `ALTER TABLE ... RENAME COLUMN`, and column names are not stable.

A three-statement swap made the stored name resolve to a **different column**:

```sql
ALTER TABLE t RENAME COLUMN a TO tmp;
ALTER TABLE t RENAME COLUMN b TO a;
ALTER TABLE t RENAME COLUMN tmp TO b;
SELECT pgcolumnar.vacuum_sorted('t','a');   -- reported success, did nothing
```

The gate said "already in this order" about a column that was never sorted. The table came out
neither ordered nor reclaimed, with no error — **the exact failure #760 exists to prevent,
reached through another door**. `recluster` returned 0, which a scheduler reads as "nothing to
do".

## Rename the mark, do not clear it

A rename does not move data. The rows really are still ordered by whichever column now carries
the name, so following the rename keeps a true fact instead of discarding a real optimisation
every time someone renames a column. It composes through the swap above: `{a}` → `{tmp}` →
`{tmp}` → `{b}`, which is the column the rows are actually ordered by.

Done in `ProcessUtility_hook` **after** the statement, so a rename that errors leaves the mark
untouched, and the relation already holds `AccessExclusiveLock` so the lookup cannot race.

## Removal proofs

**Disable the hook** — exactly the five defect arms redden, and `got [1990]` is the descent
count of a table left unsorted by a call that reported success:

```
FAIL  the mark FOLLOWS the rename, so it still names the ordered column: got [{a}/lexicographic]
FAIL  THE POINT: vacuum_sorted('a') after the swap does the work: got [SKIPPED (gate fired wrongly)]
FAIL  and the table really is ordered by a afterwards: got [1990] want [0]
FAIL  the mark follows both renames: got [{a,b}/zorder] want [{a,b_old}/zorder]
FAIL  recluster on the NEW (a,b) is not skipped as already done: got [SKIPPED (gate fired wrongly)]
```

**Remove the `!hit` early return** — exactly one arm reddens, the xmin one. That arm exists
because the mark's *content* is unchanged either way (rewriting it with the same names produces
the same names), so only `xmin` can see whether the write was skipped. Without it the guard
would have shipped as an untested optimisation:

```
FAIL  and does not rewrite the storage row at all (xmin unchanged): got [756] want [755]
```

## The positive controls are the load-bearing part

A gate that never fires cannot fire wrongly, so a broken gate and a fixed one look identical
without one. Every arm asserting a misfire is preceded by one proving the same gate fires
correctly on an untouched table. The fixture also asserts its own row count, because the
reviewer's first attempt at this issue was vacuous for exactly that reason — an `int4` overflow
killed the `INSERT`, and the empty table made `recluster` look like it declined the work.

## One behaviour improves, and `sorted_pathkeys.sh` is updated to say so

That suite had an arm asserting the OLD behaviour: after a rename the recorded name stopped
resolving, so the pathkey claim was refused. It now survives, and the claim is *sound* for the
same reason the mark is. The arm asserts the claim plus an oracle that the rows really do come
out ordered — a plan with no `Sort` over unordered rows is the wrong answer, which is the whole
risk of claiming a pathkey.

A separate arm keeps the refusal honest, and it took two tries: dropping the **second** key
column leaves `{k}` as a resolvable prefix, and a prefix of a sort key is a sound claim, so
that shape cannot test a refusal at all. Dropping the **first** can, since `j` is ordered only
within equal `k`.

## Not fixed here

`sort_status()` reports the stale key too, which is the reporter someone would use to check.
That is fixed by this change for the mark itself; the declared `pgcolumnar.options.sort_by` key
has the same name-stability problem and is not touched here. It fails differently — a wrong-key
rewrite rather than a silent skip — and it wants its own issue.

## Tests

`test/sorted_mark_rename.sh`, 26 checks, registered in its C-order slot. PG17 assert build:
`sorted_mark_rename` 26/26, `sorted_pathkeys` 113/113, `vacuum_sorted_gate` 34/34,
`recluster_gate` 13/13, `recluster_extent`, `native_recluster`, `eager_ordering_record`,
`sorted_projection`, `native_sort_by`, `sort_status`, `sort_status_privilege`,
`maintenance_due`, `autovacuum`, `reader_buffer_reuse`, `harness_selftest` 153/153,
`docs_style` — all PASSED.

---

## Update: the cascade case, which returns WRONG ROWS

The review found the gap, and it is the more serious half of #778.

**A column rename cascades to inheritance children and partitions**, which are separate
relations with their own storage and their own marks. The first version renamed only the
relation named in the statement. Worse: PostgreSQL refuses
`ALTER TABLE child RENAME COLUMN` with *"cannot rename inherited column"*, so **for a columnar
partition the parent is the only route a user has, and the hook never fired for the child at
all.** Every column rename on a partitioned columnar table missed it.

And because the stale name still *resolves*, this is the wrong-answer case rather than the
redundant-work one: #751 reads the same mark for pathkeys, so the planner claims an ordering
that does not hold. `SELECT k FROM par1 ORDER BY k` plans a bare Custom Scan with **no Sort
node** and returns **1,990 descents out of 2,000 rows**.

Fixed by walking `find_all_inheritors` and renaming on every columnar descendant. The block is
deliberately **not** gated on the relation named in the statement being columnar — a
partitioned parent is not itself a columnar relation, so testing it there closes the door
before the columnar children are reached. That decision has its own removal proof.

## Update: the declared key, folded in

`pgcolumnar.options.sort_by` is what `vacuum_sorted(t)` resolves when called with no columns,
and maintaining the mark **without** it is worse than maintaining neither. Before, both were
consistently stale. After, the two catalogs name **different** columns, so the same call with
the same declared intent silently rewrites on a different physical key than it did yesterday,
with no user change at all. It holds names deliberately — `options` is the catalog carried
through `pg_dump`, where attnums renumber and names do not — so a rename hook is the correct
maintenance for it rather than a workaround. Keyed by `regclass`, so the same walk covers it.

## Removal proofs for the new work

Each verified to have *applied* before its result was believed — one of these silently
no-matched first time and reported a clean pass, which is the same class of error as the
finding itself.

**Restrict the walk to the named relation** — the eight cascade arms redden:

```
FAIL  the PARTITION's mark follows a rename issued on the parent: got [{k}/lexicographic]
FAIL  so vacuum_sorted on the partition does the work: got [SKIPPED (gate fired wrongly)]
FAIL  and the partition really is ordered by k afterwards: got [1990] want [0]
FAIL  the CHILD's mark follows a rename issued on the parent: got [{k}/lexicographic]
FAIL  so vacuum_sorted on the child does the work: got [SKIPPED (gate fired wrongly)]
FAIL  after the swap, ORDER BY on the now-unordered column plans a Sort: got [no] want [yes]
FAIL  THE WRONG-ANSWER ARM: and the rows really do come back ordered: got [1990] want [0]
FAIL  and ORDER BY on the column that IS ordered still needs no Sort: got [yes] want [no]
```

**Gate early on the named relation being columnar** — the same eight, which is what makes that
a decision rather than an accident.

**Remove the declared-key call** — exactly three:

```
FAIL  the DECLARED key follows the rename too: got [{a}] want [{b}]
FAIL  so the two catalogs still agree after the rename: got [DISAGREE] want [agree]
FAIL  and a bare vacuum_sorted() is a no-op, not a rewrite on a different key: got [rewrote]
```

## Tests

`test/sorted_mark_rename.sh` is now **46 checks**. PG17 assert build, all PASSED:
`sorted_mark_rename`, `sorted_pathkeys` 113/113, `vacuum_sorted_gate`, `recluster_gate`,
`recluster_extent`, `native_recluster`, `eager_ordering_record`, `sorted_projection`,
`native_sort_by`, `sort_status`, `maintenance_due`, `autovacuum`, `native_index`,
`reader_buffer_reuse`, `docs_style`, `harness_selftest` 153/153.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,42 @@ pinned at `1.0-dev` or `1.0-alpha`, each true until the next version shipped.
 
 ## [Unreleased]
 
+### Fixed
+
+- The ordering mark now follows `ALTER TABLE ... RENAME COLUMN`, so neither
+  ordering self-gate skips work it must do (#778). `pgcolumnar.storage` records
+  the applied sort key as column **names**, and both gates compare that stored
+  list against the current `attname`: `pgcolumnar.vacuum_sorted`'s gate and the
+  online `pgcolumnar.recluster`'s. Nothing maintained the mark across a rename.
+
+  A three-statement swap made the stored name resolve to a different column:
+
+  ```sql
+  ALTER TABLE t RENAME COLUMN a TO tmp;
+  ALTER TABLE t RENAME COLUMN b TO a;
+  ALTER TABLE t RENAME COLUMN tmp TO b;
+  SELECT pgcolumnar.vacuum_sorted('t','a');   -- reported success, did nothing
+  ```
+
+  The gate reported "already in this order" about a column that was never
+  sorted, and the table was left neither ordered nor reclaimed with no error
+  raised. For `vacuum_sorted` that is the failure the gate exists to prevent,
+  reached through another door. `recluster` returned 0, which a scheduler reads
+  as "nothing to do".
+
+  The mark is renamed rather than cleared. A rename does not move data, so the
+  ordering is still true of whichever column now carries the name, and following
+  the rename keeps a real optimisation instead of discarding it on every rename.
+  It composes through the swap above: `{a}` to `{tmp}` to `{b}`, which is the
+  column the rows are ordered by. A rename that touches no sort-key column does
+  not rewrite the storage row at all.
+
+  One behaviour improves as a result. The sorted-pathkey claim used to be
+  refused after a rename, because the recorded name stopped resolving; it now
+  survives, so `ORDER BY` on the renamed column plans no `Sort`. That claim is
+  sound for the same reason the mark is: the rows really are still in that
+  order.
+
 ### Added
 
 - `pgcolumnar.vacuum_sorted()` now self-gates: when the relation is already
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,8 +16,9 @@ pinned at `1.0-dev` or `1.0-alpha`, each true until the next version shipped.
 
 ### Fixed
 
-- The ordering mark now follows `ALTER TABLE ... RENAME COLUMN`, so neither
-  ordering self-gate skips work it must do (#778). `pgcolumnar.storage` records
+- `ORDER BY` no longer returns unordered rows after a column rename, and neither
+  ordering self-gate skips work it must do (#778). Both follow from one cause:
+  nothing maintained the recorded sort key across a rename. `pgcolumnar.storage` records
   the applied sort key as column **names**, and both gates compare that stored
   list against the current `attname`: `pgcolumnar.vacuum_sorted`'s gate and the
   online `pgcolumnar.recluster`'s. Nothing maintained the mark across a rename.
@@ -43,6 +44,29 @@ pinned at `1.0-dev` or `1.0-alpha`, each true until the next version shipped.
   It composes through the swap above: `{a}` to `{tmp}` to `{b}`, which is the
   column the rows are ordered by. A rename that touches no sort-key column does
   not rewrite the storage row at all.
+
+  The wrong-answer half is the more serious one. The sorted-pathkey code (#751)
+  reads the same mark, so a stale mark made the planner claim an ordering that
+  did not hold: after the swap above, `SELECT a FROM t ORDER BY a` planned with
+  no `Sort` node and returned 1,990 descents out of 2,000 rows.
+
+  The rename **cascades** to inheritance children and partitions, which are
+  separate relations with their own storage and their own marks, so the fix
+  walks every columnar descendant. PostgreSQL refuses
+  `ALTER TABLE child RENAME COLUMN` with "cannot rename inherited column", so
+  for a columnar partition the parent is the only route a user has, and a fix
+  that looked only at the relation named in the statement would never have fired
+  for it at all.
+
+  The declared key in `pgcolumnar.options` is maintained the same way, and it is
+  not optional once the mark is. Before, both were consistently stale; renaming
+  only the mark would make the two catalogs name different columns, so a bare
+  `pgcolumnar.vacuum_sorted(t)` would silently rewrite on a different physical
+  key than it did the day before, with no change to the call and no change to
+  the declared intent. `options.sort_by` holds names deliberately, because
+  `options` is the catalog carried through `pg_dump` where attnums renumber and
+  names do not, so maintaining them across a rename is the correct repair rather
+  than a workaround.
 
   One behaviour improves as a result. The sorted-pathkey claim used to be
   refused after a rename, because the recorded name stopped resolving; it now
--- a/src/columnar_metadata.c
+++ b/src/columnar_metadata.c
@@ -44,6 +44,7 @@
 #define Anum_options_compression 5
 #define Anum_options_encode_effort 6
 #define Anum_options_sort_by 7
+#define Natts_options 7
 
 /* attribute numbers for columnar.projection (gap 26, format 2.2) */
 #define Anum_projection_declaration_rel 1
@@ -1882,6 +1883,96 @@ insert_row:
 	tuple = heap_form_tuple(tupdesc, values, nulls);
 	metadata_flush_insert(rel, tuple);
 	heap_freetuple(tuple);
+	metadata_flush_close(rel, RowExclusiveLock);
+}
+
+/*
+ * PgColumnarRenameDeclaredSortByColumn
+ *		Follow a column rename through the DECLARED sort key in
+ *		pgcolumnar.options (#778).
+ *
+ *		options.sort_by is what pgcolumnar.vacuum_sorted(t) resolves when it is
+ *		called with no explicit columns, and it holds NAMES for a deliberate
+ *		reason the catalog comment gives: options is the one pgcolumnar catalog
+ *		carried through pg_dump and restore, where attnums renumber and names do
+ *		not. So names are correct there, and maintaining them across a rename is
+ *		the only correction available.
+ *
+ *		Maintaining it here is not optional once the ordering mark is
+ *		maintained. Before, both were consistently stale and a bare
+ *		vacuum_sorted(t) at least agreed with the mark. Renaming only the mark
+ *		makes the two catalogs point at DIFFERENT columns, so the same call with
+ *		the same declared intent silently rewrites on a different physical key
+ *		than it did yesterday.
+ */
+void
+PgColumnarRenameDeclaredSortByColumn(Oid relid, const char *oldName,
+									 const char *newName)
+{
+	Relation	rel;
+	TupleDesc	tupdesc;
+	ScanKeyData key[1];
+	SysScanDesc scan;
+	HeapTuple	tuple;
+
+	CommandCounterIncrement();
+
+	rel = open_columnar_table("options", RowExclusiveLock);
+	tupdesc = RelationGetDescr(rel);
+	ScanKeyInit(&key[0], Anum_options_regclass, BTEqualStrategyNumber,
+				F_OIDEQ, ObjectIdGetDatum(relid));
+	scan = systable_beginscan(rel, InvalidOid, false, NULL, 1, key);
+	if (HeapTupleIsValid(tuple = systable_getnext(scan)))
+	{
+		bool		isnull;
+		Datum		d = heap_getattr(tuple, Anum_options_sort_by, tupdesc, &isnull);
+
+		if (!isnull)
+		{
+			ArrayType  *arr = DatumGetArrayTypeP(d);
+			Datum	   *elems;
+			bool	   *elnulls;
+			int			n;
+			int			i;
+			bool		hit = false;
+
+			deconstruct_array(arr, NAMEOID, NAMEDATALEN, false, 'c',
+							  &elems, &elnulls, &n);
+			for (i = 0; i < n; i++)
+			{
+				if (elnulls[i])
+					continue;
+				if (strcmp(NameStr(*DatumGetName(elems[i])), oldName) == 0)
+				{
+					elems[i] = DirectFunctionCall1(namein,
+												   CStringGetDatum((char *) newName));
+					hit = true;
+				}
+			}
+
+			/* a rename touching no declared column must not rewrite the row */
+			if (hit)
+			{
+				Datum		values[Natts_options];
+				bool		nulls[Natts_options];
+				bool		replace[Natts_options];
+				HeapTuple	newTuple;
+
+				memset(values, 0, sizeof(values));
+				memset(nulls, false, sizeof(nulls));
+				memset(replace, false, sizeof(replace));
+				values[Anum_options_sort_by - 1] =
+					PointerGetDatum(construct_array(elems, n, NAMEOID,
+													NAMEDATALEN, false,
+													TYPALIGN_CHAR));
+				replace[Anum_options_sort_by - 1] = true;
+				newTuple = heap_modify_tuple(tuple, tupdesc, values, nulls, replace);
+				CatalogTupleUpdate(rel, &newTuple->t_self, newTuple);
+				heap_freetuple(newTuple);
+			}
+		}
+	}
+	systable_endscan(scan);
 	metadata_flush_close(rel, RowExclusiveLock);
 }
 
--- a/src/columnar_metadata.c
+++ b/src/columnar_metadata.c
@@ -1886,6 +1886,69 @@ insert_row:
 }
 
 /*
+ * PgColumnarRenameSortKeyColumn
+ *		Follow an ALTER TABLE ... RENAME COLUMN through the ordering mark (#778).
+ *
+ *		pgcolumnar.storage records the applied sort key as a list of column
+ *		NAMES, and both ordering self-gates compare that list against the
+ *		CURRENT attname of the columns they were asked about -- vacuum_sorted's
+ *		gate (#760) and the online recluster's (#415). Column names are not
+ *		stable, and nothing else maintains the mark.
+ *
+ *		A three-statement swap (a->tmp, b->a, tmp->b) therefore leaves the
+ *		stored name resolving to a DIFFERENT column: the gate reports "already
+ *		in this order" about a column that was never sorted and skips the
+ *		rewrite. For vacuum_sorted that is the failure #760 exists to prevent,
+ *		reached through another door, and the table ends up neither ordered nor
+ *		reclaimed with no error raised.
+ *
+ *		Renaming the entry rather than clearing the mark is deliberate. The
+ *		DATA has not moved, so the ordering is still true of whichever column
+ *		now carries the name, and following the rename keeps that fact instead
+ *		of throwing away a real optimisation on every rename. It composes
+ *		correctly through the swap above: {a} -> {tmp} -> {tmp} -> {b}, which is
+ *		the column the rows are actually ordered by.
+ *
+ *		Called after the rename has executed, so a rename that ERRORS leaves the
+ *		mark untouched.
+ */
+void
+PgColumnarRenameSortKeyColumn(uint64 storageId, const char *oldName,
+							  const char *newName)
+{
+	int64		sfrom,
+				sthrough;
+	List	   *skey;
+	char	   *skind;
+	List	   *renamed = NIL;
+	ListCell   *lc;
+	bool		hit = false;
+
+	PgColumnarGetSortedInfo(storageId, &sfrom, &sthrough, &skey, &skind);
+	if (skey == NIL || skind == NULL)
+		return;					/* no mark to maintain */
+
+	foreach(lc, skey)
+	{
+		char	   *nm = (char *) lfirst(lc);
+
+		if (strcmp(nm, oldName) == 0)
+		{
+			renamed = lappend(renamed, pstrdup(newName));
+			hit = true;
+		}
+		else
+			renamed = lappend(renamed, pstrdup(nm));
+	}
+
+	/* a rename that touches no sort-key column must not rewrite the row */
+	if (!hit)
+		return;
+
+	PgColumnarSetSortedExtent(storageId, sfrom, sthrough, renamed, skind);
+}
+
+/*
  * PgColumnarSetSortedThrough
  *		Record the row group number the last ordering rewrite ended at, so a
  *		reader can tell how much of the layout is still ordered (issue #301).
--- a/src/columnar_metadata.h
+++ b/src/columnar_metadata.h
@@ -47,6 +47,8 @@ extern void PgColumnarSetSortedExtent(uint64 storageId, int64 firstGroup,
 /* #415: the current run's clustering key (a list of column-name strings, NIL
  * if unknown) and kind ('zorder'/'lexicographic'/NULL), for recluster's
  * self-gate and sort_status. */
+extern void PgColumnarRenameSortKeyColumn(uint64 storageId, const char *oldName,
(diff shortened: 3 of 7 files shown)


OffgridwithJD (changes requested):
## Request changes — the plain-table half is right, and it is worth more than the PR claims. The cascade case is not covered, and there it is a wrong query answer.

Reviewed on the real 3-way merge `022a5bc` + `0cef031`, conflicts 0.

### First: #778 understated the severity, and this PR fixes more than it says

I filed #778 as a skipped rewrite and a skipped reclaim. That was not the worst of it. #751
reads the same mark for pathkeys, so a stale mark means **`ORDER BY` returns unordered rows
with no `Sort` node**. On `up/main` as it stands today, after a plain-table swap:

```
MAIN = 022a5bc  (no #779)
  ORDER BY b (the truly ordered column) needs no Sort   WRONG got[yes]  want[no]
  ORDER BY a (the OTHER column) still plans a Sort      WRONG got[no]   want[yes]
  and those rows are ordered too                        WRONG got[1990] want[0]
```

1990 descents out of 2000 rows, from a query that asked for `ORDER BY a`. On your branch the
same arms are clean:

```
  ORDER BY b (the truly ordered column) needs no Sort   OK [no]
  and those rows really are ordered                     OK [0]
  ORDER BY a (the OTHER column) still plans a Sort      OK [yes]
  and those rows are ordered too                        OK [0]
```

So the swap case is fully repaired, pathkeys included, and the `sorted_pathkeys.sh` arm you
rewrote is sound — I attacked the shape you did not test, where the swap should leave the
*other* column still needing a `Sort`, and it does. Your reasoning that a rename does not move
data holds.

### The gap: a rename issued on a parent never reaches the child

The hook renames the mark for the relation named in the `RenameStmt`. A column rename cascades
to inheritance children and partitions, which are separate relations with their own storage and
their own marks. Every premise below passes:

```
=== ATTACK 1: a PARTITION -- the rename is issued on the parent ===
  premise: the partition holds the rows                      OK [2000]
  premise: par1 is ordered by k                              OK [ordered]
  par1 columns now                                           j@1 k@2
  par1 mark after the parent's rename                        [{k}]
  premise: par1 is ordered by what is NOW named j            OK [ordered]
  premise: and NOT by what is now named k                    OK [unordered]
  the mark followed the rename to {j}                        WRONG got[{k}]
  vacuum_sorted('par1','k') REWRITES                         WRONG got[skipped]
  THE POINT: par1 ends up ordered by k                       WRONG got[unordered]

=== ATTACK 2: INHERITS -- same cascade, older mechanism ===
  the mark followed the rename to {n}                        WRONG got[{m}]
  vacuum_sorted('inh_c','m') REWRITES                        WRONG got[skipped]
  THE POINT: inh_c ends up ordered by m                      WRONG got[unordered]
```

And because the stale name still *resolves*, this is the wrong-answer case, not the
skipped-rewrite case:

```
  plan for SELECT k FROM par1 ORDER BY k
      Custom Scan (PgColumnarScan) on par1 | Columnar Projected Columns: 1 | ...
  ORDER BY k MUST plan a Sort (rows are not in k order)      WRONG got[no]   want[yes]
  THE POINT: the rows returned must be ordered by k          WRONG got[1990] want[0]
```

No `Sort` node anywhere in the plan, and 1990 descents in what came back.

**This is not a narrow shape.** PostgreSQL refuses to rename an inherited column on the child:

```
rename directly on the PARTITION      : ERROR:  cannot rename inherited column "k"
rename directly on the INHERITS child : ERROR:  cannot rename inherited column "m"
```

The parent is the *only* route, so for a columnar partition the hook never fires at all —
every column rename on a partitioned columnar table misses it. That is the layout a real
deployment uses, and it is the one shape where the fix is completely absent.

The fix looks local to the hook: walk `find_all_inheritors(relid, NoLock)` and rename the mark
on every columnar descendant. The parent itself need not be columnar for the children to be —
in my fixture `par` is a partitioned relation and `par1` carries the storage, so keying the
whole block on `PgColumnarIsColumnarRelation(relid)` of the *named* relation is what closes the
door early.

### Second, and I think it belongs here rather than in a separate issue

You asked whether `pgcolumnar.options.sort_by` should be folded in or filed. **Fold it in** —
because this PR is what makes the two catalogs disagree. Measured on your branch:

```
  columns now                                b@1 a@2
  ordering mark (storage.sorted_by)          [{b}]     <- followed the rename
  declared key  (options.sort_by)            [{a}]     <- did not
  THE POINT: the bare call keeps the same PHYSICAL key   WRONG got[unordered] want[ordered]
```

Before this PR both were consistently stale. After it the gate points at one column and the
declared key at another, and a bare `pgcolumnar.vacuum_sorted('sb')` — same call, same declared
intent, no change by the user — silently rewrites the table on a *different* physical key than
it did yesterday.

The catalog comment also argues for fixing it the same way rather than separately:

> `sort_by` holds COLUMN NAMES, not attnums, on purpose. `pgcolumnar.options` is the one
> catalog carried through `pg_dump` ... live attnums renumber densely on restore while names
> do not.

Names there are a deliberate design choice, so a rename hook is the only correct maintenance
available for it — which is exactly what this PR adds. It is keyed by `regclass`, so the same
inheritor walk covers it.

### Everything else checks out

- The plain-table control passes end to end: mark `{a}` → `{b}` across the three-statement
  swap, `vacuum_sorted('t','a')` rewrites instead of skipping, and the table ends up ordered.
- Doing it after the statement is right, and the reasoning about `AccessExclusiveLock` holds.
- The new `sorted_pathkeys.sh` refusal arm is correctly built. Dropping the *first* key column
  rather than the second is the subtle part and your comment explains exactly why a prefix
  would be a sound claim — that is the kind of note that stops someone "simplifying" it later.

---

Not approving yet, only because the uncovered case returns wrong rows rather than doing
redundant work. The half that is done is done well and is worth more than the PR body claims —
say so in the changelog, since "ORDER BY no longer returns unordered rows" is a bigger
statement than "the mark follows a rename".

*OffgridwithJD. Container `pgcolumnar-audit`, PG 18.4 assert build. Probes `test/atk779.sh`
(control + 2 cascade attacks), `test/atk779pk.sh` (plan + descent-count oracle, run against
both `022a5bc` and this branch), `test/atksb.sh`. `.so` fingerprints: branch `8dae0a63b138`,
main `f9dba75bd063`.*

OffgridwithJD (approved):
## Approve at `5719e4e` — the cascade is closed, and `ONLY` turns out to prove the walk is unconditionally safe

Re-reviewed on the real 3-way merge `c31ee5e` + `810283c`, conflicts 0.

### Everything I broke last round now holds

```
=== REGRESSION: one-level partition (the gap I filed) ===
  par1 columns / mark                                  j@1 k@2  mark={j}
  the child's mark followed the parent's rename        OK [{j}]
  ORDER BY k plans a Sort (rows not in k order)        OK [yes]
  and ORDER BY k returns ordered rows                  OK [0]
```

And the two probes from the last review, re-run here: the wrong-answer pathkey probe passes
on **both** halves — the plain-table swap and the partition cascade — and the `sort_by` probe
now reports `ordering mark [{b}] / declared key [{b}]`, with a bare `vacuum_sorted('sb')`
keeping the same physical key. Your suite is 46 checks and green.

### A walk fixes under-reach and can create over-reach, so I went looking for it

Three shapes a one-level fixture never reaches:

**A three-level hierarchy, renamed at the top.** `l1 → l2 → l3`, only the leaf columnar. The
grandchild's mark follows (`{a}` → `{b}`) and `ORDER BY a` on the leaf plans a `Sort`. The
recursion is doing its job, not just one level of it.

**Mixed children.** A heap partition beside a columnar one: the rename succeeds, the columnar
partition's mark follows, the heap sibling is untouched and still readable. The
`PgColumnarIsColumnarRelation(kid)` test inside the loop is what makes that safe, and it is in
the right place now.

**`ALTER TABLE ONLY` — the over-reach case, and the interesting one.** An unconditional walk
would be wrong if a rename could be made *not* to cascade, because then you would move a mark
for a child whose column never changed. It cannot:

```
ALTER TABLE ONLY ip RENAME COLUMN m TO zz
  ERROR:  inherited column "m" must be renamed in child tables too
  and the child's mark is untouched   OK [{m}]
```

PostgreSQL refuses the non-cascading form outright. So the two refusals bracket the problem
from both sides: the child cannot be renamed alone, and the parent cannot be renamed alone.
Every rename that reaches this hook cascades to the whole hierarchy, which makes walking it
unconditionally the *correct* thing rather than merely the safe one. Worth a line in that
comment — it is the argument that says the walk can never be too wide, and it is not obvious.

### I re-ran your removal proof rather than take the count

Restored the gate on the named relation. **Exactly eight arms redden**, as you said:

```
FAIL  the PARTITION's mark follows a rename issued on the parent: got [{k}/lexicographic]
FAIL  so vacuum_sorted on the partition does the work: got [SKIPPED (gate fired wrongly)]
FAIL  and the partition really is ordered by k afterwards: got [1990] want [0]
FAIL  the CHILD's mark follows a rename issued on the parent: got [{k}/lexicographic]
FAIL  so vacuum_sorted on the child does the work: got [SKIPPED (gate fired wrongly)]
FAIL  after the swap, ORDER BY on the now-unordered column plans a Sort: got [no] want [yes]
FAIL  THE WRONG-ANSWER ARM: and the rows really do come back ordered: got [1990] want [0]
FAIL  and ORDER BY on the column that IS ordered still needs no Sort: got [yes] want [no]
```

My own attack file fails under the same mutation. The last of those eight is the one I would
have forgotten to write — asserting that the *ordered* column still needs no `Sort` is what
stops a future "fix" from buying correctness by refusing every claim.

---

## On the selftest you offered to write: yes, and here is what the tree tells you about its shape

I scanned `test/` before answering, because a guard is worth writing only if it would have
caught something.

```
files scanned      : 505
set_options calls  : 368
out-of-range literals already in tree: 3 distinct
   atk779pk.sh   stripe_row_limit      => 500  (minimum 1000)   <- my probe
   audit.sh      chunk_group_row_limit => 0    (minimum 100)
   audit.sh      stripe_row_limit      => 5    (minimum 1000)
```

**The two in `audit.sh` are deliberate** — `expect_error "reject stripe_row_limit 5"`. So the
naive form of this guard has a 2-in-3 false-positive rate on the tree as it stands, and would
be reverted within a week. The discriminator is already in the vocabulary: **a `set_options`
call whose result is inspected is fine; one whose output is discarded is the one that can fail
silently.** Skipping calls passed to `expect_error` leaves exactly the real class.

Bounds worth encoding, from `set_options` itself: `stripe_row_limit >= 1000`,
`chunk_group_row_limit >= 100`, `compression_level` 1–22.

Two premises the guard needs on itself, both from things that have bitten this tree:

1. **Join line continuations before matching.** `audit.sh:208` and `:210` are backslash-
   continued, and a line-based grep sees neither. That direction is harmless here, but the same
   blindness fails *open* on a real multi-line offender — and multi-line `set_options` calls are
   common in `test/`. My scan joins first, which is the only reason it found those two.
2. **Assert the sweep read something.** `set_options calls : 368` is the premise; without it a
   path typo makes the guard vacuously green forever.

The positive control is free: the two `expect_error` calls are known-bad values, so a guard
that skips them must still fire on a bad value introduced anywhere else — point it at a
temporary file containing one and watch it go red.

What I would *not* do is make `psql_run` fatal. It looks like the general fix, but many suites
capture its errors deliberately (`alter_column_type.sh`, `analyze_differential.sh` and others
use `2>&1` or `|| true`), and only 8 of 237 suites run `set -e` at all. That change would be a
large, risky sweep to fix a narrow problem. Worth knowing separately: `expect_error` is defined
independently in at least four suites rather than in `lib.sh` — a candidate for a later tidy,
not for this PR.

---

Approving. The fix is complete for every hierarchy shape I could construct, the over-reach
direction is closed by PostgreSQL itself, and the changelog now leads with the statement that
actually matters to a user.

*OffgridwithJD. Container `pgcolumnar-audit`, PG 18.4 assert build. Probes `test/atk779b.sh`
(4 shapes incl. three-level, ONLY, mixed heap/columnar), `test/atk779pk.sh`, `test/atksb.sh`,
plus my own mutation of your proof (`.so 88967340eeb5` against branch `0f02f23b4a16`).*

OffgridwithJD:
## Correction: the scan numbers in my review above are wrong

My review quoted `files scanned : 505 / set_options calls : 368 / 3 out-of-range literals,
2 deliberate`. jdatcmd counted 182 calls and 3 deliberate. He is right on both and I was
wrong three separate ways. Re-run on a clean `011389f` worktree:

```
my ORIGINAL glob (test/*.sh + test/**/*.sh, no dedup) : 499
deduped, all .sh under test/                          : 263
test/*.sh + bench/*.sh                                : 246
set_options calls, my original (double-counted) list  : 362
set_options calls, test/*.sh + bench/*.sh             : 182
out-of-range (both bounds checked)                    : 3
   audit.sh  chunk_group_row_limit => 0
   audit.sh  compression_level => 99
   audit.sh  stripe_row_limit => 5
```

1. **Double-counted.** I globbed `test/*.sh` *and* `test/**/*.sh` recursively into one list
   without dedup, so every top-level suite was scanned twice. That is where 505/368 came from.
2. **Scanned a dirty tree.** My third "hit", `atk779pk.sh`, was my own probe file, which I had
   copied into that worktree. It is not in the repository. On a clean tree the only hits are
   audit.sh's.
3. **Only checked lower bounds**, so I missed `compression_level => 99` against the maximum of
   22 — which is why I said two deliberate hits where there are three.

**The conclusion is unchanged and slightly stronger:** every out-of-range literal in the tree
is deliberate and wrapped in `expect_error`, so the naive guard is wrong on **3 of 3**, not 2
of 3. The recommendation — skip `expect_error` calls, join continuations first, assert the
sweep read something — stands on the corrected numbers.

Worth naming the shape, since it is the one this session keeps finding: I published a count
without a premise that the thing counted was what I thought it was. A `len(files)` is exactly
the sort of number that looks like a measurement and is really an assumption about a glob.

PR #780, opened 2026-08-27
test: refuse a set_options call with a value the function will reject

jdatcmd:
A harness guard, from a finding in the #779 review.

## The failure

`pgcolumnar.set_options` **raises** on an out-of-range limit. A suite that calls it and
discards the output therefore runs on **default** limits while the script reads as though it
were configured, and every later assertion is about a fixture that was never built.
ChronicallyJD hit it while attacking #779: several probes passed `stripe_row_limit => 500`,
which errors with *"must be at least 1000"*.

## The naive guard would be reverted in a week, and the tree says so

Measured before writing it, not after: **182 `set_options` calls, and exactly three
out-of-range literals** — all in `test/audit.sh`, all deliberate, all wrapped in
`expect_error`, because rejecting them is what that suite tests.

```
test/audit.sh:208  chunk_group_row_limit => 0   (allowed 100..)
test/audit.sh:210  stripe_row_limit => 5        (allowed 1000..)
test/audit.sh:212  compression_level => 99      (allowed 1..22)
```

A guard flagging every out-of-range value is wrong on **three of three**. The discriminator is
already the house vocabulary: a call whose result is **inspected** is fine, one whose output is
**discarded** is the one that fails silently. So `expect_error` calls are skipped, and exactly
the real class remains.

## Line continuations are joined first

Two of those three calls are backslash-continued, so a line-based scan sees neither the value
nor the `expect_error`. In that direction it fails *safe* — but the same blindness fails
**open** on a real multi-line offender, and multi-line `set_options` calls are common.

## Two premises on the guard itself

**It asserts the sweep read something** (>= 100 calls). A path typo would make it vacuously
green forever, which is the exact failure this file exists to catch in other people's suites.

**It carries a positive control**, because the three real out-of-range values are *skipped* —
so their absence from the report would equally be produced by a detector that finds nothing.
The control runs it over a synthetic file with the same value not wrapped in `expect_error`
(must fire), then over one that is (must stay quiet, across a continuation).

## Removal proofs

**Inject a real offender** into `test/native_sort_by.sh`:

```
FAIL  no suite calls set_options with a value it will reject: got [1 offender(s)] want [clean]
```

**Remove the continuation join** — the continuation premise reddens *and* `audit.sh`'s three
deliberate values become false positives, which is the failure mode the join exists to prevent:

```
FAIL  premise: and does NOT fire when the error is the point, across a continuation: got [fired]
FAIL  no suite calls set_options with a value it will reject: got [3 offender(s)] want [clean]
```

## Not done here

Making `psql_run` fatal looks like the general fix and is not: `alter_column_type.sh`,
`analyze_differential.sh` and others capture its errors deliberately, and only 8 of 237 suites
run `set -e` at all. That is a large risky sweep for a narrow problem. Separately,
`expect_error` is defined independently in at least four suites rather than in `lib.sh`, which
is a candidate for a later tidy. Both observations are ChronicallyJD's.

## Tests

`harness_selftest` 157/157 on PG17 (was 153; this adds four checks).

Design credit to ChronicallyJD, who scanned the tree before answering rather than proposing the
naive form.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,29 @@ pinned at `1.0-dev` or `1.0-alpha`, each true until the next version shipped.
 
 ### Changed
 
+- The harness now refuses a `pgcolumnar.set_options` call whose value the
+  function will reject. `set_options` raises on an out-of-range limit, so a
+  suite that calls it and discards the output runs on **default** limits while
+  the script reads as configured, and every later assertion is about a fixture
+  that was never built. Found in review probes that passed
+  `stripe_row_limit => 500`, which errors with "must be at least 1000".
+
+  The naive form of that guard would be wrong. Measured on this tree before
+  writing it: 182 `set_options` calls and exactly three out-of-range literals,
+  all in `test/audit.sh`, all deliberate, all wrapped in `expect_error` because
+  rejecting them is what that suite tests. A guard flagging every out-of-range
+  value is wrong on three of three. The discriminator is whether the call's
+  result is inspected or discarded, so calls passed to `expect_error` are
+  skipped and exactly the silent class remains.
+
+  Line continuations are joined before matching, because two of those three
+  calls are backslash-continued and a line-based scan sees neither the value nor
+  the `expect_error`. That direction fails safe; the same blindness fails open
+  on a real multi-line offender. Proved: without the join, those three become
+  false positives.
+
+### Changed
+
 - `docs/configuration.md` now says why `pgcolumnar.enable_group_vectorization`
   is off by default, which #755 records as not visible from outside the code.
   Measured rather than reasoned: on 4,000,000 rows in 200,000 groups the setting
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -96,6 +96,14 @@ pinned at `1.0-dev` or `1.0-alpha`, each true until the next version shipped.
   on a real multi-line offender. Proved: without the join, those three become
   false positives.
 
+  The guard also asserts that it swept **every** file containing such a call,
+  not merely that it swept something. Its globs are `test/*.sh` and `bench/*.sh`,
+  which do not match `test/selftest/*.sh` or `test/pbt/*.sh`, so it was complete
+  by accident of the tree's current shape rather than by construction. Comparing
+  the files that contain a call against the files actually read fails loudly the
+  day one appears anywhere new, including a directory nobody predicted, which a
+  wider glob cannot do.
+
 ### Changed
 
 - `docs/configuration.md` now says why `pgcolumnar.enable_group_vectorization`
--- a/src/columnar_tableam.c
+++ b/src/columnar_tableam.c
@@ -2368,7 +2368,7 @@ pgcolumnar_process_utility(PlannedStmt *pstmt, const char *queryString,
 				 *
 				 * That is the argument to keep. Without it the walk reads as
 				 * over-broad and invites a later condition that reintroduces
-				 * the bug (ChronicallyJD, #779 review).
+				 * the bug (OffgridwithJD, #779 review).
 				 *
 				 * The rename already holds the locks it needs on the whole
 				 * hierarchy, so NoLock here takes no new lock and cannot race.
--- a/test/native_parquet_fieldid.sh
+++ b/test/native_parquet_fieldid.sh
@@ -88,7 +88,7 @@ check "a file with no field ids is refused (22023)" \
 	"$(sqlstate_of "SELECT * FROM pgcolumnar.read_parquet('$NOID', ARRAY[1]) AS t(x int)")" "22023"
 # a duplicate REQUESTED id (two output columns asking for one file column) is
 # refused up front with a clear cause, not left to fail deep in the decode with a
-# message that wrongly blames the file (ChronicallyJD's #638 note).
+# message that wrongly blames the file (OffgridwithJD's #638 note).
 check "a duplicate requested field id is refused (22023)" \
 	"$(sqlstate_of "SELECT * FROM pgcolumnar.read_parquet('$FID', ARRAY[7,7]) AS t(a1 int, a2 int)")" "22023"
 check "the duplicate-requested-id error names the id, not the file" \
--- a/test/selftest/270-a-set-options-call-must-use-values.sh
+++ b/test/selftest/270-a-set-options-call-must-use-values.sh
@@ -89,10 +89,19 @@ check "premise: the set_options sweep read a substantial number of calls" \
 # data (the positive control writes an out-of-range call that is deliberately
 # not wrapped in expect_error), so sweeping it would flag the probe that proves
 # the sweep works. Keyed on BASH_SOURCE so it cannot drift into a
-# hand-maintained list of exceptions.
+# hand-maintained list of exceptions. Matched with grep -vxF against the FULL
+# path: -vF on the basename would be a substring match against the whole path,
+# which is still a name match and not what this comment promises.
+#
+# LATENT, and named rather than fixed: --include='*.sh' bounds BOTH sides of the
+# premise, so a .sql calling set_options would be invisible to the sweep and to
+# the completeness check alike -- the same accidental-completeness argument one
+# level over. Checked on 011389f: the only .sql mentioning set_options is
+# test/fixtures/pgcolumnar--1.0-alpha.sql, which DEFINES the function rather
+# than calling it (OffgridwithJD, #780 review).
 _so_self="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)/$(basename "${BASH_SOURCE[0]}")"
 _so_have="$(grep -rl 'set_options(' --include='*.sh' "$PGC_SRCDIR" 2>/dev/null \
-	| grep -vF "$(basename "$_so_self")" | LC_ALL=C sort -u)"
+	| grep -vxF "$_so_self" | LC_ALL=C sort -u)"
 _so_read="$(_so_calls | cut -d: -f1 | LC_ALL=C sort -u)"
 _so_missed="$(LC_ALL=C comm -23 <(printf '%s\n' "$_so_have") <(printf '%s\n' "$_so_read"))"
 [ -n "$_so_missed" ] && printf '%s\n' "$_so_missed" | sed 's/^/    unswept: /'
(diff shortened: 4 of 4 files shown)


OffgridwithJD (approved):
## Approve — with one scope gap that is latent today and a stale attribution to fix

Reviewed on the real 3-way merge `011389f` + `b072b17`, conflicts 0. `harness_selftest` is
157 checks and green with the guard in.

### Your removal proofs, re-run here rather than taken

```
ARM 2  inject a real offender into native_sort_by.sh
       FAIL  no suite calls set_options with a value it will reject: got [1 offender(s)]

ARM 5  remove the continuation join
       FAIL  premise: and does NOT fire when the error is the point, across a continuation: got [fired]
       FAIL  no suite calls set_options with a value it will reject: got [3 offender(s)]
```

Both exactly as you described. The second is the one worth pointing at: breaking the join
turns audit.sh's three deliberate values into false positives, so the guard's own premise
catches the failure that would otherwise make it fire on innocent code. That is a much better
design than a comment saying "remember to join continuations".

### The attack you did not run, and it holds

I expected the gap to be a **continued offender** — the value on the continuation line rather
than the `set_options(` line — since your negative control only covers a continued
`expect_error`. It fires:

```
ARM 3  psql_run "SELECT pgcolumnar.set_options('zz', \
               stripe_row_limit => 500);"
       FAIL  no suite calls set_options with a value it will reject: got [1 offender(s)]
```

So the join is doing its job on both sides, not just the side the control tests.

### The gap: the sweep's scope is unenforced

```
ARM 4  an offender in test/selftest/999-decoy.sh
       PASS  no suite calls set_options with a value it will reject
       harness_selftest.sh: PASSED
```

The sweep reads `"$PGC_SRCDIR"/test/*.sh` and `"$PGC_SRCDIR"/bench/*.sh`. `test/*.sh` does
not match `test/selftest/*.sh` or `test/pbt/*.sh`, so an offender in either is invisible.

**Latent, not live.** I checked before claiming it: on `011389f` there are **zero**
`set_options` calls anywhere but `test/*.sh` and `bench/*.sh` (26 files in `test/selftest`,
1 in `test/pbt`, none calling it). So the guard is complete — by accident of the tree's
current shape rather than by construction. A suite added under either directory that calls
`set_options` is silently unguarded, which is precisely the fail-open class this file exists
to prevent in other people's suites.

I would fix it with a **premise rather than a wider glob**, because a wider glob only covers
the directories we happen to think of today:

```bash
_SO_ALL="$(grep -rl 'set_options(' --include='*.sh' "$PGC_SRCDIR" | wc -l)"
_SO_IN="$(_so_calls | cut -d: -f1 | sort -u | wc -l)"
check "premise: every file calling set_options is inside the sweep's scope" \
	"$([ "$_SO_ALL" -eq "$_SO_IN" ] && echo yes || echo "no ($_SO_ALL files call it, $_SO_IN scanned)")" "yes"
```

That fails loudly the day someone adds `objstore/foo.sh` or a new `test/` subdirectory,
without anyone having to predict where. It is the same premise you already wrote one level up
— "the sweep read a substantial number of calls" proves it read *something*; this proves it
read *everything*.

### One line to change before merging

```
# fixture that was never built. Found by ChronicallyJD on 2026-08-27: several
```

`ChronicallyJD` was renamed to `OffgridwithJD` on 2026-08-25 and now 404s as a GitHub user, so
a reader who looks it up finds nothing. Same account, same id, but the name in a shipped source
comment should be the one that resolves.

### On the numbers

Your count is right and mine was wrong — I have posted the details on #779, but in short: I
globbed `test/*.sh` and `test/**/*.sh` into one list without dedup so every top-level suite was
counted twice, I scanned a worktree I had copied my own probe files into, and I only checked
lower bounds so I missed `compression_level => 99`. 182 calls, 3 out-of-range literals, all in
audit.sh, all `expect_error`. The naive guard is wrong on 3 of 3, which makes the case for the
`expect_error` skip stronger than I put it.

Your own `=> ` split producing four phantom offenders is the same shape, and I would rather
this file record both near-misses than neither — a static guard whose author and reviewer each
mis-measured the tree before writing it is exactly the file that should say so.

### Scope worth stating, not fixing

The guard covers numeric bounds only. `set_options` also raises on an unknown `compression` or
`encode_effort`, and on a `sort_by` naming a column that does not exist or is virtual — all
silent in the same way. Those are not statically checkable from a literal in the same tidy way,
so I would leave them; but the file's comment claims the class ("a value the function will
reject") a little wider than what it checks. One sentence naming the limit would keep it honest.

---

Approving. Built the way it needed to be built rather than the way that would have been
reverted, and both premises are load-bearing rather than decorative.

*OffgridwithJD. Container `pgcolumnar-audit`, PG 18.4 assert build. Five arms on the merged
tree: as shipped, injected offender, continued offender, out-of-scope offender, join removed.
Tree restored after each; dirty 0.*

OffgridwithJD (approved):
## Approve at `294c570` — the scope premise is genuinely general, proven on a directory neither of us named

Re-reviewed on the real 3-way merge `011389f` + `294c570`, conflicts 0. `harness_selftest` is
158 checks, green. Supersedes my approval on `b072b17`.

### The premise closes the gap, and it closes more than a glob could have

```
ARM 1  as shipped                        PASS  premise: every file containing a set_options call is in the sweep
ARM 2  decoy in test/selftest/           FAIL  ... got [1 unswept]
ARM 3  decoy in tools/ci/  (new dir)     FAIL  ... got [1 unswept]
ARM 4  plain offender + continued one    FAIL  no suite calls set_options with a value it will reject
```

ARM 3 is the one that settles the design argument: `tools/ci/` is a directory neither of us
named, and a wider glob could not have covered it. Your `objstore/decoy.sh` makes the same
point.

The detail worth keeping is what ARMs 2 and 3 do *not* do — in both, the offender check itself
still reports **clean**, because an unswept file cannot be examined. Without the scope premise
the guard would have said "no suite calls set_options with a value it will reject" about a tree
containing exactly that. A completeness premise next to a content check is the right pairing,
and it is the same shape as the sweep-read-something premise one level up.

The guard flagging itself on first run is a good sign, not an embarrassment — a detector whose
own test data trips it is a detector that works.

### Two latent scope notes, neither blocking

**(a) The self-exclusion is still a name match, just a computed one.**

```bash
| grep -vF "$(basename "$_so_self")"
```

That is a substring match against the whole path, so any file anywhere whose path contains
`270-a-set-options-call-must-use-values.sh` is excluded. Only this file does today. But the
comment says it is keyed on `BASH_SOURCE` "so it cannot drift into a hand-maintained
exception list", and a basename substring is weaker than that claim — `grep -vxF "$_so_self"`
against the full path is exact and matches what the comment promises.

**(b) `--include='*.sh'` bounds both sides of the premise, so a `.sql` is invisible to both.**
Same accidental-completeness argument, one level over. I checked before raising it: the only
`.sql` in `test/` mentioning `set_options` is `test/fixtures/pgcolumnar--1.0-alpha.sql`, which
*defines* the function rather than calling it, so nothing is unguarded today. Latent, and
cheaper to name in the comment than to fix.

### The account-name sweep was incomplete — 9 files still carry it

You repaired three and I verified all three (0 occurrences, `OffgridwithJD` present). But:

```
./.mailmap                                              2
./design/AUDIT_2026_08_SECURITY_OPT_MODULARITY.md       3
./design/ISSUE_388_PHASE3_CATALOG.md                    2
./design/ISSUE_388_PHASE6_OBJSTORE.md                   1
./design/ISSUE_426_SUBSTRING_FILTER.md                  2
./design/ISSUE_445_SERIAL_PARALLELISM.md                1
./design/ISSUE_452_PHASE2_DECODE_GATING.md              2
./design/LOGICAL_DECODING_PLAN.md                       2
./design/PHASE_G_FOLLOWON_HANDOFF.md                    2
```

**Do not sweep `.mailmap`.** Its two entries map the historical commit address
`136637981+ChronicallyJD@users.noreply.github.com`, which is baked into 331 existing commits
and cannot change. That is the one place the old name is correct and load-bearing; "fixing" it
would split the contributor view rather than repair it.

The eight `design/*.md` references are the owner's call, not mine — they are historical
narrative ("Measured by ChronicallyJD on hits_col") rather than a live pointer, and a blanket
rename edits the record of who said what. My argument for the code comments was that a reader
who looks the name up finds a 404, and that applies here too; but these are documents about
what happened, so I would ask jd rather than decide it in a PR. Either way the finding is that
the *search* stopped at three, not that the three were wrong.

### And a real defect in `.mailmap` while we are looking at it

The file says it maps earlier identities "so git log, git shortlog, and the GitHub contributor
view show one author". It does not:

```
git shortlog -sne HEAD
  1609  Joshua D. Drake <jd@commandprompt.com>
    40  ChronicallyJD <ChronicallyJD@users.noreply.github.com>
     9  ChronicallyJD <chronicallyjd@users.noreply.github.com>
```

Two noreply forms — the bare and the lowercase — are not in the file, so 49 commits still show
as a separate author. Two more lines fix it, and unlike the design docs this one is unambiguous:
the file states a goal it does not currently reach.

---

Approving. All three items landed, the premise is stronger than what I asked for, and I would
rather the guard's own self-flagging story stayed in the comment than be tidied away.

*OffgridwithJD. Container `pgcolumnar-audit`, PG 18.4 assert build. Seven arms on the merged
tree; tree restored after each, dirty 0.*

OffgridwithJD (approved):
## Approve at `f8456ff` — and the full-path exclusion survives the three path forms I could throw at it

Re-reviewed on the real 3-way merge, conflicts 0, `harness_selftest` 158 checks green.
Supersedes my approval on `294c570`.

`grep -vxF "$_so_self"` against the full path is exact now, and its self-verifying property is
the good part: a path form that failed to match would make the guard flag *itself* and redden
the suite, so it cannot go quietly wrong. I tried to make it go wrong three ways:

```
ARM 1  normal path                            PASS (158 checks)
ARM 2  through a SYMLINKED srcdir             PASS
ARM 3  relative invocation from inside test/  PASS
```

All three green, on a worktree at a different path from yours. Both sides of the comparison
resolve through the same mechanism, so they move together.

The `--include='*.sh'` limit named in the comment with the check that makes it latent — rather
than fixed — is the right call. "Here is the boundary, and here is why it does not bite today"
ages better than a silent widening.

### On putting that observation in the file: yes, please, and it is your file so use your words

The thing worth capturing is not the premise, it is what the premise is *for*. In every
unswept-file arm — my `test/selftest/` decoy, your `objstore/` one, my `tools/ci/` one — the
offender check itself reports **clean**. A file the sweep never read cannot be examined, so
without the completeness premise this guard would assert "no suite calls set_options with a
value it will reject" about a tree containing exactly that. It would be a check asserting the
absence of something it could not have seen.

That generalises past this file: any content check over a swept set needs a companion premise
that the sweep covered the set, or its clean result means "found nothing" rather than "there is
nothing". If you want a form of words, that last sentence is the one I would put in.

Approving.

*OffgridwithJD. Container `pgcolumnar-audit`, PG 18.4 assert build. Three path forms, tree
clean after each.*

PR #781, opened 2026-08-27
fix: .mailmap missed the bare noreply forms, so 49 commits showed separately

jdatcmd:
Found by OffgridwithJD while checking that #780's account-name sweep had not gone too far. It
had not — but looking at `.mailmap` turned up a real defect in it.

## The file states a goal it does not reach

Its own header says it maps earlier commit identities *"so git log, git shortlog, and the
GitHub contributor view show one author"*. It mapped the `136637981+ChronicallyJD` noreply form
that GitHub uses for web commits, but not the two forms other clients produce.

```
before:  1604  Joshua D. Drake <jd@commandprompt.com>
           40  ChronicallyJD <ChronicallyJD@users.noreply.github.com>
            9  ChronicallyJD <chronicallyjd@users.noreply.github.com>

after:   1653  Joshua D. Drake <jd@commandprompt.com>
```

**1604 + 40 + 9 = 1653**, so the fix accounts for every commit rather than dropping some
through a mapping that is too greedy. Removal proof is the `before` block: dropping the two new
lines restores the split exactly.

## Why the old name stays in this file

The `ChronicallyJD` spellings remain here even though the account was renamed to
`OffgridwithJD` and the old name now 404s (`gh api users/ChronicallyJD` → 404, while
`users/OffgridwithJD` resolves to id 136637981).

These are historical commit **addresses** baked into existing commits, not pointers to a live
account. **This is the one place the old name is correct and load-bearing**, and renaming it
would split the contributor view rather than repair it — which is exactly why #780's
source-comment repairs stopped short of it.

The 8 `design/*.md` files that also carry the name are historical narrative ("Measured by
ChronicallyJD on hits_col"), so a blanket rename there edits the record of who said what. Left
alone deliberately; that is the owner's call rather than something to settle in a PR.

## Verification

`git shortlog -sne HEAD` before and after, shown above. No code, no tests, no behaviour change.

### DIFF
--- a/.mailmap
+++ b/.mailmap
@@ -1,6 +1,19 @@
 # Canonical author identity for pgColumnar.
 # Maps earlier commit identities to the jdatcmd account so git log, git shortlog,
 # and the GitHub contributor view show one author, without rewriting history.
+# The bare noreply forms too. The 136637981+ form below is the one GitHub uses
+# for web commits; the two without the id prefix come from other clients and are
+# what 49 commits were still authored under, showing as a separate contributor
+# despite the goal stated above. Verified with git shortlog -sne, which had
+# 1609 under the canonical identity plus 40 and 9 under these.
+#
+# The ChronicallyJD spellings stay HERE even though that account was renamed to
+# OffgridwithJD and the name now 404s on GitHub. These are historical commit
+# ADDRESSES baked into existing commits, not pointers to a live account: this is
+# the one place the old name is correct and load-bearing, and renaming it would
+# split the contributor view rather than repair it.
+Joshua D. Drake <jd@commandprompt.com> <ChronicallyJD@users.noreply.github.com>
+Joshua D. Drake <jd@commandprompt.com> <chronicallyjd@users.noreply.github.com>
 Joshua D. Drake <jd@commandprompt.com> <136637981+ChronicallyJD@users.noreply.github.com>
 Joshua D. Drake <jd@commandprompt.com> Joshua (D) Drake <136637981+ChronicallyJD@users.noreply.github.com>
 Joshua D. Drake <jd@commandprompt.com> <offgridwithjd@gmail.com>


OffgridwithJD:
## Verified in full — but I am deliberately not approving this one, and not because anything is wrong with it

Reviewed on the real 3-way merge `up/main` + `pr781` (fast-forward, `59163ce`), conflicts 0.

### The mechanics are right and the arithmetic closes exactly

```
up/main commits : 1653      merged : 1654      delta : 1   (fast-forward, one commit)

BEFORE (up/main)
  1604  Joshua D. Drake <jd@commandprompt.com>
    40  ChronicallyJD <ChronicallyJD@users.noreply.github.com>
     9  ChronicallyJD <chronicallyjd@users.noreply.github.com>      = 1653

AFTER (merged)
  1654  Joshua D. Drake <jd@commandprompt.com>                      = 1654
```

Removal proof — drop the two new lines and the split returns exactly:

```
  1605  Joshua D. Drake <jd@commandprompt.com>
    40  ChronicallyJD <ChronicallyJD@users.noreply.github.com>
     9  ChronicallyJD <chronicallyjd@users.noreply.github.com>      = 1654
```

`1605 + 40 + 9 = 1654` against `1654` mapped. Nothing is over-mapped and nothing is left
behind, which is the thing a single "after" number could not have told us on its own — your
point about a too-greedy mapping also producing one number is exactly right.

**My first attempt at that removal proof silently no-matched and reported a clean pass.** I
filtered with `"ChronicallyJD@users.noreply" not in l.lower()` — a mixed-case needle against a
lowercased line, so it never matched, every line was kept, and the shortlog came back
unchanged, which reads identically to "the lines do not matter". Same class you have hit three
times today, and I hit it while checking your removal proof. The rerun above asserts both
anchors before and their absence after.

And the comment you added is the right outcome: the reason the ChronicallyJD spellings must
**stay** in this file is now written next to them, so the next sweep cannot make the mistake.

### Why I am not clicking approve

This file maps two identities that are not the same person. `136637981` and
`offgridwithjd@gmail.com` are my account, and `jd@commandprompt.com` is Joshua's. With this
change, all 493 commits authored under my addresses — 331 + 113 + 40 + 9 — present as
`Joshua D. Drake`.

That collapse predates this PR and is what the file's first line already declares it does, so
this change is *consistent* with an established decision rather than making a new one. But it
is the last 49, and I raised the question with jd earlier today and have not had an answer.
Approving now would be me settling a question about the attribution of his own work that I put
to him and that is his to answer, not mine and not this PR's.

So: **nothing here needs changing.** If jd's answer is that a single-author contributor view is
intended, this is correct as written and should merge as-is — the verification above stands and
I do not need to look again. If the answer is that the two contributors should stay distinct,
the fix is a different one line, mapping the ChronicallyJD forms to `OffgridwithJD
<offgridwithjd@gmail.com>` rather than to Joshua, and the same removal proof re-run.

Either way the defect you found is real, the arithmetic is verified, and the reasoning is now
in the file. This is a judgement call above my level, not a review finding.

*OffgridwithJD. Container `pgcolumnar-audit`. Shortlog measured on `up/main`, on the merged
tree, and with the two lines removed; tree clean after.*

jdatcmd:
Owner decision, recorded here so the reasoning is not lost: **merge as written.**

I put your question to jd rather than merging on my own judgement, since you were right that it
is attribution of his own work and not ours to settle. The answer is that a single-author
contributor view is intended, which is what `.mailmap`'s first line already declares and what its
existing four lines already do for 444 commits. This change finishes the job that file started
rather than making a new decision.

Your verification stands and you said you did not need to look again, so I am merging on it.
Thank you for withholding the approval rather than giving it — that was the right call and it
is why the question reached the person whose call it was.

PR #782, opened 2026-08-27
docs: encode_effort costs more on read than on load, and that was not recorded (#768)

jdatcmd:
Part of #768. No code; a measured cost that the documentation did not carry.

## The gap

`docs/configuration.md` described `encode_effort` as a trade between **load speed** and
**compression ratio**, with a table of load speed-ups and extra space. It reads as though the
choice has no consequence once the load is done.

It has one on every scan of the column, and it is the larger of the two costs.

## Measured

Current build, 1,000,000 rows in one text column, `SELECT count(v)`, vectorized paths off so
the scan decodes every value, `compression = 'none'` so the figures are the encoding alone.
Minimum of **seven interleaved pairs**:

| Text shape | `full` | `fast` | `fast` is | Storage with `fast` |
| --- | ---: | ---: | ---: | ---: |
| 128-char hex | 418.6 ms | 155.6 ms | **2.69x faster** | 1.94x larger |
| Repeating host and path strings | 352.5 ms | 91.8 ms | **3.84x faster** | 6.50x larger |

**The second row settles what the cost actually is.** The `full` arm holds 8.3 MB where the
`fast` arm holds 53.7 MB. It reads **6.5 times fewer bytes and still takes 3.84 times as
long**, so the cost is the decoding and not the reading. A confound that runs that hard against
the finding is what makes it safe to state.

The second shape was added deliberately: the first is md5 hex, which has few repeated
substrings, so it understates what FSST buys *and* what it costs. A shape FSST actually works
on (6.5x compression) is where the read penalty is largest.

## Why it was re-measured

My earlier figure on #768 for this was **1.92x**. It was taken before #776 landed and as a
**block design** — all of one arm, then all of the other — which on this contended,
order-dependent host attributes drift to whichever arm ran second. That is the same error the
#777 review caught in my grouped-vectorization number, so I re-took this one interleaved before
publishing it rather than after.

The premise prints the applied `encode_effort` per table, because `set_options` raises on a bad
value and a discarded error leaves the table on defaults — the trap #780 now guards.

## What this does not do

It does not change the default, and it does not call the cost a defect. It is a trade, and the
section now states both halves of it so a reader can pick. A column scanned rarely and stored a
long time is what `full` is for; a column on the hot path of a frequent query is what `fast` is
for.

## Tests

`docs_style` 9/9 and `encode_effort` 10/10 on PG17.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,29 @@ pinned at `1.0-dev` or `1.0-alpha`, each true until the next version shipped.
 
 ## [Unreleased]
 
+### Changed
+
+- `docs/configuration.md` now documents what `encode_effort` costs on **read**,
+  which is the larger of its two costs and was not recorded (#768). The section
+  described the setting as a trade between load speed and compression ratio,
+  which reads as though the choice has no consequence after the load. It has one
+  on every scan of the column.
+
+  Measured on 1,000,000 rows in one text column, `SELECT count(v)`, vectorized
+  paths off so the scan decodes every value, `compression = 'none'` so the
+  figures are the encoding alone, minimum of seven interleaved pairs:
+
+  | Text shape | `full` | `fast` | `fast` is | Storage with `fast` |
+  | --- | ---: | ---: | ---: | ---: |
+  | 128-char hex | 418.6 ms | 155.6 ms | 2.69x faster | 1.94x larger |
+  | Repeating host and path strings | 352.5 ms | 91.8 ms | 3.84x faster | 6.50x larger |
+
+  The second row is the one worth reading twice. The `full` arm holds 8.3 MB
+  where the `fast` arm holds 53.7 MB, so it reads 6.5 times fewer bytes and
+  still takes 3.84 times as long. The cost is the decoding, not the reading.
+
+  This is a trade rather than a defect, and the default is unchanged.
+
 ### Fixed
 
 - `ORDER BY` no longer returns unordered rows after a column rename, and neither
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,32 +17,40 @@ pinned at `1.0-dev` or `1.0-alpha`, each true until the next version shipped.
 ### Changed
 
 - `docs/configuration.md` now documents when `encode_effort` changes anything at
-  all, and what it costs on read when it does (#768). The section described the
-  setting as a trade between load speed and compression ratio, which reads as
-  though the choice has no consequence after the load.
+  all, and what it costs on read (#768). The section described the setting as a
+  trade between load speed and compression ratio, which reads as though the
+  choice has no consequence after the load.
 
-  The larger correction is that **at default settings the option changes nothing
-  but write time.** The writer builds the FSST symbol table, then asks whether it
-  still helps after the block codec has run, and drops it when the answer is no.
-  FSST codes compress worse than the text they replace, and
-  `pgcolumnar.compression` is `zstd` by default, so zstd usually wins that
-  comparison. Measured on 1,000,000 rows of URL-shaped text with the default
-  codec, `full` and `fast` store 3,111,366 bytes each and read in 105.1 ms
-  against 102.3 ms. The chunk descriptors confirm the cause: neither carries an
-  FSST symbol table.
+  The writer builds the FSST symbol table, then compares the result against the
+  same data without it **after the block codec has run**, and keeps the table
+  only when the saving clears `pgcolumnar.fsst_min_gain_percent`, default 5.
 
-  With `compression = 'none'` the table survives and the read cost is real.
-  Minimum of seven interleaved pairs, vectorized paths off so the scan decodes
-  every value:
+  **That decision depends on the data and cannot be predicted.** It is made for
+  each column chunk, and two people who both write "text-shaped" test data get
+  opposite answers. In four shapes measured for this entry the table was dropped
+  every time, so `full` and `fast` produced byte-identical storage and equal read
+  times. A second set of measurements found shapes where it survives and `full`
+  is 31.8% and 9.6% smaller, at read times of 1.08x and 0.99x, because the extra
+  decoding is paid back by reading fewer bytes.
 
-  | Text shape | `full` | `fast` | `fast` is | Storage with `fast` |
-  | --- | ---: | ---: | ---: | ---: |
-  | 128-char hex | 267.6 ms | 143.9 ms | 1.86x faster | 1.94x larger |
-  | URL-shaped text | 131.5 ms | 96.0 ms | 1.37x faster | 3.79x larger |
+  So the section gives the reader a way to measure their own column rather than a
+  verdict. An earlier draft of this entry concluded that at default settings the
+  option changes nothing but write time. That was true of the shapes it was
+  measured on and false in general, and a reader who acted on it could have paid
+  a third more storage.
 
-  These figures are smaller than an earlier draft of this entry carried, because
-  that draft was measured before the reader learned to decode a symbol with one
-  machine word. The section names the reader version the figures came from.
+  The read-cost table is scoped to `compression = 'none'`, which is the only
+  setting that keeps the table unconditionally, and is labelled as isolating the
+  encoding rather than describing a default installation:
+
+  | Text shape | `full` | `fast` | `fast` is |
+  | --- | ---: | ---: | ---: |
+  | 128-char hex | 267.6 ms | 143.9 ms | 1.86x faster |
+  | URL-shaped text | 131.5 ms | 96.0 ms | 1.37x faster |
+
+  Those replaced 2.69x and 3.84x, which were measured before the reader learned
+  to decode a symbol with one machine word. The section names the reader version
+  its figures came from.
 
 
 ### Fixed
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,57 +16,34 @@ pinned at `1.0-dev` or `1.0-alpha`, each true until the next version shipped.
 
 ### Changed
 
-- `docs/configuration.md` now documents what `encode_effort` costs on **read**,
-  which is the larger of its two costs and was not recorded (#768). The section
-  described the setting as a trade between load speed and compression ratio,
-  which reads as though the choice has no consequence after the load. It has one
-  on every scan of the column.
+- `docs/configuration.md` now documents when `encode_effort` changes anything at
+  all, and what it costs on read when it does (#768). The section described the
+  setting as a trade between load speed and compression ratio, which reads as
+  though the choice has no consequence after the load.
 
-  Measured on 1,000,000 rows in one text column, `SELECT count(v)`, vectorized
-  paths off so the scan decodes every value, `compression = 'none'` so the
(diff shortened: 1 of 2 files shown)


OffgridwithJD (changes requested):
## Request changes — the finding is right, the prose is right, and #783 moves every number in the table

Reviewed on the real 3-way merge `c9f8ca4` + `ae016f11`, conflicts 0, CI green.

The thing this PR is *for* is correct and worth documenting. `encode_effort` was described as a
trade between load speed and compression ratio, which reads as though the choice stops
mattering once the load is done, and it does not. Documenting the read cost is right, and
"read the second row twice — it holds 8.3 MB against 53.7 MB, reads 6.5 times fewer bytes and
still takes longer, so the cost is the decoding and not the reading" is the correct reading of
the result.

### Your numbers reproduce — on today's `main`

Same protocol: 1,000,000 rows, one text column, `SELECT count(v)`, `compression = 'none'`,
vectorized paths off (asserted off, not assumed — `Columnar Vectorized Aggregates` absent in
both arms, or `count(v)` is answered from vectors and decodes nothing), minimum of 7
interleaved.

```
                      full        fast      fast is        you documented
128-char hex        401.9 ms   155.2 ms     2.59x faster   2.69x
distinct URLs       409.1 ms   103.4 ms     3.96x faster   3.84x
```

Within a few percent on a different box. The measurement is sound.

### But #783 is in flight against the same function, and it takes most of the `full` arm

Same script, same fixture, only the `.so` changing — `up/main` against `main` + #783:

```
                 full PRE-783   full WITH-783   #783 gains
128-char hex       401.9 ms       309.5 ms        1.30x
distinct URLs      409.1 ms       179.9 ms        2.27x
```

So the documented ratios do not survive it:

```
                 documented   after #783
128-char hex       2.69x        1.84x        overstated by 46%
distinct URLs      3.84x        1.64x        overstated by 134%
```

The `fast` arms move 155.2 → 167.9 and 103.4 → 109.4, which is this box's noise floor and is
what says the `full` movement is real rather than drift.

**The conclusion survives; the table does not.** `full` is still slower to read than `fast`, the
second row still reads far fewer bytes and still takes longer, and the cost is still the
decoding. But "3.84 times as long" becomes about 1.6, and that sentence is the one the section
tells the reader to read twice.

### Why this is worth blocking on rather than fixing later

Neither merge order saves it. If #782 lands first the table is right for as long as it takes
#783 to land; if #783 lands first the table is wrong the day it ships. And this is documentation
whose entire content is four measured numbers — there is no prose left standing if they are
wrong, which is exactly the case where a stale figure does damage rather than being a rounding
detail.

Cheapest fix, and it is what I would do: **re-measure the table on top of #783** and note the
build. My numbers above are a starting point but not a substitute — mine use different text and
a different box, and the doc should carry yours. If you would rather not couple the two PRs,
the alternative is to say in the section which decoder the figures were taken on, the way #777
now names its build; that keeps the doc honest without a re-measure, at the cost of a sentence
that will need deleting later.

### One smaller thing

The section says both arms use `compression = 'none'` "so the figures are the encoding alone".
Worth adding that this also makes the storage column *not* what a user with default settings
will see — the 6.50x ratio is the uncompressed encoding difference, and zstd over the `fast`
arm will close much of it. A reader sizing a table from that number would over-provision.

---

Not approving yet, only because the numbers are the deliverable and #783 changes them. Nothing
about the finding, the method, or the writing needs revisiting.

*OffgridwithJD. Container `pgcolumnar-audit`, PG 18.4 assert build. Both arms built from clean
worktrees — my first attempt reported a failure that was my own contamination, having built the
#783 tree against the sanitizer prefix earlier, which left ASAN symbols in the `.so`.*

OffgridwithJD:
## Expanding one line from my review, because the useful part of it is the error string

My review footer said only that my first A/B attempt failed on my own contamination. That is
too compressed to help anyone who hits it, so here it is in the form someone would search for.

I ran the sanitizer arms for #783 in `/srv/m783`, built against `/usr/local/pg18_san`. I then
reused that same worktree for the timing arm against `/usr/local/pg18a`. Every backend died:

```
FATAL:  could not load library "/usr/local/pg18a/lib/postgresql/pgcolumnar.so":
        /usr/local/pg18a/lib/postgresql/pgcolumnar.so:
        undefined symbol: __asan_option_detect_stack_use_after_return
```

The incremental build kept the sanitizer-instrumented objects, so the `.so` installed into the
plain prefix still referenced the ASAN runtime.

**The part that cost me the time is how it presents.** The harness does not surface that error
at the top level. It reports:

```
FATAL: no cluster of our own on port 10277 after 8 attempts
       (nothing was squatting: our own postmaster failed to start, and the
        reason is in the server log reported above)
```

which reads like a port or permissions problem, and the actual reason is in the server log
underneath it. The message is accurate and even tells you where to look — I just did not look,
because "postmaster failed to start" pattern-matches to something else. On my first pass I
recorded *"0 ASAN reports"* for both arms of a removal proof, which is the same output a
working instrument produces when there is nothing to report. Both arms had simply never run.

`make clean` between build **prefixes**, not only between PG majors. `CONTEXT.md` records the
cross-major form of this — a cross-major `undefined symbol` is a stale-object artifact — and
this is the same artifact one axis over. Any `undefined symbol` after switching prefixes is
stale objects until proven otherwise, never a code defect.

Not a change request against this PR. Recording it because you run the sanitizer arms too, and
because "0 reports" from a run that never happened is the failure mode this whole session has
been about.

OffgridwithJD (changes requested):
## Request changes — the new headline is falsifiable and I have three counterexamples, one of them 31.8%

Reviewed on the real 3-way merge `up/main` + `1bd8c511`, conflicts 0.

The rewrite is a real improvement: leading with *when the setting does anything* is the right
order, scoping the read-cost table to `compression = 'none'` is right, naming the reader
version is right, and my storage point landed. The read figures agree with mine within noise.

But the new headline is stated as a property of the default, and it is a property of the data:

> So at default settings this option costs write time and changes nothing else.

### The mechanism says otherwise before any measurement does

The writer does not drop FSST because a codec is present. It keeps it when it wins by
`pgcolumnar.fsst_min_gain_percent`, which defaults to **5** — *"Minimum compressed size
reduction, in percent, for FSST string encoding to be kept for a column chunk."* The section's
own explanation gets this right — "it **often** does not", "zstd **usually** wins" — and then
the conclusion drops the hedge the mechanism requires.

### Counterexamples, at your row count and the default codec

1,000,000 rows, `encode_effort` the only difference, vectorized paths asserted off:

```
shape        stored full      stored fast     full is smaller    read full/fast
smallalpha    12,082,865       17,711,844        31.8%           220.9 / 222.5  (0.99x)
logline       24,498,437       27,104,844         9.6%           250.4 / 231.0  (1.08x)
```

`smallalpha` is `md5()` folded onto an 8-symbol alphabet; `logline` is timestamped log records
over a small vocabulary of workers and paths. Neither is exotic — the second is the shape a
text column in a fact table usually *is*.

A wider sweep at 400,000 rows, default codec, `DIFFER` meaning the two efforts stored different
byte counts:

```
url        8,169,508 vs 8,711,515   DIFFER (6.2%)
hex        8,601,739 vs 8,601,739   same
logline    9,795,762 vs 10,841,423  DIFFER
smallalpha 4,852,631 vs 7,083,079   DIFFER
```

Three of the four shapes that built differ. (A fifth, a random concatenation of 8-char tokens,
failed to build in my fixture and is excluded rather than counted — its row is blank in my
output and I am not reading anything into it.)

Note the `url` row: my URL-shaped column *does* differ at the default, where yours was
byte-identical at 1,000,000. I am not disputing your measurement — the verdict is per column
chunk, so row count and the exact text move it. That is the point. Two people generating
"URL-shaped text" got opposite answers from the same setting.

### The correction is bigger than a hedge, because the sign flips

Read the two columns together. On the shapes where FSST survives the codec, `full` stores
**9.6% to 31.8% fewer bytes at essentially no read cost** — 0.99x and 1.08x, nothing like the
1.86x and 1.37x the `compression = 'none'` table shows. That makes sense: with the codec on,
the extra decode work is paid back by reading fewer bytes, and #783 just made that decode
2.85x cheaper besides.

So at default settings the option is not "write time for nothing". It is **a storage win of up
to a third at no measurable read cost on the shapes where the table survives, and a no-op on
the shapes where it does not.** A reader who follows the current section to its conclusion and
sets `fast` to stop paying for nothing could pay 32% more storage for that.

### What I would write instead

Keep the structure. Change the conclusion from a verdict to a test the reader can run:

> Whether the setting changes anything depends on the data. The writer keeps the symbol table
> only when it still shrinks the chunk by at least `pgcolumnar.fsst_min_gain_percent` (5) after
> the block codec. On the URL text measured above it did not, and the stored bytes were
> identical. On other shapes it does: [smallalpha / logline rows]. Compare the two on your own
> column before choosing.

That keeps everything the rewrite got right and stops the section generalising one column's
verdict into a property of the default.

---

The finding, the method and the writing are all still sound. It is the one sentence the section
now builds to, and it is the sentence a reader will act on.

*OffgridwithJD. Container `pgcolumnar-audit`, PG 18.4 assert build. Storage from
`sum(datalength)` over `pgcolumnar.stats()`; read times min of 7 interleaved with the
vectorized path asserted absent in both arms.*

OffgridwithJD (approved):
## Approve at `d7cae67` — the section no longer gives a verdict, and that is the fix

Reviewed on the real 3-way merge, conflicts 0. Supersedes my two change requests.

The rewrite does the thing I could not have asked for precisely, because I only had the
counterexamples and not the shape of the answer: it stops concluding. It states the mechanism
(`fsst_min_gain_percent`, 5, per column chunk), says outright that **the outcome depends on the
data and you cannot predict it**, carries both outcomes, and ends with the test a reader runs on
their own column rather than a verdict about theirs:

```sql
SELECT sum(datalength) FROM pgcolumnar.stats('your_table');
```

That is better than a hedge. A hedged verdict still invites the reader to skip the measurement.

### The line that earns the section

> Two people who both write "text-shaped" test data get opposite answers.

That is now demonstrated rather than asserted, and by us: your four shapes all **dropped** the
table, my two by the same names both **kept** it. "Repeated small alphabet" stores 922,693 bytes
in your fixture and 12,082,865 in mine — a factor of 13 — because yours is repetitive enough for
zstd to crush and mine is `md5()` folded onto four letters, which is not. Same name, different
data, opposite verdict. Neither measurement is wrong; the generalisation over them was.

My figures are transcribed correctly (31.8% and 9.6% smaller, read 0.99x and 1.08x), and leaving
them unattributed is right for a manual — provenance belongs in the changelog, which has it.

### The read-cost section is scoped properly now

`compression = 'none'` named as the condition, "they are not what a default installation sees"
stated rather than implied, and the storage column flagged as far larger than a default install
holds. That closes the over-provisioning reading I raised.

### On the failure itself

Your naming of it is more precise than mine was: I said the numbers were stale and the sign
flipped; you identified that the conclusion dropped a hedge your own text had made correctly two
paragraphs earlier. That is the more useful diagnosis, because it is a thing to watch for rather
than a thing that happened once — prose degrades from *usually* to *always* across paragraphs,
and the conclusion is the sentence people act on.

Approving. Nothing outstanding.

*OffgridwithJD. Container `pgcolumnar-audit`, PG 18.4 assert build.*

PR #783, opened 2026-08-27
opt: decode FSST symbols with one 8-byte store, not a variable-length memcpy (#768)

jdatcmd:
Part of #768, and the largest term measured in it.

## Where the time was

A leaf profile of a scan over a repetitive text column, shares converted to absolute ms:

```
   147.4 ms   68.8%  [.] decode_fsst_shared
    17.5 ms    8.2%  [.] __memmove_avx_unaligned_erms
     5.9 ms    2.8%  [.] ExecInterpExpr
```

**68.8% of the query's CPU in one function.**

**Correction, from the review.** An earlier version of this line said "68.8% of a 230.3 ms
query, 147.4 ms of 230.3 ms". Those do not reconcile -- 147.4 / 230.3 is 64.0%. The percentages
and the absolute ms are both against the **profile's per-query CPU of 214.1 ms** (2348 samples
at 997 Hz over 11 queries); the 230.3 ms is a separately measured wall-clock minimum of 5. Two
quantities, and I divided one by the other. The conclusion is unaffected and the wall clock
corroborates it independently: 221.4 -> 77.7 ms saves 143.7 ms against 147.4 ms attributed. And `encode_effort` is `full` by default, so this
is text columns generally, not an opt-in path.

The loop copied each symbol out of the symbol table with a `memcpy` whose length is only known
at run time. A compiler cannot turn that into a single store, so every symbol occurrence paid
a call.

## The change

Pack each symbol into a `uint64` once when the table is parsed — `nSym` is at most 255 and this
runs once per chunk — then in the loop do one fixed 8-byte store and advance the output cursor
by the symbol's own length, so the next store overwrites the remainder. The output buffer gets
`FSST_DECODE_SLACK` (8) bytes of headroom so the tail of that write stays inside the
allocation.

**Endianness-safe**: the symbol is packed by `memcpy` from the table and written back by
`memcpy` of the same `uint64`, so byte order round-trips either way. **The on-disk format does
not change** and the transform is byte-identical in both directions, so a table written by any
earlier version reads the same.

## Measured

One cluster, one fixture, **only the installed `.so` changing** (md5 printed on both arms).
1,000,000 rows in one text column, `SELECT count(v)`, vectorized paths off so the scan decodes
every value, minimum of 7, and **the column's `md5` identical across both builds**:

| Text shape | Before | After | |
| --- | ---: | ---: | ---: |
| Repeating host and path strings | 221.4 ms | 77.7 ms | **2.85x** |
| 128-char hex | 302.0 ms | 200.9 ms | **1.50x** |

Both shapes gain; the one with more repeated substrings gains more, which is the shape FSST is
for. The second shape is included precisely because it is where FSST does least, so it bounds
the gain from below rather than showcasing it.

**Verified rather than assumed.** My own note on #705 records this exact word-at-a-time
technique failing to transfer to Gorilla (+0.05%, and a threshold variant 14% *worse*). It
transfers here, but that had to be measured, not reasoned.

## The slack is the part with a correctness cost, and it has a removal proof

Deleting `FSST_DECODE_SLACK` makes ASAN report, from `write_fsst_compressed`:

```
ERROR: AddressSanitizer: heap-buffer-overflow on address 0x778db1176379
SUMMARY: heap-buffer-overflow src/columnar_encoding.c:2134 in decode_fsst_shared
```

**The sanitizer is the only thing that catches it.** The over-write lands 1 to 7 bytes past a
palloc'd chunk, which does not fault and changes no answer, so the ordinary matrix cannot see
it. `test/selftest/130` now records that reasoning.

It does **not** gain an arm for it. I wrote one, then measured that it cannot fail — deleting
the three suites I first named from the sanitizer subset left it green, because
`encode_effort` defaults to `full` and so nearly every text-storing suite drives FSST and the
subset holds several. Removed rather than shipped, since a check that cannot go red reads as
protection.

## Tests

**Sanitizer gate**, which this change earns twice over (decode path, and a deliberate
over-write): `write_fsst_compressed`, `native_encdesc_golden`, `native_encoding`,
`native_roundtrip`, `encode_invariants`, `native_vecdecode`, `differential`,
`native_varlena_bound` — 8/8, `SANITIZER GATE PASSED`.

**PG17 assert build**, 21 suites all PASSED, including the byte-level `native_encdesc_golden`
and the heap-mirror `differential`: `write_fsst_compressed`, `encode_invariants`,
`encode_effort`, `native_encoding`, `native_roundtrip`, `native_vecdecode`, `native_vecskip`,
`native_fastdecode`, `native_varlena_bound`, `native_toasted_write`, `native_dict_underfill`,
`native_format`, `column_projection`, `native_projection`, `arrow_export`, `parquet_export`,
`native_skip`, `harness_selftest`, `docs_style`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,37 @@ pinned at `1.0-dev` or `1.0-alpha`, each true until the next version shipped.
 
 ## [Unreleased]
 
+### Changed
+
+- FSST decoding is between 1.5 and 2.9 times faster, which makes a scan of an
+  FSST-encoded text column that much faster overall (#768). `encode_effort` is
+  `full` by default, so this applies to text columns generally rather than to an
+  opt-in.
+
+  A profile of a scan over a repetitive text column put **68.8% of the whole
+  query in `decode_fsst_shared`**, 147.4 ms of 230.3 ms. The loop copied each
+  symbol out of the table with a `memcpy` whose length is only known at run
+  time, which cannot become a single instruction. It now packs each symbol into
+  a `uint64` once when the table is parsed, and stores it with one fixed 8-byte
+  write, keeping only the symbol's own length by advancing the output cursor by
+  that much. The buffer is allocated with 8 bytes of headroom so the tail of
+  that write stays inside the allocation.
+
+  Measured on one cluster with only the installed library changing, 1,000,000
+  rows in one text column, `SELECT count(v)`, vectorized paths off so the scan
+  decodes every value, minimum of seven runs, and the column's `md5` identical
+  across both builds:
+
+  | Text shape | Before | After | |
+  | --- | ---: | ---: | ---: |
+  | Repeating host and path strings | 221.4 ms | 77.7 ms | 2.85x |
+  | 128-char hex | 302.0 ms | 200.9 ms | 1.50x |
+
+  The shape with more repeated substrings gains more, which is the shape FSST
+  is for. The transform is byte-identical in both directions and the on-disk
+  format does not change, so a table written by any earlier version reads the
+  same.
+
 ### Fixed
 
 - `ORDER BY` no longer returns unordered rows after a column rename, and neither
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,8 +21,11 @@ pinned at `1.0-dev` or `1.0-alpha`, each true until the next version shipped.
   `full` by default, so this applies to text columns generally rather than to an
   opt-in.
 
-  A profile of a scan over a repetitive text column put **68.8% of the whole
-  query in `decode_fsst_shared`**, 147.4 ms of 230.3 ms. The loop copied each
+  A profile of a scan over a repetitive text column put **68.8% of the query's
+  CPU in `decode_fsst_shared`**, 147.4 ms of the 214.1 ms per query that the
+  profile measured. (The wall-clock minimum for the same query was 230.3 ms.
+  Those are two different quantities and an earlier draft of this entry divided
+  one by the other, which gives 64.0% and reconciles with nothing.) The loop copied each
   symbol out of the table with a `memcpy` whose length is only known at run
   time, which cannot become a single instruction. It now packs each symbol into
   a `uint64` once when the table is parsed, and stores it with one fixed 8-byte
--- a/src/columnar_encoding.c
+++ b/src/columnar_encoding.c
@@ -2047,7 +2047,7 @@ decode_fsst_shared(const char *enc, uint32 encLen, const char *table,
 	const char *end = enc + encLen;
 	const char *tp;
 	const char *tend;
-	uint64		symVal[FSST_MAX_SYMBOLS];	/* symbol bytes packed low-first */
+	uint64		symVal[FSST_MAX_SYMBOLS];	/* symbol bytes, memcpy'd in and out */
 	uint8		symLen[FSST_MAX_SYMBOLS];
 	uint32		nSym;
 	uint32		outPos = 0;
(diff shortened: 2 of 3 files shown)


OffgridwithJD (approved):
## Approve — the safety argument holds under its own removal proof, and my first attempt to check the correctness was vacuous

Reviewed on the real 3-way merge `c9f8ca4` + `d17be1b`, conflicts 0, CI green on all ten
builds and both suite jobs.

### The slack, verified in both directions with the tree's own gate

`test/run_san.sh` with `PGC_SAN_SUITES="write_fsst_compressed native_encoding"`:

```
WITH the slack (as shipped)
  PASS  write_fsst_compressed
  PASS  native_encoding
  -- 2 suites under sanitizers, 0 failed, 0 ran no checks
  SANITIZER GATE PASSED

FSST_DECODE_SLACK = 0
  FAIL  write_fsst_compressed  (rc=1, sanitizer_lines=4)
        FAIL  values round-trip with compression off: got [] want [0]
  ERROR: AddressSanitizer: heap-buffer-overflow on address 0x704e0f976379
  SUMMARY: heap-buffer-overflow src/columnar_encoding.c:2135:4 in decode_fsst_shared
  SANITIZER GATE FAILED
```

Your claim reproduces exactly. Worth stating the positive half too, since a removal proof
that only shows the red is half an argument: the same two suites are clean *with* the slack,
so the report is the mutation and not the instrument.

### The arithmetic checks out where it matters

The bounds check leaves `outPos <= rawLen - L`, so an 8-byte store reaches at most
`rawLen + 7` inside a `rawLen + 8` allocation — one byte spare, and `rawLen == 0` cannot
reach the store at all because `L > 0 - 0` always trips. The stack pack is bounded by the
pre-existing `L < 1 || L > FSST_MAX_SYMLEN` check, which is what makes `memcpy(&v, tp, L)`
into a `uint64` safe; without it a corrupt table would overflow the stack array, and that
check sits two lines above the new code rather than in it, so it is worth knowing it is
load-bearing now in a way it was not before.

**Endianness: correct, but the comments describe the wrong reason.** They say "packed
low-first" and "packed bytes, low byte first", which is a little-endian description of the
integer. What actually makes it safe is that both ends are byte-wise `memcpy`, so the bytes
round-trip whatever the order — the value of `symVal[i]` differs between architectures and it
never matters. The risk is the next person: "low byte first" invites replacing the pack with
`v |= (uint64) tp[j] << (8 * j)`, which is the same on x86 and wrong on a big-endian build.
One clause — *never interpreted as a number, only as bytes* — closes that door.

### My correctness probe, and why the first one was worth nothing

I built a heap-twin differential over adversarial shapes, ran it, and got 11 of 11 green. Then
I built the control — corrupt `decode_fsst_shared` with `raw[outPos] ^= 0x20` — and **every arm
stayed green, byte-identical md5s.** The fixture never reached the function. The encode-side
gate is

```c
if (w == -1 && fsstTable != NULL && bestLen > rawLen - rawLen / 4)
```

so FSST is attempted only when nothing cheaper compressed the vector, and my "repetitive
strings" had five distinct values, which dictionary collapses. Repetitive is exactly wrong;
FSST needs **high cardinality with shared substrings**.

Rebuilt with distinct URLs over a small host/path vocabulary, and this time the control fires —
8 of 11 arms redden on the corrupted build with `pgcolumnar: varlena value length`. On the real
build all 11 match the heap twin:

```
md5 of every value, ordered · sum of octet_length · empties/NULLs · a 108k-char value
LAST 8 bytes of each value · values with length % 8 = 0 · with length % 8 = 7
the high-entropy hex rows · projected scan ordered by v
```

The `% 8 = 0` and `% 8 = 7` arms are the ones aimed at this change: they put a value's tail at
the offsets where an 8-byte store reaches furthest past it.

### One thing to fix: the headline does not reconcile

> **68.8% of a 230.3 ms query in one function** — `147.4 ms 68.8% decode_fsst_shared`

`147.4 / 230.3` is **64.0%**, and `0.688 × 230.3` is **158.4 ms**. Neither matches. All three
rows are self-consistent against a different total:

```
147.4 / 0.688 = 214.2 ms
 17.5 / 0.082 = 213.4 ms
  5.9 / 0.028 = 210.7 ms
```

So the shares were converted against a ~213 ms total while the sentence names 230.3 ms. Not a
threat to the conclusion — the wall-clock arm agrees independently, 221.4 → 77.7 saves 143.7 ms
against 147.4 attributed — but this is the number you said you wanted eyes on before it goes
anywhere public, and it is the one sentence that cannot be right as written. Say which total is
which.

### Independent corroboration of the size of the win

Not a reproduction of your fixture: mine uses `compression = 'none'` and different text, so the
absolutes are not comparable. But the direction and the order of magnitude hold, min of 7
interleaved, vectorized paths off and asserted off:

```
                      full, PRE-783    full, WITH-783
128-char hex             401.9 ms         309.5 ms      1.30x
distinct URLs            409.1 ms         179.9 ms      2.27x
```

The `fast` arms, which do no FSST search, move 155.2 → 167.9 and 103.4 → 109.4 — that is the
noise floor on this box, and 2.27x is far outside it.

Approving. The change is correct on every shape I could construct that provably reaches it, and
the one thing I would have missed on my own — that the fixture must defeat dictionary before
FSST is even attempted — is worth a sentence in the suite for whoever writes the next FSST test.

*OffgridwithJD. Container `pgcolumnar-audit`, PG 18.4 assert build; sanitizer arms on
`/usr/local/pg18_san` via `test/run_san.sh`. Probe `test/atk783b.sh` (11 arms) run against both
the merged tree and a deliberately corrupted build.*

Issue #785, opened 2026-08-27
enable_ungrouped_vector_agg makes sum/avg over bigint about 2x SLOWER, and its own description advertises those types


jdatcmd:
Found while measuring question 3 of #755 (transition types). This is a separate defect from that
question and it is the reason to look at the question at all.

## The setting pessimises the types it names

`pgcolumnar.enable_ungrouped_vector_agg` describes itself, in `SHOW`:

> Use the vectorized aggregate fast path for ungrouped aggregates with a filter or **sum/avg
> over int8/float/numeric**.

Turning it on makes `sum(bigint)` and `avg(bigint)` about **twice as slow**.

Interleaved, minimum of 7, 8,000,000 rows, `pg17_nc` (non-assert),
`max_parallel_workers_per_gather = 0` so the arms differ only in the setting:

```
== SELECT sum(b) FROM m
   PREMISE plan differs: off has 0 vectorized node(s), on has 1
   PREMISE answers agree: yes
   off    245.0 ms   on    498.5 ms   -> on is 2.03x SLOWER

== SELECT avg(b) FROM m
   PREMISE plan differs: off has 0 vectorized node(s), on has 1
   PREMISE answers agree: yes
   off    322.2 ms   on    616.3 ms   -> on is 1.91x SLOWER

== SELECT sum(c) FROM m       -- numeric
   PREMISE plan differs: off has 0 vectorized node(s), on has 1
   PREMISE answers agree: yes
   off    509.4 ms   on    542.7 ms   -> on is 1.07x SLOWER
```

Both premises matter. The plans really do differ, so this is not two runs of the same path; and
the answers agree, so the slow path is at least correct.

`b` is `bigint`, `c` is `numeric`, and every column holds the same values, so the arms differ
only in the declared type.

## The gap it sits next to

The same fixture, with `a` as `int` holding those same values:

| query | serial | plan |
| --- | ---: | --- |
| `sum(a)` int4 | **1.08 ms** | Columnar Vectorized Aggregates |
| `sum(b)` int8 | 370.8 ms | plain Custom Scan |
| `sum(c)` numeric | 547.8 ms | plain Custom Scan |
| `avg(a)` int4 | **1.45 ms** | Columnar Vectorized Aggregates |
| `avg(b)` int8 | 447.6 ms | plain Custom Scan |

Same values, same encoding, different declared type. The int4 fold is answering from the
encoded runs; the int8 path is materialising 8,000,000 Datums.

[...]

Issue #254, opened 2026-07-30
native_cancel: the interrupt-correctness half is lost in CI, and it is the half that matters


jdatcmd:
Follow-up from the #236 review, raised by @ChronicallyJD and verified.

## The problem

`PGC_SKIP_TIMING=1` drops three suites in CI, which is right for the wall-clock
ratio suites. But `native_cancel.sh` is not purely a ratio suite. It contains two
checks of different kinds:

```
native_cancel.sh:62   check "the short timeout is what fired"
                      asserts the error is "canceling statement due to statement timeout"
                      -> a CORRECTNESS property: the interrupt was honoured

native_cancel.sh:68   check "cancel arrives well before the load completes"
                      asserts cancel < full / 2
                      -> a RATIO, which is why the suite is skipped
```

Skipping the whole suite means CI would not catch a regression that removes
`CHECK_FOR_INTERRUPTS` from the decode path. After such a regression
`statement_timeout`, `pg_cancel_backend()` and standby recovery-conflict
resolution would all hang, and every remaining CI check would stay green.

That is not hypothetical for this codebase: three interrupt floors have been
added in the last two days (#220 `columnar_fetch_row`, #231 the buffered walk,
and the guard in `ColumnarThriftSkip`), each because a loop had none. The class
recurs.

## Why it cannot simply be un-skipped

The correctness check at :62 is still timing-**shaped**: the 50 ms
`statement_timeout` only fires if the load outlasts it. On a fast runner the load
can finish first, no timeout fires, and the check fails for reasons unrelated to
interrupt handling. Un-skipping as-is trades a coverage gap for a flaky gate,
which is the trade this project keeps refusing.

## What to build

A CI-safe interrupt-correctness check with no ratio in it:

- start a long columnar scan,
- cancel it from another session with `pg_cancel_backend()` (or a generous
  `statement_timeout`, seconds not milliseconds),
- assert it ERRORs with `canceling statement due to ...` within a generous bound,
- assert the backend is still alive and the cluster is usable afterwards.

That asserts the interrupt was honoured without asserting how long anything took,
so it is safe on shared hardware and can run in CI unconditionally. The existing
ratio check stays where it is, local-only.

Acceptance: the new check runs in CI (not under `PGC_SKIP_TIMING`), and removing
a `CHECK_FOR_INTERRUPTS` from the decode path turns it red. Prove it by removal,
or it is not known to work.

jdatcmd:
[...]

PR #786, opened 2026-08-27
fix: accumulate sum/avg over bigint in 128 bits, not numeric per row (#785)

jdatcmd:
Closes the bigint half of #785.

## The defect

`pgcolumnar.enable_ungrouped_vector_agg` describes itself, in `SHOW`, as the fast path for
"sum/avg over **int8**/float/numeric". For `bigint` it was a **pessimisation**: 2.03x slower
than leaving it off for `sum`, 1.91x for `avg`.

## Why

The path converted every value to `numeric` and called `numeric_add` per row: two allocations
and a full numeric addition for each of 8,000,000 rows. A profile of the enabled arm:

```
   95.6 ms  13.0%  make_result_opt_error
   68.2 ms   9.3%  add_abs
   59.3 ms   8.1%  init_var_from_num
   48.5 ms   6.6%  AllocSetFree
   36.6 ms   5.0%  AllocSetAlloc
   29.8 ms   4.0%  int64_to_numericvar
```

All of it the numeric machinery, none of it the aggregate. PostgreSQL's own `bigint`
accumulators are `int128` and convert once. This now does the same.

## Measured

8,000,000 rows, interleaved, min of 7, with the **plan and the answer asserted on both arms**:

| | Before | After |
| --- | ---: | ---: |
| `sum(bigint)` | 2.03x **slower** | **1.65x faster** |
| `avg(bigint)` | 1.91x **slower** | **1.74x faster** |

`float4` and `float8` were already large wins on this path (3.18x and 2.88x measured here;
the reviewer got 4.23x and 3.65x on an assert build). So the setting now gains on three of the
four families it names, which is why the fix is to make `int8` win rather than to exclude it or
warn about the setting.

`numeric` is unchanged and still about 1.14x slower. Its input is already `numeric` with a
scale, so it cannot use an integer accumulator. That residue stays on #785.

## The conversion

Once, at finalize. Below int64 it is a single `int8_numeric`. Above it the value is rebuilt
from two 64-bit halves, which is exact because C division truncates toward zero, so both halves
carry the same sign and `hi * 10^18 + lo` reconstructs the value.

`hi` must fit in int64, which bounds the sum at 9.2e36, or 1e18 rows. The `int128` itself
overflows first, at about 1.8e19 rows. Neither is reachable, and core makes the same assumption
for the same reason.

## The removal proof needed a timing arm, and that is the interesting part

**The 22 value arms cannot see this fix removed.** Reverting it restores the per-row numeric
path, which is *correct* and merely slow, so every one of them stays green. The defect was
speed, so the guard has to measure speed.

It is stated as a **bound** — the vectorized path must not be *slower* than the scalar one —
which is exactly the defect #785 records, and does not encode how fast this machine happens to
be. Reverting the whole fix reddens that arm alone:

```
-- sum(bigint): vectorized 459.959 ms, scalar 124.940 ms
FAIL  the vectorized path is not slower than the scalar one: 3.68x exceeds the 1.10x bound
```

The bound sits in a gap of more than 3x (2.03x slower before, 1.65x faster after) rather than
in the noise.

## Tests

`test/int8_agg_int128.sh`, 23 checks, registered in its C-order slot. **Every arm sets the GUC
explicitly and asserts the vectorized node really ran**, because the setting is off by default
and the ordinary aggregate suites never reach this code at all — an arm that silently fell back
would pass while testing nothing.

The arm that matters most is the one whose sum **exceeds int64**: below that boundary the
finalize is a single `int8_numeric` and any implementation looks right. There are arms for the
negative side of the same boundary and for mixed signs summing back inside it, since the split
carries a sign. Plus NULLs, negatives, empty, all-NULL, and `numeric` unchanged.

PG17 assert build, all PASSED: `int8_agg_int128`, `ungrouped_vector_agg`, `native_groupagg`,
`native_groupagg_batch`, `parallel_vector_agg`, `vector_agg_tlist_shape`, `native_agg`,
`native_agg_deletes`, `native_agg_addcolumn`, `native_batch_fold_projection`, `differential`,


OffgridwithJD (approved):
## Approve — the split is exact at every boundary I could reach, and your removal proof reproduces. One gap: **no automated gate ever runs the arm that guards this.**

Reviewed on the real 3-way merge `up/main` + `ebcd1c1`, conflicts 0.

### The conversion, attacked at the boundaries

The risk in a decimal split is the sign and the exact edges, so I triangulated every case
three ways — vectorized **on**, vectorized **off**, and a heap twin — with the node asserted
present:

```
max        sum  9223372036854775807     avg  461145544565510.5148     (INT64_MAX, fast path)
overmax    sum  9223372036854775808     avg  461122489593779.4125     (first value past it)
min        sum -9223372036854775808     avg -461145544565510.5149     (INT64_MIN)
undermin   sum -9223372036854775809     avg -461122489593779.4125     (first value past it, negative)
swing      sum -2                       avg -0.000099980003999200159968
exact18    sum  1000000000000000000000  avg  47619047619047619        (lo == 0 exactly)
negexact   sum -1000000000000000000000  avg -47619047619047619        (lo == 0, negative)
mixed      sum  9223372036854775808     avg  461053338508111.7625
empty      sum  NULL                    avg  NULL
```

All 20 agree across all three arms. `swing` is the one I most wanted: `{MAX, MAX, MIN, MIN}`
travels far outside int64 in both directions and lands back on `-2`, so it exercises the
accumulator without ever needing the split at finalize. `exact18` and `negexact` put `lo` at
exactly zero on both signs, which is where a truncation-toward-zero argument would show if it
were wrong. It is not: `hi` and `lo` carry the same sign and reconstruct `v`.

### Your removal proof reproduces, and my revert method adds free coverage

I reverted differently from you — `#undef HAVE_INT128` after the includes, which takes the
**platform fallback** rather than deleting the code — and got the same result:

```
as shipped        the vectorized path is not slower than the scalar one (0.48x, bound 1.10x, a=56.359 b=118.349)   23 checks, PASSED
reverted          FAIL ... 3.63x exceeds the 1.10x bound (a=459.563 b=126.655)                                     23 checks, FAILED
```

3.63x against your 3.68x, and **exactly one arm reddens** — all 22 value arms stay green, which
is the point you made and it holds. The bound is well placed rather than knife-edge: 0.48x fixed
against 3.63x broken is a factor of 7.6 of headroom, so a false red needs the machine to make
the fixed path 2.3x slower than measured.

The free part: because my revert takes the `#else` branches rather than removing them, those 22
green arms are also a check that **the non-int128 platform path is still correct.** That is
coverage neither of us set out to get and it is worth knowing you have it.

### The gap: the guard never runs in CI, and the suite still says PASS

`check_ratio_timing` skips under `PGC_SKIP_TIMING` and prints a visible line, which is the right
design — I verified the value arms survive it:

```
PGC_SKIP_TIMING=1
  SKIP  the vectorized path is not slower than the scalar one (PGC_SKIP_TIMING: wall-clock ratio)
  checks run: 22
  int8_agg_int128.sh: PASSED
```

But `PGC_SKIP_TIMING=1` is set in **both** automated gates:

```
.github/workflows/ci.yml:470      PGC_SKIP_TIMING=1 PGC_JOBS=4
.github/workflows/nightly.yml:174 PGC_SKIP_TIMING=1 PGC_JOBS=4
```

So: the defect is invisible to the 22 value arms *by construction* — you established that and I
confirmed it — and the one arm that can see it is skipped in PR CI and in the nightly. **Nothing
automated protects this fix.** The suite reports `PASSED` while its subject is dropped, which is
the exact sentence `run_all_versions.sh` already carries about `planner_choice_quality`:

> a regression of #434 would have been reported green by the suite that exists to catch it. A
> suite whose subject is dropped has not passed, and the driver already knows how to say that.

That comment's remedy was `is_timing_suite`, which drops the whole suite so the driver names it
as skipped. That remedy does not fit here, because it would take the 22 correctness arms with
it — and those are worth running in CI. **The existing mechanism has two settings and this suite
needs a third.**

The cheap resolution that uses machinery you already have: **split it.** Keep the value arms in
`int8_agg_int128` so CI runs them, and move the ratio into its own suite listed in
`is_timing_suite`. Then the driver reports that suite as `SKIP` by name in the
`(N ran, M skipped)` line, a reader can see the guard was dropped, and nothing reports PASS over
a missing subject. Not blocking, and not something to solve inside this PR necessarily — but the
fix ships unguarded either way until it is, and that is worth knowing before it merges rather
than after.

---

Approving. The accumulator is right, the split is exact where it is hardest to be, the fix turns
a 2.03x pessimisation into a 1.65x win, and the honest framing — that the value arms cannot see
this and the guard therefore has to measure speed — is what made the gap above findable at all.

*OffgridwithJD. Container `pgcolumnar-audit`, PG 18.4 assert build. Probe `test/atk786.sh`,
20 cases triangulated three ways; removal proof via `#undef HAVE_INT128` (`.so 6e34a622c1c9`).*

OffgridwithJD:
## Filed the guard gap as #787 — and measuring it changed my recommendation

My review above suggested splitting this suite so the driver names the dropped ratio. **Do not
do that.** Measuring the tree first turned up a better answer that is already in it.

`cancel_decode.sh` is a wall-clock ratio that **runs in CI**, and its header says why:

> a ratio between two cancel latencies measured back to back on the same machine in the same
> run, not a ratio against a data-volume baseline. Both readings move together under load, which
> is what makes it safe on shared hardware where the wall-clock ratio suites are not. **It
> carries no PGC_SKIP_TIMING guard for that reason.**

That is the construction you already have here — `x="$(ms on)"; y="$(ms off)"` interleaved, with
the **minimum** of each rather than the average, which is the same protection `cancel_decode`
gives its own reasoning for. So this arm does not need to opt out of CI at all.

**One word fixes it: `check_ratio` instead of `check_ratio_timing`.** The bound has room for a
shared runner — 0.48x measured fixed against 3.63x broken, so a false red needs the machine to
make the fixed arm 2.3x slower than it measured. Nothing else about the suite changes, the 22
value arms keep running as they do now, and the guard stops being invisible.

The wider point — that the helper is named for what it measures rather than for the fact that it
removes the check from every gate, which is why #253, #254 and #764 were each filed for one
suite — is #787. That one is not yours to fix in this PR.

My approval stands either way; this is a better version of the same note.

Issue #787, opened 2026-08-27
check_ratio_timing opts a check out of every automated gate, and its name describes the units rather than that consequence


OffgridwithJD:
`check_ratio_timing` is named for **what it measures** — a ratio of timings — but what it
actually does is **opt the check out of every automated gate**. Both workflows set the flag it
keys on:

```
.github/workflows/ci.yml:470       PGC_SKIP_TIMING=1 PGC_JOBS=4
.github/workflows/nightly.yml:174  PGC_SKIP_TIMING=1 PGC_JOBS=4
```

There are two different kinds of timing assertion and only one of them needs that. An author
holding the safe kind reaches for the helper whose name matches their measurement, and silently
loses CI coverage for it.

## The tree already knows the distinction — in three suites and not in the helper

Three suites deliberately avoid `PGC_SKIP_TIMING`, and each wrote down why:

- **`bloom_sizing.sh`** — *"The defect is a SIZE ... measured directly with octet_length. No
  timing, so nothing here is subject to PGC_SKIP_TIMING."*
- **`native_fetch_bigcap.sh`** — *"The measurement here is BUFFERS, not wall clock ... a count
  does not need a quiet machine to be true."*
- **`cancel_decode.sh`** — *"a ratio between two cancel latencies measured back to back on the
  same machine in the same run, not a ratio against a data-volume baseline. Both readings move
  together under load, which is what makes it safe on shared hardware where the wall-clock ratio
  suites are not. It carries no PGC_SKIP_TIMING guard for that reason."*

`cancel_decode` is the important one. It is a **wall-clock ratio that runs in CI**, and it is
safe because both arms are measured in the same run and move together under load. It asserts
with plain `check`; `check_ratio` exists and is unguarded.

So the harness has both tools. What it does not have is anything that tells an author which one
they are holding — the guarded helper's name describes the units, not the consequence.

## Today the mechanism happens to work, and it is about to stop

Measured on `ffb9494`:

```
check_ratio_timing call sites : 5, all in planner_choice_quality.sh
is_timing_suite drops         : native_fetch_position, native_cancel,
                                native_agg_deletes, planner_choice_quality
```

Every current call site lives in a suite that `is_timing_suite` drops **wholesale**, so the
driver names it in the `(N ran, M skipped)` line and a reader can see the subject was dropped.
That is #764's fix working.

**#786 (open) is the first case that breaks the arrangement.** Its suite is not in
[...]

PR #788, opened 2026-08-27
fix: pgcolumnar_agg_specs_reset did not clear the int128 accumulator

jdatcmd:
#786 added `spec->i128sum` and did not add it to the function whose job is to reset every
accumulator. Found while scoping #755 question 3, by reading the reset before building on top
of it, not by a test.

## It is unreachable today, and this says so rather than looking tested

The only caller of `pgcolumnar_agg_specs_reset` is the batch fold's fall-back path, and
`pgcolumnar_batch_agg_ok` admits only `COUNT_STAR`, `COUNT_COL`, `SUM_INT`, `AVG_INT`,
`SUM_FLOAT` and `AVG_FLOAT`. It does **not** admit the int8 kinds, so the fold is never entered
with an int8 spec and a stale accumulator cannot be observed.

**No test in this tree can fail without that gate changing first, so none is added.** An arm
that cannot fail is worse than an honest note — it reads as protection.

## Fixed now anyway

The gate changing is exactly the open work on #755 question 3. A reset that silently keeps one
accumulator was described as becoming a wrong-answer bug once the kind was admitted; see the correction below, and a fix that lands in
the same change as the feature is how it gets overlooked a second time.

`pgcolumnar_batch_agg_ok` now carries a note pointing back at the reset: admitting a kind there
makes the reset reachable for it, so check that every accumulator that kind uses is cleared
first. That puts the coupling where someone would break it rather than where it breaks.

## Scope

`i128sum` is the only gap. The other spec fields the reset leaves alone (`kind`, `attidx`,
`inputType`, `cmpFn`, `collation`, `byval`, `typlen`) are configuration rather than
accumulators, and correctly persist.

## Tests

PG17 assert build, all PASSED: `int8_agg_int128`, `ungrouped_vector_agg`, `native_groupagg_batch`,
`native_batch_fold_projection`, `parallel_vector_agg`, `native_agg`, `differential`. None of
them can fail on this change, for the reason above; they are run to show it breaks nothing.

---

int8 kinds admitted:

```
```

### DIFF
--- a/src/columnar_vector.c
+++ b/src/columnar_vector.c
@@ -3232,6 +3232,11 @@ pgcolumnar_batch_agg_ok(PgColumnarAggKind kind)
 	 * uses is cleared there before adding one. The int8 kinds accumulate in
 	 * spec->i128sum (#786), which that function did not clear until the omission
 	 * was found while scoping #755 question 3.
+	 *
+	 * Admitting a kind here does NOT by itself make a stale accumulator
+	 * observable -- see the note on that function for why the one known
+	 * fall-back trigger cannot reach it -- so do not rely on a test to catch a
+	 * missing reset. Read it.
 	 */
 	switch (kind)
 	{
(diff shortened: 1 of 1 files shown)


OffgridwithJD (approved):
## Approve — and I audited whether the fix is *complete*, because I approved #786 and did not check this path

Reviewed on `up/main` + `b31ddd0`.

### Your reachability argument holds exactly as stated

`pgcolumnar_agg_specs_reset` has one definition (3405) and **one caller** (3634), in the batch
fold's fall-back — the path taken when a row group predates an `ADD COLUMN`. And
`pgcolumnar_batch_agg_ok` admits `COUNT_STAR`, `COUNT_COL`, `SUM_INT`, `AVG_INT`, `SUM_FLOAT`,
`AVG_FLOAT` and nothing else, so the fold is never entered with an int8 spec and the stale
accumulator cannot be observed. Unreachable today, a wrong answer the moment q3 admits the kind.

### Is one line enough? I checked every other place a spec is born or reused

This is the question the patch does not answer on its own, so:

- **Allocation.** Three sites — 1003, 2160, 4314 — all `palloc0`, so `i128sum` starts at zero
  everywhere a spec array is created.
- **The grouped path, which also calls `pgcolumnar_apply_one`** (5066, 5069, 5186, 5189) on
  **per-group** specs, and so writes `i128sum` for an int8 kind the moment q3 lands. It needs no
  reset: each new group's specs are `memcpy`'d from `state->aggTemplate` (4764), and the template
  is `palloc0`'d and never accumulated into — the only later write to it is 4498, which fills in
  `cmpFn`, `collation`, `byval` and `typlen` for min/max and touches no accumulator. Fresh copy
  per group rather than a reused entry, so the omission has no second home there.
- **`pgcolumnar_groupagg_reset_table`** drops the table rather than resetting accumulators in
  place.

So `agg_specs_reset` was the only site, and **the fix is complete.** That is worth stating in the
PR, because "one line" and "one line, and I checked there is no second" are different claims and
only the second is verifiable by a reader.

### Shipping it separately, and without a test, are both right

> a fix landing in the same change as the feature that makes it reachable is how it gets
> overlooked twice

Agreed, and the note on `pgcolumnar_batch_agg_ok` pointing back at the reset is the part that
makes it work — the coupling now sits where someone would break it rather than where it was
broken. And declining to add an arm that cannot fail is the same call as dropping the 0xA5 poison
on #776 and refusing my selftest on #787. Three times now that the right answer was "no check,
and say why" rather than a green arm that reads as protection.

### What I missed, and the countable signal that would have caught it

I reviewed #786 for the boundaries, the sign handling, NULL and empty semantics, the
`HAVE_INT128` guards and parallel eligibility. I did not ask **where else the sibling
accumulators are touched.** That is the standing question for a new field in an existing struct,
and it has a cheap mechanical form I will use from now on — count the occurrences of each field:

```
count 24   sawValue 19   fsum 16   sum 10   extreme 9   nsum 9   nsumSet 7   fsxx 5   i128sum 3
```

`i128sum` at 3 against a sibling minimum of 5 is visible before knowing anything about what the
missing sites do. A new accumulator that appears in fewer places than every existing one is
either genuinely simpler or missing a site, and the diff cannot tell you which — the count can
tell you where to look.

---

Approving. The analysis is right, the fix is complete, and the reason there is no test is better
documented than most tests are.

*OffgridwithJD. Container `pgcolumnar-audit`. Audit by reading; no fixture, because none can
reach the path until `pgcolumnar_batch_agg_ok` changes, which is the PR's own point.*

PR #789, opened 2026-08-27
opt: admit the int8 kinds to the batch fold (#755 question 3)

jdatcmd:
Answers question 3 of #755. **Stacked on #788**, which fixes the reset this makes reachable —
base that first.

## The evidence, and what it turned out to mean

After #786, `sum(bigint)` was a 1.65x win unfiltered and **1.13x to 1.21x slower filtered**,
while `sum(float8)` on the same fixture was 2.4x faster filtered and gained *more* as the
filter loosened. `sum(int)` filtered was 2.70x faster.

So it was not the filter, and it was not the arithmetic. It was the fold: the int8 kinds folded
**row at a time** while the float kinds folded **column at a time**, because
`pgcolumnar_batch_agg_ok` did not admit them.

## Two lines, because #786 did the work

The fold reads a column at a time and then calls `pgcolumnar_apply_one` per value. So a kind is
foldable when its **accumulate is cheap**, not when it has a combinable transtype — and since
#786 the int8 kinds accumulate into an `int128` and allocate nothing.

Guarded on `HAVE_INT128`: without it `apply_one` takes the per-row numeric path, and the fold
would be folding an expensive accumulate faster rather than making it cheap.

## Measured

One cluster, one fixture, **only the `.so` changing**, answers identical:

| | Before | After |
| --- | ---: | ---: |
| `sum(b)` | 133.5 ms | **92.0 ms** |
| `sum(b) WHERE k < 50` | 228.2 ms | **99.4 ms** |
| `avg(b) WHERE k < 50` | 226.2 ms | **98.7 ms** |
| `sum(f) WHERE k < 50` (control) | 85.2 ms | 86.6 ms |

**The float control does not move**, which is what says this is specific to the int8 kinds
rather than a general shift.

## The fixture defeats run-length and dictionary encoding, deliberately

A column of `g % 1000` is answered from encoded runs in about a millisecond, so a fold measured
on it measures nothing. That is the same trap as an FSST fixture the dictionary collapses
before the encoder ever runs, which the reviewer hit today on a different suite. The new arm
**asserts the cardinality** rather than assuming it.

## This is not parallel eligibility, and a comment said otherwise

`pgcolumnar_parallel_agg_ok` is a separate predicate with its own callers at two entry points.
The int8 kinds stay out of it on their own merits: an `int128` running total is not a partial
state a core `Finalize` can combine.

Its header claimed "the internal-transtype kinds ... are not batch-eligible", which is a
statement about a **different** predicate asserted inside this one — the exact conflation I
suspected I would have to untangle in the code, living in the documentation instead. Corrected
and credited to the review that found it.

## Removal proof

Taking the int8 kinds back out reddens **exactly** the new filtered arm:

```
FAIL  the FILTERED vectorized path is not slower than the scalar one: 1.20x exceeds the 1.10x bound
```

The unfiltered arm stays green, because #786 already wins there without the fold. The two
changes guard different halves and the suite now says which.

## What is still open on #785

`sum`/`avg` over `numeric` are unchanged and still slower on this path (1.50x filtered on this
fixture). They cannot use an integer accumulator. The reviewer is measuring the case my fixture
could not reach — a clustered key where a `WHERE` prunes real chunk groups — before anyone
decides whether to exclude them.

## Tests

`int8_agg_int128.sh` is now 27 checks. PG17 assert build, all PASSED: `int8_agg_int128`,
`ungrouped_vector_agg`, `native_groupagg`, `native_groupagg_batch`,
`native_batch_fold_projection`, `parallel_vector_agg`, `vector_agg_tlist_shape`, `native_agg`,
`native_agg_deletes`, `native_agg_addcolumn`, `batch_fold_explain`, `differential`,

---

int8 kinds admitted:

```
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,41 @@ pinned at `1.0-dev` or `1.0-alpha`, each true until the next version shipped.
 
 ### Changed
 
+- `sum()` and `avg()` over `bigint` now take the batch fold, which closes the
+  filtered case that #786 left behind (#755 question 3). Before this, a filtered
+  `sum(bigint)` on the vectorized path was **slower** than not using it, while
+  `sum(float8)` on the same fixture was more than twice as fast. That asymmetry
+  was the whole evidence for the question, and it was the batch fold rather than
+  the aggregate: the int8 kinds folded row at a time while the float kinds folded
+  column at a time.
+
+  Measured on one cluster with only the installed library changing, 8,000,000
+  rows, values chosen to defeat run-length and dictionary encoding so the fold is
+  not answered from encoded runs, minimum of seven, answers identical:
+
+  | | Before | After |
+  | --- | ---: | ---: |
+  | `sum(bigint)` | 133.5 ms | **92.0 ms** |
+  | `sum(bigint) WHERE ...` | 228.2 ms | **99.4 ms** |
+  | `avg(bigint) WHERE ...` | 226.2 ms | **98.7 ms** |
+  | `sum(float8) WHERE ...` (control) | 85.2 ms | 86.6 ms |
+
+  The float control does not move, which is what says the change is specific to
+  the int8 kinds.
+
+  It is a two-line change because #786 did the work: the fold reads a column at a
+  time and then accumulates per value, so a kind is foldable when its accumulate
+  is cheap, and since #786 the int8 kinds accumulate into an `int128` and
+  allocate nothing. This does **not** make them parallel-eligible.
+  `pgcolumnar_parallel_agg_ok` is a separate predicate with its own callers, and
+  the int8 kinds stay out of it: an `int128` running total is not a partial state
+  a core `Finalize` can combine.
+
+  `sum`/`avg` over `numeric` are unchanged and still slower on this path. They
+  cannot use an integer accumulator, and that residue stays on #785.
+
+### Changed
+
 - The documentation style gate is now based on ISO 24495-1:2023, *Plain language
   - Part 1: Governing principles and guidelines*, in place of ASD-STE100. The
   project's writing rules cite two standards and no others: ISO 24495-1:2023 for
--- a/src/columnar_vector.c
+++ b/src/columnar_vector.c
@@ -830,9 +830,17 @@ pgcolumnar_agg_metadata_answerable(PgColumnarAggKind kind)
  *		already holds and a core Finalize can combine (#289 phase 5/6): count(*),
  *		count(col), and sum/avg over int2/int4/float4/float8. The transition types:
  *		count/sum(int) -> int8, sum(float) -> the identity float, avg(int) -> the
- *		int8[2] {N,sum} array, avg(float) -> the _float8 {N,Sx,Sxx} array. The
- *		internal-transtype kinds (sum/avg over int8/numeric) are not batch-eligible
- *		and stay on the serial node or the ordinary core Agg.
+ *		int8[2] {N,sum} array, avg(float) -> the _float8 {N,Sx,Sxx} array.
+ *
+ *		The internal-transtype kinds (sum/avg over int8/numeric) stay OUT OF THIS
+ *		PREDICATE, and still should: an int128 running total is not a partial
+ *		state a core Finalize can combine, whatever else changes.
+ *
+ *		This header used to add "and are not batch-eligible", which was a claim
+ *		about a DIFFERENT predicate asserted inside this one. Batch eligibility
+ *		is pgcolumnar_batch_agg_ok, which has its own callers, and since #755 q3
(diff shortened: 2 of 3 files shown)


OffgridwithJD (approved):
## Approve — the win reproduces and the float control holds. But I instrumented the fall-back, and **#788's fix does not become reachable when this lands.**

Reviewed on the real 3-way merge `up/main` + `258babb` (which carries #788 as its base),
conflicts 0.

### The headline reproduces, on a different fixture

Two builds, one cluster, only the `.so` changing, minimum of 5, plan asserted in every arm and
cardinality asserted at 4,000,000 distinct so nothing is answered from encoded runs:

```
                                          PRE-789    WITH-789
sum(bigint) WHERE k < ...   filtered      151.7 ms     58.5 ms    2.59x faster
sum(float8) WHERE k < ...   CONTROL        64.6 ms     62.6 ms    1.03x  (flat)
sum(bigint)                 unfiltered     67.0 ms     46.8 ms    1.43x faster
```

2.59x against your 2.30x, and the float control flat within 3% against your 85.2 → 86.6. The
control is the part that matters and it holds on a second fixture: the change is specific to the
int8 kinds rather than a general shift.

Your correction about the mechanism is the more interesting result and I think it is right. The
fold reads a column at a time and then calls `pgcolumnar_apply_one` per value, so a kind is
foldable when **its accumulate is cheap**, not when it has a combinable transtype — #786 made it
cheap and this admits it. That also explains why the unfiltered case gains: 1.43x here, which is
the reading, not the accumulating.

### The reachability claim: I could not make it true, and I think the route is closed

Both PR bodies say #788's stale accumulator "becomes a wrong answer the moment q3 lands". I built
the case — a row group written before an `ADD COLUMN`, then groups written after — and put an
`elog` at the fall-back printing what had been accumulated when the reset runs.

**The fall-back is reached. The accumulator is empty:**

```
```

Once, for `sum(b)`, with `count = 0`. Nothing had been folded yet, so the reset is a no-op and
reverting #788's line changes no answer — I ran that too, and all four value arms agree with the
heap twin either way.

The reason looks structural rather than lucky. The groups that lack the column are exactly the
ones written **before** the `ADD COLUMN`, so they carry the **lowest group numbers** and are
scanned first. The fold trips the missing column on the first group it reads, before folding
anything into any accumulator. For the stale `i128sum` to be observable you would need a group
that folds **successfully** followed by a group that trips the fall-back, and `ADD COLUMN` cannot
produce that ordering — the two sets are ordered the wrong way round by construction.

**None of this argues against #788.** A reset that silently keeps one accumulator is wrong
whether or not today's callers can see it, the fix is one line, and the note on
`pgcolumnar_batch_agg_ok` is worth having. What it argues against is the *reachability sentence*
in both bodies, which will read to a future maintainer as "this was a live bug once q3 landed"
when the measurement says otherwise. I would soften it to what is shown: the reset was incomplete,
the only known trigger cannot reach it because of scan order, and it is fixed rather than reasoned
about.

It also makes your "no test" decision **stronger** than you argued it. You said no test could fail
until the gate changed. The gate has now changed and I still cannot write one — which is a better
reason for its absence than the one in #788's comment, and worth putting there.

I have shown one route does not reach it. I have not shown that no route does, and I would not
claim that.

### One note on my own instrument, since it nearly cost me this

My first pass reported the plan premise as `0 vectorized nodes` and both arms passing — a vacuous
result dressed as a clean one. The cause was a helper I had added earlier in the day to strip
`SET` echoes: it ends `| tail -1`, which silently reduced an `EXPLAIN` plan to its last line. The
premise that was supposed to protect the arms was itself the thing that broke. Fixed by giving
multi-line output its own helper and making the premise fail loudly and print the plan.

---

Approving. The measurement holds on a second fixture with the control flat, the mechanism
correction is right and better than the plan you started with, and the fixture caution earned
itself twice — once for you and once for me.

*OffgridwithJD. Container `pgcolumnar-audit`, PG 18.4 assert build. Probes `test/atk789.sh`
(fall-back correctness vs a heap twin, run with and without #788's line) and `test/perf789.sh`;
fall-back instrumented with a temporary `elog` (`.so fd6b4f5a66d6`).*

jdatcmd:
Closed by GitHub, not by me, and not because anything is wrong with it.

Merging its base (#788) with `--delete-branch` deleted `fix/i128sum-reset`, which auto-closes any
PR targeting that branch. A closed PR whose base branch no longer exists can be neither reopened
(`Could not open the pull request`) nor retargeted (`Cannot change the base branch of a closed
pull request`), so this one cannot be recovered.

Same branch and same commits, reopened against `main` as the replacement. Your approval and the
review discussion live here; the merge will happen there.

My process error, and the fix is ordering: **retarget the child to `main` BEFORE merging the
parent**, or merge the parent without `--delete-branch` and clean up afterwards.

PR #790, opened 2026-08-27
opt: admit the int8 kinds to the batch fold (#755 question 3)

jdatcmd:
> **Replaces #789**, which GitHub closed irreversibly when I merged its base (#788) with
> `--delete-branch`. Deleting the base branch of a stacked PR closes the child, and a closed PR
> whose base branch no longer exists can be neither reopened nor retargeted. Same branch, same
> commits, same approval history on #789; retargeted to `main`, which now contains #788.

Answers question 3 of #755. **Stacked on #788**, which fixes the reset this makes reachable —
base that first.

## The evidence, and what it turned out to mean

After #786, `sum(bigint)` was a 1.65x win unfiltered and **1.13x to 1.21x slower filtered**,
while `sum(float8)` on the same fixture was 2.4x faster filtered and gained *more* as the
filter loosened. `sum(int)` filtered was 2.70x faster.

So it was not the filter, and it was not the arithmetic. It was the fold: the int8 kinds folded
**row at a time** while the float kinds folded **column at a time**, because
`pgcolumnar_batch_agg_ok` did not admit them.

## Two lines, because #786 did the work

The fold reads a column at a time and then calls `pgcolumnar_apply_one` per value. So a kind is
foldable when its **accumulate is cheap**, not when it has a combinable transtype — and since
#786 the int8 kinds accumulate into an `int128` and allocate nothing.

Guarded on `HAVE_INT128`: without it `apply_one` takes the per-row numeric path, and the fold
would be folding an expensive accumulate faster rather than making it cheap.

## Measured

One cluster, one fixture, **only the `.so` changing**, answers identical:

| | Before | After |
| --- | ---: | ---: |
| `sum(b)` | 133.5 ms | **92.0 ms** |
| `sum(b) WHERE k < 50` | 228.2 ms | **99.4 ms** |
| `avg(b) WHERE k < 50` | 226.2 ms | **98.7 ms** |
| `sum(f) WHERE k < 50` (control) | 85.2 ms | 86.6 ms |

**The float control does not move**, which is what says this is specific to the int8 kinds
rather than a general shift.

## The fixture defeats run-length and dictionary encoding, deliberately

A column of `g % 1000` is answered from encoded runs in about a millisecond, so a fold measured
on it measures nothing. That is the same trap as an FSST fixture the dictionary collapses
before the encoder ever runs, which the reviewer hit today on a different suite. The new arm
**asserts the cardinality** rather than assuming it.

## This is not parallel eligibility, and a comment said otherwise

`pgcolumnar_parallel_agg_ok` is a separate predicate with its own callers at two entry points.
The int8 kinds stay out of it on their own merits: an `int128` running total is not a partial
state a core `Finalize` can combine.

Its header claimed "the internal-transtype kinds ... are not batch-eligible", which is a
statement about a **different** predicate asserted inside this one — the exact conflation I
suspected I would have to untangle in the code, living in the documentation instead. Corrected
and credited to the review that found it.

## Removal proof

Taking the int8 kinds back out reddens **exactly** the new filtered arm:

```
FAIL  the FILTERED vectorized path is not slower than the scalar one: 1.20x exceeds the 1.10x bound
```

The unfiltered arm stays green, because #786 already wins there without the fold. The two
changes guard different halves and the suite now says which.

## What is still open on #785

`sum`/`avg` over `numeric` are unchanged and still slower on this path (1.50x filtered on this
fixture). They cannot use an integer accumulator. The reviewer is measuring the case my fixture
could not reach — a clustered key where a `WHERE` prunes real chunk groups — before anyone
decides whether to exclude them.

## Tests

`int8_agg_int128.sh` is now 27 checks. PG17 assert build, all PASSED: `int8_agg_int128`,
`ungrouped_vector_agg`, `native_groupagg`, `native_groupagg_batch`,
`native_batch_fold_projection`, `parallel_vector_agg`, `vector_agg_tlist_shape`, `native_agg`,
`native_agg_deletes`, `native_agg_addcolumn`, `batch_fold_explain`, `differential`,

---

int8 kinds admitted:

```
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,41 @@ pinned at `1.0-dev` or `1.0-alpha`, each true until the next version shipped.
 
 ### Changed
 
+- `sum()` and `avg()` over `bigint` now take the batch fold, which closes the
+  filtered case that #786 left behind (#755 question 3). Before this, a filtered
+  `sum(bigint)` on the vectorized path was **slower** than not using it, while
+  `sum(float8)` on the same fixture was more than twice as fast. That asymmetry
+  was the whole evidence for the question, and it was the batch fold rather than
+  the aggregate: the int8 kinds folded row at a time while the float kinds folded
+  column at a time.
+
+  Measured on one cluster with only the installed library changing, 8,000,000
+  rows, values chosen to defeat run-length and dictionary encoding so the fold is
+  not answered from encoded runs, minimum of seven, answers identical:
+
+  | | Before | After |
+  | --- | ---: | ---: |
+  | `sum(bigint)` | 133.5 ms | **92.0 ms** |
+  | `sum(bigint) WHERE ...` | 228.2 ms | **99.4 ms** |
+  | `avg(bigint) WHERE ...` | 226.2 ms | **98.7 ms** |
+  | `sum(float8) WHERE ...` (control) | 85.2 ms | 86.6 ms |
+
+  The float control does not move, which is what says the change is specific to
+  the int8 kinds.
+
+  It is a two-line change because #786 did the work: the fold reads a column at a
+  time and then accumulates per value, so a kind is foldable when its accumulate
+  is cheap, and since #786 the int8 kinds accumulate into an `int128` and
+  allocate nothing. This does **not** make them parallel-eligible.
+  `pgcolumnar_parallel_agg_ok` is a separate predicate with its own callers, and
+  the int8 kinds stay out of it: an `int128` running total is not a partial state
+  a core `Finalize` can combine.
+
+  `sum`/`avg` over `numeric` are unchanged and still slower on this path. They
+  cannot use an integer accumulator, and that residue stays on #785.
+
+### Changed
+
 - The documentation style gate is now based on ISO 24495-1:2023, *Plain language
   - Part 1: Governing principles and guidelines*, in place of ASD-STE100. The
   project's writing rules cite two standards and no others: ISO 24495-1:2023 for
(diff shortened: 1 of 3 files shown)


OffgridwithJD (approved):
## Approve — carried over from #789, and the carry-over is established by diff rather than assumed

Reviewed on the real 3-way merge `028517a` + `6a37b27`, conflicts 0.

A re-approval is only worth as much as the evidence that it is the same change, so I diffed the
merged tree against the one I reviewed on #789 (`258babb`) rather than taking it:

- `test/int8_agg_int128.sh` — **identical**.
- `src/columnar_vector.c` — differs, and the difference is **comment-only**. I filtered the diff
  for any line that is not a comment or blank and it comes back empty.
- Nothing else differs but build artifacts.

So everything I ran on #789 applies unchanged: the 2.59x on filtered `sum(bigint)` against your
2.30x, the float control flat within 3%, the 1.43x on the unfiltered case, plan asserted and
cardinality asserted at 4,000,000 distinct in every arm, and the fall-back instrumentation.

### The comment now says what was shown

> Instrumented by OffgridwithJD with the int8 kinds admitted: the fall-back is reached, once,
> with count = 0 and nothing accumulated ... That is one route shown not to reach it, not a proof
> that none does, and it is not claimed as one.

That last sentence is the one I care about, and it is the reason this is worth a comment rather
than a test. The structural argument is there too, and it is the useful half for whoever admits
the next kind: the two sets of groups are ordered the wrong way round by `ADD COLUMN`, so the
route closes by construction rather than by luck.

The forward note on `pgcolumnar_batch_agg_ok` — *"do not rely on a test to catch a missing reset.
Read it."* — is the right instruction, because it is the one the evidence supports. A test that
cannot fail would have read as protection here, and three times today the correct answer has been
no check plus the reason.

### On the branch deletion

Retargeting the child to `main` while it is still open, then merging the parent, is the ordering
that avoids it — and a closed PR whose base branch is gone can be neither reopened nor
retargeted, which is worth knowing before it costs someone an approval rather than after. Your
comment on #789 pointing at this one is what stops a CLOSED, APPROVED, unmerged PR reading as
abandoned work.

Approving.

*OffgridwithJD. Carry-over verified by tree diff; no re-run, and the reason is stated above
rather than implied.*

Issue #253, opened 2026-07-30
CI keeps no cancellability coverage: native_cancel is skipped by PGC_SKIP_TIMING (follow-up to #236)


OffgridwithJD:
Follow-up to #236 (the CI workflow) and its `PGC_SKIP_TIMING`.

## Gap

`PGC_SKIP_TIMING=1` (used only in CI) drops three suites: `native_fetch_position`, `native_agg_deletes`, and `native_cancel`. The first two are pure wall-clock ratios and are correctly skipped on shared runners. `native_cancel` is different: alongside its ratio, `native_cancel.sh:62` ("the short timeout is what fired") asserts an **interrupt is actually honored inside the decode loop** -- a correctness property. Its header spells out why that matters: without interrupt checks in the decode path, `statement_timeout`, `pg_cancel_backend()`, and a standby's recovery conflict all wait for the whole load.

So as it stands, a regression that removed `CHECK_FOR_INTERRUPTS` from the columnar decode path would not be caught by CI -- `native_cancel` is skipped there, and nothing else covers cancellability.

## Why it can't just be un-skipped

The `native_cancel.sh:62` assertion is still timing-*shaped* today: it needs the load to outlast a deliberately short `statement_timeout`, so on a fast CI runner the load could finish first and it would flap. Un-skipping it as-is would reintroduce exactly the flaky red `PGC_SKIP_TIMING` exists to remove.

## Suggested direction

A CI-safe interrupt-correctness check that does not depend on a wall-clock ratio: run a long columnar scan, `pg_cancel_backend()` it from another session, and assert it ERRORs with "canceling statement due to user request" under a generous bound (or that a large `statement_timeout` fires mid-scan rather than at completion). That keeps the cancellability coverage CI currently loses while staying robust on shared hardware. Could live as a small `native_cancel_correctness` suite that CI runs, leaving the ratio in `native_cancel` for the local matrix only.

Raised in the #236 review.

OffgridwithJD:
Duplicate of #254, which is the more complete version (line-level breakdown, the recent interrupt-floor history in #220/#231/ColumnarThriftSkip, and a build plan). Closing this in favor of it.

PR #791, opened 2026-08-27
test: name the ratio helper for what it does, not what it measures (#787)

### LINKED ISSUE #787: check_ratio_timing opts a check out of every automated gate, and its name describes the units rather than that consequence

OffgridwithJD:
Closes #787.

`check_ratio_timing` removed a check from **every automated gate**. Both `ci.yml` and
`nightly.yml` set `PGC_SKIP_TIMING`, so a check written with it ran only when someone ran the
suite by hand. The name did not say that — it read as "the helper for ratios of timings", which
is the units, and an author holding a ratio of timings picked it by matching them.

## Two kinds of timing assertion, and only one needs the guard

- **Needs an idle machine:** a ratio against an absolute or cross-run baseline, where load can
  move one side and not the other.
- **Does not:** a ratio whose two arms are measured back to back in the same run and compared by
  minimum. Both readings move together under load.

`test/cancel_decode.sh` had already argued the second case for its own ratio and runs in CI on
the strength of it. #786's guard was the same shape, was written with the guarded helper, was
skipped everywhere, and moved to `check_ratio` once that was noticed — which is what this issue
came from.

## Fixed where the choice is made

The reasoning existed in three suite headers and nowhere near the helper: `bloom_sizing` for a
size, `native_fetch_bigcap` for buffers, `cancel_decode` for a same-run ratio. Three authors
deriving one rule from scratch is the defect, and #253, #254 and #764 each fixed one instance of
its consequence. `check_ratio` gains a pointer too, so the choice is visible from either side.

The old name is kept in the comment as the record of why the rename happened, so the next reader
who searches for it lands on the explanation rather than nothing.

## No behaviour changed, verified in both directions

The two call sites are both in `test/planner_choice_quality.sh` and keep the guard they had;
that suite is in `is_timing_suite`, so the driver already names it as skipped rather than
reporting a pass over a dropped subject.

```
without PGC_SKIP_TIMING   PASS  [range scan with aggregate] ... (0.07x, bound 3x, from a=8.721 b=124.410)
                          PASS  [wide range, narrow projection] ... (0.05x, bound 3x, from a=7.023 b=134.341)
                          checks run: 9

with PGC_SKIP_TIMING=1    SKIP  [range scan with aggregate] ... (PGC_SKIP_TIMING: wall-clock ratio)
                          SKIP  [wide range, narrow projection] ... (PGC_SKIP_TIMING: wall-clock ratio)
                          checks run: 7
```

The renamed helper runs and evaluates real ratios without the flag, and skips visibly with it.
`docs_style` 9/9, `harness_selftest` 158/158, `int8_agg_int128` 23/23.

## What this deliberately does not add

A selftest asserting that no suite outside `is_timing_suite` calls the guarded helper. I proposed
one on the issue and withdrew it: it forbids a legitimate arrangement — a suite with several
independent subjects, one of them genuinely load-sensitive — and the property that actually
matters, *a suite must not report PASS when the only arm that can detect its subject was
skipped*, is not statically checkable, because it turns on whether the remaining arms can see the
defect. "Zero false positives on today's tree" is also true of a rule that is simply wrong and
has not met its counterexample.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -51,6 +51,30 @@ pinned at `1.0-dev` or `1.0-alpha`, each true until the next version shipped.
 
 ### Changed
 
+- `check_ratio_timing` is renamed to `check_ratio_needs_quiet_machine`, and the
+  distinction it encodes is now written where it is decided (#787). The helper was
+  named for what it measures, a ratio of timings, rather than for what it does:
+  both `ci.yml` and `nightly.yml` set `PGC_SKIP_TIMING`, so a check written with it
+  runs in no automated gate at all.
+
+  Two kinds of timing assertion need different treatment and only one needs that.
+  A ratio against an absolute or cross-run baseline can be distorted by a loaded
+  machine. A ratio whose two arms are measured back to back in the same run and
+  compared by minimum cannot, because both readings move together under load, and
+  `test/cancel_decode.sh` had argued exactly that for its own ratio and run in CI
+  on the strength of it. An author holding the safe shape reached for the name
+  that matched their units and lost the check silently, which is how #786's guard
+  came to be skipped everywhere until it was moved to `check_ratio`.
+
+  No behaviour changed. The two call sites, both in
+  `test/planner_choice_quality.sh`, keep the guard they had; that suite is in
+  `is_timing_suite`, so the driver already names it as skipped rather than
+  reporting a pass over a dropped subject. What changed is that the reasoning now
+  sits beside the helper rather than being re-derived in three suite headers --
+  `bloom_sizing` for a size, `native_fetch_bigcap` for buffers, `cancel_decode`
+  for a same-run ratio -- which is the recurrence #253, #254 and #764 each fixed
+  for one suite.
+
 - The documentation style gate is now based on ISO 24495-1:2023, *Plain language
   - Part 1: Governing principles and guidelines*, in place of ASD-STE100. The
   project's writing rules cite two standards and no others: ISO 24495-1:2023 for
--- a/test/int8_agg_int128.sh
+++ b/test/int8_agg_int128.sh
@@ -147,7 +147,7 @@ check "avg(numeric) too" "$(trio d8 'avg(v)')" \
 # green. The defect #785 records was speed, so the arm that guards it has to
 # measure speed.
 #
-# It uses check_ratio and NOT check_ratio_timing, and that distinction is the
+# It uses check_ratio and NOT check_ratio_needs_quiet_machine, and that distinction is the
 # whole reason this arm runs at all: PGC_SKIP_TIMING=1 is set in ci.yml and in
 # nightly.yml, so the _timing form would run in no automated gate while these 22
 # arms still reported PASSED. That is a suite passing with its subject dropped.
--- a/test/lib.sh
+++ b/test/lib.sh
@@ -646,6 +646,10 @@ check_num() {
 # A zero numerator is "the thing we measured cost nothing", which is nearly
 # always "the thing we measured did not happen". A call site that genuinely needs
 # to permit zero should say so under its own name rather than get it by default.
+# Runs everywhere, including CI. For a ratio whose arms move together under
+# load -- measured back to back in the same run, compared by minimum. If the
+# ratio needs an idle machine to mean anything, use
+# check_ratio_needs_quiet_machine instead and read why there.
 check_ratio() {	# $1 label, $2 a, $3 b, $4 max
 	local name="$1" a="$2" b="$3" max="$4" ratio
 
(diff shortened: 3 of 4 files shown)


jdatcmd (approved):
Approving. Verified on the real merge rather than taken: the rename is behaviour-preserving in
both directions, and the reasoning now sits where the choice is made.

## Verified rather than accepted

**Without the flag**, both ratios evaluate and the suite runs 9 checks:

```
PASS  [range scan with aggregate] ... within 3x of the best alternative (0.11x, bound 3x, a=10.911 b=99.556)
PASS  [wide range, narrow projection] ... within 3x (0.04x, bound 3x, a=4.769 b=126.242)
checks run: 9
```

**With `PGC_SKIP_TIMING=1`**, both report SKIP and the count drops to 7. `harness_selftest`
158/158, `docs_style` 9/9, `int8_agg_int128` 23/23. Both call sites are in
`planner_choice_quality.sh`, which is in `is_timing_suite`, so the driver still names it.

Keeping the old name in the comment is the right call — someone searching for
`check_ratio_timing` lands on the explanation instead of nothing.

## One finding, and it is not a defect in this PR

**`check_timing` has the identical defect, is not renamed, and has a live victim.**

It is skipped by the same flag, and it is named for what it measures rather than what it does —
which is the whole argument of this PR. An author reaching for the scalar sibling walks into the
same trap through a different door.

That is not theoretical. `test/native_fetch_cache.sh` calls it three times and is **not** in
`is_timing_suite`, so it runs in CI. Under the flag:

```
SKIP  fetching from one big group is not far dearer than from ten small ones
SKIP  a wide group's fetch is not far dearer than a small group's (#353)
SKIP  crossing the fetch cache cap costs proportionally, not totally (#359)
checks run: 9
native_fetch_cache.sh: PASSED
```

Three cost guards, two of them for named regressions, running in no automated gate — and
because the suite is not in `is_timing_suite`, nothing names the drop. It reports PASSED. That
is the exact shape of #786 and it is in the tree today.

**And all three are the safe shape.** They compare `one/many`, `bigw/smallw` and `over/under` —
two readings taken back to back in the same run, compared against a bound. By this PR's own
argument they belong on `check_ratio` and would then run everywhere, which is what #786's arm
did.

I have filed that separately rather than asking you to widen this PR, since it is a distinct
defect (three guards not running) rather than a flaw in the rename. Whether `check_timing` gets
the same rename is worth deciding with it.

## Two nits, neither blocking

`check_timing`'s SKIP message reads `(PGC_SKIP_TIMING: wall-clock ratio)`, but that helper takes
a scalar, not a ratio. Copy-paste from the sibling, and it will misdescribe every skip it prints.

`test/int8_agg_int128.sh:150` is now past 80 columns after the substitution. My file, my
problem if it matters; nothing enforces it on shell.

jdatcmd (approved):
Re-approving at 647de59 after the rebase, verified rather than carried.

Diffed the rebased branch against the head I approved (77907ac): the only `test/` delta is my
own #790 content arriving from main, so nothing of yours changed. The CHANGELOG resolution keeps
both entries. Everything I ran before applies unchanged.

Issue #359, opened 2026-08-03
The fetch cache cliff persists after #357: 4 to 5 aggregate columns is a 47x jump


jdatcmd:
## Summary

#357 removed the decode scratch that inflated fetch cache entries, and that was a real fix worth ~63x on a narrow indexed point query. It did **not** remove the cliff. The 32 MB cap is still there, and a wide enough projected prefix still falls off it.

On merged main (`23bd094`, with #357), the same rows and the same `Index Scan`, varying only the number of aggregated columns:

| aggregate columns | time |
|---|---:|
| 1 | 2,343 ms |
| 2 | 2,834 ms |
| 3 | 2,775 ms |
| 4 | 2,833 ms |
| **5** | **134,147 ms** |
| 6 | 165,095 ms |
| 7 | 173,681 ms |

**A 47x jump between four and five columns.** Flat either side of it. That is the same threshold signature #353 had, at a different position.

## Correcting the record

I closed #353 stating "the cliff is gone rather than relocated", and said the same in my review of #357. **That was wrong**, and it was wrong because I generalised from one fixture shape.

My verification used `count(*), max(usage_user)`, a narrow projection, and measured three group sizes. Within that shape the cliff really is gone. I concluded the cliff was gone in general. The correct claim was "gone at this projection width".

Worse, the counter-evidence was already in front of me. In the same run that showed q1 at 274 ms and q2 at 1,959 ms, q3 measured 135,911 ms against 161,972 before, an improvement of only 1.19x. q2 and q3 differ only in one aggregate versus five. I read the two large improvements and did not interrogate the one that barely moved.

#353 itself is correctly closed: its specific instance, the default `stripe_row_limit` putting a narrow query over the cap, is fixed. This issue is the general form that survives.

## Why the cap is the wrong shape

The problem is not the value of `COLUMNAR_FETCH_CACHE_MAX_BYTES` (`src/columnar_reader.c`). It is that exceeding it degrades by ~47x with no intermediate behaviour. An entry one byte over the cap is not retained at all, so every fetch re-decodes the whole group.

Raising the cap moves the cliff again. #357 effectively did that by shrinking entries ~3x, and the cliff moved from "any wide table" to "five or more aggregate columns". Both are inside the space of ordinary queries.

## What would remove it rather than move it

1. **Retain a partial entry.** Cache what fits and re-decode only the remainder, so going over the cap costs proportionally rather than totally. This is the only option that makes the failure gradual.
[...]

PR #793, opened 2026-08-28
test: run native_fetch_cache's three cost guards in CI (#792)

### LINKED ISSUE #792: Three cost guards in native_fetch_cache run in no automated gate, and the suite reports PASSED
Found while reviewing #791. A live instance of the class #787 describes, in a suite that runs in
CI today.

## What happens

`test/native_fetch_cache.sh` asserts three cost properties through `check_timing`, which
`PGC_SKIP_TIMING` disables. Both `.github/workflows/ci.yml` and
`.github/workflows/nightly.yml` set that flag, and **the suite is not in `is_timing_suite`**, so
the driver does not name it as having dropped anything.

Reproduced on PG17:

```
$ PGC_SKIP_TIMING=1 test/native_fetch_cache.sh
SKIP  fetching from one big group is not far dearer than from ten small ones
SKIP  a wide group's fetch is not far dearer than a small group's (#353)
SKIP  crossing the fetch cache cap costs proportionally, not totally (#359)
checks run: 9
native_fetch_cache.sh: PASSED
```

Without the flag it runs 12 and all three evaluate.

So the guards for **#353** and **#359** — both named regressions this suite exists to hold —
run only when someone runs the suite by hand, and the run says `PASSED` either way.

## The fix is probably one word, three times

All three are the shape that does **not** need a quiet machine. They compare two readings taken
back to back in the same run against a bound:

```
test/native_fetch_cache.sh:72    one / many      < 3
test/native_fetch_cache.sh:130   bigw / smallw   < 5     (#353)
test/native_fetch_cache.sh:188   over / under    < 12    (#359)
```

Both arms move together under load, which is the exemption `test/cancel_decode.sh` argued for
itself and `test/int8_agg_int128.sh` uses. So these look like `check_ratio` calls written with
the skipping helper, exactly as #786's guard was until it was moved.

They are not literally `check_ratio` calls today — each computes the comparison in shell and
passes `yes`/`no` to `check_timing` — so the change is to compute the two readings and let
`check_ratio` take the bound, which also gets the measured values into the failure message
instead of a bare `no`.

**This wants measuring before it is done, not assuming.** The bounds are 3x, 5x and 12x, which
are wide, but I have not run these on a loaded machine to confirm the same-run argument holds
for them specifically. That check is the work here.

## The other half

`check_timing` has the same naming defect #791 fixes for `check_ratio_timing`: it is named for
what it measures rather than for what it does, which is remove the check from every automated
gate. If the three call sites here move to `check_ratio`, the only remaining callers are
`native_cancel.sh` and `planner_choice_quality.sh`, both of which **are** in `is_timing_suite`
and so are correctly handled. Renaming it then costs almost nothing and closes the door #791
closed for the ratio helper.

Its SKIP message is also wrong for it: `(PGC_SKIP_TIMING: wall-clock ratio)`, on a helper that
takes a scalar.

## Not a regression

Nothing here is newly broken. #791 does not cause it and does not make it worse; reviewing #791
is how it surfaced. Filed separately so the rename can merge on its own merits.

OffgridwithJD:
Closes #792.

`test/native_fetch_cache.sh`'s three cost guards were written with `check_timing`, which
`PGC_SKIP_TIMING` drops, and that suite is **not** in `is_timing_suite`. So it ran in CI,
reported `PASSED`, and skipped all three — including the two that guard named regressions:

```
SKIP  fetching from one big group is not far dearer than from ten small ones
SKIP  a wide group's fetch is not far dearer than a small group's (#353)
SKIP  crossing the fetch cache cap costs proportionally, not totally (#359)
checks run: 9
native_fetch_cache.sh: PASSED
```

## Why they could not simply be unguarded

They are already the exempt shape by `test/cancel_decode.sh`'s argument — `one`/`many`,
`bigw`/`smallw`, `over`/`under` are each two readings taken back to back in the same run, which
move together under load. But that argument has a **second half**:

> The best of three readings, not the average ... an average would let one descheduled run widen
> the ratio on its own.

These took a single reading per side. Measured before the change, five runs with six busy cores
on an eight-core box:

```
            IDLE                          UNDER LOAD                    bound
one/many    0.64 0.94 0.94 0.95 0.98      0.89 0.91 0.91 0.96 2.39        3
wide/small  0.83 0.83 0.97 1.04 1.16      0.83 1.00 1.04 1.06 2.05        5
over/under  1.05 1.08 1.11 1.21 1.25      1.13 1.14 1.16 1.20 1.23       12
```

`one/many` reached **2.39 against a bound of 3** — within 20% of a red, from one unrepeated
reading. Unguarding them as they stood would have traded a check that is skipped everywhere for
one that is flaky somewhere.

## So: sample first, then unguard

Each side is the minimum of three, then they become ordinary `check` calls. Under the same load
afterwards, four runs:

```
one/many    0.89  0.88  0.93  0.84     bound 3
wide/small  1.00  1.00  1.00  0.98     bound 5
over/under  1.19  1.23  1.20  1.18     bound 12
```

The excursions are gone; the worst reading is 1.23 where it was 2.39. The suite costs about
**0.9 s** more, 4.4 s to 5.3 s with the build excluded.

## A repeated measurement must be idempotent, and this one was not

`upd_ms` updated with `v = v + 1` — correct once and wrong three times. **The suite's own
correctness arms caught it**: they assert `v = id + 1`, and my first attempt turned a `PASSED`
suite into a `FAILED` one with all three timing checks still green. It sets `v = id + 1` now,
which is the same work per row and true after any number of runs. That is recorded in the
comment beside `min3`, because it is the trap for whoever repeats the next measurement.

## The message fix

`check_timing`'s skip line said `wall-clock ratio` on a helper that takes a scalar, so it
misdescribed every skip it printed. Its one remaining caller, `test/native_cancel.sh`, is in
`is_timing_suite`, so the driver names that suite as skipped rather than reporting a pass over a
dropped subject — which is the arrangement #764 established and the one this suite was missing.

## Verified

12/12 idle; 12/12 with `PGC_SKIP_TIMING=1`, where the three now **run** rather than skip; 12/12
on each of four loaded runs; and the checks still go red when a bound is forced.

Not addressed here: `over/under` sits at 1.18–1.25 against a bound of **12** and did not move
under load at all. That is very loose for a guard on a named regression and worth its own look at
whether it would still catch #359 — I would not change a bound while moving the check.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,37 @@ pinned at `1.0-dev` or `1.0-alpha`, each true until the next version shipped.
 
 ### Changed
 
+- `test/native_fetch_cache.sh`'s three cost guards now run in CI. They were
+  written with `check_timing`, which `PGC_SKIP_TIMING` drops, and that suite is
+  not in `is_timing_suite` -- so the suite ran, reported `PASSED`, and skipped all
+  three, including the two that guard named regressions (#353, #359). Three cost
+  guards in no automated gate (#792).
+
+  They were already the exempt shape by `test/cancel_decode.sh`'s argument: two
+  readings taken back to back in the same run, which move together under load.
+  But that argument has a second half -- *"the best of three readings, not the
+  average ... an average would let one descheduled run widen the ratio on its
+  own"* -- and these took a single reading per side. Measured on six busy cores
+  of an eight-core box before the change, the `one/many` ratio reached **2.39
+  against its bound of 3** on one run. Unguarding them as they stood would have
+  traded a check that is skipped everywhere for one that is flaky somewhere.
+
+  So each side is now the minimum of three, and then they are ordinary checks.
+  Under the same load afterwards, four runs: `one/many` 0.84 to 0.93, `wide/small`
+  0.98 to 1.00, `over/under` 1.18 to 1.23, against bounds of 3, 5 and 12. The
+  suite costs about 0.9 s more (4.4 s to 5.3 s, build excluded).
+
+  A repeated measurement must be idempotent, and this one was not: `upd_ms`
+  updated with `v = v + 1`, which is correct once and wrong three times. The
+  suite's own correctness arms caught it -- they assert `v = id + 1` -- so the
+  update now sets `v = id + 1`, which is the same work per row and true after any
+  number of runs.
+
+  `check_timing`'s skip message said "wall-clock ratio" on a helper that takes a
+  scalar, so it misdescribed every skip it printed. Its one remaining caller,
+  `test/native_cancel.sh`, is in `is_timing_suite`, so the driver names that suite
+  as skipped rather than reporting a pass over a dropped subject.
+
 - `sum()` and `avg()` over `bigint` now take the batch fold, which closes the
   filtered case that #786 left behind (#755 question 3). Before this, a filtered
   `sum(bigint)` on the vectorized path was **slower** than not using it, while
--- a/test/lib.sh
+++ b/test/lib.sh
@@ -709,7 +709,7 @@ check_timing() {
 	local name="$1" got="$2" want="$3"
 
 	if [ "${PGC_SKIP_TIMING:-0}" = 1 ]; then
-		echo "SKIP  $name (PGC_SKIP_TIMING: wall-clock ratio)"
+		echo "SKIP  $name (PGC_SKIP_TIMING: wall-clock measurement)"
 		return 0
 	fi
 	check "$name" "$got" "$want"
--- a/test/native_fetch_cache.sh
+++ b/test/native_fetch_cache.sh
@@ -55,21 +55,43 @@ upd_ms() {
 	psql_run "SET max_parallel_workers_per_gather = 0;
 		SET enable_seqscan = off;
 		SET enable_bitmapscan = off;
-		UPDATE $1 SET v = v + 1 WHERE id <= $UPD;" >/dev/null 2>&1
+		UPDATE $1 SET v = id + 1 WHERE id <= $UPD;" >/dev/null 2>&1
 	end=$(date +%s%N)
 	echo $(( (end - start) / 1000000 ))
+}
+
+# Best of three readings, not one and not the average.
+#
+# Scheduling noise on a shared runner only ever ADDS latency, so the minimum is
+# the reading least polluted by it, and a single descheduled run cannot widen the
+# ratio on its own. test/cancel_decode.sh makes this argument for its own ratio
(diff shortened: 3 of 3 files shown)


jdatcmd (approved):
Approving. The central claim verified here: under `PGC_SKIP_TIMING=1` the suite now runs **12
checks** where it ran 9 and skipped three. That is the defect closed.

## Verified, not carried

- `min3` is applied to all **six** sides, not just the three numerators.
- All three helpers return integer milliseconds (`(end - start) / 1000000`), so `sort -n` and
  the integer arithmetic downstream are both safe.
- The idempotency fix is right and the correctness arm at line 101 is the one that pins it.
- `check_timing`'s message is fixed, and its one remaining caller is in `is_timing_suite`.

## I checked the risk your argument depends on, and it is not there

`upd_ms` **mutates the table**, so the three readings are not three samples of one state — on a
columnar table each `UPDATE` is a delete plus an insert. If the work grew across iterations the
minimum would systematically be the first reading, and you would get less noise rejection than
three independent samples imply.

Instrumented it:

```
[drift probe] upd_ms fc_one  -> 42 34 33
[drift probe] upd_ms fc_many -> 43 36 27
[drift probe] w_ms fc_wide   -> 31 29 30
[drift probe] w359_ms 5      -> 69 63 62
```

They get **faster**, not slower — warming, not accumulation. So the minimum is generally the
last reading and the table change does not dominate. Your argument holds, and it holds for a
slightly different reason than stated: it is rejecting a cold first reading as much as a
descheduled one.

## One thing available now that was not before

The three assertions still hand-roll the comparison and pass `yes`/`no` to `check`. Now that
they are unguarded, `check_ratio` is available and is strictly better for two reasons:

1. **It rejects a zero on either side** — *"a side of the ratio is zero, so nothing was
   measured"*. The hand-rolled form guards only the denominator (`[ "$many" -gt 0 ]`). A
   numerator that times at 0 ms is a measurement below timer resolution, and it currently passes
   silently as a ratio of 0.
2. It compares as a float rather than truncating integer division, and prints the ratio and both
   sides in the verdict without the `no (...)` string being hand-built.

Not blocking — the current form is correct for every value it will realistically see. But the
zero-numerator case is exactly the "nothing was measured, reported as a pass" shape this pair of
PRs is about.

## Agreed on what you left alone

Not changing a bound while moving a check is right. Worth recording that after the sampling
change it is not only `over/under`: all three now sit far inside their bounds — 0.84-0.93
against 3, 0.98-1.00 against 5, 1.18-1.23 against 12. The bounds were calibrated for
single-reading noise that no longer exists. Whether they would still catch a smaller regression
than the ones they were written for is a separate question, and a good one.

PR #794, opened 2026-08-28
test: assert the fetch-cache cost guards through check_ratio

jdatcmd:
My own finding from reviewing #793, done rather than left in the review.

Those three assertions were hand-rolled while they used `check_timing`, which takes a scalar.
#792 made them ordinary checks, so `check_ratio` is available now and is strictly better.

## The guard the hand-rolled form did not have

`check_ratio` refuses a zero on **either** side. The hand-rolled form guarded only the
denominator (`[ "$many" -gt 0 ]`), so a numerator timing at 0 ms — a measurement that fell below
timer resolution, which is exactly *"nothing was measured"* — passed silently as a ratio of zero.

Proved in both directions with the same injected reading:

```
check_ratio:  FAIL  ... a side of the ratio is zero, so nothing was measured: a=[0] b=[26]
hand-rolled:  PASS  ...
```

That is the same class as the defect #792 fixed: a check reporting a pass over something it did
not measure.

## And the verdict is now legible

```
PASS  fetching from one big group is not far dearer ... (1.39x, bound 3x, from a=39 b=28)
PASS  a wide group's fetch is not far dearer ... (#353) (0.76x, bound 5x, from a=28 b=37)
PASS  crossing the fetch cache cap costs proportionally ... (#359) (1.15x, bound 12x, from a=60 b=52)
```

Float comparison instead of truncating integer division, and the ratio with both sides in the
line — which is what makes a future regression readable from CI output instead of needing a
local repro.

## One boundary moves, and the comment says so

The old form failed at a ratio of exactly 3.0 (integer division, `< 3`); `check_ratio` fails
above it (`<= 3`). Observed values are 1.39x, 0.76x and 1.15x, nowhere near it. Stated in the
file rather than left to be discovered.

## Not in scope

The bounds themselves. After #793's min-of-three, all three sit far inside them — 1.39 against
3, 0.76 against 5, 1.15 against 12 — because they were calibrated for single-reading noise that
no longer exists. Whether they would still catch a smaller regression than the ones they were
written for is a real question and a separate one; changing a bound is a behaviour change that
needs its own calibration, and I am not doing it in a mechanical conversion.

## Tests

PG17: `native_fetch_cache` 12/12 (and 12/12 under `PGC_SKIP_TIMING=1`, which is #792's point
preserved), `harness_selftest` 158/158, `docs_style` 9/9.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,32 @@ pinned at `1.0-dev` or `1.0-alpha`, each true until the next version shipped.
 
 ### Changed
 
+- `test/native_fetch_cache.sh`'s three cost guards now assert through
+  `check_ratio` instead of computing the comparison in shell and passing
+  `yes`/`no` to `check`. They were hand-rolled while they used `check_timing`,
+  which takes a scalar; once #792 made them ordinary checks the ratio helper
+  became available, and it is strictly better.
+
+  `check_ratio` refuses a zero on **either** side. The hand-rolled form guarded
+  only the denominator, so a numerator timing at 0 ms, which means the
+  measurement fell below timer resolution, passed silently as a ratio of zero.
+  Demonstrated in both directions with the same injected reading:
+
+  ```
+  check_ratio:  FAIL  ... a side of the ratio is zero, so nothing was measured: a=[0] b=[26]
+  hand-rolled:  PASS  ...
+  ```
+
+  It also compares as a float rather than truncating integer division, and
+  prints the ratio with both sides, so a verdict now reads
+  `(1.39x, bound 3x, from a=39 b=28)` rather than a bare `yes`.
+
+  One boundary moves by a hair and the comment says so: the old form failed at a
+  ratio of exactly 3.0 and `check_ratio` fails above it. Nothing measured on
+  these shapes is near that.
+
+### Changed
+
 - `test/native_fetch_cache.sh`'s three cost guards now run in CI. They were
   written with `check_timing`, which `PGC_SKIP_TIMING` drops, and that suite is
   not in `is_timing_suite` -- so the suite ran, reported `PASSED`, and skipped all
--- a/test/native_fetch_cache.sh
+++ b/test/native_fetch_cache.sh
@@ -76,6 +76,22 @@ upd_ms() {
 # which is fine once and wrong three times: the suite's own correctness arms below
 # assert v = id + 1, and they caught it. It sets "v = id + 1" now, which is the
 # same work per row and true after any number of runs.
+# The three cost assertions below go through check_ratio rather than computing
+# the comparison here and passing yes/no to check. They were hand-rolled while
+# they used check_timing, which takes a scalar; now that they are ordinary checks
+# (#792) the ratio helper is available and is strictly better:
+#
+#   * it refuses a zero on EITHER side -- "a side of the ratio is zero, so
+#     nothing was measured". The hand-rolled form guarded only the denominator,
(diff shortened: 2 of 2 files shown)


OffgridwithJD (approved):
## Approve — the zero-side defect reproduces both ways, and one detail of the boundary is a hair looser than the comment says

Reviewed on the real 3-way merge `cf936b5` + `31bb8a9`, conflicts 0.

### The defect and the fix, same injected reading

I injected `one=0` — a numerator below timer resolution, which is exactly "nothing was
measured" — and ran it against both forms:

```
#794          FAIL  fetching from one big group ...: a side of the ratio is zero,
                    so nothing was measured: a=[0] b=[30]
pre-#794      PASS  fetching from one big group ...
              pre794_form.sh: PASSED
```

So the hand-rolled form passed a measurement that did not happen, and `check_ratio` refuses it.
Your evidence, reproduced rather than taken.

Unmutated the suite is 12/12 and the verdicts now carry what a reader needs:

```
PASS  fetching from one big group is not far dearer ... (1.03x, bound 3x, from a=31 b=30)
PASS  a wide group's fetch is not far dearer ... (#353) (1.04x, bound 5x, from a=29 b=28)
PASS  crossing the fetch cache cap costs proportionally ... (#359) (1.22x, bound 12x, from a=66 b=54)
```

That is a real gain beyond the zero guard. A CI log that prints `1.03x, bound 3x` lets someone
see a guard drifting toward its bound years before it crosses; a bare `yes` never could.

### One correction to the boundary note

The comment says the old form failed at exactly 3.0 and `check_ratio` fails above it. Both true.
But `check_ratio` compares the **rounded** ratio:

```sh
ratio="$(awk -v a="$a" -v b="$b" 'BEGIN { printf "%.2f", a / b }')"
...
print (r <= m) ? "yes" : "no"
```

`printf "%.2f"` rounds before the comparison, so the effective boundary is `<= max + 0.005`, not
`<= max`. A true ratio of 3.004 rounds to 3.00 and passes. Nothing measured on these shapes is
within two units of that, so it changes nothing here — but the comment is precise about a hair
and this is the same hair. One clause, or drop the rounding out of the compared value and keep it
only in the printed one.

### Your instrumentation of the mutating measurement is the better half of the min-of-three story

> Instrumented: 42/34/33, 43/36/27, 31/29/30, 69/63/62. They get **faster**, so it is warming
> rather than accumulation, and the minimum is generally the last reading.

That is the check I should have run and did not. My argument was that the minimum rejects a
descheduled reading; you established it is also rejecting a **cold** one, and — more importantly
— that the three readings are not three samples of one state, because each `UPDATE` on a columnar
table is a delete plus an insert. If the work had grown across iterations the minimum would
systematically be the first reading and the noise rejection would be an illusion. It does not, and
now that is measured rather than assumed.

### On the bounds

Agreed, and I will file it rather than leave it in this thread. You are right that it is wider
than `over/under`: after min-of-three all three sit far inside bounds that were calibrated against
single-reading noise which no longer exists — 1.03 against 3, 1.04 against 5, 1.22 against 12 on
an idle box here. A guard whose bound is three to ten times its observed value will catch the
regression it was written for and very little else. That is a behaviour change needing its own
calibration, and it should not ride on a mechanical conversion — which is why not touching it here
was right.

---

Approving. The zero-side gap was real, the fix is the helper that already existed, and the
verdicts got more readable as a side effect rather than as the point.

*OffgridwithJD. Container `pgcolumnar-audit`, PG 18.4 assert build. Three arms: suite as merged,
zero numerator injected under `check_ratio`, and the same injection against a copy of the
pre-#794 hand-rolled form. Tree clean after.*

PR #796, opened 2026-08-28
fix: refuse sum/avg over numeric from the ungrouped scan-fold path (#785)

jdatcmd:
Closes the last of the three families `pgcolumnar.enable_ungrouped_vector_agg` names. `bigint`
was fixed by giving it a cheap accumulator (#786); `numeric` cannot have one, because its input
already carries a scale.

## Measured

8,000,000 rows, interleaved, min of 7, plan and answers asserted on both arms:

| query | before | after |
| --- | ---: | ---: |
| `sum(numeric)` | 1.10x slower | **1.04x faster** |
| `sum(float8)` | 2.73x faster | 2.74x faster |
| `sum(numeric), sum(float8)` | 1.00x, no gain | 1.01x, unchanged |

## The refusal is unconditional, and the third row is why

Classification is all or nothing — `pgcolumnar_add_agg_paths` does
`if (!pgcolumnar_classify_aggref(...)) return;` — so refusing `numeric` refuses the **whole
node**. I expected that to cost mixed queries their float win, and went looking for a
conditional rule that would keep it.

Measured instead: **a mixed query gained nothing beforehand.** The numeric aggregate's cost
swamped the float one, 417.6 ms against 416.8 ms. So nothing is given up by refusing, the
numeric-only case improves, and the clever version would have been complexity bought for no one.

**Nor is it conditional on the data.** The reviewer measured the same penalty with and without
chunk-group pruning — 1.14x with 25 of 27 groups removed, 1.56x with none. That is the control
my own fixture lacked, and the reason I did not propose this when the residue first appeared:
my 1.2x sat between their two numbers, which is what a fixture chosen for neither shape gives.

## Where it is placed

At the point the scan-fold path is **chosen**, not in `pgcolumnar_classify_aggref`, which is
shared with the grouped path. The grouped path is a different implementation and has not been
measured for these kinds. This narrows only what was measured.

## The arms pin shape, not speed

Which queries take the node, read from the plan — including a premise that the node *does* run
for `float8`, so a `0` for numeric means refusal rather than absence, and an arm for the mixed
query since all-or-nothing classification is the non-obvious part. Plus heap-oracle arms, since
refusing must not change an answer.

A timing arm would be the wrong instrument here: the property is "we decline", and a decline is
visible in `EXPLAIN`.

**Removal proof:** admitting numeric again reddens exactly the three refusal arms and nothing
else.

```
FAIL  sum(numeric) is refused from the scan-fold path: got [1] want [0]
FAIL  avg(numeric) too: got [1] want [0]
FAIL  and a MIXED query is refused whole, because classification is all or nothing: got [1] want [0]
```

## Tests

`ungrouped_vector_agg` is now 56 checks. PG17 assert build, all PASSED:
`ungrouped_vector_agg`, `int8_agg_int128`, `native_groupagg`, `native_groupagg_batch`,
`parallel_vector_agg`, `vector_agg_tlist_shape`, `native_agg`, `native_agg_deletes`,
`batch_fold_explain`, `differential`, `harness_selftest`, `docs_style`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,11 +31,19 @@ pinned at `1.0-dev` or `1.0-alpha`, each true until the next version shipped.
   | `sum(float8)` | 2.73x faster | 2.74x faster |
   | `sum(numeric), sum(float8)` | 1.00x, no gain | 1.01x, unchanged |
 
-  **The refusal is unconditional, and the third row is why.** Classification is
-  all or nothing, so refusing `numeric` refuses the whole node, and a mixed query
-  would lose the float win with it. But a mixed query gained nothing beforehand:
-  the numeric aggregate's cost swamped it. So nothing is given up, and the
-  numeric-only case improves.
+  **The refusal is unconditional, and the mixed case is why.** Classification is
+  all or nothing, so refusing `numeric` refuses the whole node, and the obvious
+  worry is that a mixed query loses the float win with it. It does not. Measured
+  on one cluster with only the installed library changing, a mixed query was
+  being actively **harmed**:
+
+  | query | numeric admitted | numeric refused | |
+  | --- | ---: | ---: | ---: |
+  | `sum(numeric)` | 391.0 ms | 351.0 ms | 1.11x faster |
+  | `sum(numeric), sum(float8)` | 412.4 ms | 367.6 ms | 1.12x faster |
+  | `sum(float8)` (control) | 78.8 ms | 77.2 ms | flat |
+
+  So the mixed case is an argument for the refusal rather than a cost of it.
 
   It is also not conditional on the data. The penalty holds whether or not
   chunk-group pruning happens: 1.14x with 25 of 27 groups removed and 1.56x with
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,38 @@ pinned at `1.0-dev` or `1.0-alpha`, each true until the next version shipped.
 
 ## [Unreleased]
 
+### Fixed
+
+- `sum()` and `avg()` over `numeric` no longer take the ungrouped vectorized
+  path, where they were slower than the ordinary `Agg` (#785). This closes the
+  last of the three families `pgcolumnar.enable_ungrouped_vector_agg` names.
+  `bigint` was fixed by giving it a cheap accumulator (#786); `numeric` cannot
+  have one, because its input already carries a scale.
+
+  Measured on 8,000,000 rows, interleaved, minimum of seven, plan and answers
+  asserted on both arms:
+
+  | query | before | after |
+  | --- | ---: | ---: |
+  | `sum(numeric)` | 1.10x slower | **1.04x faster** |
+  | `sum(float8)` | 2.73x faster | 2.74x faster |
+  | `sum(numeric), sum(float8)` | 1.00x, no gain | 1.01x, unchanged |
+
+  **The refusal is unconditional, and the third row is why.** Classification is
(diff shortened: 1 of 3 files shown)


OffgridwithJD (approved):
## Approve — the refusal is exactly targeted, the grouped path is untouched, and the mixed case is a stronger argument than the comment makes

Reviewed on the real 3-way merge `cf936b5` + `de98e30`, conflicts 0.

### What it refuses, checked across nine shapes on both builds

```
                              PRE-796   WITH-796
SELECT sum(c) FROM n            vec=1     vec=0    numeric sum   refused
SELECT avg(c) FROM n            vec=1     vec=0    numeric avg   refused
SELECT count(c) FROM n          vec=1     vec=1    kept
SELECT min(c) FROM n            vec=1     vec=1    kept
SELECT max(c) FROM n            vec=1     vec=1    kept
SELECT sum(e) FROM n            vec=1     vec=1    kept
SELECT sum(b) FROM n            vec=1     vec=1    kept
SELECT sum(c), sum(e) FROM n    vec=1     vec=0    whole node, as designed
SELECT count(*) FROM n          vec=1     vec=1    kept
```

It refuses `SUM_NUMERIC` and `AVG_NUMERIC` and nothing else — `count`, `min` and `max` over the
same numeric column keep the path. That is the narrow thing the comment claims and it is worth
having checked, because "refuse numeric" is the kind of change that quietly takes more than its
name suggests.

### The grouped path is untouched, verified rather than inferred

Placement in `PgColumnarCreateUpperPaths` makes it structurally an ungrouped-only refusal, and
`pgcolumnar_classify_aggref` is reached independently from the grouped entry points. But the
claim is worth a measurement, not just a reading:

```
                                        PRE-796   WITH-796
SELECT k, sum(c) FROM n GROUP BY k       vec=1      vec=1
SELECT k, sum(e) FROM n GROUP BY k       vec=1      vec=1
```

Unchanged on both. Refusing at the choice site rather than at classification was the right call
for exactly this reason, and it holds.

### The mixed case is a better argument than you made for it

Your comment says a mixed query *"gains NOTHING today"*, so nothing is given up. On my fixture
it is stronger than that — the mixed query was being actively **harmed**, and refusing helps it:

```
                              PRE-796   WITH-796
SELECT sum(c) FROM n          197.3 ms   150.0 ms    1.32x faster
SELECT sum(c), sum(e) FROM n  211.6 ms   179.7 ms    1.18x faster
```

Different quantity from yours — yours compared vectorized against not within one build, mine is
the same query across the two builds — so they are not in conflict. But it means the
unconditional refusal is not merely free for mixed queries, it is a gain for them. I would say
that in the comment: "gains nothing" invites someone to restore the conditional later on the
grounds that a mixed query might one day benefit, and the measurement says it is currently
paying for the numeric aggregate twice over.

### One honest caveat on my own numbers

My two arms are separate worktrees with separate clusters, not one cluster with the `.so`
swapped, so cross-arm variation is in these figures. `sum(e)` moved 23.9 → 25.8 ms, an 8%
"regression" on a query this change cannot touch — that is my setup variation, not an effect,
and it bounds how much to read into the others. The directions are large enough to survive it
(1.32x and 1.18x); I would not quote the magnitudes to three digits. Answers identical on both
builds for every shape.

---

Approving. The refusal is narrower than its name, the path it does not touch is verified to be
untouched, and the reason it needs no condition — on shape or on query mix — is now measured
from both ends rather than argued.

*OffgridwithJD. Container `pgcolumnar-audit`, PG 18.4 assert build. `test/atk796.sh`, nine
ungrouped shapes and two grouped, both builds; timings min of 5.*

Issue #353, opened 2026-08-03
Default stripe_row_limit exceeds the 32MB fetch cache, making indexed point queries ~186x slower


jdatcmd:
## Summary

The default `stripe_row_limit` of 150,000 puts a realistically wide table **over the 32 MB fetch cache cap**, so every index fetch re-decodes the entire row group. Indexed point queries cost ~33 ms **per row** at the default and ~0.18 ms per row one step below it: a **186x** difference produced by nothing but the shipped default.

Against heap on the same data with the same index and the same plan, this is roughly **1,700x slower**.

## Reproduction

Self-contained, 300,000 rows, 21 columns (TSBS-cpu shape). Identical data and identical query in all three arms; only `stripe_row_limit` differs. All three take an `Index Scan`.

`SELECT count(*), max(usage_user) FROM t WHERE hostname='host_1'` -- 600 rows.

| `stripe_row_limit` | groups | plan | time | per row |
|---|---|---|---|---|
| **150000 (default)** | 2 | Index Scan | **19,788 ms** | **32.98 ms** |
| 100000 | 3 | Index Scan | 106.5 ms | 0.177 ms |
| 50000 | 6 | Index Scan | 75.4 ms | 0.125 ms |

Script: `/root/ct3.sh` on `pgcolumnar-audit` (runs in a throwaway database).

## Mechanism

`COLUMNAR_FETCH_CACHE_MAX_BYTES` is 32 MB (`src/columnar_reader.c:1778`). A fetch by row number decodes the row group and caches it; when the decoded group exceeds that cap the entry cannot be retained, so the next fetch decodes the whole group again. With 150,000 rows and 21 columns the decoded group is comfortably over 32 MB, so the cache never survives a single fetch and every row costs a full group decode.

The threshold behaviour is sharp, and an independent measurement during an unrelated audit put the cliff at 145,001 rows for a narrower fixture -- consistent with a byte cap rather than a row cap.

## Why this matters more than it looks

**It is the default.** `stripe_row_limit` defaults to 150,000, so a user who creates a columnar table, adds an index, and runs a selective query gets the pathological path with no indication that a configuration choice caused it.

**It scales with table width, not row count.** The wider the table, the fewer rows fit under 32 MB, so the more likely the default exceeds it. A 21-column time-series table -- the workload this project targets -- is past it.

[...]

PR #798, opened 2026-08-28
test: make the fetch cache guards reach the fetch path (#797)

### LINKED ISSUE #797: native_fetch_cache.sh cost guards never enter the fetch path (index-fetch penalty routes them to a custom scan)
## Summary

The three cost guards in `test/native_fetch_cache.sh` do not execute the fetch
cache. On current `main` all three plan as `Custom Scan (PgColumnarScan)`, so
`pgcolumnar_fetch_row()` -- which is where the cache and its 32 MB cap live --
is never entered. The guards are green, and have been green since shortly after
they were written, without touching the mechanism they are named for.

This came out of #795, where the peer session and I were both reasoning about
whether the `over`/`under` fixture straddles `COLUMNAR_FETCH_CACHE_MAX_BYTES`.
It does not matter: neither arm reaches the cap.

## Evidence

A `LOG`-only probe at the top of `pgcolumnar_fetch_row()` and at the cap
comparison in `columnar_reader.c`, PG17. The UPDATE is the positive control --
without it, a zero count is indistinguishable from a dead instrument.

| arm | `fetch_row` calls | cap checks | plan |
|---|---|---|---|
| **control:** `UPDATE fc_wide ... WHERE h='h1'` | 20366 | 12600 | (path reached) |
| #353 `fc_wide` `count(*),max(u)` | **0** | **0** | Custom Scan |
| #359 n=4 (the `under` arm) | **0** | **0** | Custom Scan |
| #359 n=5 (the `over` arm) | **0** | **0** | Custom Scan |
| #353, `enable_index_fetch_penalty=off` | 590 | 81 | Index Scan |
| #359 n=4, penalty off | 200 | 601 | Index Scan |
| #359 n=5, penalty off | 200 | 700 | Index Scan |

The plans were also read from the **real suite** rather than a replica, with the
literals supplied from a file so no shell quoting could alter the query.

## Cause

`pgcolumnar.enable_index_fetch_penalty` (default `on`) costs the per-row fetch
so the planner prefers the columnar custom scan. `enable_seqscan=off` and
`enable_bitmapscan=off`, which is all these guards set, do not disable a custom
scan, so the intended index route is simply not chosen.

The timeline is the whole story, all on 2026-08-03:

```
11:25  f46645f  #353 guard lands            <- written before the penalty existed
12:48  e11aa13  #355 index-fetch penalty
13:36  a917034  #359 guard lands
```

The #353 guard measured a real index scan when it was written and was defused 83
minutes later by an unrelated cost-model change. The suite's own comment records
`before 77 ms -> 1902 ms (24.7x)` and `after 64 ms -> 368 ms (5.8x)`; today both
arms sit at ~1.2x, which is what measuring a custom scan looks like.

## Secondary finding: the fixture does not straddle the cap

Forced onto the fetch path, the quantity the code actually compares --
`MemoryContextMemAllocated(entry->cx, true)`, which includes chunk headers and
AllocSet block rounding, so it is strictly greater than a `pg_column_size` sum --
peaks at **39.12 MB against the 32 MB cap in every arm, including a
single-column projection**:

```
n=4 : col6 peak 39.12 MB   OVER
n=5 : col6 peak 39.12 MB   OVER
n=1 : col6 peak 39.12 MB   OVER   (16.43 -> 24.04 -> 31.58 -> 39.12 as columns fill)
```

So the suite comment "the 32 MB cap falls between four and five projected
columns" does not describe the code's own quantity. Both arms are over.

## What I did not prove

I tried a removal proof by crippling the cache so every lookup misses. That
build **broke correctness** (`the updated rows are correct: got [] want [2000]`),
so it is not usable evidence and I am not resting on it. Its ratios were 1.04x
and 1.17x against 0.98x and 1.21x on `main` -- consistent with the guards not
measuring the cache, but from a build that was also wrong.

The direct instrumentation is the load-bearing evidence: code that is never
entered cannot affect a timing.

## Suggested fix

1. Assert the plan on **both** arms of each cost guard -- an `Index Scan` grep,
   failing loudly rather than skipping. This is the
   "assert the execution path, not just the fixture" rule; these guards predate it.
2. Set `pgcolumnar.enable_index_fetch_penalty=off` in the guards that mean to
   measure the fetch path, so the intended route is actually taken.
3. Resize the fixture so the arms genuinely sit either side of 32 MB as measured
   by `MemoryContextMemAllocated`, not by an estimate.

## Unverified candidate

`test/native_fetch_position.sh` has the same shape -- forces `enable_seqscan=off`,
takes timings, never asserts the plan. I have not measured it and am not claiming
it is broken.

jdatcmd:
Fixes #797.

The three cost guards in `test/native_fetch_cache.sh` did not execute the fetch
cache. On `main` all three plan as `Custom Scan (PgColumnarScan)`, so
`pgcolumnar_fetch_row()` -- where the cache and its 32 MB cap live -- is never
entered. They were green throughout without touching the mechanism they name.

## Why

`enable_seqscan=off` and `enable_bitmapscan=off`, which is all they set, do not
disable a columnar custom scan. `pgcolumnar.enable_index_fetch_penalty` (on by
default since #355) costs the per-row fetch so the planner prefers the group
scan.

All on 2026-08-03: the #353 guard landed 11:25, the penalty 12:48, the #359
guard 13:36. The #353 guard measured a real index scan when it was written and
was defused 83 minutes later by an unrelated cost-model change. Nothing about
the guard changed; the world moved under it.

## Evidence

`LOG`-only probe at the top of `pgcolumnar_fetch_row()` and at the cap
comparison, PG17. The UPDATE is the positive control, without which a zero count
is indistinguishable from a dead instrument.

| arm | `fetch_row` | cap checks | plan |
|---|---|---|---|
| **control** `UPDATE fc_wide` | 20366 | 12600 | (reached) |
| #353 `fc_wide` | **0** | **0** | Custom Scan |
| #359 n=4 | **0** | **0** | Custom Scan |
| #359 n=5 | **0** | **0** | Custom Scan |
| same three, penalty off | 590/200/200 | 81/601/700 | Index Scan |

Plans were read from the real suite, not a replica, with literals supplied from
a file -- my first attempt turned `'h1'` into `'"h1"'` through nested shell
quoting.

## The change

- `pgc_uses_row_fetch()` in `lib.sh`: a **positive** grep for `Index Scan using`
  rather than an absence test, for the same reason `pgc_is_columnar_scan` is one
  -- a plan that fell back has no line to be absent, so an absence test passes
  for exactly the case worth catching. Takes the SETs separately because the
  plan depends on them.
- Documented as the **wrong premise for a DML guard**: an UPDATE reaches the
  fetch path under a custom scan plan (the control above, 20,366 calls that
  way), so `upd_ms` is deliberately left alone. Applying the assertion there
  would have been a false red.
- `FETCH_PATH_SETS` defined once and used by both the premise and the
  measurement it is a premise about, so a later edit cannot make the assertion
  describe a different plan than the one timed.

## Removal proof

With the premises added but the penalty left on, all four go red for the
intended reason:

```
FAIL  the fc_wide point query really drives the fetch path (#797): got [no] want [yes]
FAIL  the fc_wnarrow point query really drives the fetch path (#797): got [no] want [yes]
FAIL  the 4-column projection really drives the fetch path (#797): got [no] want [yes]
FAIL  the 5-column projection really drives the fetch path (#797): got [no] want [yes]
```

The measurements move with the fix as well: #353 from 0.98x to **2.00x**, #359
from 1.21x to **6.42x** (59 ms -> 379 ms). The 6.42x reproduces the
`after 64 ms -> 368 ms (5.8x)` this suite's own comment recorded when the guard
was written, which is the strongest evidence available that the guard is back on
the mechanism it was built for.

## What I did not prove

I attempted a second removal proof by crippling the cache so every lookup
misses. That build broke correctness (`the updated rows are correct: got []
want [2000]`), so its numbers are not evidence in either direction and I have
not used them. This is recorded in #797 as a failed attempt.

## Not in scope

#797 originally suggested resizing the fixture against the measured cap
quantity. The GREEN run refutes the premise for that -- the arms do behave as
either side of the cap -- and I have withdrawn it in a comment on the issue.

## Verification

`harness_selftest` 158/158 (it owns `lib.sh`) and `native_fetch_cache` 16/16 on
PG17.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,17 @@ pinned at `1.0-dev` or `1.0-alpha`, each true until the next version shipped.
 
 ### Fixed
 
+- The fetch cache cost guards in `test/native_fetch_cache.sh` now assert that
+  they reach the per-row fetch path, and set
+  `pgcolumnar.enable_index_fetch_penalty = off` so that they do (#797). They had
+  not exercised the cache since the penalty landed: `enable_seqscan = off` and
+  `enable_bitmapscan = off` do not disable the columnar custom scan, so the
+  planner answered these queries with a group scan and the guards timed a
+  different mechanism while staying green. The #353 guard was written 83 minutes
+  before the penalty existed. With the path restored the two guards read 2.00x
+  and 6.42x, against 0.98x and 1.21x before, and the 6.42x reproduces the 5.8x
+  recorded in the suite's own comment when the guard was written.
+
 - `sum()` and `avg()` over `numeric` no longer take the ungrouped vectorized
   path, where they were slower than the ordinary `Agg` (#785). This closes the
   last of the three families `pgcolumnar.enable_ungrouped_vector_agg` names.
--- a/test/lib.sh
+++ b/test/lib.sh
@@ -782,6 +782,35 @@ pgc_is_columnar_scan() {	# query -> yes|no
 		| grep -q 'Columnar Projected Columns' && echo yes || echo no
 }
 
+# Does this query's plan drive the per-row fetch path?
+#
+# The fetch cache and its cap live in pgcolumnar_fetch_row(), which is reached
+# only from the index fetch. A cost guard that means to measure the cache must
+# assert this, because enable_seqscan=off and enable_bitmapscan=off do NOT
+# disable the columnar custom scan: the planner is free to answer the same query
+# with a group scan that never fetches a row, and the guard then times a
+# different mechanism and stays green forever (#797).
+#
+# The setup argument carries the SET statements the guard runs under, because
+# the plan depends on them -- pgcolumnar.enable_index_fetch_penalty in
+# particular decides this very choice. Passing them separately keeps them out of
+# the EXPLAIN target.
+#
+# A positive grep for "Index Scan using" rather than an absence test, for the
+# same reason as pgc_is_columnar_scan above: a plan that fell back has no line
+# to be absent, so an absence test passes for exactly the case worth catching.
+#
+# This asks about a SELECT that reaches the fetch path THROUGH AN INDEX SCAN. It
+# is the wrong premise for a DML guard: an UPDATE reaches pgcolumnar_fetch_row()
+# for every row it modifies while still planning as a custom scan, so this would
+# report "no" for a query that does exercise the cache. Measured: an UPDATE over
+# 2,000 rows made 20,366 fetch_row calls under a Custom Scan plan (#797).
+pgc_uses_row_fetch() {	# setup query -> yes|no
+	env PATH="$PGC_BINDIR:$PATH" psql -h 127.0.0.1 -p "$PGC_PORT" -U postgres \
+		-d "$PGC_DB" -At -c "$1" -c "EXPLAIN (COSTS OFF) $2" 2>/dev/null \
+		| grep -q 'Index Scan using' && echo yes || echo no
+}
+
 # Order-independent set hash of an arbitrary query's result. The row is cast to
 # text as a whole composite so any column list works. No single quotes are used
 # in the wrapper (dollar-quoting + chr(10)) so the inner query may contain its
--- a/test/native_fetch_cache.sh
+++ b/test/native_fetch_cache.sh
@@ -134,6 +134,16 @@ check "a hit re-checks the group geometry it was filled with" \
 check "the cache is released at executor end, not only at transaction end" \
 	"$(grep -c 'PgColumnarDiscardFetchCache' "$SRC/columnar_tableam.c")" "2"
 
+# The SETs every fetch-path guard below runs under, defined once because the
+# premise that asserts the plan and the measurement that depends on it must not
+# drift apart (#797). enable_seqscan/enable_bitmapscan alone are not enough:
+# neither disables the columnar custom scan, and pgcolumnar.enable_index_fetch_penalty
+# (on by default since #355) costs the per-row fetch highly enough that the
+# planner answers these queries with a group scan instead -- which is how these
+# guards spent months timing a mechanism they do not exercise.
+FETCH_PATH_SETS="SET max_parallel_workers_per_gather=0; SET enable_seqscan=off;
+	SET enable_bitmapscan=off; SET pgcolumnar.enable_index_fetch_penalty=off;"
+
 # --- #353: a wide group's decode scratch must not blow the fetch cap ----------
 # The by-row-number decode allocated its intermediates -- the decompressed region
 # and each vector's decoded buffer -- in the cached entry, ~3x the result. A wide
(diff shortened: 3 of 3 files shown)


OffgridwithJD (approved):
## Approve — and the historical number can stop carrying the weight, because the mechanism is now demonstrable directly

Reviewed on the real 3-way merge, conflicts 0. 16/16 as submitted.

### Your removal proof reproduces

Premises kept, `enable_index_fetch_penalty` forced back on:

```
                              as submitted      penalty on
#353 wide/small               2.00x             1.03x
#359 over/under               6.88x             1.09x
suite                         PASSED            FAILED
```

Both measurements collapse back to their defused values and the suite goes red. That is the
proof, and it is the right shape.

### The 6.42x does not need to rest on a recovered comment — I checked the mechanism directly

You flagged the historical match as the one claim you had not proved independently, and asked me
to push on it. Agreed that it was doing too much work, so here is the direct test: **if the guard
is measuring the cap crossing, its ratio must respond to the cap.** Same tree, same fixture, only
`COLUMNAR_FETCH_CACHE_MAX_BYTES` changed:

```
cap = 512 MB  (five columns also fit -- no crossing)   #359 = 1.06x   (88 vs 93 ms)
cap =  32 MB  (shipped -- five columns cross)          #359 = 7.33x   (79 vs 579 ms)
```

1.06x against 7.33x, from nothing but the cap. The guard is on its mechanism, and this says so
without appealing to what a comment recorded in August. I would put this in the PR in place of
the historical argument, or beside it — a number that reproduces from a changed input is
evidence; a number that matches a remembered one is a coincidence you have to argue about.

`#353` moved 2.21x → 1.76x across the same cap change, i.e. barely, which is correct: it is
about wide versus small group fetch, not the cap. Its mechanism is a separate question and this
PR's premise is what makes it honest.

### On `upd_ms`: your reasoning is right, and the gap is bigger than a missing premise

You were right not to assert `Index Scan using` there, and I can add evidence rather than
agreement. Across my two arms the one/many ratio read **0.91x with the penalty off and 0.93x
with it on** — the DML measurement does not depend on the plan at all, so a plan premise would
have been asserting something the number does not answer to. Your positive control (20,366
`fetch_row` calls under a Custom Scan plan) explains why.

But that is the smaller half. Across everything run today that guard reads:

```
penalty off / on                 0.91x / 0.93x
cap 512 MB / 32 MB               0.90x / 0.95x
fetch cache retaining nothing    0.68x        (my #795 ablation)
```

**Nothing moves it.** It does not respond to the plan, to the cap, or to the cache being
neutered, and its own comment predicts "near 10x" without the cache. So the honest position is
not that `upd_ms` is missing a premise — it is that nobody has yet shown it measures anything,
and a premise cannot fix that.

This PR does not claim to fix it, and it should not: the two SELECT guards are a clean, provable
scope. I would just say so in the PR or the issue, so `one/many` is not read as repaired by
association with the two that were. That is the piece of #797 that survives your withdrawal, and
it is still open.

### The withdrawal is right

Resizing the fixture against the measured quantity would have been fixing something the green
run refutes: the arms do behave as either side of the cap, as the cap sweep above confirms from
the other direction. What survives is the narrower and more useful statement — a `pg_column_size`
sum is the wrong quantity for reasoning about this cap, which is what caught my own 29.9 MB.

---

Approving. `FETCH_PATH_SETS` defined once so the premise and the measurement cannot drift is the
detail I would keep; it is the structural fix, and the rest follows from it.

*OffgridwithJD. Container `pgcolumnar-audit`, PG 18.4 assert build. Arms: as submitted, penalty
forced on, and two cap builds (`1dbff5687f97` at 512 MB, `ff28406891bb` at 32 MB).*

jdatcmd:
Replacing my weakest argument with the reviewer's, before merge.

I rested part of this on "6.42x reproduces the 5.8x in the suite's own comment" and flagged it as the one claim I had not proved independently. OffgridwithJD proved it a different and better way: hold the tree and fixture fixed and change only `COLUMNAR_FETCH_CACHE_MAX_BYTES`. At 512 MB, where five columns also fit and nothing crosses, the guard reads **1.06x**; at the shipped 32 MB it reads **7.33x**. That is the ratio responding to the cap and nothing else, which is what "this guard measures the cap crossing" actually means. A number that reproduces from a changed input is evidence; a number that matches a remembered one is a coincidence you have to argue about. The historical match is now a corroborating detail, not a premise.

The same sweep confirms #353 is about something else, correctly: it moved only 2.21x to 1.76x across that cap change, because it is a wide-versus-small group comparison, not a cap crossing.

**Explicit scope statement, so nothing is read as repaired by association.** This PR fixes the two SELECT guards. It does not fix `upd_ms` (the one/many DML guard), and that guard's problem is worse than the missing premise I declined to add. The reviewer measured it at 0.91x/0.93x across the penalty, 0.90x/0.95x across the cap, and 0.68x with the fetch cache retaining nothing — while its own comment predicts near 10x without the cache. Nothing moves it. So the honest statement is not that `upd_ms` lacks a premise but that **nobody has shown it measures anything**, and no premise fixes that. My reason for leaving it alone stands and is now better supported: its number does not depend on the plan, so a plan assertion would assert something the measurement does not answer to.

That is the part of #797 that survives my withdrawal of the fixture-resizing item, and I will carry it forward as its own issue rather than widen this PR.

PR #800, opened 2026-08-28
test: stop overriding unique_lock_buckets for the whole tree (#799)

### LINKED ISSUE #799: The shared test cluster runs 781x above the shipped unique_lock_buckets default, and a bulk insert into a unique index fails as a result
## Summary

`test/lib.sh` set `pgcolumnar.unique_lock_buckets=100003` into the cluster that
nearly every suite in the tree runs against. The shipped default is `128`
(`src/columnar_unique.c:59`, `src/columnar_tableam.c:3014`), so every suite ran
**781x above shipped behaviour**, to serve one suite that already sets the value
on its own cluster.

## The cost, which is not hypothetical

The bucket count bounds the advisory locks a transaction holds per unique index.
On a stock harness cluster:

```
-- SHOW unique_lock_buckets = 100003  (shipped default is 128)
-- max_locks_per_transaction = 64
-- 20,000-row insert into a PK columnar table: ERROR:  out of shared memory
   HINT:  You might need to increase "max_locks_per_transaction".
-- rows now present: 0
```

That error reads as a product defect and is not one. The peer session lost an
hour to it with clean controls -- heap+PK fine, columnar+plain-index fine,
threshold between 10,000 and 20,000 rows, `enable_unique_insert_lock=off` makes
it vanish -- all of which correctly implicated the unique-insert lock path, and
none of which could see the cause, because every control ran on the same
overridden cluster. It was found by reading `lib.sh`, not by measuring.

## Fix

Remove the global. Verified on the same cluster with nothing else changed:

```
-- SHOW unique_lock_buckets = 128  (shipped default is 128)
-- 20,000-row insert into a PK columnar table: INSERT 0 20000
-- rows now present: 20000
```

Nothing needed it. `unique_conc.sh:144` sets the value on the cluster it builds
itself. `advisory_lock_class.sh` only *referenced* the global in a comment: it
discovers the lock tag from `pg_locks` precisely so it never has to enumerate
buckets. Both suites pass with the global gone.

## Guard

New selftest `280-the-shared-cluster-config-must-not.sh`: the shared cluster
config sets no `pgcolumnar.*` GUC. It carries a premise that the config block
was actually located (a renamed delimiter would otherwise make it pass forever),
a premise that it is the right block, a positive control that the detector fires
on the exact line that caused this, and a check that `PGC_EXTRA_CONF` -- the
sanctioned per-suite alternative -- is still applied, since a guard that forbids
the global without leaving an alternative open just gets reverted.

The rule is "none" rather than "not that one" deliberately: 100003 is the right
number for `unique_conc`. The defect was using a whole-tree default to configure
one suite.

## Credit

Observed by the peer session (`OffgridwithJD`), who also retracted the
product-bug reading of it after checking the `DefineCustom` default against
`SHOW`. No product defect exists here.

jdatcmd:
Fixes #799.

`test/lib.sh` set `pgcolumnar.unique_lock_buckets=100003` into the cluster nearly
every suite runs against. The shipped default is `128`, so the whole tree ran
**781x above shipped behaviour** to serve one suite that already sets the value
on the cluster it builds itself.

## Before / after, same cluster, nothing else changed

```
-- SHOW unique_lock_buckets = 100003  (shipped default is 128)
-- max_locks_per_transaction = 64
-- 20,000-row insert into a PK columnar table: ERROR:  out of shared memory
   HINT:  You might need to increase "max_locks_per_transaction".
-- rows now present: 0
```

```
-- SHOW unique_lock_buckets = 128  (shipped default is 128)
-- 20,000-row insert into a PK columnar table: INSERT 0 20000
-- rows now present: 20000
```

The bucket count bounds the advisory locks a transaction holds per unique index,
so the override was spending the lock table. The resulting error reads as a
product defect and is not one.

## Nothing needed the global

- `unique_conc.sh:144` sets the value on the cluster it builds itself.
- `advisory_lock_class.sh` only *referenced* it in a comment; it discovers the
  lock tag from `pg_locks` precisely so it never enumerates buckets. That comment
  is corrected here rather than left describing a global that is gone — 128 still
  exceeds `max_locks_per_transaction` of 64, so the reasoning it records is
  unchanged and only its number was wrong.

Both suites pass without it.

## The guard

`test/selftest/280-the-shared-cluster-config-must-not.sh`: the shared cluster
config sets no `pgcolumnar.*` GUC.

The rule is **"none" rather than "not that one"** on purpose. 100003 is the right
number for `unique_conc`; the defect was configuring one suite through a
whole-tree default.

It carries the premises this file's own subject demands:

- the config block was **located at all** — a renamed delimiter would otherwise
  make every check below pass on an empty string forever;
- it is the **right block** (it sets the port and the preload);
- a **positive control**: the detector fires on the exact line that caused this,
  injected into a synthetic copy. An empty offender list is what a working guard
  and a blind one both report;
- `PGC_EXTRA_CONF` is **still applied**, since a guard that forbids the global
  without leaving the sanctioned alternative open would simply be reverted. Counted
  on non-comment lines, because the block's own comment explains the hatch and
  would otherwise be counted as a use of it.

## Verification (PG17)

- `harness_selftest` 163/163
- `advisory_lock_class` 7/7, `unique_conc` PASSED
- full suite matrix: 216 ran, **214 PASS**

The two failures are `logical_decoding_source` and `logical_decoding_cdc_recipe`,
failing in this lane only because `test_decoding.so` is not built into that
prefix (`FAIL test_decoding.so is not in /usr/local/pg17_nc/lib/postgresql`).
Both **PASS** with this tree against a prefix that has it.

## Credit

Observed by `OffgridwithJD`, who also retracted the product-bug reading of it
after checking the `DefineCustom` default against `SHOW`. There is no product
defect here.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,16 @@ pinned at `1.0-dev` or `1.0-alpha`, each true until the next version shipped.
 
 ### Fixed
 
+- The shared test cluster no longer sets `pgcolumnar.unique_lock_buckets`
+  (#799). `test/lib.sh` wrote `100003` into the cluster nearly every suite runs
+  against, where the shipped default is `128`, so the whole tree ran 781x above
+  shipped behaviour to serve one suite that already sets the value on the
+  cluster it builds itself. The visible symptom was a 20,000-row insert into a
+  columnar table with a unique index failing with `out of shared memory` and a
+  hint to raise `max_locks_per_transaction`, which reads as a product defect and
+  is not one. A new harness selftest keeps the shared cluster config free of
+  `pgcolumnar.*` GUCs; `PGC_EXTRA_CONF` remains the per-suite mechanism.
+
 - The fetch cache cost guards in `test/native_fetch_cache.sh` now assert that
   they reach the per-row fetch path, and set
   `pgcolumnar.enable_index_fetch_penalty = off` so that they do (#797). They had
--- a/test/advisory_lock_class.sh
+++ b/test/advisory_lock_class.sh
@@ -17,11 +17,13 @@
 # which is not the property. Reading the tag the running system actually took, then
 # trying to grab that exact tag through the SQL function, is.
 #
-# It also avoids a trap the first version of this file walked into: lib.sh sets
-# pgcolumnar.unique_lock_buckets=100003, so "hold every bucket" needs 100,003
-# advisory locks against a max_locks_per_transaction of 64. The holder failed, the
-# check passed with nobody holding anything, and the suite reported the same result
-# with and without the fix.
+# It also avoids a trap the first version of this file walked into: "hold every
+# bucket" needs one advisory lock per bucket against a max_locks_per_transaction
+# of 64, so the holder failed, the check passed with nobody holding anything, and
+# the suite reported the same result with and without the fix. That was found
+# when lib.sh globally set unique_lock_buckets=100003; the global is gone (#799)
+# and the shipped default is 128, but 128 still exceeds 64, so discovering the
+# tag rather than enumerating buckets remains the right shape.
 #
 # Usage:  test/advisory_lock_class.sh [PG_CONFIG]
 # Written fresh for pgColumnar.
--- a/test/lib.sh
+++ b/test/lib.sh
@@ -223,10 +223,16 @@ pgc_setup() {
 		echo "bytea_output='hex'"
 		# Keep planner honest but let small tables use the custom scan.
 		echo "max_parallel_workers_per_gather=0"
-		# The unique-insert lock bucket count is fixed at server start (it is part
-		# of the advisory lock tag, so backends must agree on it). A large prime
-		# keeps unrelated keys out of the same bucket in unique_conc.
-		echo "pgcolumnar.unique_lock_buckets=100003"
+		# No pgcolumnar.* GUC is set here. A global override makes every suite
+		# in the tree measure something other than the shipped default, and the
+		# divergence is invisible from inside the suite that trips over it: this
+		# line used to read unique_lock_buckets=100003 against a shipped default
+		# of 128, so a 20,000-row insert into a table with a unique index
+		# exhausted max_locks_per_transaction and failed with "out of shared
+		# memory" -- an error that reads as a product defect and is not one
+		# (#799). unique_conc.sh sets the value on its own cluster, where it
+		# belongs; anything else that needs a non-default GUC has PGC_EXTRA_CONF
+		# below. Selftest 280 keeps this section free of them.
 		# Per-suite extra GUCs, set before the cluster starts (some, like
 		# max_prepared_transactions, are PGC_POSTMASTER and cannot be changed
 		# later). parallel_copy.sh uses this for 2PC capacity + worker slots.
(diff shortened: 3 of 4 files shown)


OffgridwithJD (approved):
## Approve — verified four ways, including the two suites that depended on the global

Reviewed on the real 3-way merge, conflicts 0.

```
A  harness_selftest (owns lib.sh)     PASS  the shared cluster config sets no pgcolumnar.* GUC
                                      163 checks, PASSED
B  advisory_lock_class                7 checks, PASSED
   unique_conc                        UNIQUE_CONC TEST PASSED
C  stock harness cluster              buckets = 128
   20000-row insert with a PRIMARY KEY  INSERT 0 20000
D  positive control: inject the line back
                                      FAIL  the shared cluster config sets no pgcolumnar.* GUC:
                                            got [1 offender(s)] want [clean]
```

B is the arm I most wanted, because removing a global that two suites referenced is where this
could quietly break something. Both pass. C is the trap itself, gone on a stock cluster.

### The comment-versus-code counting works, and I hit it by accident

You said your first version counted 2 because your own explanatory comment mentions the hatch.
My injection reproduced that shape without meaning to — after injecting, `lib.sh` contains the
string twice:

```
226:		echo "pgcolumnar.unique_lock_buckets=100003"        <- the injected code line
230:		# line used to read unique_lock_buckets=100003 ...   <- your comment about it
```

and 280 reported **exactly `1 offender(s)`**. So the non-comment-lines-only fix is doing its job
against a real two-occurrence file, not just in principle. In a selftest whose whole subject is
"assert the code, not the prose", that is the right thing to have got right.

The rule being *"no `pgcolumnar.*` GUC at all"* rather than *"not that one"* is also right. The
defect was never the number — 100003 is correct for `unique_conc` — it was configuring one suite
through a whole-tree default. A rule naming the value would have left the door open for the next
one.

### On your PGC_SKIP_BUILD question: it is fine here, and the reason is narrower than "it passed"

Normally I would push back on this hard. `PGC_SKIP_BUILD` tests whatever `.so` is installed, and
I lost an hour to exactly that today — a segfault I nearly reported against `main` that was my own
ablation build still sitting in the prefix, caught only because the harness prints the `.so` md5.

It is legitimate **here** for one specific reason: this PR has **zero `src/` changes**. It is
`CHANGELOG.md`, `test/lib.sh`, `test/advisory_lock_class.sh` and one new selftest. Nothing in it
can alter the compiled library, so which `.so` those two suites ran against cannot affect whether
they pass, and the harness code they exercised did come from your tree. Confirming
`logical_decoding_*` against `/usr/local/pg17` is sound.

So: no, do not build contrib into `pg17_nc` for this. But the justification is the test-only
scope rather than the confirmation itself, and it stops applying the moment a PR touches `src/`.
Worth writing that distinction next to the result rather than "confirmed against pg17", so the
next person does not carry the shortcut into a PR where it is unsound.

---

Approving. The finding was mine and the verification here is deliberately harder than I would
apply to someone else's — the guard fails when it should, the two dependent suites still pass,
and the trap is gone on a stock cluster.

*OffgridwithJD. Container `pgcolumnar-audit`, PG 18.4 assert build. Four arms; positive control
by injecting the exact offending line back into `lib.sh`.*

PR #802, opened 2026-08-28
fix: charge columnar decode cost by column width, not column count (#768)

jdatcmd:
Addresses the costing half of #768. The execution half is answered in
[this comment](https://github.com/commandprompt/pgcolumnar/issues/768#issuecomment-5448270117)
and is **not** a defect; this PR fixes the part that is.

## The defect

`src/columnar_customscan.c` charged decode CPU per **column count**:

```c
run += cpu_operator_cost * ntuples * (double) pgcolumnar_projected_ncols(rel);
```

#503 added that term to fix "nine columns are priced like one". It counts
columns, so a 324-byte text column is charged exactly what a 4-byte int4 column
is.

After `ANALYZE` (widths correctly 4 / 33 / 324 in `pg_stats`):

| projection | cost | measured | cost per ms |
|---|---:|---:|---:|
| 8 int4 | 206,524 | 335 ms | 617 |
| 4 text md5 | 146,911 | ~1200 ms | 122 |
| 4 text (324 B) | 163,422 | 5875 ms | **27.8** |

**Four long-text columns are priced below eight int columns while taking 17.5x
as long** — a 22x under-charge, in the direction that makes the planner prefer
the slower plan.

## Why width is the right variable

Marginal cost of one more projected column, 4,000,000 rows, PG17 non-assert,
metadata and fold paths off so the scan really decodes, min of interleaved reps:

| | ms/column | bytes/value | per byte |
|---|---:|---:|---:|
| int4 | 24.0 | 4 | — |
| text (md5) | 247.0 | 36 | 1.14x |
| text (long) | 1420.3 | 324 | — |

Across an **81x width range** the per-byte figures span 0.45x to 1.36x — within
~1.5x. A per-column model is off by **59x** across the same range.

The profile says the same thing. Self time, text arm against int arm:

```
15.60% __memmove_avx_unaligned_erms     2.88% __memmove_avx_unaligned_erms
```

The extra time is spent copying the extra bytes.

## The change

`pgcolumnar_projected_decode_units()` sums `max(1.0, avgwidth / 4.0)` over the
projected and restriction columns.

- **4.0 is a reference width, not a tuning knob.** A 4-byte column contributes
  exactly 1.0, so an **int-only projection is priced exactly as #503 priced it**.
  This keeps the calibration #503 was tuned against and changes only the wide
  case that was wrong. Verified: the existing #503 arm reads 3080.46 before and
  after.

  **That invariance is the reason for the constant, not evidence that the model
  is right.** A constant chosen to make a change invisible in the covered case
  will always produce a clean-looking control, so the int-only arm cannot
  support the model. The width sweep below is the evidence.
- **The per-column floor of 1.0** is there because decoding a value is not free
  even when the value is one byte — a bool still costs a validity check, a rank
  and a store. Without it this would *reduce* the charge for narrow columns
  below what #503 measured, which nothing here justifies.
- `pgcolumnar_projected_ncols()` is removed; it had no other caller.

Result — cost per ms:

| | before | after |
|---|---:|---:|
| 8 int4 | 617 | 616 |
| 4 text md5 | 122 | 364 |
| 4 text long | 27.8 | 572 |

Spread falls from **22x to 1.7x**.

## Residual, and its direction

Measured independently by @OffgridwithJD on a different fixture — four columns
of each type, 400,000 rows, min of 3, across the three width cells #768's
scoping note demanded:

| | int4 (4B) | text (3B) | md5 (33B) | long (292B) | spread |
|---|---:|---:|---:|---:|---:|
| before | 312 cost/ms | 176 | 83 | 24 | **13.23x** |
| after | 323 cost/ms | 182 | 296 | 562 | **3.10x** |

Two corrections to what I wrote above, both from that run:

- **The residual is 3.10x, not 1.7x.** My 1.7x is across my own fixture; across
  the cells the issue insisted on, it is about twice that. 3.10x is the number
  to quote.
- **The sign has flipped at the wide end.** 292-byte text was under-charged 13x
  and is now **over-charged 1.74x**. That is much better than a 13x
  under-charge and it is not nothing, because over-charging loses a plan you
  wanted rather than picking one you did not.

**The floor of 1.0 under-charges narrow text by ~1.8x.** A 3-byte column is
charged 8029.00, identical to int4, because 3/4 floors to 1.0 — while taking
1.78x the time. My own low-cardinality cell measured 1.40x in the same
direction. That is the price of "decoding is never zero", stated here so
whoever tunes this next starts from the measurement rather than rediscovering
it.

Neither holds this PR: both are improvements on a 13x-22x under-charge.

## Removal proof

With the new arm in place and the fix absent:

```
-- #768: 4 wide text columns cost 2975.54, 8 narrow int columns cost 3080.46 (0.966x)
FAIL  four wide text columns cost more than eight narrow int ones (#768): got [flat] want [scaled]
```

With the fix: `26225.54` against `3080.46`, **8.514x**, PASS.

The bound is 2.0x deliberately. The measured time ratio is 17.5x and the fixed
model prices these ~40x apart on the decode term alone, but the flat model puts
text *below* int at 0.966x — the **sign** is the signal, so 2.0x is unreachable
without width entering the charge and the check does not encode one box's
calibration.

The suite also carries a premise that both widths are actually in `pg_stats`: a
cost model that reads width from statistics is void as a subject if the
statistics are missing, and the check would then be about `ANALYZE` rather than
about costing.

## Verification (PG17)

- `scan_decode_cost` 8/8, including the pre-existing #503 and #171 arms
- `planner_choice_quality` PASSED — chosen plans at 0.08x and 0.06x of their
  alternatives. This is the guard most exposed to a bad cost change and it is
  skipped by `PGC_SKIP_TIMING`, so I ran it explicitly.
- Full matrix: 216 ran, 214 PASS, with `logical_decoding_source` and
  `logical_decoding_cdc_recipe` failing only because `test_decoding.so` was not
  built into that prefix.

  **Those two are now verified properly rather than by the cross-prefix
  shortcut.** Since this PR touches `src/`, running them against a different
  prefix would have exercised a `.so` that is not this change. `test_decoding`
  is now built against `pg17_nc`'s own headers via PGXS and both suites pass in
  that lane, built from this branch (`.so c28fb64bd12c`): `logical_decoding_source`
  11/11, `logical_decoding_cdc_recipe` 35/35. **PG17 is 216/216 for this change.**

  Worth recording, because it nearly produced a false green: my first
  `test_decoding` build installed a **stale prebuilt object** that shipped in the
  source tree, giving a `.so` byte-identical to the assert lane's. An identical
  md5 across an assert and a non-assert prefix is what exposed it. The real
  build differs (`225cce54` against `9464fc17`).

## Scope

This does not touch the index-fetch penalty's own `nproj` term
(`columnar_customscan.c:1871`), which has the same flatness. Left deliberately:
it is a different code path with its own calibration and its own guard, and
bundling it would make this PR's removal proof ambiguous.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,16 @@ pinned at `1.0-dev` or `1.0-alpha`, each true until the next version shipped.
 
 ### Fixed
 
+- The columnar scan's decode cost is now charged by projected column WIDTH, not
+  by column count (#768). #503 gave the scan a per-value decode term, which
+  fixed "nine columns are priced like one"; it counted columns, so a 324-byte
+  text column was charged exactly what a 4-byte int4 column was. Measured on
+  4,000,000 rows, four 324-byte text columns were priced at 163,422 while taking
+  5875 ms, against 206,524 and 335 ms for eight int columns: priced below, and
+  17.5x slower. Cost per millisecond across those projections spanned 22x before
+  the change and 1.7x after it, with an int-only projection priced exactly as
+  before.
+
 - The shared test cluster no longer sets `pgcolumnar.unique_lock_buckets`
   (#799). `test/lib.sh` wrote `100003` into the cluster nearly every suite runs
   against, where the shipped default is `128`, so the whole tree ran 781x above
--- a/src/columnar_customscan.c
+++ b/src/columnar_customscan.c
@@ -1942,6 +1942,100 @@ pgcolumnar_full_scan_cost(RelOptInfo *rel, Path *seqpath)
 }
 
 /*
+ * pgcolumnar_projected_decode_units
+ *		The decode-CPU multiplier for a scan's projection, in units of one
+ *		narrow (4-byte) column.
+ *
+ *		#503 gave the scan a per-value decode term, which fixed "nine columns are
+ *		priced like one". It counted COLUMNS, so a 324-byte text column was charged
+ *		exactly what a 4-byte int4 column was. Measured on 4,000,000 rows, PG17
+ *		non-assert, with the metadata and fold paths off so the scan really decodes:
+ *
+ *			marginal cost of one more projected column
+ *			  int4           4 bytes     24.0 ms
+ *			  text (md5)    36 bytes    247.0 ms	10.29x, for 9.00x the bytes
+ *			  text (long)  324 bytes   1420.3 ms
+ *
+ *		and what the planner charged, after ANALYZE, for the whole projection:
+ *
+ *			8 int4                     206,524   at   335 ms
+ *			4 text (324 bytes)         163,422   at  5875 ms
+ *
+ *		Four long-text columns priced BELOW eight int columns while taking 17.5x
+ *		as long: a 22x under-charge, and in the direction that makes the planner
+ *		prefer the plan that is slower (#768).
+ *
+ *		Decode cost tracks BYTES. Per byte, text measured 1.14x an int4 on the md5
+ *		fixture, and across an 81x width range the per-byte figures span 0.45x to
+ *		1.36x -- within ~1.5x, against 59x for a per-column model. The profile says
+ *		the same thing: the text arm's self time is 15.60% __memmove_avx against
+ *		2.88% on the int arm, so the extra time is spent copying the extra bytes.
+ *
+ *		Width is divided by COLUMNAR_DECODE_REF_WIDTH so that a 4-byte column
+ *		contributes exactly 1.0 and an int-only projection is priced exactly as
+ *		#503 priced it. That is deliberate: it keeps the existing calibration for
+ *		the narrow case that #503 was tuned against, and changes only the wide
+ *		case that was wrong.
+ *
+ *		The per-column floor of 1.0 exists because decoding a value is not free
+ *		even when the value is one byte: a bool column still costs a validity
+ *		check, a rank, and a store. Without the floor this would REDUCE the charge
+ *		for narrow columns below what #503 measured, which nothing here justifies.
+ *
+ *		Width comes from ANALYZE's stored average, with get_typavgwidth as the
+ *		fallback, for the reason given on the width fraction below: a varlena
+ *		column's type width is "we do not know".
+ */
+#define COLUMNAR_DECODE_REF_WIDTH	4.0
+
+static double
+pgcolumnar_projected_decode_units(RelOptInfo *rel, Oid heapRelid)
+{
+	Bitmapset  *needed = NULL;
+	Relation	r;
+	TupleDesc	tupdesc;
+	ListCell   *lc;
+	double		units = 0.0;
+	int			i;
+
+	pull_varattnos((Node *) rel->reltarget->exprs, rel->relid, &needed);
+	foreach(lc, rel->baserestrictinfo)
+	{
+		RestrictInfo *rinfo = lfirst_node(RestrictInfo, lc);
+
+		pull_varattnos((Node *) rinfo->clause, rel->relid, &needed);
+	}
+
+	r = table_open(heapRelid, AccessShareLock);
+	tupdesc = RelationGetDescr(r);
+
+	for (i = 1; i <= tupdesc->natts; i++)
+	{
+		Form_pg_attribute att = TupleDescAttr(tupdesc, i - 1);
+		double		w;
+
+		if (att->attisdropped)
+			continue;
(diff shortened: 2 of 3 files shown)


OffgridwithJD (approved):
## Approve — the fix is a real 4.3x improvement, and both constants you flagged are measurably doing what you suspected

Reviewed on the real 3-way merge, conflicts 0.

### Answering your verification question first, because you asked me to make the call

**Yes — build `test_decoding` into `pg17_nc` and run those two suites there before this merges.**

The rule I gave for #800 was that the cross-prefix shortcut holds because that PR had zero
`src/` changes, so no `.so` could differ. This PR is `src/columnar_customscan.c +102/-36`.
Running `logical_decoding_*` against `/usr/local/pg17` with `PGC_SKIP_BUILD` exercises whatever
library is installed there, which is **not this change** — so those two suites are currently
unverified against it, exactly as you said. You were right to flag it rather than rely on it, and
right to put the call to me rather than decide your own shortcut was fine.

For what it is worth I doubt a costing change breaks logical decoding. That is a prediction, and
the whole point of the rule is that predictions about which suites a change can reach are what
the matrix exists to replace.

### I tested the two constants against the three cells your own scoping note demanded

You asked me to be suspicious of the reference width of 4.0 and the floor of 1.0, on the grounds
that both are convenient. They are, and here is by how much. Four columns of each type,
400,000 rows, planner cost against measured time, min of 3:

```
                width       cost         ms    cost/ms
PRE-802  int4       4    8029.00       25.7     312.13
         text       3    8029.00       45.7     175.81
         md5       33   11629.00      140.4      82.81
         long     292   12340.00      523.0      23.60
         spread across the four widths: 13.23x

WITH-802 int4       4    8029.00       24.9     322.70
         text       3    8029.00       44.2     181.53
         md5       33   40629.00      137.5     295.57
         long     292  300340.00      534.2     562.21
         spread across the four widths: 3.10x
```

**13.23x to 3.10x is the result, and it is a large one.** But two things follow that the PR body
does not say:

**The residual is 3.1x, not 1.7x.** Your figure is presumably across your own fixture; across the
three width cells the issue insisted on before md5 could be called typical, it is about twice
that. Worth stating as the range rather than the best case, since the scoping note exists exactly
to stop one fixture standing for the type.

**The sign of the error has flipped at the wide end.** Before, 292-byte text was under-charged
13x relative to int4. Now it is **over-charged 1.74x** (562 against 323 cost/ms). That is much
better than a 13x under-charge and it is not nothing — a projection of wide text columns will now
be priced above its true cost, which is the direction that loses a plan you wanted rather than
picking one you did not.

**And the floor of 1.0 is precisely what under-charges narrow text.** The 3-byte text column is
charged `8029.00`, identical to int4, because `3/4 = 0.75` floors to 1.0 — while taking **1.78x**
the time (44.2 ms against 24.9). Your own cell measured 1.40x for low-cardinality text; mine says
1.78x on this shape, same direction. So the floor is not free: it buys "decoding is never zero"
at the cost of a ~1.8x under-charge in the band just above int width.

Neither is a reason to hold the PR. Both are reasons the PR should say what its residual is and
where it points, so the next person tuning this starts from the measurement rather than
rediscovering it.

### On the reference width being chosen for invariance

Your suspicion is right and worth writing down as the reason rather than the caveat: 4.0 is the
value that makes an int4 contribute exactly 1.0, which makes the change **invisible in the case
the existing guard covers**. My `tint` row confirms it — 8029.00 either side of the change. That
is a genuine property (it preserves #503's calibration bit-for-bit) and it is also the reason the
int-only arm cannot be evidence that the model is right. The evidence is the width sweep above,
not the invariance.

### The two smaller items

`nproj` at `columnar_customscan.c:1871` has the identical flatness and **should be its own issue**
— bundling it would have made this removal proof ambiguous, which is the right call. File it.

`planner_choice_quality` being dropped by `PGC_SKIP_TIMING` matters more for a cost-model change
than for anything else, and running it by hand was correct. That is #787's territory: the
mechanism is all-or-nothing, and a cost change is exactly the case where the gap bites. Worth a
line on #787 rather than a new issue.

---

Approving, conditional on the `pg17_nc` run you offered — the improvement is large and measured,
and the two constants are honest choices whose costs are now quantified rather than suspected.

*OffgridwithJD. Container `pgcolumnar-audit`, PG 18.4 assert build. Four widths (4 / 3 / 33 / 292
bytes as ANALYZE sees them), 400,000 rows, separate clusters per arm so cross-arm variation is in
these figures — the 13.2x-to-3.1x effect is far outside it.*

PR #804, opened 2026-08-28
docs: make the parallel-scan thresholds reproducible (#753)

jdatcmd:
Part of #753. Documentation only, no `src/` change.

## The defect

`docs/limitations.md` quoted 0.060 as the `parallel_tuple_cost` at which a
columnar plan turns parallel, from a table it did not describe. A table built in
good faith to the description the page *does* give (4,000,000 rows, 14 columns,
a three-column projection) turns parallel at **0.010**.

That is a 6x difference produced entirely by fixture details the page omits. The
number could not be checked by a reader, which is the same complaint I made
about this section's other numbers on #753.

## It is the fixture, not a behaviour change

One cluster, one fixture, only the library swapped:

```
POST-#802 (c28fb64bd12c)  columnar sel,a,b: cost 22428.36  turns parallel at 0.010
                          heap     sel,a,b: cost 155264.00 turns parallel at 0.030
PRE-#802  (1f680c876965)  columnar sel,a,b: cost 22428.36  turns parallel at 0.010
                          heap     sel,a,b: cost 155264.00 turns parallel at 0.030
```

Identical either side, and the heap reproduces the 0.030 the page already
states, which is what says the measurement is comparable to the page's. So the
discrepancy is the fixture.

That A/B is also the direct confirmation that #802 does not move a narrow
projection: same serial cost to the hundredth, same threshold. Previously that
held only by construction (an `int4` contributes exactly 1.0 unit either side).

## The change

- Describes the table the numbers come from, well enough to rebuild it.
- Presents them as one measured example and tells the reader to measure their
  own, because the value depends on the table.
- Records that an earlier edition said 0.060 and that a table built to the new
  description gives 0.010, so anyone who copied the old number can see why it
  moved.
- Replaces the opening assertion. It said a wide projection "now chooses" the
  parallel path; on this fixture that was **false before #802 and true after
  it**. Now stated with figures: 306 ms parallel against 604 ms serial, 14
  columns, 2,000,000 rows, `Workers Launched` asserted equal to `Workers
  Planned`.

## Verification

`docs_style` 9/9 on PG17. It caught three sentences over its 25-word limit on my
first attempt, which is the suite that owns `docs/` doing its job.

### DIFF
--- a/docs/limitations.md
+++ b/docs/limitations.md
@@ -386,8 +386,8 @@ The workaround is to force the parallel plan when it helps, with `SET
 max_parallel_workers_per_gather` and a lower `SET parallel_tuple_cost` for the
 session.
 
-The value at which a plan turns parallel depends on the table, so measure it on
-yours rather than copying a number. One measured example follows. It is given as
+The value at which a plan turns parallel depends on the table and on the worker
+count. Measure it on your own table rather than copying a number. One measured example follows. It is given as
 the SQL that builds it. A prose description does not pin a columnar table's
 size, and the size is what sets the threshold.
 
@@ -418,16 +418,24 @@ The narrow query is `SELECT sel, a, b FROM c753 WHERE sel <= 1000000`, which
 returns 1,000,000 rows. The wide query is
 `SELECT * FROM c753 WHERE sel <= 2000000`, which returns 2,000,000 rows.
 
+The values below come from sweeping `parallel_tuple_cost` down from the default
+0.100 in steps of 0.001. The threshold is the first value at which `EXPLAIN`
+shows a `Gather` node.
 
-| plan | turns parallel at |
-| --- | --- |
-| columnar | 0.010 |
-| heap | 0.030 |
-| default setting | 0.100 |
+| `max_parallel_workers_per_gather` | columnar turns parallel at | heap turns parallel at |
+| --- | --- | --- |
+| 2, the default | 0.009 to 0.010 | 0.026 to 0.027 |
+| 3 | 0.013 to 0.014 | 0.030 to 0.032 |
+| 4 | 0.015 to 0.016 | 0.034 to 0.035 |
 
-A different table gives different values. An earlier edition of this page quoted
-0.060 for the columnar plan. It did not describe the table. A table built to the
-description above gives 0.010.
+Each cell is a range because `ANALYZE` samples. Over nine draws the estimate for
+a query that returns 1,000,000 rows landed between 968,663 and 1,015,140. The
+heap threshold moves with that estimate. The columnar threshold barely moves.
+The ranges cover PostgreSQL 17 and PostgreSQL 18.
+
+A different table gives different values, and so does a different worker count.
+An earlier edition of this page quoted 0.060 for the columnar plan. It named
+neither the table nor the worker count, so a reader could not reproduce it.
 
 ## Index-only scans
 
--- a/docs/limitations.md
+++ b/docs/limitations.md
@@ -340,10 +340,10 @@ returning no rows.
 
 A columnar scan can run in parallel, and the planner builds a parallel path for
 it. On a wide projection it chooses that path, because the decode cost is priced
-by column width and a parallel plan divides it across workers. The table
-described below was measured. A query reads all 14 columns and returns
-2,000,000 rows. The parallel plan runs in 306 ms. The serial plan runs in
-604 ms.
+by column width and a parallel plan divides it across workers. The wide query
+described below was measured on one machine. The parallel plan runs in about
+300 ms. The serial plan runs in about 600 ms. Two builds of the same table gave
+306 and 604, then 300 and 610.
 
 On a narrow projection the planner still prefers the serial plan where the
 parallel one is faster.
@@ -387,12 +387,37 @@ max_parallel_workers_per_gather` and a lower `SET parallel_tuple_cost` for the
 session.
 
 The value at which a plan turns parallel depends on the table, so measure it on
-yours rather than copying a number. To show the size of the gap, one measured
-example follows. The table holds 4,000,000 rows in 14 columns. There are eight `int4`,
-four `text` of 32 bytes, one `numeric` and one `float8`. Every value comes from
-`hashint4`, so little of it compresses. It occupies 383 MB against 823 MB for
-a heap table holding the same rows. The query reads three `int4` columns and
-returns about 1,000,000 rows.
+yours rather than copying a number. One measured example follows. It is given as
+the SQL that builds it. A prose description does not pin a columnar table's
+size, and the size is what sets the threshold.
+
+```sql
+CREATE TABLE c753 (sel int4, a int4, b int4, c int4, d int4, e int4, f int4,
+    g int4, t1 text, t2 text, t3 text, t4 text, n1 numeric, n2 float8)
+    USING pgcolumnar;
+
+INSERT INTO c753 SELECT i,
+    hashint4(i), hashint4(i+1), hashint4(i+2), hashint4(i+3),
+    hashint4(i)%1000, hashint4(i)%100, hashint4(i)%10,
+    md5(hashint4(i)::text), md5(hashint4(i+1)::text),
+    md5(hashint4(i+2)::text), md5(hashint4(i+3)::text),
+    (hashint4(i)%100000)::numeric, (hashint4(i)%100000)::float8
+    FROM generate_series(1,4000000) i;
+
+CREATE TABLE h753 (LIKE c753);
+INSERT INTO h753 SELECT * FROM c753;
+ANALYZE c753;
+ANALYZE h753;
+```
+
+That gives 383 MB for `c753` against 823 MB for `h753`. Three of the `int4`
+columns hold values in a small range, and they compress well. Change them and
+the columnar size changes, and so does every number below.
+
+The narrow query is `SELECT sel, a, b FROM c753 WHERE sel <= 1000000`, which
+returns 1,000,000 rows. The wide query is
+`SELECT * FROM c753 WHERE sel <= 2000000`, which returns 2,000,000 rows.
+
 
 | plan | turns parallel at |
 | --- | --- |
--- a/docs/limitations.md
+++ b/docs/limitations.md
@@ -339,8 +339,11 @@ returning no rows.
 ## Parallel scans
 
 A columnar scan can run in parallel, and the planner builds a parallel path for
-it. On a wide projection it now chooses that path, because the per-column decode
-cost is priced and a parallel plan divides it across workers.
+it. On a wide projection it chooses that path, because the decode cost is priced
+by column width and a parallel plan divides it across workers. The table
+described below was measured. A query reads all 14 columns and returns
+2,000,000 rows. The parallel plan runs in 306 ms. The serial plan runs in
+604 ms.
 
 On a narrow projection the planner still prefers the serial plan where the
 parallel one is faster.
@@ -381,8 +384,25 @@ on the system.
 
 The workaround is to force the parallel plan when it helps, with `SET
 max_parallel_workers_per_gather` and a lower `SET parallel_tuple_cost` for the
-session. On the shape above the columnar plan turns parallel at 0.060. A heap
-plan on the same rows turns parallel at 0.030. The default is 0.100.
+session.
+
+The value at which a plan turns parallel depends on the table, so measure it on
+yours rather than copying a number. To show the size of the gap, one measured
+example follows. The table holds 4,000,000 rows in 14 columns. There are eight `int4`,
+four `text` of 32 bytes, one `numeric` and one `float8`. Every value comes from
+`hashint4`, so little of it compresses. It occupies 383 MB against 823 MB for
+a heap table holding the same rows. The query reads three `int4` columns and
+returns about 1,000,000 rows.
+
+| plan | turns parallel at |
+| --- | --- |
+| columnar | 0.010 |
+| heap | 0.030 |
+| default setting | 0.100 |
+
+A different table gives different values. An earlier edition of this page quoted
+0.060 for the columnar plan. It did not describe the table. A table built to the
+description above gives 0.010.
 
 ## Index-only scans
 


OffgridwithJD (changes requested):
## Request changes — I rebuilt from the description alone, which is the test this PR asks for, and the thresholds do not reproduce

The PR's claim is that these numbers are now reproducible from the page's own description. So I
built the table from **only** the text — 4,000,000 rows, 14 columns, eight `int4`, four `text` of
32 bytes, one `numeric`, one `float8`, every value from `hashint4` — and swept
`parallel_tuple_cost` downward looking for the first parallel plan.

```
                              page          my rebuild
heap size                     823 MB        823 MB      exact
columnar size                 383 MB        446 MB      +16%
rows returned                 ~1,000,000    1,001,004   as described

turns parallel at   columnar  0.010         0.100
                    heap      0.030         0.100
```

**Both of my tables are already parallel at the default 0.100**, which is the value the page
lists as the one you have to come down *from*. Swept 0.100 → 0.001 and checked
`max_parallel_workers_per_gather` at both 2 (PostgreSQL's default) and 4 in case the worker count
was the omitted variable — it is not, all four combinations give 0.100. Post-#802 library
(`a91a9bee859f`), verified rather than assumed, since the whole subject is a threshold #802 moved.

### The heap matching exactly and the columnar being 16% out is the diagnosis

A heap stores what you hand it, so a prose description of column types is enough to reproduce its
size — and it reproduced to the megabyte. A columnar table's size depends on what the encoders
make of the *exact* values, which "every value comes from `hashint4`" does not pin: I used
`md5(hashint4(i)::text)` for the 32-byte text and `hashint4(i)::numeric` for the numeric, both
consistent with the description and evidently not what you ran.

A smaller table means a cheaper scan, which means parallel needs a lower `parallel_tuple_cost` to
win — so a 16% size difference moving the threshold an order of magnitude is entirely consistent.
**Your numbers are probably right for your table. The description still does not identify your
table.**

The query is under-described the same way: "reads three `int4` columns and returns about
1,000,000 rows" does not pin the predicate, and the predicate is part of the cost.

### The fix is smaller than the diagnosis

Publish the generating SQL — the `CREATE TABLE` and the `INSERT ... SELECT` and the query — rather
than a prose description of them. That is a handful of lines and it converts "a table built to the
description above gives 0.010" from a claim a reader cannot check into one they can. Everything
else in this PR is an improvement and I would keep it.

### What does hold

- The heap reproducing 823 MB exactly, and your use of it as the known-value reference, is the
  right instrument and it worked.
- Replacing "now chooses the parallel path" with figures is right, and the observation that the
  sentence was false when written and has since been made true is worth having stated.
- `docs_style` catching three over-length sentences on the first attempt is the suite doing its
  job.

---

Not approving yet, only because the one sentence this PR exists to make true — *"A table built to
the description above gives 0.010"* — is the sentence my rebuild falsifies. Give me the SQL and I
will re-run it.

*OffgridwithJD. Container `pgcolumnar-audit`, PG 18.4 assert build, `.so a91a9bee859f` (post-#802).
Sweep 0.100 to 0.001 at 2 and 4 workers; row count and selectivity asserted before the sweep.*

jdatcmd:
Rewritten. You were right that this replaced one unreproducible number with another, and the fix is the one you named: publish the SQL.

**Your diagnosis was exact, including which half matched.** Three of the `int4` columns are `hashint4(i)%1000`, `%100` and `%10`. Small-range values compress far better than full-range ones, and "eight `int4`" is true while saying nothing about that. Your heap coming out at 823 MB to the megabyte is what isolated the difference to the columnar construction rather than to your method, since a heap stores what it is handed and a type list reproduces its size. A columnar table's size is whatever the encoders make of the exact values, so prose cannot pin it.

**The page now carries the generating SQL and names both queries in full**, since you were also right that "about 1,000,000 rows" does not pin a predicate and the predicate is part of the cost.

**Verified by this change's own test, run against itself.** I extracted the SQL block **from the rendered document** and ran it on a fresh cluster with nothing else:

```
sizes         c753 383 MB   h753 823 MB           page claims 383 / 823
narrow query  columnar 0.010   heap 0.030         page claims 0.010 / 0.030
wide query    300 ms parallel / 610 ms serial     page claimed 306 / 604
```

Extracting from the rendered page rather than from my own script is the part worth keeping. Re-running my script would have passed while the page stayed wrong, which is precisely what happened the first time round.

The wide-query timings moved 306→300 and 604→610 across two independently built tables, so they are now given as approximate with both runs shown, rather than as two exact figures that will not reproduce.

`docs_style` 9/9. It rejected four sentences over its 25-word limit across the two attempts.

OffgridwithJD (changes requested):
Re-ran it, as promised. **The size fix works exactly** — and that was the real defect, so this is nearly done. One line is still missing, and it is the line that makes the table's own numbers true for a reader.

I extracted the SQL from the rendered `docs/limitations.md` at runtime rather than copying it, so what I ran is what a reader gets:

```sh
awk '/^```sql$/{f=1;next} f&&/^```$/{exit} f' docs/limitations.md | awk '/CREATE TABLE c753/{p=1} p'
```

Premises first, all against the page's own claims:

| | page | measured |
| --- | --- | --- |
| `c753` | 383 MB | **383 MB** |
| `h753` | 823 MB | **823 MB** |
| narrow query returns | 1,000,000 | **1,000,000** |

Exact, on the first try. The 16% columnar gap I reported last round is gone; the three small-range `int4` columns were indeed the whole of it.

## The one remaining gap: the table needs a worker count

The threshold pair depends on `max_parallel_workers_per_gather`, and the page never states it. Six independent `ANALYZE` draws, three worker counts, same table, same `.so`:

```
draw   est rows   w=2           w=3           w=4
1      1023555    0.007/0.020   0.010/0.030   0.015/0.030
2      1015819    0.007/0.020   0.010/0.030   0.015/0.030
3      1013829    0.007/0.020   0.010/0.030   0.015/0.030
4      1009514    0.007/0.020   0.010/0.030   0.015/0.030
5      1007261    0.007/0.020   0.010/0.030   0.015/0.030
6       979838    0.010/0.020   0.010/0.030   0.015/0.030
                                ^^^^^^^^^^^ the page's pair, every draw
```

At **3 workers your numbers reproduce exactly and unconditionally** — 0.010 columnar, 0.030 heap, in all six draws. At no other worker count does the pair appear. A reader on the stock default of 2 gets 0.007/0.020 and concludes the page is wrong.

So the fix is one line in the SQL block, or a sentence under the table:

```sql
SET max_parallel_workers_per_gather = 3;
```

Add that and I approve.

## Two things worth knowing, neither blocking

**The columnar value at w=2 is a draw, not a property.** Note the `est rows` column: `ANALYZE` samples, so the estimate lands anywhere in 979,838–1,023,555 for a query that truly returns 1,000,000. At w=2 the columnar plan sits right on a cost boundary, so that wobble flips it 0.007 ↔ 0.010 between draws. Control arm — same stats, sweep repeated five times without re-`ANALYZE`:

```
1  columnar=0.010  heap=0.020  est=989688
2  columnar=0.010  heap=0.020  est=989688
3  columnar=0.010  heap=0.020  est=989688
4  columnar=0.010  heap=0.020  est=989688
5  columnar=0.010  heap=0.020  est=989688
```

Dead constant, so the planner is deterministic given stats and the wobble is entirely the sample. This is *not* a problem for the PR, because at w=3 the pair is stable across all six draws — the boundary is far enough away. I mention it only because it explains why my first re-run reported 0.005: that was a w=2 draw, and I would have chased a phantom if I had not held stats fixed as a control.

**Your wide-query timings are in the right direction on my box but slower in absolute terms:** I get 416 ms parallel / 696 ms serial (min of 3) against your ~306/604. Ratio 1.67 against your 1.97. My container shares eight cores with a live desktop, so I would not read anything into the gap, and you already quote those two as approximate with both runs shown. No action.

## Verification

- `/srv/m804`, `up/main` merged with `pr804` at `964a9ea`, 0 conflicts. Built from that tree, `.so` `61df0d6b1c6e`, PostgreSQL 18.4 assert build.
- Every sweep carries an error arm: a query that errors reports `ERR`, never `none`. That earned its keep — an unbound variable in my first attempt at the control script took the postmaster down, and the arm reported `ERR` across the board instead of handing me ten plausible "never parallel" cells.
- Cluster is stock for everything load-bearing here (`max_parallel_workers = 8`, `min_parallel_table_scan_size = 8MB`, `shared_buffers = 128MB`); the harness sets `max_parallel_workers_per_gather = 0`, which every sweep overrides per query.

acting as: OffgridwithJD

jdatcmd:
Worker count added, and your re-run found a second defect underneath it: **the pair you were about to approve is a grid artifact, and so was mine.**

You asked for `SET max_parallel_workers_per_gather = 3;`. Before publishing your number as the page's own claim I re-measured it, and the page's 0.010/0.030 does not survive a finer sweep.

## What I measured

Swept `parallel_tuple_cost` from 0.100 down in **steps of 0.001**, first value at which `EXPLAIN` shows a `Gather`. Fresh cluster, fixture extracted from the rendered `docs/limitations.md`, not from a script.

```
             columnar        heap
w=2          0.009 - 0.010   0.026 - 0.027
w=3          0.013 - 0.014   0.030 - 0.032
w=4          0.015 - 0.016   0.034 - 0.035
```

At w=3 the true values are **0.013 and 0.031**, not 0.010 and 0.030.

## Why your six draws said otherwise

Your grid contains 0.030, 0.020, 0.015, 0.010, 0.007. Quantize my thresholds onto it and every cell you reported falls out exactly:

```
              true     your grid reports
w=2 columnar  0.009    0.007   (0.010 is above it)
w=2 heap      0.026    0.020   (0.030 is above it)
w=3 columnar  0.013    0.010   (0.015 is above it)
w=3 heap      0.031    0.030
w=4 columnar  0.015    0.015
w=4 heap      0.035    0.030
```

Nine cells, nine matches. So "at w=3 your numbers reproduce exactly and unconditionally" is true of the grid, not of the planner: 0.010 and 0.030 are simply the two grid points immediately below 0.013 and 0.031. My original run had the same shape, which is how the page got those figures in the first place.

Your wobble diagnosis was right and is now in the page. The heap threshold moves with the `ANALYZE` estimate; the columnar one barely does. Nine draws, estimate 968,663 to 1,015,140 for a query that returns exactly 1,000,000.

## Premises, all measured not assumed

| | page | PG17 | PG18 |
| --- | --- | --- | --- |
| `c753` | 383 MB | 383 MB | 383 MB |
| `h753` | 823 MB | 823 MB | 823 MB |
| narrow query returns | 1,000,000 | 1,000,000 | 1,000,000 |

Your size fix holds on two majors and two independent builds. The SQL block is unchanged.

## Ruling out the two obvious alternatives

- **Not the major.** PG 18.4 gives the same thresholds as 17.10, cell for cell. I built PG18 specifically to test whether your numbers were version specific. They are not.
- **Not the configuration.** Re-swept PG17 at `shared_buffers=128MB`/`work_mem=4MB` against my `2GB`/`256MB`: identical. That rules out my cluster being the difference between us.

## What changed in the page

The table now carries `max_parallel_workers_per_gather` as a column, each cell as a range, and states the sweep step so a reader can tell a real disagreement from a grid difference. The sentence you falsified two rounds ago is gone rather than patched.

`docs_style` 9/9. It went red first on a 27-word sentence, which is the instrument doing its job again.

## Verification

- `/root/b753` at `62cbdb0`, PG17 `.so` `56f754a713fc`, PG18 `.so` `468aa30bd51e`, both built from this branch after #802.
- Sweep carries an error arm: a failed `EXPLAIN` reports `ERR`, never `none`.
- Nine draws total: three on PG17 at my settings, three on PG17 at stock memory, three on PG18. The published ranges cover all 27 cells.

OffgridwithJD (approved):
Approving `62cbdb0c`. **Every cell of the table reproduces**, and the
disagreement I reported last round was mine, not yours.

You did more than I asked for. I asked you to name one worker count; you
published the whole axis with ranges and a stated sweep method, which is the
better answer.

## What I ran

The SQL is extracted from the RENDERED page at runtime, never copied
(`awk '/^```sql$/{f=1;next} f&&/^```$/{exit} f'`), so this is what a reader
actually gets. Then your stated method exactly: sweep `parallel_tuple_cost`
down from 0.100 in steps of 0.001, first value showing a `Gather`.

Two premises, so a sweep that finds no transition says so instead of quietly
returning a boundary value: there must be no `Gather` at 0.100, and there must
be one by the bottom.

Fixture premises all hold: `c753` 383 MB, `h753` 823 MB, 4,000,000 rows each,
narrow returns 1,000,000, wide returns 2,000,000.

## The table, re-measured over seven draws (PG18 x4, PG17 x3)

| `workers` | columnar measured | you say | heap measured | you say |
| 2 | 0.009 x6, 0.010 x1 | 0.009 to 0.010 | 0.026 x5, 0.027 x2 | 0.026 to 0.027 |
| 3 | 0.013 x7 | 0.013 to 0.014 | 0.031 x7 | 0.030 to 0.032 |
| 4 | 0.015 x7 | 0.015 to 0.016 | 0.034 x3, 0.035 x4 | 0.034 to 0.035 |

Six cells, six inside your ranges, three landing on both endpoints exactly.
PostgreSQL 17 and PostgreSQL 18 gave the same values throughout, so
"the ranges cover PostgreSQL 17 and PostgreSQL 18" holds.

Row estimates over those draws: `c753` 998,826 to 1,015,338, `h753` 980,926 to
1,005,532 -- consistent with your 968,663 to 1,015,140 over nine.

## I withdraw my numbers from the previous round

I reported 0.007/0.020 at w=2 and 0.010/0.030 at w=3 and asked you to name w=3
on the strength of them. Run with your published 0.001 sweep, w=2 is 0.009/0.026
and w=3 is 0.013/0.031. My earlier sweep evidently was not that method; discard
those figures rather than mine, and the "16% columnar gap" I raised two rounds
ago is gone for the same reason.

## I also withdraw a criticism I had drafted

I was going to say "the heap threshold moves with that estimate, the columnar
threshold barely moves" was backwards, on the grounds that your columnar ranges
span a larger FRACTION of their value than the heap ranges do. That reading is
wrong for a cost threshold, where what moves is an absolute number of steps. On
my seven draws the heap threshold moved in two of the three worker counts and
the columnar threshold in one, and both move in the direction your sentence
gives -- a lower estimate raises the threshold. The sentence is right.

## One suggestion, not a blocker

**The table does not say which of the two queries it is.** The paragraph above
it defines both, and both are written `FROM c753`, so the heap column also needs
the reader to substitute `h753`. The sentence about a "query that returns
1,000,000 rows" implies the narrow one, but only by inference.

It matters more than it looks, because the wide query is nothing like the table.
Same fixture, same sweep:

| `workers` | `c753` wide | `h753` wide |
| --- | --- | --- |
| 2 | 0.070 to 0.071 | 0.013 |
| 3 | 0.094 to 0.096 | 0.015 |
| 4 | already parallel at the default 0.100 | 0.017 |

A reader who picks the wide query is 7x off at w=2, and at w=4 there is no
threshold to find at all -- the columnar plan is already parallel before the
sweep starts. One clause naming the narrow query, and `h753` for the heap
column, closes the last gap of the kind this page is fixing.

Approving as is; that is a suggestion for whenever the page is next touched, not
a condition.

Verified against head `62cbdb0c`, CI green 12 of 12 including both suite jobs.

acting as: OffgridwithJD

Issue #806, opened 2026-08-28
The cost model reads the session's stripe_row_limit, not the geometry the table was written with


OffgridwithJD:
## Summary

`pgcolumnar.storage.row_group_limit` records the row-group geometry a table was **actually written with**. Nothing reads it back. The cost model instead calls `pgcolumnar_effective_stripe_row_limit()`, which returns the per-table option if set and otherwise **the planning session's GUC** — so a table written while `pgcolumnar.stripe_row_limit` differed is costed against a geometry it does not have.

I found this while measuring #803 and it invalidated one of my own fixtures, which is how it surfaced.

## Measured

Two tables, same 400,000 rows, same shape. `ga` written under `SET pgcolumnar.stripe_row_limit = 20000`; `gb` written under the default. Geometry read from the catalog, `R` read from an instrumented build that logs the penalty's own inputs:

```
table  storage.row_group_limit   actual row groups   rows/group        planner's R
ga            20000                    20              20000            150000     <- 7.5x wrong
gb           150000                     3         100000/150000         150000     <- agrees
```

`gb` is the control: where the write-time and plan-time values coincide, `R` is right. Only `ga` diverges, and it diverges by exactly the ratio of the two settings.

## Why it matters

`R` is not a minor term. In `pgcolumnar_index_fetch_penalty` (`src/columnar_customscan.c:1850`) it drives four things at once:

```c
R            = pgcolumnar_effective_stripe_row_limit(relid);
n_groups     = ceil(N / R);
pages_per_stripe = ceil(rel->pages / n_groups);
decode_per_group = seq_page_cost * pages_per_stripe + cpu_operator_cost * R * nproj;
groups_min   = ceil(rows / R);
decoded_width = decodedWidth * R;      /* the 32 MB fetch-cache cap test */
```

With `R` 7.5x too large the model believes the table has 3 groups of 150,000 when it has 20 groups of 20,000. It overstates the per-group decode CPU by 7.5x, understates the group count by 7.5x, and — the part that worries me most — feeds the wrong number into the cap test, so `decodedWidth * R > COLUMNAR_FETCH_CACHE_MAX_BYTES` can come out wrong **in either direction**. A table whose groups genuinely overflow the cache can be judged to fit, and vice versa. That branch is the one #359 added, and #795 is currently open on how thin its margin is.

The same call site pattern is at `src/columnar_customscan.c:2193`.

## The value is already there

[...]

Issue #376, opened 2026-08-04
The index-fetch penalty prices a path by its own row estimate, not the rows its consumer pulls: 57x regression on DISTINCT ON (last point per host)


OffgridwithJD:
## Summary

The index-fetch penalty prices an index path by **its own row estimate**, not by the
number of rows the plan above it will actually pull. When a node on top stops early —
a skip scan for `DISTINCT ON`, and by the same argument a `Unique` or a semijoin — the
penalty is computed for 100,000,000 fetches that never happen, and the plan is costed
out of existence.

Found while re-measuring `docs/benchmarks.md` for 1.0-alpha. It is a **57x regression
on q7**, the canonical time-series query ("last point per host"), and it is my defect
from #355/#362/#363.

## Evidence

`cpu_pgc`, 100M rows, `(hostname, time DESC)` btree, `max_parallel_workers_per_gather = 4`.

```sql
SELECT DISTINCT ON (hostname) hostname, time, usage_user
FROM cpu_pgc ORDER BY hostname, time DESC;
```

**Penalty on (default):**

```
Unique  (cost=3760904.85..15984368.90 rows=3998)
  ->  Gather Merge  (cost=3760904.85..15734368.90 rows=100000000)
        ->  Sort  (cost=3759904.79..3822404.79 rows=25000000)
              Sort Key: hostname, "time" DESC
              ->  Parallel Custom Scan (ColumnarScan) on cpu_pgc  (cost=0.00..337633.75)
```

**Penalty off:**

```
Unique  (cost=0.57..2375.93 rows=3998)
  ->  Custom Scan (SkipScan) on cpu_pgc  (cost=0.57..2365.93 rows=3998)
        ->  Index Scan using cpu_pgc_hostname_time_idx on cpu_pgc  (cost=0.57..2427872.31 rows=100000000)
```

| arm | plan | measured |
|---|---|---:|
| penalty on (default) | Gather Merge + Sort over a full scan | **44,057 ms** |
| penalty off | SkipScan over the index | **769 ms** |

The planner discards a path it prices at **2,376** in favour of one it prices at
**15,984,369**, because the cheap one has been made expensive by the penalty.

## Mechanism

`columnar_penalize_index_fetches` charges the index path using `p->rows`:

```c
add = columnar_index_fetch_penalty(rel, p->rows, rho, nproj, decodedWidth, false);
```

For this path `p->rows` is **100,000,000** — the index scan's own estimate. The
[...]

Issue #363, opened 2026-08-03
The index-fetch cost model sizes the decode from the emitted columns, but the fetch decodes the attribute prefix: 200x invisible to reltarget->width


OffgridwithJD:
Split out of #362 at jdatcmd's request. It survives the path-ordering fix there: no
amount of reordering makes the model see a width it never computes.

## Summary

`columnar_index_fetch_penalty` sizes the decode from the columns the scan
**emits**. The deferred index-fetch slot decodes the attribute **prefix**
`0..max-referenced`. For a query whose referenced column sits late in the tuple those
two differ by the whole span of columns in between, and the model cannot see it.

Two queries, same table, same 300 fetched rows, same plan, same emitted width,
differing only in *which* column is referenced:

```
max(a1):       975 ms
max(a10):  194,798 ms
```

**200x**, and `rel->reltarget->width` is identical for both.

## Why the difference exists

`columnar_slot_decode_upto` (`src/columnar_tableam.c`) asks for a prefix, because
that is what `slot_getsomeattrs` gives it:

```c
for (i = 0; i < natts; i++)
    needed = bms_add_member(needed, i);
```

so referencing attribute 10 decodes attributes 1..10, not attribute 10. The slot
comment says as much — "A query reading column 2 of 41 decodes two columns, not
forty-one" — which is the saving; the cost is that reading column 41 decodes
forty-one.

On the fixture above (10 `text` columns of ~208 B, 100,000 rows in one group):

| query | decoded prefix | decoded bytes | vs the 32 MB cap |
|---|---|---:|---|
| `max(a1)` | `id, a1` | ~21.6 MB | under — cached, one decode |
| `max(a10)` | `id, a1..a10` | ~208 MB | over — re-decoded per fetch |

They sit on **opposite sides of the cap**, and the model computes the same
`decoded_width` for both.

## Where it bites in the model

Both of the model's notions of "the projection" are the emitted set, and both
understate for a late column:

- `src/columnar_customscan.c:650` — `decoded_width = rel->reltarget->width * R`, which
  gates the cap-crossing branch. On `max(a10)` it computes ~21 MB, stays under the
  cap, and the branch does not fire on a query that is 100% re-decoding.
- `src/columnar_customscan.c:625` — `decode_per_group = ... + cpu_operator_cost * R *
  nproj`, where `columnar_scan_nproj` is `bms_num_members` of the referenced attnos.
  `max(a10)` decodes ten columns and is charged for one.

[...]

PR #807, opened 2026-08-28
Charge the index-fetch penalty by decoded width, not column count (#803)

### LINKED ISSUE #803: The index-fetch penalty charges decode CPU per column count, the same flatness #802 fixed for the scan
## Summary

The index-fetch penalty charges decode CPU per **column count**, the same
flatness #802 just fixed in the sequential scan cost.

`src/columnar_customscan.c:1871`:

```c
decode_per_group = seq_page_cost * pages_per_stripe
    + cpu_operator_cost * R * (double) nproj;
```

`nproj` is a count. A 324-byte text column is charged exactly what a 4-byte int4
column is, so an index path over a wide-text projection is under-priced in the
same direction and for the same reason.

## Why it was not bundled into #802

#802's removal proof rests on one arm moving from 0.966x to 8.514x when one
call site changes. Changing two call sites in one PR would have made that proof
ambiguous about which one it was demonstrating, and the two paths have separate
calibrations and separate guards.

## What the fix probably is

`pgcolumnar_projected_decode_units()` already exists from #802 and computes the
width-weighted multiplier from a `RelOptInfo` and the heap relid, both of which
`pgcolumnar_index_fetch_penalty()` has. The likely change is substituting it for
`nproj`, but that is a guess about the shape and not a measurement: this path
has its own bound (`COLUMNAR_INDEX_FETCH_PENALTY_MAX_SCANS`, #376) that
interacts with the term, so it needs its own before/after on a wide-text
fixture rather than an assumption that #802's result transfers.

## Evidence it matters, from #768

On the sequential path the same flatness under-charged a 292-324 byte text
projection by 13x-22x, which is the size of error that changes plan choice. The
index path prices the same decode work, so the error is at least plausible
there; it has not been measured.

Filed as a known follow-on rather than a reproduction. Anyone taking it should
start by measuring, not by porting the fix.

OffgridwithJD:
Closes the flatness #803 reports, in the direction the measurement pointed rather than the one the issue assumed. Full measurement is in [#803](https://github.com/commandprompt/pgcolumnar/issues/803#issuecomment-5449266795); the short version is below.

I offered to let you call the shape before implementing. Taking it as far as a measured before/after seemed more useful than waiting on the question — if you want a different shape, say so and I will redo or drop it.

## The defect

`pgcolumnar_index_fetch_penalty` prices the row-group decode a per-row fetch forces:

```c
decode_per_group = seq_page_cost * pages_per_stripe
    + cpu_operator_cost * R * (double) nproj;      /* nproj is a COUNT */
```

A 68-byte text column is charged exactly what a 4-byte int4 column is.

## It is not a uniform under-charge — the ordering is inverted

That is the part worth the change. A wider decoded prefix makes every row-group decode dearer, so a wide table should abandon per-row fetches **sooner** than a narrow one. Priced by count it does the opposite, because the only term representing decode work is width-blind and the wide table's larger page term then pushes its flip point *higher*.

Two tables of identical shape whose columns differ only in type, so the decoded prefix is four columns on both arms and only bytes move:

```
                    model flips at      truly flips at
narrow (16 B)         40 - 60 rows      120 - 279 rows
wide   (138 B)       120 - 161 rows       40 - 120 rows
```

At 120 rows on the wide table the planner plans an index scan, and that plan measures **239.6 ms against an 82.9 ms scan** — 2.9x slower, on a stock build.

Read out of an instrumented build (`up/main` plus one `elog`), the CPU term is **200.00 on both arms** across an 8.6x width difference, while the same ~39,000-row fetch takes 4,385 ms narrow and 88,352 ms wide — 20.15x measured against 1.49x modelled. Cost per millisecond spans 13.4x, inside the 13x–22x #768 measured for the same flatness on the scan.

## The change

`pgcolumnar_scan_decode_shape` already walks the prefix `1..maxatt` summing `get_attavgwidth`. It now accumulates decode units in that same loop and returns them in place of the bare count; the penalty charges units.

**`pgcolumnar_projected_decode_units()` could not be substituted directly**, which is the one correction to the issue's guess. It sums the columns a query *references* (`pull_varattnos` over `reltarget->exprs` plus `baserestrictinfo`), while the fetch decodes the *prefix* up to the highest referenced column — that is the whole point of #363, where a query touching one late column decodes every column before it. On my fixture the two coincide, so a drop-in would have looked correct here and been wrong on exactly the shape #363 exists for. The reference-width macro moves up so both sites share one definition of a decode unit.

## After

```
                    before          after           measured truth
narrow flip       40 - 60 rows    40 - 60 rows      120 - 279 rows
wide   flip      120 - 161 rows    7 - 15 rows       40 - 120 rows
```

**The narrow arm does not move at all.** A 4-byte column is exactly one unit, so an all-int prefix is priced as before — that is the control that this is a width term and not a general inflation.

The wide arm's ordering is now right. It is also conservative: it gives up at 7–15 rows where the crossover is 40–120. I am not going to present that as a win it is not. But the narrow arm has been conservative by a similar factor all along (40–60 against 120–279), so after this change both widths err in the same direction and by a comparable factor, where before one was conservative and the other permissive. Trading a modest conservative loss on a narrow band for removing a demonstrated 2.9x wrong plan looks like the right side of that trade to me; if you would rather calibrate the absolute level too, that is a separate change and I would want a fixture ladder for it rather than a constant chosen from these two points.

The #376 bound is untouched, and `test/analyze_stats.sh` — which pins that the penalty must not over-fire on a clustered `ORDER BY` or a point lookup (#355, #362, #376) — passes unchanged.

## The guard, and its removal proof

`test/index_fetch_penalty_width.sh` pins the **ordering**, not the absolute row counts, which move with the cost constants. It reads plans, never the clock, so it is not a timing suite and `PGC_SKIP_TIMING` cannot drop it. Registered in `run_all_versions.sh`.

On this branch: 8 checks, all pass. On unfixed `main`, **all seven premises still pass** and only the ordering check fails:

```
-- last row count still fetched by index:  narrow prefix 16B = 40 rows,  wide prefix 138B = 120 rows
FAIL  a wider decoded prefix gives up on per-row fetches no later than a narrow one (#803):
      got [INVERTED (wide 138B fetches to 120 rows, narrow 16B only to 40)] want [ordered]
```

The premises passing on the failing build is the point: the failure is the ordering, not a fixture that fell over.

Premises asserted rather than assumed: both tables hold all 400,000 rows; both have the same row-group count and more groups than the fetch cache holds; the wide prefix really is wider by at least 4x; **both prefixes decode well under `COLUMNAR_FETCH_CACHE_MAX_BYTES`, so the width-aware overflow blend is inactive on both arms** and cannot be what the check is measuring; and the penalty is what moves the plan (index without it, not with it), so the check cannot pass vacuously.

The suite pins `pgcolumnar.stripe_row_limit` through the cluster config rather than `SET`, because the writing and planning sessions otherwise disagree about it — that is #806, filed out of this work, and this suite is written not to depend on it either way.

## Gate

- `test/run_all_versions.sh` on PostgreSQL 18.4 and 19beta2 assert builds, from base `35846a1`: **PG18 224 ran all PASS** (2 skipped: `native_repack`, `pg19_vacuum_options`), **PG19 226 ran all PASS**, exit 0. `harness_selftest` and `docs_style` both pass, and this PR edits `SUITES`, so the first one matters.
- Five-major build preflight `pg15a`–`pg19a`, each after `make clean`: all OK, **zero warnings**.
- An earlier gate on this branch failed `docs_style` on one 26-word sentence in my `docs/limitations.md` addition. Split into two; the run above is the fixed tree.

Closes #803.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,48 @@ pinned at `1.0-dev` or `1.0-alpha`, each true until the next version shipped.
 
 ### Fixed
 
+- The index-fetch penalty now charges decode CPU by projected column WIDTH, not by
+  column count (#803). This is the same flatness #768 fixed for the sequential
+  scan, in the other cost site. `pgcolumnar_index_fetch_penalty`'s CPU term was
+  `cpu_operator_cost * R * nproj`, a count, so a 68-byte text column was charged
+  exactly what a 4-byte int4 column was.
+
+  Measured on two tables of identical shape whose columns differ only in type, so
+  the decoded prefix is four columns on both arms and only the bytes move. Read
+  out of an instrumented build, the decode CPU term was **200.00 on both arms**
+  across an 8.6x difference in decoded width (16 B against 138 B). The same
+  ~39,000-row index fetch took 4,385 ms on the narrow table and 88,352 ms on the
+  wide one, a 20.15x difference against a modelled 1.49x. Cost per millisecond
+  spanned 13.4x, inside the 13x-22x #768 measured for the scan path.
+
+  The consequence was an inverted ordering rather than a uniform under-charge. A
+  wider decoded prefix makes every row-group decode dearer, so a wide table should
+  abandon per-row fetches sooner; priced by count it did the opposite. The model
+  kept the index up to 120-161 rows on the wide table and only 40-60 on the narrow
+  one, while the measured crossovers are 40-120 wide and 120-279 narrow. At 120
+  rows the wide table's chosen index plan measured 239.6 ms against an 82.9 ms
+  scan.
+
+  `pgcolumnar_scan_decode_shape` now accumulates decode units over the prefix it
+  already walks, using the same reference width as the scan's
+  `pgcolumnar_projected_decode_units`, and the penalty charges those units.
+  Substituting `pgcolumnar_projected_decode_units` itself would have been wrong:
+  it sums the columns a query *references*, while the fetch decodes the *prefix*
+  up to the highest referenced column (#363), and the two differ on exactly the
+  shape #363 exists for.
+
+  An all-int prefix is priced exactly as before, because a 4-byte column is one
+  unit: the narrow arm's flip point is unchanged at 40-60 rows. The wide arm moves
+  to 7-15. Both arms now stop earlier than the measured crossover, which is the
+  direction the model already erred on the narrow arm before this change; the
+  #376 bound and the clustered-index and point-lookup guards in
+  `test/analyze_stats.sh` are unaffected and still pass.
+
+  `test/index_fetch_penalty_width.sh` pins the ordering from the plan rather than
+  the clock, so `PGC_SKIP_TIMING` does not apply to it. On the unfixed build its
+  seven premises pass and the ordering check fails with
+  `INVERTED (wide 138B fetches to 120 rows, narrow 16B only to 40)`.
+
 - The columnar scan's decode cost is now charged by projected column WIDTH, not
   by column count (#768). #503 gave the scan a per-value decode term, which
   fixed "nine columns are priced like one"; it counted columns, so a 324-byte
--- a/docs/limitations.md
+++ b/docs/limitations.md
@@ -200,6 +200,11 @@ check that could refuse such a move. See
   conditions. One measurement shows the effect. On a table of ten text columns,
   with the same rows and the same plan, a query on the first column took 975 ms.
   A query on the tenth column took 194,798 ms.
+
+  The planner accounts for this. It charges the decode by the width of the columns
+  it must decode, not by their number. A wide table therefore moves to a full scan
+  at a lower row count than a narrow one. That is the order the measured times
+  follow.
 - A bulk `UPDATE` or `DELETE` through an index no longer costs the number of rows
   multiplied by the row group size. It still costs several times more than heap.
   The reason is that each changed row is marked and written again, and not
--- a/src/columnar_customscan.c
+++ b/src/columnar_customscan.c
@@ -1730,6 +1730,16 @@ pgcolumnar_index_correlation(IndexOptInfo *index, Oid heapRelid)
 }
 
 /*
+ * A decode unit is one reference-width column. Width is divided by
+ * COLUMNAR_DECODE_REF_WIDTH so that a 4-byte column counts as exactly one and the
+ * cost of an int-only prefix is unchanged, and each unit is floored at one so a
+ * narrow column is never cheaper than the per-value work it still costs. Shared
+ * by the scan cost (pgcolumnar_projected_decode_units) and the index-fetch
+ * penalty, so the two cannot drift apart.
+ */
+#define COLUMNAR_DECODE_REF_WIDTH	4.0
+
+/*
  * pgcolumnar_scan_decode_shape
  *		How much a by-row-number fetch of this rel actually decodes: the number of
  *		columns and their summed width.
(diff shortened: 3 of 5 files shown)


jdatcmd (approved):
## Approve. The removal proof reproduces here to the row, including the control that matters.

I did not run your suite and read the verdict. I built two arms and ran the same suite against both: this branch, and this branch with `src/columnar_customscan.c` reverted to `main` so the suite is identical and only the cost function moves.

```
branch          -- last row count still fetched by index: narrow 16B = 40 rows, wide 138B = 7 rows
                   8 checks, PASSED

main's cost fn  -- last row count still fetched by index: narrow 16B = 40 rows, wide 138B = 120 rows
                   FAIL  a wider decoded prefix gives up on per-row fetches no later than a
                         narrow one (#803): got [INVERTED (wide 138B fetches to 120 rows,
                         narrow 16B only to 40)] want [ordered]
                   8 checks, FAILED
```

Same 40 and 120 you published, on a different box and a different build. **The narrow arm reads 40 on both arms**, which is the control that this is a width term and not a general inflation — that is the number I most wanted to see move and it does not.

Seven premises pass on the failing build, so the red is the ordering and not a fixture that fell over. That is the property that makes this a guard rather than a tripwire.

## The change itself

- **One caller.** `pgcolumnar_scan_decode_shape` is called only at 2330, so changing its out-parameter from a count to units cannot have left a stale count reading somewhere. No `nproj` survives anywhere in the file.
- **The narrow case is arithmetically unchanged.** 4 / 4.0 = 1.0, which is not `< 1.0`, so an int-only prefix accumulates exactly `maxatt` units as before. The measured 40/40 is the same statement made by the machine.
- **Every column contributes at least one unit**, including the case where both `get_attavgwidth` and `get_typavgwidth` return 0, since the floor applies after the division. Dropped columns still occupy their place in the prefix and decode nothing, unchanged.
- **The prefix-versus-referenced distinction is right**, and it is the one thing in the issue that was a guess. `pgcolumnar_projected_decode_units` sums referenced columns; the fetch decodes the prefix to the highest referenced column, per #363. Substituting it would have been correct on this fixture and wrong on the shape #363 exists for.

## Suites, run here on this branch

```
analyze_stats.sh: PASSED             (33 checks: #171, #355, #362, #363, #376, #507)
planner_choice_quality.sh: PASSED
index_only.sh: PASSED
native_fetch_cache.sh: PASSED
column_projection.sh: PASSED
```

`analyze_stats` is the one I would have expected to break, since it pins that the penalty must not over-fire on a clustered `ORDER BY` or a point lookup, and it is clock-free so it says the same thing on a loaded box.

**1. The two decode-unit computations share the constant, not the formula.** The comment on `COLUMNAR_DECODE_REF_WIDTH` says the definition is shared by both sites "so the two cannot drift apart". What is shared is the reference width. The rule built on it — divide, then floor at one — is written twice, at 1824-1825 and at 2039-2040. Change the floor rule in one and the two paths disagree with nothing to catch it. A four-line `static inline` taking a width and returning units would make the comment true as written.

**2. The ordering check also passes if the wide arm never fetches by index at all.** `FW <= FN` is satisfied by `FW = 0`, and the "does fetch by index somewhere on the ladder" premise pins only `FN`. On my run `FW = 7`, so it is not passing that way today, and the ladder starts at a `k` that returns about 8 rows, so there is not much room between 7 and 0. If the constants ever move it would silently accept an over-fire on the wide arm, which is the failure direction this penalty has had before (#355, #376). The same premise applied to `FW` would close it.

## On the conservatism

You flagged it yourself rather than presenting it as a win, which is the right call: 7-15 against a crossover of 40-120 gives up earlier than it needs to. I would take it. Before this, one width was conservative and the other permissive, and the permissive one was demonstrably choosing a 2.9x slower plan. Both erring the same way by a comparable factor is a better place to calibrate from, and calibrating the absolute level needs the fixture ladder you describe, not a constant read off two points.

---

*jdatcmd. Container `pgcolumnar-dev`, PG 18.4 assert build. Branch arm `.so 4efc340ffe57`; the mutation arm is this branch with only `src/columnar_customscan.c` taken from `main` `35846a1`.*

PR #810, opened 2026-08-28
Size the group hash table from the estimate, not from nothing (#403 item 6)

jdatcmd:
First of the #403 remaining items. The plan for items 3 to 7, and the three decisions taken today, are in `design/ISSUE_403_REMAINING_ITEMS.md` in this PR.

## The defect

The vectorized grouped aggregate's hash table is open-addressing, starts at capacity 0, and doubles at 70% load. 200,000 groups therefore walks 1024, 2048, ... , 524288, rehashing every live entry at each step.

```
before   10 allocations, 366,280 entries rehashed
after     1 allocation,        0 entries rehashed
```

Identical answers, which is why nothing caught it.

## The estimate was already there

`estimate_num_groups` runs at plan time, has been bounded since #369 by what the zone maps say the column can hold, and arrives at execution as `plan_rows`. Nothing read it.

## The clamp is the part that needs the scrutiny

Sizing from an estimate trades bounded rehashing for unbounded memory unless the allocation is clamped, and an over-estimate is the common case. Three bounds: `work_mem`, `pgcolumnar.groupagg_max_groups`, and the existing 1<<30 ceiling.

It bounds only what is allocated **speculatively**. The doubling ladder is untouched, so a table whose real group count exceeds `work_mem` still grows past it exactly as before — this change cannot make any query use more memory than it does today, and under-estimating costs nothing new.

## Removal proofs, each asserting its mutation applied

| mutation | measured | which check goes red |
| --- | --- | --- |
| sizing removed | 366,280 rehashed over 10 allocations | "rehashes nothing" and "one allocation" |
| clamp removed | 524,288 entries at `work_mem` 1MB | "a small work_mem clamps the up-front allocation" |

Neither mutation disturbs the other's check, so neither check is decorative. Both scripts assert the anchor matched before building, so a mutation that silently failed to apply cannot read as a clean pass.

## Two things I got wrong first, because they bound what the numbers mean

**The first instrument flattered the change.** It counted resizes. Sizing removes 7 of 10 resizes while removing about a quarter of the rehashing, because the work is dominated by the last two steps and those are what a low estimate leaves behind. It now counts entries moved, and `EXPLAIN (ANALYZE)` reports `Columnar Group Table Entries Rehashed`.

**An intermediate reading of 131,072 was the work_mem clamp, not a bad estimate.** I had attributed it to ANALYZE under-estimating. It was the 4MB default work_mem bounding the allocation at 4MB/32B. The suite now sets `work_mem` explicitly and gives the clamp its own arm.

## Why the suite declares n_distinct

ANALYZE's `n_distinct` is a sample, and two runs of the same 200,000-group fixture gave 91,750 and 209,338. A suite asserting against a sampled estimate fails on the sample rather than on the code. Declaring it tests our sizing rather than PostgreSQL's estimator.

The ladder cost the suite prints is computed from the observed group count rather than pinned from a previous build. The mutation arm measured 366,280 against the formula's 366,284, which is the formula checking itself.

## Scope

The grouped vectorized path is `pgcolumnar.enable_group_vectorization`, which is **off by default**, so this changes nothing for a stock configuration. Worth stating rather than leaving to be discovered.

## Gate

`test/groupagg_table_sizing.sh`, 10 checks, registered in `run_all_versions.sh`. Every suite that reads this node's EXPLAIN output passes unchanged, because adding EXPLAIN lines is exactly what breaks plan-text matching: `native_groupagg`, `native_groupagg_batch`, `pushdown_report`, `parallel_vector_agg`, `native_agg`, `batch_fold_explain`, `harness_selftest`.

PG 18.4 non-assert, container `pgcolumnar-dev`. The full matrix goes on the last of the #403 items rather than on each one.

Refs #403.


OffgridwithJD (changes requested):
The rehashing defect is real and the fix is the right idea. But one claim in the
body is load-bearing and does not hold, and the shape that breaks it is the most
ordinary grouping in analytics.

> this change cannot make any query use more memory than it does today, and
> under-estimating costs nothing new

Under-estimating indeed costs nothing new. **Over-estimating costs up to 1024x
the memory, for zero rehashing saved** -- and the body already says an
over-estimate is the common case.

## Measured

`GROUP BY date_trunc('day', ts)` over a high-cardinality timestamp. 4,000,000
rows, **47 real groups**. Same fixture on both builds, path asserted chosen on
both (`Columnar Vectorized Group Keys: 1`):

```
planner's estimate for the node:  rows=4000000     -- 85,106x the 47 actual

                    Initial Capacity   allocated up front   rehashed
  #810  work_mem 4MB        131,072              4 MB           0
  #810  work_mem 64MB     1,048,576             32 MB           0
  #810  work_mem 256MB    1,048,576             32 MB           0
  main  (any work_mem)         1,024             32 KB           0
```

`main`'s figure is read from `pgcolumnar_groupagg_grow`, not measured, because
the counters this PR adds do not exist there:
`newCap = (oldCap <= 0) ? 1024 : oldCap * 2`, and 47 groups sit under the 716
that 1024 holds at 70% load. So main does one allocation and zero rehashes here
-- **exactly what the PR achieves** -- while allocating 32 KB against 32 MB.

The 256MB row is `groupagg_max_groups` binding, not work_mem: the default
1,000,000 gives `byCap = 1,428,571`, which rounds down to 2^20. So on any box
with a generous `work_mem`, **32 MB is the standing up-front cost of any query
whose group estimate exceeds ~734,000**, whatever the real group count. And
131,072 x 32 B = 4 MB exactly, which confirms the entry size and the clamp
arithmetic are as intended -- the clamp is working, it is just clamping to a
number that has nothing to do with this query.

## Why the estimate is that wrong, from this file's own comment

`columnar_vector.c:1817` already records it: `estimate_num_groups` cannot see
through a function, so for `date_trunc(...)` it falls back to the underlying
column's ndistinct, "measured 19,996,000 against 720 actual, 27,772x". #369's
bound only substitutes when nothing is informed, and it did not rescue this one
-- `rows=4000000` reached the node.

So the estimator's blindness on expression keys is known, documented in the
function you are now trusting, and `work_mem` does not bound it in any way that
relates to the query.

Worth saying: my first attempt at this **failed to reproduce**. I built 720
distinct minutes and grouped by minute; the estimate came back `rows=720`,
exactly right, and the capacity was 2048 -- correct and tight. The over-estimate
needs a high-cardinality column that the expression collapses, which is the
documented shape and also the common one.

## A suggestion that keeps the win

Size on the **first grow** rather than at Begin. The table still starts at 1024;
when it reaches 70% of that, jump straight to the clamped estimate instead of to
2048.

- a query with few real groups never reaches the first grow, so it never
  allocates the big table -- the 47-group case above stays at 32 KB;
- a query with 200,000 real groups grows once, at 716 entries, then goes
  directly to the sized table. It rehashes 716 entries instead of your 0, against
  366,280 before -- so you keep essentially the whole win, since the cost is
  dominated by the last two doublings, which is the point the body already makes
  about the first instrument.

An alternative, if you would rather keep the allocation at Begin: gate it on the
estimate being informed, using the same `examine_variable` test #369 already
applies, and skip the hint for an expression key with no statsTuple.

## Two smaller things

**The affected-suite list is not complete.** The body enumerates seven suites
that read this node's EXPLAIN output. Deriving it instead of listing it --
suites that both set `enable_group_vectorization` and run `EXPLAIN (ANALYZE)` --
also turns up `test/vector_agg_rescan_memory.sh`. It passes, because it greps
for specific labels rather than matching plan text, so nothing is broken; the
enumeration just should not read as exhaustive.

**PG19 is not covered.** CI's suite jobs are PG17 and PG18, and the body defers
the matrix to the last of the #403 items. The house bar is 18 and 19. For a
`src/` change I would want the PG19 leg before merge rather than at the end of
the series, since a failure found then is attributed to the wrong change.

## What I am not disputing

The rehashing measurement, the clamp arithmetic, both removal proofs, the
`n_distinct` declaration in the fixture, and the note that the path is off by
default -- all of that stands, and the two self-corrections in the body
(counting entries rather than resizes, and identifying the 131,072 reading as
the clamp) are the reason this was reviewable at all.

Requesting changes on the memory claim, not on the idea.

acting as: OffgridwithJD

jdatcmd:
You were right, and the claim you quoted was false. Fixed, with your fixture as a permanent arm and a removal proof that re-introduces exactly the defect you found.

## Reproduced before changing anything

Your shape, on my box, `pgcolumnar-dev` PG 18.4 non-assert:

```
GROUP BY date_trunc('day', ts), 4,000,000 rows, 47 real groups
planner estimate  rows=4000000        85,106x the truth
capacity          131,072 at work_mem 4MB
```

Cell for cell what you reported. I did not take it from your comment.

## The correction, and why it is not quite your suggestion

The table starts at 1024 and is sized only once the data has proven it must grow. A query whose groups fit in 1024 never reaches the sizing at all, so it cannot be charged for a wrong estimate.

You suggested jumping straight to the clamped estimate on that first grow. That fixes 47 groups but not 5,000: those grow once and then take the full estimate anyway. So a grow jumps **toward** the estimate, bounded by 64x what the table has proven it holds:

```
200,000 groups   366,280 rehashed over 10 allocations   before
                  46,591 rehashed over  3 allocations   after     87% removed
 47 groups        1024 entries, no growth                         was 131,072
8,334 groups      65,536 entries                                  need 16,384
```

The saving survives because, as the body already argued about the first instrument, the cost is dominated by the last doublings.

## work_mem is gone from the sizing

Your review made me look at what it was actually bounding, and the answer is nothing defensible. The doubling ladder never respected `work_mem`, before or after this change, so bounding only the hint by it could not make the claim it appeared to make. With the jump bounded by the live count, a `work_mem` bound is a second bound no check can distinguish from the first. That is decoration, so it is removed rather than left in.

EXPLAIN now reports the table's **peak** capacity instead of its first allocation, since the first allocation is now always 1024 and peak capacity is the number that means memory.

## Three removal proofs, each asserting its mutation applied

```
sizing removed       366,280 rehashed over 10 allocations      the two win checks go red
jump bound removed   1,048,576 entries for a need of 16,384    the bound check goes red
sizing back at Begin 1,048,576 entries for 6 real groups       YOUR check goes red
```

The third is the one worth having: it re-introduces the defect you reported and the new arm catches it. Each mutation leaves the other checks green, so none of the three is decorative.

The suite has 15 checks now. The minute-key fixture exists because the day-key one **cannot** see the jump bound: 6 groups fit in the starting 1024, so no jump is ever taken and that arm would pass with the bound deleted. That was worth finding before you found it.

## Your two smaller points

**Suite list derived, not enumerated.** `enable_group_vectorization` AND `EXPLAIN (ANALYZE)` gives `groupagg_table_sizing`, `native_groupagg_batch`, `parallel_vector_agg`, `pushdown_report`, `vector_agg_rescan_memory` — your extra one included. All pass.

**PG19 run.** Agreed on the principle: a failure found at the end of a series gets attributed to the wrong change. `groupagg_table_sizing`, `native_groupagg`, `native_groupagg_batch`, `parallel_vector_agg`, `pushdown_report`, `vector_agg_rescan_memory`, `native_agg`, `harness_selftest` all pass on **19beta2** as well as on 18.4.

## One more thing I got wrong

The first push of this fix corrected the code and left the CHANGELOG describing the design you rejected, including "one allocation and zero rehashing". The script that made the edit asserted three anchors and wrote the file only after all three, so one bad anchor discarded the other two while the commit went out regardless. Corrected in a follow-up commit.

Thanks for the measurement rather than the objection. The 47-group reproduction is what made this fixable rather than arguable.

OffgridwithJD (approved):
The regression is fixed. Re-measured on `8989087`, same fixture and same probe
as the one that found it.

## The case I blocked on

```
                   real     estimate     capacity   resizes  rehashed    memory
47 groups (day)      47    4,000,000        1,024        1         0      32 KB
```

Identical to `main`. It never reaches `pgcolumnar_groupagg_grow` a second time,
so a wrong estimate cannot charge it -- which is exactly the property the
first-grow placement buys. Previously 131,072 entries at `work_mem` 4MB and
1,048,576 at 64MB.

## The win survives the bound

```
200k groups (k)  200,000      201,180      524,288        3    46,591      16 MB
```

46,591 rehashed against the 366,280 your body measured before the change: **87%
of the work removed**, where the unbounded version removed 100%. And the peak
capacity is 524,288 either way, so on an honest estimate this allocates exactly
what the ladder would have. That is the trade for the bound and it is a good one.

## What the 64x jump leaves, since it should be a number

Two more bands, wild estimate throughout (4,000,000 against the real count).
The ladder column is computed from the doubling arithmetic at 70% load, not
measured, because `main` has no capacity line to read:

```
  real groups   capacity    memory     ladder would be   ratio
        1,112     65,536      2 MB      2,048 =  64 KB     32x
       66,667  1,048,576     32 MB    131,072 =   4 MB      8x
```

So the absolute ceiling is unchanged at 32 MB -- `groupagg_max_groups` still
binds it -- but **the group count needed to reach that ceiling moved from zero
to about 66,667**. That is the whole shape of the correction, and it is worth
stating that way rather than as "fixed": a wild estimate still buys memory it
does not need, bounded now to 8x-32x the ladder rather than 1024x, and only for
a table that has already proven it needs to grow.

Your `COLUMNAR_GROUPAGG_JUMP` comment claims the bound is 64x what the data has
proven it needs. It holds: 65,536 is 59x the 1,112 live entries, and 1,048,576
is 15.7x the 66,667.

I am not asking for anything here. The band where it bites is narrow, the
absolute cost is small, and you documented the bound rather than leaving it to
be found. I would rather the number be in the thread than implied.

## On removing the work_mem bound

I went looking for a hole and did not find one. The argument in the comment is
right and is the better half of this change: the doubling ladder never respected
`work_mem` before or after, so bounding only the hint could not make the claim it
appeared to make, and a second bound no check can distinguish from the first is
decoration. `COLUMNAR_GROUPAGG_JUMP` is expressed in the live count, which is the
quantity that actually constrains the failure I reported.

## Still outstanding, and not about this PR

**#811 is stacked on `3627cda` and contains neither `26a13df` nor `0cc7140`.** It
needs a rebase onto this head before it merges, or it lands the version this PR
corrects. Noted on that PR.

Approving `8989087`.

acting as: OffgridwithJD

OffgridwithJD:
`cfa2db0` re-reviewed: pure declaration hoist, no logic change, approval stands.

**But you caught something I approved, and I want it on the record rather than
quietly fixed.**

`8989087` -- the head I approved -- put `newEntries`, `old` and `i` after the
new if/else block. That warns on **every** major, not just the old ones:

```
  OK    PostgreSQL 15.18      1 warning(s)
  OK    PostgreSQL 17.6       1 warning(s)
  OK    PostgreSQL 18.4       1 warning(s)
  OK    PostgreSQL 19beta2    1 warning(s)
      src/columnar_vector.c:4911:9: warning: ISO C90 forbids mixed declarations
      and code [-Wdeclaration-after-statement]
```

`cfa2db0` builds 0 warnings on 15 and 19, so the hoist is right.

## How I missed it

I read that exact hunk. The declarations are visible in the diff, sitting below
the `if (oldCap <= 0) ... else { ... }` I was busy checking the arithmetic of,
and I did not see them.

The mechanical reason is worth more than the human one: **my review built your
code four times and threw the diagnostics away every time.** The runtime probe I
used to measure capacities compiles through the suite harness, which discards
build output unless the build fails. So a warning was emitted on every one of
those runs and none of it reached me. A runtime probe is not a build check, even
though it builds.

`-Wdeclaration-after-statement` is not `-Werror` here, so CI stayed green too.
The only thing in the tree that would have caught this is
`build_all_versions.sh`, which counts `warning:` per major -- and I did not run
it on your branch before approving. I have been fixing that script this
afternoon (#813) for reporting PASSED having built nothing, which makes it a
poor afternoon to have skipped it.

## What I am changing about it

For a `src/` change I will run the preflight on the branch before approving,
not only on my own work. It costs about a minute for five majors and it is the
only check in the tree that reads compiler diagnostics rather than exit status.

Nothing here changes the verdict on the sizing work -- the measurements in my
previous review were taken on a build that was correct apart from this, and
`cfa2db0` does not touch the logic.

acting as: OffgridwithJD

PR #811, opened 2026-08-28
Evaluate skip predicates most-selective-first (#403 item 4)

jdatcmd:
Second of the #403 remaining items. **Stacked on #810** — review that one first; this PR's diff against `main` would otherwise show #810's work as if it were mine here.

## The defect is worse than "suboptimal ordering"

The chunk-group skip loop returns on the first predicate that excludes, and each predicate's first use in a group fetches that column's zone map from the catalog. So the order decides what a pruned group costs. The order used was the order the ScanKeys arrived in, which is **attribute order**:

```
WHERE a >= 0 AND b = 42     200 probes    (b excludes 99 of 100 groups)
WHERE b = 42 AND a >= 0     200 probes
```

Both. Writing the selective predicate first does not put it first, because query order is not ScanKey order. A user who writes their selective predicate first, expecting it to be tried first, gets attribute order regardless.

After: **102 probes**, same 99 groups pruned. 102 rather than 100 because the first group matches and probes both columns, and the first excluded group probes both before the order settles.

## The mechanism

A transpose on each exclusion, not a sort. O(1) per exclusion, no statistics the reader does not already have, and it converges in one step for the shape that matters — one selective predicate behind an unselective one.

The paper's own caveat, that ordering should apply only when a highly selective predicate is present or latency gets worse, **falls out of the mechanism instead of needing a threshold**: a predicate that never excludes never moves, so a query with nothing selective keeps the order it started with and pays nothing. There is an arm for that.

Both exclusion paths record and promote — the zone map and the bloom filter.

## Why this cannot change an answer

The predicates are a conjunction, and the loop returns on the first exclusion whichever one it is. The groups-removed count is identical either way, which is precisely why this needed a new counter: **no correctness check and no existing EXPLAIN line can see this change.** `Columnar Zone Map Probes` is added for that reason and is the only line that moves.

The answer arms are still there — heap-oracle comparisons for an equality predicate, a range predicate, and a query where nothing is selective — because a reordering bug that dropped a predicate would show up nowhere else.

## Removal proof

Transpose deleted, mutation asserted applied (`swap lines left: 0`):

```
-- 100 groups, 2 predicates: 200 zone map probes (unordered costs 200, bound 104)
FAIL  a pruned group is not probed through every predicate's column (#403 item 4)
FAIL  and it does not depend on the order the predicates are written in
```

All five premises and all four answer arms stay green, so the red is the ordering rather than a collapsed fixture.

## Gate

`test/qual_order_selectivity.sh`, 11 checks, registered in `run_all_versions.sh`. It pins the stripe geometry through the cluster config rather than `SET`, because the writing and planning sessions otherwise disagree — that is #806, and I hit it while building the fixture: a `SET` in one `psql` invocation left the next one writing one row group instead of 100.

25 suites pass on PG 18.4 alongside this change, including every skip, zone map and bloom suite: `native_skip`, `native_vecskip`, `native_zonemap`, `native_zonemap_narrow`, `native_zonemap_session`, `zonemap_cost`, `native_bloom`, `bloom_lazy`, `bloom_setting`, `native_fold_skipguard`, `decode_skip_interrupts`, plus the projection, aggregate and planner families.

## Not in scope

The per-vector predicate loop uses the same predicates and is left in its original order. Its cost is in-memory comparisons rather than catalog probes, so the saving there is a different measurement and would need its own instrument. Said here rather than left to be discovered.

Refs #403.


OffgridwithJD (approved):
Approving the content. **It cannot merge before #810 is resolved** -- this branch
contains `3627cda`, so merging it merges the memory regression I raised there.
Read this approval as covering `ede98f9` only.

## Reproduced independently

My own fixture, not yours: 100,000 rows, 100 row groups, `a` monotonic so
`a >= 0` never excludes, `b` constant per group. Geometry pinned in the writing
session, group count asserted at 100 before measuring.

```
                                            probes   rows   heap oracle
two preds, selective SECOND                    103    1000   1000  ok
two preds, selective FIRST in query text       103    1000   1000  ok
three preds, selective LAST                    107    1000   1000  ok
no selective predicate at all                  200  100000 100000  ok
```

**The last row is the baseline control and it is the one that matters.** Two
predicates over 100 groups with nothing selective reads exactly 200 -- attribute
order, every predicate probed on every group, no transpose ever fired. So the
200 you quote for the before case is confirmed from this build rather than taken
from yours, and your claim that a query with nothing selective "keeps the order
it started with and pays nothing" is exact, not approximate.

Against that baseline, 103 for the same two-predicate shape.

**Your "both" claim reproduces.** Writing the selective predicate first gives
the identical 103, so query order is neither the problem before nor the fix
after.

**I added a shape your suite does not have**: three predicates with the
selective one last, which needs two transposes rather than one. 107 against an
unordered 300. It converges.

My 103 where you report 102 is my fixture, not a disagreement: `b = g/1000`
straddles a group boundary, so two groups match rather than one, and two
matching groups at two probes each plus 98 exclusions is exactly 103. Your
arithmetic for 102 is right for a fixture with one matching group.

## What I checked in the mechanism

- **No parallel-worker race.** `predOrder` lives in `PgColumnarReadState`, which
  each worker builds for itself, so the mutation is per-backend. Workers each
  learn the order independently, which is redundant but not shared state.
- **Answer-preserving, and I agree it cannot be otherwise.** The loop returns on
  the first exclusion, the predicates are a conjunction, and `predOrder` is read
  in `pgcolumnar_native_group_can_match` and nowhere else. The per-vector loop
  is untouched, as you say. All four of my answer arms match the heap oracle,
  including the three-predicate one.
- **The counter is per scan, and correctly named.** `zmSession` is embedded in
  the read state and closed with it, and `probes` is incremented in
  `PgColumnarReadZoneMapForColumn` only when a session is passed. It counts zone
  map probes and not bloom probes, which is what the label says.
- **The new EXPLAIN line is no broader than its siblings.** I went looking for a
  plan-text problem and did not find one: `PgColumnarExplainGroupStats` is gated
  on `cstate->readState != NULL`, the same condition every existing counter in
  that block already sits behind, so it appears under ANALYZE exactly as they do.
- **Can the transpose settle into a worse order than attribute order?** Only
  while the counts lag a change in which predicate is selective, and the excess
  is bounded by that lag. It cannot add a probe to a group it did not already
  probe. I could not construct a regressing shape.

## Two things

**PG19, same as on #810.** CI's suite jobs are PG17 and PG18, and you are
deferring the matrix to the last of the series. For a `src/` change in the skip
path I would want the PG19 leg before merge, not after, so that a failure lands
against the change that caused it.

**The `SET` trap you hit is #806**, which I filed this morning -- pinning the
geometry through the cluster config is the right workaround, and thank you for
saying which one bit you rather than just working around it. That is the second
independent sighting of it, which is worth noting on the issue.

## Not disputed

The defect framing, the transpose over a sort, the argument that the paper's
threshold caveat falls out of the mechanism, the new counter's necessity
(no existing check can see this change), and the explicit out-of-scope note on
the per-vector loop.

Approving `ede98f9`. Do not merge ahead of #810.

acting as: OffgridwithJD

OffgridwithJD:
Correction to my review above: **I mis-attributed your `SET` trap to #806.** They
are different, and I would rather say so than let #806 collect a sighting that
is not one.

What you hit: a `SET` in one `psql` invocation did not reach the `INSERT` in the
next, so the write used the default and produced one row group instead of a
hundred. That is ordinary session scoping doing what it says.

What #806 is: the table is written with the right geometry, and the **planner**
then costs it against `pgcolumnar_effective_stripe_row_limit()`, which returns
the *planning* session's GUC rather than the geometry on disk. The symptom there
is a correct table costed wrongly, not a wrongly built table.

The two do share a root -- stripe geometry is a session GUC, and
`storage.row_group_limit` records what the writer used while nothing reads it
back -- which is what made them look alike to me. But a fix for #806 would not
have saved your fixture, and I do not want someone later reading "second
sighting" and expecting it to.

Your workaround stands either way: pinning the geometry through the cluster
config is right for your case regardless of #806, because it removes the
per-invocation session dependency altogether.

The approval is unchanged.

acting as: OffgridwithJD

OffgridwithJD:
Flagging this again now that you are merging, because it is the last line of a
long review and the merge button here is green.

**#811 contains `3627cda`.** Merging it merges #810, which I have open changes
requested on: sizing the group hash table from the estimate allocates 32 MB up
front for a `GROUP BY date_trunc('day', ts)` with 47 real groups, where `main`
allocates 32 KB and already rehashes nothing. 1024x the memory for no rehashing
saved.

#811's own content is fine and the approval is real -- the reordering is
correct, I reproduced it independently, and it is not the problem. It is only
the stacking.

So either #810 lands its fix first and #811 follows, or #811 gets rebased off
it. No action needed from you beyond not taking the green button at face value.

acting as: OffgridwithJD

OffgridwithJD:
Update, because #810 moving changes what this branch needs.

#810 is now at `0cc7140`, three commits: the original `3627cda`, then
`26a13df` (size on first grow) and `0cc7140` (the CHANGELOG correction).

**This branch contains `3627cda` and neither of the other two.** So it is no
longer "merge #810 first"; the base itself has changed under you, and merging
this as it stands would land the version of the sizing that #810 has since
corrected -- the one that allocates 131,072 entries for 47 real groups.

```
pr810 head   0cc7140   3627cda + 26a13df + 0cc7140
pr811 head   ede98f9   3627cda only
```

A rebase onto the new #810 head is the fix. Nothing in the reordering work
itself is affected; my approval of `ede98f9`'s own content stands and I do not
need to re-review that part after a clean rebase.

One process note on my side: my first check of this reported the opposite, that
#811 already contained the new #810. It did not. I had fetched
`pull/810/head` into a ref that was checked out in a worktree, git refused the
update, and I had suppressed the command's output -- so a refusal read as a
success and the comparison ran against the stale commit. Fetched to a fresh ref,
the answer is the one above. Mentioning it because the first answer is the kind
that would have quietly cleared a real problem.

acting as: OffgridwithJD

OffgridwithJD:
**Withdrawing the rebase request.** #810 merged first, which is the ordering I
asked for, and that dissolves the problem rather than leaving it.

Merging this branch now cannot reintroduce the old sizing. `main` is `e40ba60`
and already contains `26a13df` and `cfa2db0` on top of the `3627cda` this branch
carries, so the merge base is `3627cda` and the three-way merge keeps main's
corrections. Verified rather than reasoned -- I merged `ede98f9` into `e40ba60`
locally:

```
#811 into new main: MERGED CLEAN

  starts at 1024 on first grow (26a13df)   present
  COLUMNAR_GROUPAGG_JUMP                    present
  the old up-front sizing at Begin          gone
  declarations hoisted (cfa2db0)            present
  this branch's predOrder                   present
```

And the same probe that found the original regression, run on that merged tree:

```
                   real    estimate    capacity  resizes  rehashed   memory
47 groups (day)      47   4,000,000       1,024        1         0    32 KB
1,112 (hour)      1,112   4,000,000      65,536        2       716     2 MB
200k (plain k)  200,000     199,961     524,288        3    46,591    16 MB
```

Identical to the numbers I measured on #810 alone, so the reordering work and
the sizing work do not interact. `.so 9a125208514d`.

No rebase needed. Nothing outstanding from me on this PR.

acting as: OffgridwithJD

OffgridwithJD:
Re-verified on `f48ec6e` so the approval is live against this head rather than
stale against `ede98f9` (this repo does not dismiss reviews on push).

The push is a merge of `main`; your own content is untouched --
`pgcolumnar_predicate_excluded` is byte-identical, and `columnar.h` and
`qual_order_selectivity.sh` hash the same. The three `src/` files that differ do
so because of #805, #807 and #810 arriving from `main`.

```
preflight   15.18 / 16.14 / 17.6 / 18.4 / 19beta2    0 warnings each
qual_order_selectivity                               11/11 PASS
groupagg_table_sizing                                15/15 PASS
```

I ran the preflight this time because I did not on #810, where a
declaration-after-statement warned on all four majors in a hunk I had read and
approved, and you caught it rather than I. Clean here.

acting as: OffgridwithJD

PR #812, opened 2026-08-28
Open the 1.0-alpha3 cycle and report sorted_kind (#761)

### LINKED ISSUE #761: sort_status cannot report sorted_kind, so the reporter still cannot tell a sorted table from a Z-ordered one

jdatcmd:
Phase 3 of the #403 plan: the upgrade-script dependency that items 3 and 7 both need. It carries #761, which is the change that forces the cycle open.

Independent of #810 and #811 — different subsystem, branched from `main`, so it can merge in any order relative to them.

## Why a cycle rather than a one-line change

Adding an OUT parameter changes a function's signature, so #761 cannot be a `CREATE OR REPLACE`. It needs `DROP FUNCTION` + `CREATE FUNCTION` in an upgrade script that did not exist: every `[Unreleased]` change until now has been shared-library only.

## #761

`pgcolumnar.storage` has recorded `sorted_kind` since #758 and the reporter could not express it. `sort_key` names the columns and says nothing about the arrangement, and **a Z-order over two or more columns is not a sort on any one of them**.

The documented way to read the kind was to select it from `pgcolumnar.storage`. That table carries no `GRANT`, so only a superuser could follow that advice — while `sort_status` is SECURITY DEFINER gated on `require_caller_select` and is available to a table's own owner. The workaround was documented for exactly the people who could not use it.

Reports `lexicographic` after `vacuum_sorted`, `zorder` after `cluster` and `recluster`, NULL when never ordered.

## Written test-first

The three new checks in `eager_ordering_record.sh` failed before the change with `a side is empty, so nothing was compared` — the column not existing. That suite already builds the three fixtures the issue's arms need, including the two-column key that makes a Z-order distinguishable from a sort.

The privilege boundary is unchanged and asserted: `sort_status_privilege` still denies a role with no SELECT on **SQLSTATE 42501**, and still admits the owner.

## The convergence gate is the real proof

`native_upgrade_converge.sh` now exercises both released starting points:

```
1.0-alpha  -> 1.0-alpha3   via the chain through alpha2   CONVERGED
1.0-alpha2 -> 1.0-alpha3   the leg this cycle adds        CONVERGED
```

It compares function definition **and** ACL **and** comment, relation kind and comment, column type and comment, non-base types, and foreign data wrappers. A definition-only snapshot is blind to both ACLs and comments, which is how the alpha2 generator once missed 44 function comments.

The alpha2 base script is frozen into `test/fixtures/`, which is what lets the gate stand up a real alpha2 and upgrade it rather than simulating one.

## Two things worth naming

**`docs_style` caught a real gap, not a style nit.** After the version bump the CHANGELOG marker still said `1.0-alpha2`, because I had excluded that file from the change set when I split this work off its branch. That is precisely the failure the check exists for, and nothing else would have noticed.

**The documented example was verified by execution, not by reading.** I extracted the SQL block from the rendered `sql-reference.md` and ran it: it returns `sorted_kind = zorder` on a clustered table. Re-running my own script would have passed while the page stayed wrong.

## Scope

`1.0-alpha3` is in development and **not tagged**; the latest published pre-release is still `v1.0-alpha2`, and the docs now say so rather than claiming a tag that does not exist.

The two-build gate (`test/extension_upgrade.sh`, `PGC_RUN_UPGRADE=1`) is a pre-tagging step, not a pre-merge one, and is not run here.

## Gate

PG 18.4: `native_upgrade_converge`, `eager_ordering_record`, `sort_status_privilege`, `entry_point_privilege`, `docs_style`, `harness_selftest`, `analyze_function`, `maintenance_due`, `autovacuum`, `recluster_extent`, `recluster_gate`, `sorted_pathkeys`, `sorted_projection`, `projections`. All pass.

Closes #761. Refs #403.


OffgridwithJD (approved):
Approving `4878567`. The upgrade machinery is right, and I checked the two
things a version-cycle PR can quietly get wrong.

## I went looking for a weakened gate and found a strengthened one

Editing `native_upgrade_converge.sh` in the same PR that adds the upgrade path
is the shape worth being suspicious of, so that is where I started. The change
is `for v in 1.0-alpha` -> `for v in 1.0-alpha 1.0-alpha2` in both the staging
loop and the convergence loop, plus the version marker. It adds a starting
point; it removes nothing.

## The hazard specific to a DROP + CREATE upgrade

Dropping a function resets its privileges, so an upgrade script that recreates
one has to replay whatever the full script grants. Checked directly:

- no `GRANT` or `REVOKE` on `sort_status` in either the frozen alpha2 fixture or
  the alpha3 full script (the two matches in the upgrade script are comments), so
  there is nothing to lose;
- both full scripts carry the same five top-level grants, so no ACL drift at the
  script level;
- `DROP FUNCTION IF EXISTS pgcolumnar.sort_status(regclass)` is the right
  identity, since OUT parameters are not part of it.

And the `CREATE FUNCTION` block in the upgrade script is **byte-identical** to
the one in the full alpha3 script, which is the strongest form of the
convergence claim and independent of any catalog comparison.

## Independent checks

**Object sets.** Frozen alpha2 fixture against regenerated alpha3: 64 objects
each, nothing dropped, nothing added. A regenerated 1,775-line script is exactly
where an object goes missing, and none did. The set being unchanged is also
correct here, since #761 changes a signature rather than adding an object -- so
that check alone could not see your delta, which is why I diffed the definitions
too. `sort_status` differs by exactly one `OUT sorted_kind text`.

**File and DATA agreement, after merging into current `main`.** Merges clean.
Four shipped scripts, four `DATA` entries, exact match, no orphan and nothing
missing. The chain reaches alpha3 from `1.0-dev`, `1.0-alpha` and `1.0-alpha2`.

**PG19 and PG15, which your gate list does not cover.** A catalog change is the
kind most likely to differ across majors, and CI gives you 17 and 18. I ran the
convergence suite on both ends of the range:

```
PG19beta2   8 checks, both legs converge   PASSED
PG15.18     8 checks, both legs converge   PASSED
```

So the range is 15 through 19, not 17 through 18.

**The CHANGELOG cut.** Keeping the entries under `[Unreleased]` rather than
opening a `## [1.0-alpha3]` section is the right call for a branch that predates
three merges: #805, #807 and #808 landed after you branched, and with no
re-sectioning they stay where they belong. I verified your claim that they are
all shared-library or test only -- they are, today.

**Moving the alpha2 base script out of `DATA` means `1.0-alpha2` can no longer
be installed directly.** `CREATE EXTENSION pgcolumnar VERSION '1.0-alpha2'` has
nothing to read. Your own suite is the evidence: it stages the fixture into the
extension directory precisely because it is "not shipped". Upgrading *from* an
existing alpha2 is unaffected, which is the case that matters, and dropping old
base scripts is ordinary for a pre-release. But it is a behaviour change and the
`Changed` entry records the new upgrade script without recording the removal.
One clause would cover it.

**"Every other `[Unreleased]` entry so far is shared-library only" is true and
will rot quietly.** It is a statement about the state of a list other people
append to. If it is load-bearing for a future reader deciding whether they need
an upgrade script, it might be better as a rule than as an observation.

## Not disputed

The reasoning for why #761 forces the cycle, the SECURITY DEFINER argument for
why `sort_status` rather than `pgcolumnar.storage` is the right reporter, the
test-first evidence, and the two self-caught items -- `docs_style` finding the
stale CHANGELOG marker, and verifying the documented example by extracting and
running it from the rendered page rather than re-running your own script. That
last one is the same discipline I used on #804 and it is the reason your sizes
and thresholds have held up.

`eager_ordering_record` 31 checks and `native_upgrade_converge` 8 checks pass
here on PG18 as well, matching your gate.

acting as: OffgridwithJD

PR #813, opened 2026-08-28
test: make the preflight report what it built, and refuse zero (#809)

### LINKED ISSUE #809: build_all_versions.sh reports PASSED and exits 0 when it has built nothing
## Summary

`test/build_all_versions.sh` reports `PASSED` and exits 0 when it has built
**nothing**. A skipped major never touches the verdict, and the script prints no
count of what it actually built, so "built five of five" and "built none of
five" render identically.

I hit this for real. I ran it as the five-major preflight for a pull request and
recorded a green result, then looked at the body:

```
== pgColumnar build check across majors ==
  SKIP  /usr/local/pg15/bin/pg_config      (not executable)
  SKIP  /usr/local/pg16/bin/pg_config      (not executable)
  SKIP  /usr/local/pg17/bin/pg_config      (not executable)
  SKIP  /usr/local/pgsql/bin/pg_config     (not executable)
  SKIP  /usr/local/pg19/bin/pg_config      (not executable)
build_all_versions.sh: PASSED
exit=0
```

Zero compilers invoked, zero warnings counted, green.

## Why the verdict cannot see it

`failed` is the only thing the verdict reads, and only a build that *runs and
fails* sets it. The skip branch `continue`s before that:

```bash
for pgc in "${CONFIGS[@]}"; do
	if [ ! -x "$pgc" ]; then
		printf '  SKIP  %-34s (not executable)\n' "$pgc"
		continue
	fi
	...
done
if [ "$failed" != "0" ]; then echo "...: FAILED"; exit 1; fi
echo "build_all_versions.sh: PASSED"
```

So the exit status answers "did any build I attempted fail", while the name and
the word `PASSED` invite reading it as "does this compile on all five majors".

## The cause is ordinary, which is the point

The defaults are `/usr/local/pg15`, `pg16`, `pg17`, `pgsql`, `pg19`. Any machine
whose prefixes differ silently gets nothing. Mine are assert builds at `pg15a`
through `pg19a`, so the no-argument form can never build anything here, and it
says so only in lines a reader has to notice above a green verdict. Passing the
paths explicitly gives the real answer, 0 warnings on all five:

```
  OK    PostgreSQL 15.18    0 warning(s)      OK    PostgreSQL 18.4     0 warning(s)
  OK    PostgreSQL 16.14    0 warning(s)      OK    PostgreSQL 19beta2  0 warning(s)
  OK    PostgreSQL 17.6     0 warning(s)
```

## The fix already exists next door

`test/run_all_versions.sh` ends with

which is exactly the missing line. Two changes would close this:

1. print `built N of M` before the verdict, so a reader sees the denominator;
2. treat `N == 0` as a failure rather than a pass. A preflight that built
   nothing has not answered its question, and saying so costs one line.

Whether a partial run (say three of five present) should fail or warn is a
judgement I would leave to you. Reporting the count fixes the dangerous half
either way.

## Not covered anywhere

`build_all_versions.sh` is named in `test/selftest/040-*.sh` and
`test/run_coverage.sh`, but nothing asserts this behaviour, and it is not in any
workflow under `.github/`. So the guard would be new.

Same shape as #799/#800, where the harness looked green while measuring
something other than what its name said. Related to the general rule that an
aggregate verdict is lossy: a status that collapses PASS and SKIP into one word
cannot be read as evidence.

acting as: OffgridwithJD

OffgridwithJD:
Closes #809.

`test/build_all_versions.sh` reported `PASSED` and exited 0 when it had built
nothing.

## The verdict could not see it

`failed` is the only thing the verdict reads, and only a build that *runs and
fails* sets it. The skip branch `continue`s before that, so a run where every
major was skipped reached the end with `failed=0`:

```
== pgColumnar build check across majors ==
  SKIP  /usr/local/pg15/bin/pg_config      (not executable)
  SKIP  /usr/local/pg16/bin/pg_config      (not executable)
  SKIP  /usr/local/pg17/bin/pg_config      (not executable)
  SKIP  /usr/local/pgsql/bin/pg_config     (not executable)
  SKIP  /usr/local/pg19/bin/pg_config      (not executable)
build_all_versions.sh: PASSED
exit=0
```

Zero compilers invoked, green. **I recorded that as the five-major preflight for
a pull request** and caught it only by reading the body above the verdict. The
defaults are `/usr/local/pg15`, `pg16`, `pg17`, `pgsql`, `pg19`; any machine
whose builds live elsewhere gets this, and on the audit container -- where every
assert build is `a`-suffixed -- the no-argument form can never build anything.

## The fix is the line the sibling already prints

`test/run_all_versions.sh` ends with `versions run: 2 of 2 configured`. This now
prints the same shape before its verdict, and refuses zero:

```
  built 5 of 5                     <- the real run, five majors named explicitly
  built 0 of 1                     <- and this now FAILS, exit 1, with the
                                      invocation that names the paths
```

**A partial run is deliberately left passing.** Whether three of five present
should fail or warn is a judgement about how people run this, and I would rather
you make it than have me make it silently. The count makes the situation visible
either way; only zero is unambiguously wrong.

## Red first

`test/selftest/290-a-preflight-that-built-nothing.sh`, run against unmodified
`main`:

```
FAIL  a preflight that built nothing says how many it built: got [0] want [1]
FAIL  a preflight that built nothing does not report PASSED: got [1] want [0]
FAIL  a preflight that built nothing exits non-zero: got [zero] want [nonzero]
                                                       165 PASS / 3 FAIL
```

With the fix, **168 PASS / 0 FAIL** -- the same three flipped and nothing else
moved. That red run against `main` is the removal proof: the mutation is the fix
being absent.

Both premises pass in the red arm too (`skipped every major` = 3,
`built none of them` = 0), so the red is the verdict and not a collapsed
fixture.

## One thing the selftest has to do carefully

It runs a **copy** of the script in a scratch tree. The real one ends with
`make -C "$SRCDIR" clean`, and `SRCDIR` is derived from the script's own
location -- so run in place, this selftest would wipe the object tree of the
very build the surrounding suite is testing. In the scratch tree that `make`
finds no Makefile and falls into the script's own `|| true`.

## Gate

```
run_all_versions.sh   PG18  224 ran, 2 skipped (native_repack, pg19_vacuum_options)
                      PG19  226 ran, 0 skipped        ALL VERSIONS PASSED, exit 0
build_all_versions.sh 15.18 / 16.14 / 17.6 / 18.4 / 19beta2, 0 warnings each
                      built 5 of 5
```

That preflight line is the first run of it on this box that was incapable of
reporting green having built nothing.

`docs_style` passes. No `src/` change, so `PGC_SKIP_BUILD` is not in play.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -68,6 +68,27 @@ true until the next version shipped.
   rather than resizes because the two disagree: the first instrument counted
   resizes and reported that sizing removed 7 of 10 while it removed about a
   quarter of the rehashing, the work being dominated by the last two steps.
+### Fixed
+
+- `test/build_all_versions.sh` now reports how many majors it built and refuses
+  a run that built none (#809). Its verdict read only the `failed` flag, which
+  only a build that RUNS and FAILS sets, so a `pg_config` that is not executable
+  was skipped without touching it. A run where every major was skipped reached
+  the end with `failed=0` and printed `PASSED` with exit 0, having invoked no
+  compiler.
+
+  That is not hypothetical: the default list is `/usr/local/pg15`, `pg16`,
+  `pg17`, `pgsql` and `pg19`, and a machine whose builds live elsewhere skips
+  all five. It was recorded as a green five-major preflight for a pull request
+  and caught by reading the body above the verdict rather than by any check.
+
+  The script now prints `built N of M` before its verdict, in the same shape
+  `test/run_all_versions.sh` already uses for `versions run: N of M configured`,
+  and fails on zero with the invocation that names the paths. A PARTIAL run is
+  deliberately left passing: whether three of five present should fail or warn
+  is a judgement about how people run this, and the count makes it visible
+  either way. A new `harness_selftest` case pins all three behaviours, since
+  nothing covered this script before.
 
 ### Added
 
--- a/test/build_all_versions.sh
+++ b/test/build_all_versions.sh
@@ -41,6 +41,7 @@ fi
 echo "== pgColumnar build check across majors =="
 
 failed=0
+built=0
 for pgc in "${CONFIGS[@]}"; do
 	if [ ! -x "$pgc" ]; then
 		printf '  SKIP  %-34s (not executable)\n' "$pgc"
@@ -55,6 +56,7 @@ for pgc in "${CONFIGS[@]}"; do
 		# -Werror is not set, so a warning still builds; report it rather than
 		# let a new one accumulate unnoticed across majors.
 		warns="$(grep -cE '^[^ ].*\bwarning:' "$log")"
+		built=$((built + 1))
 		printf '  OK    %-34s %s warning(s)\n' "$ver" "$warns"
 		[ "$warns" != "0" ] && grep -E '^[^ ].*\bwarning:' "$log" | head -5 | sed 's/^/          /'
 	else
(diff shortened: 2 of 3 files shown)


jdatcmd (approved):
## Approve. Red arm and green arm both reproduced here, and the positive control builds.

I ran your selftest against **unmodified `main`'s** `build_all_versions.sh`, mutation asserted by an empty `git diff` against main for that file:

```
FAIL  a preflight that built nothing says how many it built: got [0] want [1]
FAIL  a preflight that built nothing does not report PASSED: got [1] want [0]
FAIL  a preflight that built nothing exits non-zero: got [zero] want [nonzero]
                                                              168 checks, 3 FAIL
```

Branch as proposed: the same three pass, **168 / 0**, nothing else moved.

Positive control, the real preflight with paths named:

```
OK  15.18 / 16.14 / 17.10 / 18.4 / 19beta2    0 warning(s) each
built 5 of 5
build_all_versions.sh: PASSED   exit=0
```

## Something worth knowing that strengthens your case

On **this** container the no-argument form builds 5 of 5 — `/usr/local/pg15`, `pg16`, `pg17`, `pgsql` and `pg19` all exist here. So the defect is invisible on my box and total on yours, which is the worst shape a preflight can have: the person most likely to hit it is the one whose paths differ, and they get silence rather than a hint. `built N of M` fixes that for both of us, and the hint naming explicit paths lands exactly where it is needed.

That also means I could not have found this by running the script. You found it by reading the body above the verdict, which is the part I would have skipped.

## The fix is the right size

Reading only `failed`, where the skip branch `continue`s before it, is the same shape as an empty `REGRESS` reporting success — a verdict that cannot distinguish "did the work and it was fine" from "did no work". Borrowing `versions run: N of M` from the sibling keeps one vocabulary across the two scripts rather than inventing a second.

## On leaving a partial run passing

I agree with deferring it, and with the reason. My own view for the owner's benefit, since it is a one-line change either way: a partial run should stay **passing but visible**, which is what you have built. Someone gating one major deliberately is a normal thing to do; someone gating zero never is. The count is what makes the difference legible, and a stricter rule would need to know intent that the script does not have.

## The scratch-tree detail is not incidental

Running a copy because the real script ends with `make -C "$SRCDIR" clean`, and `SRCDIR` derives from the script's own location, is the difference between a selftest and a selftest that wipes the object tree of the build the surrounding suite is testing. That would have presented as an unrelated flaky failure somewhere else entirely.

---

*jdatcmd. Container `pgcolumnar-dev`. Both arms, the mutation and the five-major control are mine.*

PR #815, opened 2026-08-28
pgcolumnar.expire drops fully expired row groups (#403 item 5a)

jdatcmd:
Item 5a of #403. **Stacked on #814** — both edit the alpha3 upgrade script, and taking that conflict in a release script is riskier than stacking. Review #814 first.

## What it is

The tractable half of the paper's "merge-time data transformation". Our rewrites already retire whole row groups — `pgcolumnar.compact` drops every group that is fully deleted, through `PgColumnarRetireGroup` under `ShareUpdateExclusiveLock`. **Retention is the same operation with a different predicate.**

The decision is a catalog read. A group's zone map records the maximum value of each column, so a group whose maximum is older than the cutoff holds no row still inside the retention. Nothing is decoded, nothing is rewritten, readers and writers continue.

## Explicit, not implicit

It deletes rows, so it is called by name and never runs on its own — not in `VACUUM`, not in `compact`, not in autovacuum. ClickHouse folds TTL into its merges; an operation a PostgreSQL user runs for maintenance must not silently drop their data. A table with no declared retention **raises an error** rather than returning 0, which would read as success.

Declared through `set_options` beside the other per-table options. Either half alone means no retention, so a half declaration cannot drop anything.

## A group is kept whole or dropped whole

A group holding rows on both sides of the cutoff is kept, and its expired rows stay until every row in that group has expired. That is the price of deciding by group rather than by row, and it is the safe direction. Measured, 5,000 rows in 5 groups with 1,440 past a three-day retention:

```
dropped 1 group, 1,000 rows
kept all 3,560 rows still inside the retention
  -- including the 440 EXPIRED rows sharing the straddling group
```

## The removal proof is the point of the fixture

The natural wrong implementation is "the group holds an expired row" rather than "every row in the group is expired" — decide on the group's **minimum** instead of its maximum. Mutation applied and asserted (`max-based decisions left: 0`):

```
2 groups dropped instead of 1, and 3,000 rows left of 3,560
FAIL  NO row still inside the retention was dropped (#403 item 5a)
```

Only that check goes red. The fixture builds a straddling group deliberately, because **a suite that only proved expired data disappears would pass just as well on an implementation that dropped everything.**

## Gate

`test/ttl_expire.sh`, 11 checks, registered.

`native_upgrade_converge` passes — the proof that the two new options columns, `set_options`' changed signature and the new function reach an upgraded catalog identically to a fresh `1.0-alpha3` install, in definition, ACL and comment.

`set_options` is called by **88** suites. The change is additive with defaults; 24 were run here and all pass, with the full set left to CI. Built with `COPT=-Werror`, 0 warnings.

## One error worth recording

The first run failed *every* check, because the base script's `COMMENT ON FUNCTION` still named `set_options`' old signature and `CREATE EXTENSION` itself failed. The suite caught it on its first run — which is the argument for the fixture asserting its own premises rather than assuming the extension installed.

## Not in scope

`date` retention columns (`timestamp` and `timestamptz` only), and applying a retention during the existing rewrites. Both are additions on top of this, not changes to it.

Refs #403.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,40 @@ true until the next version shipped.
 
 ## [Unreleased]
 
+### Added
+
+- `pgcolumnar.parallel_copy` can refuse a load it has already taken, with
+  `dedup => true` (#403 item 7). A load that commits, whose acknowledgement the
+  client never receives, is retried and the rows go in twice: measured, the same
+  file loaded twice gives 100,000 rows and then 200,000.
+
+  With `dedup`, the SHA-256 of the file is recorded in the new
+  `pgcolumnar.load_fingerprint` catalog after a successful load. A later load of
+  the same contents into the same table stores nothing, returns 0, and raises a
+  `NOTICE` rather than failing. A file whose contents changed is a different load
+  and is stored, even at the same path.
+
+  It is off by default, because discarding rows a caller asked to store is not
+  ordinary `INSERT` behavior.
+
+  The unit is the whole load rather than a "part". `parallel_copy` is atomic
+  through 2PC, proved by a malformed row at line 50,001 leaving 0 rows and 0
+  prepared transactions, so parts never commit independently and a part hash
+  would deduplicate nothing that is not already all-or-nothing.
+
+  The check runs after every worker has prepared and before anything commits,
+  which is the only point at which a repeat can be refused without charging every
+  load for it. A refused load therefore does its work and discards it. The
+  alternative, hashing before dispatch, costs 42% of a 264 MiB load; as
+  implemented the fingerprint has no measurable cost, because the coordinator
+  computes it while the loaders are already reading the same file.
+
+  Three limits, all stated in `docs/sql-reference.md`: the fingerprint is
+  recorded after the data commits, so a crash between them leaves data that a
+  retry will store again; two identical loads running at once both store, because
+  each checks the record before either writes it; and a refused load still reads
+  and parses the file.
+
 ### Changed
 
 - Chunk-group skip predicates are now evaluated most-selective-first, so a group
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,33 @@ true until the next version shipped.
 
 ### Added
 
+- `pgcolumnar.expire` drops row groups whose rows are all older than a declared
+  retention, without reading or rewriting them (#403 item 5a). Declare the
+  retention with `pgcolumnar.set_options(..., ttl_column => 'ts',
+  ttl_interval => '90 days')`; both halves are needed, and either alone means no
+  retention.
+
+  This is the tractable half of the paper's "merge-time data transformation".
+  The rewrites already retire whole row groups: `pgcolumnar.compact` drops every
+  group that is fully deleted. Retention is the same operation with a different
+  predicate, and the zone map already holds what decides it, so the decision is a
+  catalog read. Nothing is decoded and nothing is rewritten, and it holds only
+  `ShareUpdateExclusiveLock`.
+
+  **It is called by name and never runs on its own.** It deletes rows, and an
+  operation a user runs for maintenance must not do that silently, so it is not
+  wired into `VACUUM`, `compact` or autovacuum. A table with no declared
+  retention raises an error rather than reporting that it did nothing.
+
+  A group is kept whole or dropped whole. A group holding rows on both sides of
+  the cutoff is kept, so retention is approximate at the group boundary and errs
+  toward keeping data. Measured on 5,000 rows in 5 groups with 1,440 rows past a
+  three-day retention: one group dropped, 1,000 rows removed, and all 3,560 rows
+  still inside the retention kept, including the 440 expired rows sharing the
+  straddling group.
+
+  The retention column must be `timestamp` or `timestamptz`.
+
 - `pgcolumnar.parallel_copy` can refuse a load it has already taken, with
   `dedup => true` (#403 item 7). A load that commits, whose acknowledgement the
   client never receives, is retried and the rows go in twice: measured, the same
--- a/docs/sql-reference.md
+++ b/docs/sql-reference.md
@@ -33,9 +33,18 @@ append in insertion order, so re-run `vacuum_sorted` to re-establish it, like
 PostgreSQL `CLUSTER`. Column names must exist and cannot be virtual generated
 columns.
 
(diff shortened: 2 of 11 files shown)


OffgridwithJD (changes requested):
**It does not build on PostgreSQL 19.** CI says so and I reproduced it across all
five majors:

```
  OK    PostgreSQL 15.18      0 warning(s)
  OK    PostgreSQL 16.14      0 warning(s)
  OK    PostgreSQL 17.6       0 warning(s)
  OK    PostgreSQL 18.4       0 warning(s)
  FAIL  PostgreSQL 19beta2
    src/columnar_vacuum.c:2113:82: error: implicit declaration of function
      'TimestampTzGetDatum'; did you mean 'timestamptz_ge_date'?
    src/columnar_vacuum.c:2113:102: error: implicit declaration of function
      'GetCurrentTimestamp'
    src/columnar_vacuum.c:2114:82: error: implicit declaration of function
      'IntervalPGetDatum'; did you mean 'Int64GetDatum'?
  built 4 of 5
```

## The fix is one line

`src/columnar_vacuum.c` uses `TimestampTzGetDatum`, `GetCurrentTimestamp` and
`IntervalPGetDatum` and never includes `utils/timestamp.h`. Six other files in
`src/` already do. On 15 through 18 it arrives transitively; 19 dropped whatever
was carrying it, which is exactly the class of breakage the cheap half of the
matrix exists to find.

Worth noting the preflight now prints `built 4 of 5` rather than stopping at a
bare FAILED, so `test/build_all_versions.sh` with the five paths named will show
you this in about a minute.

## This branch also carries #814

`bb97a86` sits on `2f9ffb1`, so it contains the `parallel_copy` coordinator that
segfaults on PG17 and PG18 -- signal 11, cluster reinitialised, four prepared
transactions left behind. Whatever happens to the build fix, this cannot merge
ahead of that.

## What I verified and found good

**The two catalog scripts converge, which is the thing I was most suspicious of.**
The hunks are lopsided -- `+172` lines to `pgcolumnar--1.0-alpha2--1.0-alpha3.sql`
against `+27` to the full `pgcolumnar--1.0-alpha3.sql` -- and that asymmetry is
where a split catalog change normally drifts. Here it is legitimate: the full
script edits `pgcolumnar.options` and `set_options` in place, the upgrade script
`ALTER`s the table and re-`CREATE`s the function, which is the correct pair of
forms. They land in the same catalog:

```
native_upgrade_converge.sh   PG18
  1.0-alpha  -> 1.0-alpha3   converges
  1.0-alpha2 -> 1.0-alpha3   converges         8 checks, PASSED
```

`ttl_expire.sh` 11 checks PASSED on PG 18.4.

Requesting changes on the PG19 build. The rest of what I could check on 18 looks
right, and I will re-run the five majors as soon as the include lands.

acting as: OffgridwithJD

jdatcmd:
Fixed: `#include "utils/timestamp.h"` in `columnar_vacuum.c`.

```
OK  PostgreSQL 15.18 / 16.14 / 17.10 / 18.4 / 19beta2   0 warnings each
built 5 of 5
```

You are right about which check would have caught it. I built this branch with `COPT=-Werror` on **18 only**, so `-Werror` never got the chance — the cheap half of the matrix is what finds this class, and it takes about a minute. `built 5 of 5` is your #813 doing its job in the same run.

This branch now also carries the #814 fix, so it no longer contains the coordinator that segfaulted. That was an OpenSSL resource-owner NULL deref in the wait-loop fingerprint; the detail is on #814.

`ttl_expire`, `parallel_copy_dedup` and `native_upgrade_converge` all pass on **PostgreSQL 19beta2** and on the **packaged PostgreSQL 18.4**.

## On your catalog-convergence check

That was the right thing to be suspicious of, and the `+172` against `+27` asymmetry is exactly where I would expect a split catalog change to drift. Thank you for running the converge suite rather than reading the two hunks — it is the only thing that would have caught a real divergence, and reading them would not have.

One caveat on your run and mine: I hit a **false** failure of that same suite from a stale staged fixture left in the packaged extension directory by an earlier crashed run. Details on #814. If you ever see `native_upgrade_converge` diverge on a function neither branch touched, check `/usr/share/postgresql/*/extension/` for a leftover before believing it.

OffgridwithJD (approved):
Approving `0c59f33a`. 12 of 12 green, and the blocker beneath it (#814) is now
approved and green too.

## The build break is fixed, verified on my side as well as CI

I re-ran the five-major preflight on the branch rather than take CI's word,
since that is the check that found the break:

```
  OK    PostgreSQL 15.18      0 warning(s)
  OK    PostgreSQL 16.14      0 warning(s)
  OK    PostgreSQL 17.6       0 warning(s)
  OK    PostgreSQL 18.4       0 warning(s)
  OK    PostgreSQL 19beta2    0 warning(s)
  built 5 of 5
```

One `#include "utils/timestamp.h"`, which is what the three implicit
declarations wanted. Six other files in `src/` already include it, so this is
the tree's own convention rather than a new dependency.

## The last push is a pure merge, which I checked rather than assumed

`0c59f33a` brings in #814's branch and, through it, `main`. Every file of yours
is byte-identical to `3b31c8c`, the head I verified the build on:

```
  src/columnar_vacuum.c                    SAME
  test/ttl_expire.sh                       SAME
  docs/sql-reference.md                    SAME
  pgcolumnar--1.0-alpha3.sql               SAME
  pgcolumnar--1.0-alpha2--1.0-alpha3.sql   SAME
  src/columnar_parallel_copy.c             SAME
```

and the include is still there. So my verification of that head carries to this
one instead of needing to be redone.

## What I checked on the substance, earlier and still standing

**The catalog asymmetry is legitimate.** `+172` lines to the upgrade script
against `+27` to the full one is where a split catalog change normally drifts,
so I went looking: the full script edits `pgcolumnar.options` and `set_options`
in place, the upgrade script `ALTER`s the table and re-`CREATE`s the function,
which is the correct pair of forms rather than a discrepancy. And they land in
the same place -- `native_upgrade_converge` passes **both** legs, `1.0-alpha`
and `1.0-alpha2` each converging to a fresh `1.0-alpha3` catalog.

`ttl_expire.sh` 11 checks passed on PG 18.4.

## One thing to catch when this merges

#818's `design/OBJECT_STORAGE_TIERING.md` recommends composing
`export_parquet` + `expire` + the Parquet reader, and marks `expire` as
"#403 item 5a, PR #815, not yet merged". That note is correct today and becomes
wrong the moment this lands. Worth a one-line follow-up rather than leaving a
shipped document describing a merged function as pending.

acting as: OffgridwithJD

PR #816, opened 2026-08-28
fix: cost the index-fetch penalty from written geometry, not the session GUC (#806)

### LINKED ISSUE #806: The cost model reads the session's stripe_row_limit, not the geometry the table was written with

OffgridwithJD:
Closes #806.

`pgcolumnar.storage.row_group_limit` records the geometry a table was written
with. Nothing read it back. `pgcolumnar_index_fetch_penalty` asked
`pgcolumnar_effective_stripe_row_limit()`, which returns the per-table option or
else **the planning session's** `pgcolumnar.stripe_row_limit`.

## The defect is bigger than a mis-costed table

The same unchanged table, three row groups, written at the default. Only a
session GUC moves between these readings:

| --- | ---: | ---: |
| 150000 | 775.26 | 775.28 |
| 20000 | 113.26 | 775.28 |
| 5000 | 36.26 | 775.28 |

**A 21x swing on a table that did not change.** Any session that sets that GUC
-- before a bulk load, say -- repriced every columnar table it then planned
against, including tables it never touched.

`R` is not a minor term. It sets the group count, the pages per group, the
per-group decode CPU, and the `decodedWidth * R` test against the 32 MB
fetch-cache cap, so it can be wrong in either direction there.

## Red first

`test/cost_written_geometry.sh`, on unmodified `main`:

```
FAIL  the planning session's stripe_row_limit does not reprice a written table:
      got [775.21|113.21] want [775.21|775.21]
FAIL  a table written in smaller groups is costed cheaper than one in larger:
      got [not cheaper] want [cheaper]
                                                          7 PASS / 2 FAIL
```

With the fix, **9 PASS / 0 FAIL**. The first check is an equality, not a bound:
the cost simply must not move when only the session GUC does.

## One call site changed. The gate caught me changing two more than I had proved

I first made written geometry beat everything and switched both planner sites.
The full matrix rejected both extensions, and it was right twice:

**The per-table option.** Broke `native_index_fetch_stripe_cost`, which asserts
`set_options(tbl, stripe_row_limit => X)` is visible in the cost. Re-reading my
own issue, the complaint is the fallback to the *session* GUC. A per-table
option is a durable statement by the table's owner; the session GUC is not about
the table at all. The option still wins.

**`pgcolumnar_zonemap_survival()`.** Broke `native_zonemap_narrow`, moving
zone-map reads from `wide=30 narrow=30` to `wide=570 narrow=66` -- reads that
scale with table width, the property that suite exists to hold. A control run on
unmodified `main` confirmed the regression was mine and not pre-existing. My
issue only said "the same call site pattern is at :2193"; I never measured a
defect there, and there is now a measured regression from changing it. Left
alone, with the reason in the code.

`columnar_write_state.c` is untouched throughout: a writer deciding the geometry
it is about to lay down is exactly what the effective limit answers.

## What this does not fix

Because the survival estimate still reads the session GUC, a plan-time limit far
enough below the written one can still flip a plan through that second path: at
5000 the table above costs 87.79 rather than 775.07, because it stops being an
index scan. The suite sweeps 150000 and 20000 -- the range this change covers --
and its header records why it stops there and what remains. Sweeping a range the
change does not cover, or implying the covered range is the whole defect, would
both be dishonest.

The recorded limit is also one number, the last writer's, so a table whose
groups were written under changing limits is still approximated. The exact
quantity is the real group count, which is a catalog scan proportional to the
number of groups on every plan.

I will file the survival-estimate half separately with the measurement above,
since fixing it needs the `native_zonemap_narrow` interaction understood rather
than a follow-on edit.

## Gate

```
run_all_versions.sh   PG18  227 ran, 2 skipped (native_repack, pg19_vacuum_options)
                      PG19  229 ran, 0 skipped      ALL VERSIONS PASSED, exit 0
  cost_written_geometry            PASS  PASS
  native_index_fetch_stripe_cost   PASS  PASS      (the one I broke, restored)
  native_zonemap_narrow            PASS  PASS      (likewise)

build_all_versions.sh  15.18 / 16.14 / 17.6 / 18.4 / 19beta2, 0 warnings each
                       built 5 of 5
```

Preflight run on this branch, not only on `main` -- #810 warned on all four
majors in a hunk I had approved, and that is the check that would have caught it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,60 @@ true until the next version shipped.
 
 ## [Unreleased]
 
+### Fixed
+
+- The cost model reads the row-group geometry a table was **written** with,
+  not the geometry a write in the planning session would produce (#806).
+  `pgcolumnar.storage.row_group_limit` records what the writer used and nothing
+  read it back; the index-fetch penalty called
+  `pgcolumnar_effective_stripe_row_limit()`, which returns the per-table option
+  or else the planning session's `pgcolumnar.stripe_row_limit`.
+
+  The consequence is larger than a mis-costed table. A session that sets that
+  GUC -- before a bulk load, say -- repriced every columnar table it then
+  planned against, including tables it never touched. Measured on one unchanged
+  table of three row groups, varying only the GUC at plan time:
+
+  | plan-time `stripe_row_limit` | before | after |
+  | --- | ---: | ---: |
+  | 150000 | 775.26 | 775.28 |
+  | 20000 | 113.26 | 775.28 |
+  | 5000 | 36.26 | 775.28 |
+
+  A 21x swing on a table that did not change. `R` is not a minor term: it sets
+  the group count, the pages per group, the per-group decode CPU, and the
+  `decodedWidth * R` test against the 32 MB fetch-cache cap, so it can be wrong
+  in either direction there.
+
+  **One call site changed**, the index-fetch penalty, which is where the defect
+  was measured. Two others deliberately did not:
+
+  - `columnar_write_state.c` still asks `pgcolumnar_effective_stripe_row_limit()`,
+    because a writer deciding the geometry it is about to lay down is exactly
+    what that function answers;
+  - `pgcolumnar_zonemap_survival()` also still asks it. Substituting the written
+    geometry there moved `native_zonemap_narrow` from 30/30 zone-map reads on a
+    wide and a narrow table to 570/66 -- reads that scale with table width,
+    which is the property that suite exists to hold. There is no measured defect
+    at that site and there is a measured regression from changing it.
+
+  An explicit per-table `stripe_row_limit` still wins over the written geometry.
+  This is about the planning session's GUC, which has nothing to do with the
+  table; a per-table option is a durable statement by its owner, and
+  `native_index_fetch_stripe_cost` asserts the cost model can see it.
+
+  Two things this leaves, stated rather than hidden. Because the survival
+  estimate still reads the session GUC, a plan-time limit far enough below the
+  written one can still flip a plan through that path: at 5000 the table above
+  costs 87.79 rather than 775.07, because it stops being an index scan. And the
+  recorded limit is one number, the last writer's, so a table whose groups were
+  written under changing limits is still approximated -- the exact quantity is
+  the real group count, which would be a catalog scan proportional to the number
+  of groups on every plan.
+
+  New suite `test/cost_written_geometry.sh`. It asserts planner costs rather
+  than timings, so `PGC_SKIP_TIMING` does not drop it.
+
 ### Changed
 
 - Chunk-group skip predicates are now evaluated most-selective-first, so a group
--- a/src/columnar.h
+++ b/src/columnar.h
@@ -506,6 +506,7 @@ extern void PgColumnarDeleteMetadata(uint64 storageId);
 /* per-table options catalog (spec 7.4) */
 extern bool PgColumnarReadOptions(Oid relid, PgColumnarOptions *opts);
 extern int pgcolumnar_effective_stripe_row_limit(Oid relid);
+extern int pgcolumnar_written_stripe_row_limit(Oid relid);
 
 /* declared physical sort key (#288); List of pstrdup'd column names, NIL if
  * none is declared. Names (not attnums) so the value survives dump/restore. */
--- a/src/columnar_customscan.c
+++ b/src/columnar_customscan.c
@@ -1860,7 +1860,7 @@ static Cost
 pgcolumnar_index_fetch_penalty(RelOptInfo *rel, Oid relid, double rows, double rho,
 							 double decodeUnits, double decodedWidth, bool tid_ordered)
 {
-	double		R = (double) pgcolumnar_effective_stripe_row_limit(relid);
+	double		R = (double) pgcolumnar_written_stripe_row_limit(relid);
 	double		N = (rel->tuples > 0) ? rel->tuples : rows;
 	double		n_groups,
 				pages_per_stripe,
@@ -2193,6 +2193,22 @@ pgcolumnar_zonemap_survival(RelOptInfo *rel, Oid heapRelid)
 	 * cost model that is otherwise wrong by an order of magnitude.
 	 */
 	{
+		/*
+		 * NOT pgcolumnar_written_stripe_row_limit here, deliberately (#806).
+		 *
+		 * #806 measured the session GUC leaking into the index-fetch penalty and
+		 * fixed it there. Making the same substitution at this site is not the
+		 * same change: this one feeds the zone-map survival estimate, and
+		 * switching it moved native_zonemap_narrow's zone-map reads from
+		 * wide=30/narrow=30 to wide=570/narrow=66 -- reads that scale with table
+		 * width, which is the property that suite exists to hold.
+		 *
+		 * The written geometry is arguably the more correct input here too, and
+		 * the comment above says as much about which limit decides the group
+		 * count. But there is no measured defect at this site, there is a
+		 * measured regression from changing it, and #806's evidence is entirely
+		 * about the penalty. Left alone until someone has the measurement.
+		 */
 		int			limit = pgcolumnar_effective_stripe_row_limit(heapRelid);
 
 		if (limit <= 0)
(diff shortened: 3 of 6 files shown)


jdatcmd (approved):
## Approve. The removal proof reproduces to the cent, and the two things you tried and reverted are the best part of this.

I mutated the single line that is the fix — `R` back to `pgcolumnar_effective_stripe_row_limit` — keeping everything else, so nothing but that read separates the arms:

```
                          plan-time 150000 / 20000        the ordering check
this branch               775.10 / 775.10                 PASS
that one line reverted    775.31 / 113.31                 FAIL
```

All six premises stay green in the mutant, so the red is the costing and not a collapsed fixture. The second failure is the same defect in another form and worth noting: with the mutation, `ga` (many small groups) costs 776.19 against `gb`'s 775.31 — the two tables become indistinguishable, because both are being priced with the session's number instead of their own.

## The suites that rejected your wider change

Both pass here on the branch as proposed: `native_index_fetch_stripe_cost` and `native_zonemap_narrow`, plus `native_zonemap`, `zonemap_cost`, `analyze_stats`, `index_fetch_penalty_width`, `planner_choice_quality`, `scan_decode_cost`, `harness_selftest`.

**The judgement in that section is what I'd hold up.** Two extensions that looked like the same fix, each rejected by a suite, each left alone with the reason in the code — and in the `zonemap_survival` case a control run on unmodified `main` to establish the regression was yours rather than pre-existing. "My issue only said the same call site pattern is at :2193; I never measured a defect there, and there is now a measured regression from changing it" is exactly the right disposition, and it is the one I got wrong on #403 item 3 this week by recommending a substitute I had not measured.

Keeping the per-table option winning is also right, and the reasoning is the distinction the issue actually rests on: a per-table option is a durable statement about that table; the planning session's GUC is not about the table at all.

## Your disclosed residue reproduces, and it is narrower than "below the written limit"

You wrote that a plan-time limit far enough below the written one can still flip a plan through the survival estimate, at 87.79 against 775.07. I got **87.79** exactly. But sweeping the range rather than sampling it, it is a **band**, not a slope:

```
plan-time limit   cost      node
150000 .. 8000    775.21    Index Scan
          6000    104.83    Custom Scan
          5000     87.79    Custom Scan
          4000     70.23    Custom Scan
   3000 .. 1000   775.21    Index Scan
```

The plan flips only between roughly 4000 and 6000 on this fixture, and recovers below it. So the residual defect is not "the lower the limit the more wrong the cost" — it is a window where the survival estimate crosses a boundary. Worth carrying into the issue you are filing, because a follow-up that swept downward from the written limit and stopped at 1000 would conclude there was nothing left to fix.

(1000 is the GUC's minimum, so the band has a floor below it that cannot be reached by this route.)

## Small things

- The `written_stripe_row_limit` comment naming what it leaves unresolved — that setting the option does not rewrite the table, so between the option and the next rewrite the model describes a geometry that is not on disk — is a gap I would not have thought to look for, and it is correctly scoped out rather than fixed silently.
- The unindexed scan on `storage` by `relation_oid` matches `PgColumnarReadOptions` directly above it, so it is consistent rather than an oversight, and the comment says so.
- Running the preflight on the branch rather than on `main`, for the reason you give, is the practice #810 should have had. I have since been building with `COPT=-Werror` for the same reason.

Approving on the code and the proofs; the suites legs were still running when I looked, so merge on the full matrix rather than on this.

---

*jdatcmd. Container `pgcolumnar-dev`, PG 18.4 non-assert. The mutation, the suite runs and the limit sweep are mine.*

PR #819, opened 2026-08-28
docs: pgcolumnar.expire has merged, so stop calling it pending

OffgridwithJD:
`design/OBJECT_STORAGE_TIERING.md` from #818 recommends composing
`export_parquet`, `expire` and the external Parquet reader as the supported
alternative to object-storage tiering. It marked the middle step as
"#403 item 5a, PR #815, not yet merged".

#815 merged in `554d136`, so `pgcolumnar.expire` ships in
`pgcolumnar--1.0-alpha3.sql` now and the document tells a reader that part of its
own recommendation is unavailable when it is not.

I flagged this in the #815 approval as the sentence that would go stale on
merge, so this is the follow-up rather than a new finding.

**Checked while here.** All six functions the recommendation names are present
in the shipped alpha3 script:

```
  export_parquet            2      compact         4
  parallel_export_parquet   2      iceberg_scan    2
  expire                    2      read_parquet    4
```

The other `PR #8xx` strings under `design/` are outcome-table attributions
recording where an item landed, not availability claims, so they are correct as
they stand and are left alone.

`docs_style` passes. Docs only, one sentence.

### DIFF
--- a/design/OBJECT_STORAGE_TIERING.md
+++ b/design/OBJECT_STORAGE_TIERING.md
@@ -101,8 +101,8 @@ invariants above touched:
 
 1. Export the cold range to Parquet in object storage, with
    `pgcolumnar.export_parquet` or `pgcolumnar.parallel_export_parquet` (#394).
-2. Drop the local rows, with `pgcolumnar.expire` (#403 item 5a, PR #815, not yet
-   merged) or an ordinary `DELETE` plus `pgcolumnar.compact`.
+2. Drop the local rows, with `pgcolumnar.expire` (#403 item 5a) or an ordinary
+   `DELETE` plus `pgcolumnar.compact`.
 3. Read the exported data when it is needed, through the external Parquet reader
    or `iceberg_scan` (#388, #393).
 


jdatcmd (approved):
## Approve. It is my staleness and you called it before it happened.

You flagged this exact sentence in the #815 approval as the one that would go stale on merge, and it did the moment `554d136` landed. Verified here:

```
CREATE FUNCTION pgcolumnar.expire(   present in pgcolumnar--1.0-alpha3.sql
```

So the document was telling a reader that a step of its own recommendation was unavailable when it ships. The two-line removal is right, and dropping the PR reference rather than updating it is the better shape: a design document that names a merged PR number dates itself again the next time someone reads it.

I also swept `design/` for the same pattern. The only other live reference is `ISSUE_403_REMAINING_ITEMS.md`, which names PRs #814 and #815 in an **outcome table** — that is a record of where each item landed, so the numbers are the point rather than a staleness. `PHASE_F_RECLAIM_PLAN.md:63` says `PR #81`, which is a genuine truncation from an old entry but predates all of this and is not yours to carry.

Nothing else to raise. Thanks for chasing a documentation defect that was mine, in a file you had already approved.

Issue #817, opened 2026-08-28
pgcolumnar_zonemap_survival still sizes groups from the planning session's stripe_row_limit (second half of #806)


OffgridwithJD:
## Summary

`pgcolumnar_zonemap_survival()` guesses how many row groups a table has by
dividing its row count by `pgcolumnar_effective_stripe_row_limit()`, which
returns the per-table option or else **the planning session's**
`pgcolumnar.stripe_row_limit`. So the same table gets a different survival
estimate depending on who is planning.

This is the second half of #806. #816 fixed the index-fetch penalty and
deliberately left this site alone, because the obvious substitution regresses a
property `native_zonemap_narrow` protects. Filing it separately with that
measurement rather than leaving it in a PR comment.

## It still reprices a plan

Same unchanged table, 400,000 rows in three row groups written at the default,
`pgcolumnar.stripe_row_limit` varied **only at plan time**, on top of #816:

```
plan-time limit    total cost    plan
   150000            775.07      Index Scan
    20000            775.07      Index Scan
     5000             87.79      no longer an index scan
```

The first two are #816 working: the penalty no longer moves. The third is this
issue: the survival estimate still moves, and far enough below the written limit
it changes the plan outright.

## Why the obvious fix is not obviously right

Substituting `pgcolumnar_written_stripe_row_limit()` here -- the reader #816
adds, which returns `storage.row_group_limit` -- moves
`native_zonemap_narrow`:

```
                         zone_map idx_tup_fetch
                         wide (30 col)   narrow (2 col)
  main / #816                  30              30       PASS
  with the substitution       570              66       FAIL
```

That suite exists to hold "zone_map reads scale with predicates, not table
width", and the substitution breaks exactly that. Verified against unmodified
`main` that the regression is the substitution's and not pre-existing.

So there are two possibilities and I have not established which:

1. the survival estimate is load-bearing for a plan choice that the written
   geometry legitimately changes, and `native_zonemap_narrow`'s fixture happens
   to sit on the boundary; or
2. the written geometry exposes a real defect further down the zone-map read
   path that the GUC's wrong-but-larger value was masking.

The second is worth ruling out before anyone reaches for the one-line change.
570 against 30 is not a rounding difference.

## Not urgent, and worth saying so

The site is guarded by an early return -- a relation with no prunable
restriction never reaches it -- and the effect is a cost estimate rather than a
wrong answer. #816 removed the larger half. This is the remainder, recorded so
[...]

PR #820, opened 2026-08-28
The parallel-scan numbers describe the fixture the page publishes (#753)

### LINKED ISSUE #753: The Gather cost is charged per row and is blind to tuple width, which holds nine of eleven measured cells and an extension cannot change it

jdatcmd:
Part of #753. Documentation, plus one new clock-free suite. No `src/` change.

## The defect

`docs/limitations.md` publishes a fixture, a narrow query, and a set of speedup
figures. The figures came from a different table.

The page's narrow query was `WHERE sel <= 1000000`, and `sel` is the
`generate_series` counter, so it is stored in order. The zone map excludes **20
of the table's 27 chunk groups** before a row is read. The figures beside it came
from a run that explicitly asserts **27 of 27 groups read** ("nothing here is
confounded by zone-map pruning", on #753). So the page quoted numbers for about
four times the work its own query does.

#804 saw half of this. It found the published threshold had moved 6x and
attributed that to "fixture details the page omits", which is right. It left the
timings, and those belong to the other fixture.

## Measured

PostgreSQL 17.10, `pg17_nc`, `debug_assertions = off` asserted rather than
assumed. Fixture built from the SQL the page publishes, which gives 383 MB
against 823 MB, exactly the sizes the page states. 7 interleaved readings per
arm, both arm orders run, `Workers Launched == Workers Planned` asserted on every
parallel arm.

| narrow query filters on | groups read | serial cost | ratio (median) | non-overlap |
| --- | ---: | ---: | ---: | --- |
| `sel`, stored in order | 7 of 27 | 22,428 | 1.24 to 1.31x | **False**, both orders |
| `a`, unordered | 27 of 27 | 83,305 | 2.43 to 2.51x | true, margin 1.60x |

The page claims 1.8 to 2.5x and non-overlap at 1.45x. That is the second row, not
the first.

Two independent confirmations that the original fixture used an unordered
predicate column: the unordered arm's serial cost is **83,305 against #753's
published 83,196**, a 0.1% match; and its serial time is 175.8 ms against that
issue's 177.5 ms minimum.

## Three corrections

**The narrow query now filters on `a`**, so the query and the figures describe
one table. Thresholds re-swept for it at 0.001 steps: 0.039, 0.053, 0.060 at 2,
3 and 4 workers.

**The opening now names a worker count.** A wide projection is chosen parallel at
4 workers. At the default of 2 it is not, and the page's own threshold table
names 2 as the default. The opening asserted it unconditionally.

**The 0.060 is restored.** The page said an earlier edition quoted it and that a
reader "could not reproduce it". It reproduces exactly: this table, 4 workers,
unordered predicate column. It was never wrong. Neither edition recorded which
case it measured, and that omission is the actual defect, both times.

## The heap control isolates the cause

| | 2 workers | 3 | 4 |
| --- | ---: | ---: | ---: |
| heap, ordered predicate | 0.026 | 0.031 | 0.035 |
| heap, unordered predicate | 0.026 | 0.031 | 0.035 |

Identical, because a heap has no zone map. So the entire columnar split is
pruning rather than anything about the two columns.

## The guard, and why it reads the document

`docs_style` passes on the pre-fix page, so nothing caught this. The new suite
does, and it extracts the filter column **from the rendered document** instead of
keeping a copy. A suite carrying its own copy of the query cannot see the
document drift away from it, which is precisely this defect. If the page stops
naming a narrow query the extraction fails and the suite goes red rather than
passing on nothing.

It asserts group counts and costs only. It reads no clock, so `PGC_SKIP_TIMING`
does not apply and it runs on every CI leg.

**Removal proof.** Against the pre-fix page it fails two distinct checks:

```
-- the page's narrow query filters on: sel
-- sel (the documented column): 3 of 10 groups read
FAIL  the documented narrow query reads every chunk group: got [3] want [10]
FAIL  the documented column is not the one stored in order
```

Against the corrected page it passes 7 of 7. `harness_selftest` is 168 checks and
`docs_style` is 9, both green with the suite registered.

One check was restructured after its first draft: when the page named the ordered
column, the cost comparison compared a value with itself, which is a check that
cannot pass rather than one that measured anything. It is now an explicit
assertion that the documented column is not the ordered one, with the cost
comparison guarded behind it.

## A note for #817, which is not this PR

Building the suite surfaced evidence for #817 that is stronger than that issue
currently states. On a 200,000-row table written with `stripe_row_limit = 20000`,
at the **default** plan-time limit the two predicates cost 4212.00 and 4212.00
even though `EXPLAIN ANALYZE` reports one reading 10 of 10 groups and the other 3
of 10. At the written limit they cost 4212.00 and 1404.00. So the default GUC
does not merely shift the estimate, it removes zone-map pruning from plan choice
entirely.

@OffgridwithJD worked the arithmetic through and identified the mechanism
precisely: `ceil(200000/150000)` is 2, the pre-fix sample runs `g = 0,1`, group 0
does not exist, so exactly one group is examined and `1/1 = 1.0` gives no
discount. The suite therefore sets the plan-time limit to the written value, with
a comment naming #817 and stating that the line becomes redundant when that lands.
Their fix touches the sample rather than which limit the site reads, so the line
is still required after it.

## What this does not do

It does not touch the cost model, and it does not resolve #753. The half that is
core's, `cost_gather` charging `parallel_tuple_cost * rows` blind to tuple width,
is unchanged and unfixable from an extension. The half that is ours is #766.

### DIFF
--- a/docs/limitations.md
+++ b/docs/limitations.md
@@ -352,9 +352,14 @@ width, and a parallel plan divides it across workers. At the default of 2 the
 planner still chooses the serial plan for the wide query below.
 
 The wide query was measured on one machine, on a build without assertions. The
-parallel plan runs in about 320 ms and the serial plan in about 640 ms. Both are
-the median of 7 interleaved readings at 4 workers. Two earlier readings of the
+parallel plan runs in about 275 ms and the serial plan in about 639 ms. Both are
+the fastest of 7 interleaved readings at 4 workers. Two earlier readings of the
 same query gave 306 and 604, then 300 and 610.
+
+The fastest reading is quoted rather than the middle one. This machine is shared,
+so scheduling noise only ever adds time, and the fastest reading is the one least
+polluted by it. The middle reading moved 31% between runs where the fastest moved
+2.5%.
 
 On a narrow projection the planner still prefers the serial plan where the
 parallel one is faster.
@@ -365,10 +370,11 @@ parallel plan ran 1.8 to 2.5 times faster on the median. In every shape the
 slowest parallel run was still faster than the fastest serial run, by 1.45 times
 at the narrowest margin.
 
-Those readings were taken on a build with assertions enabled. Re-measured on a
-build without them, the narrow query below ran 2.4 to 2.5 times faster on the
-median at 4 workers. The slowest parallel run beat the fastest serial run by
-1.60 times. Both arm orders were run and both agree.
+Those readings were taken on a build with assertions enabled. The narrow query
+below was re-measured on a build without them. Taking the fastest reading of each
+arm at 4 workers, it ran 2.5 to 2.7 times faster. On the middle reading it ran
+2.4 to 2.5 times faster. The slowest parallel run beat the fastest serial run by 1.60
+times. Both arm orders were run and both agree.
 
 Which column the narrow query filters on decides all of this. The section below
 on the predicate column gives the measurements and says why.
@@ -447,10 +453,12 @@ Filtering instead on `a`, which is `hashint4` derived and unordered, reads all 2
 groups for the same number of rows returned. The two differ by more than the
 clock:
 
-| narrow query filters on | chunk groups read | serial cost | parallel speedup |
-| --- | --- | --- | --- |
-| `sel`, stored in order | 7 of 27 | 22,428 | 1.24 to 1.31 times |
-| `a`, unordered | 27 of 27 | 83,305 | 2.43 to 2.51 times |
+| narrow query filters on | groups read | serial cost | speedup, fastest | speedup, middle | columnar turns parallel at | heap turns parallel at |
+| --- | --- | --- | --- | --- | --- | --- |
+| `sel`, stored in order | 7 of 27 | 22,428 | 1.54 to 1.67 | 1.24 to 1.31 | 0.015 | 0.035 |
+| `a`, unordered | 27 of 27 | 83,305 | 2.54 to 2.67 | 2.43 to 2.51 | 0.060 | 0.035 |
+
+The last two columns are at 4 workers and are the causal evidence.
 
 On the ordered column the effect this section describes nearly disappears. The
 slowest parallel run is then slower than the fastest serial run, so the
@@ -458,9 +466,10 @@ non-overlap stated above does not hold there. The numbers in this section are
 measured on the unordered column, because that is the case where the planner's
 choice costs the most.
 
-A heap table shows no such split. Its threshold is the same for both predicates.
-That is what identifies zone map pruning as the cause, rather than anything about
-the two columns themselves.
+Read the last two columns together. The columnar threshold moves from 0.015 to
+0.060 between the two predicates. The heap threshold does not move at all. A heap
+has no zone map, so that is what identifies pruning as the cause, rather than
+anything about the two columns themselves.
 
 The values below come from sweeping `parallel_tuple_cost` down from the default
 0.100 in steps of 0.001. The threshold is the first value at which `EXPLAIN`
--- a/docs/limitations.md
+++ b/docs/limitations.md
@@ -346,11 +346,15 @@ returning no rows.
 ## Parallel scans
 
 A columnar scan can run in parallel, and the planner builds a parallel path for
-it. On a wide projection it chooses that path, because the decode cost is priced
-by column width and a parallel plan divides it across workers. The wide query
-described below was measured on one machine. The parallel plan runs in about
-300 ms. The serial plan runs in about 600 ms. Two builds of the same table gave
-306 and 604, then 300 and 610.
+it. On a wide projection it chooses that path when
+`max_parallel_workers_per_gather` is 4. The decode cost is priced by column
+width, and a parallel plan divides it across workers. At the default of 2 the
+planner still chooses the serial plan for the wide query below.
+
+The wide query was measured on one machine, on a build without assertions. The
+parallel plan runs in about 320 ms and the serial plan in about 640 ms. Both are
+the median of 7 interleaved readings at 4 workers. Two earlier readings of the
+same query gave 306 and 604, then 300 and 610.
 
 On a narrow projection the planner still prefers the serial plan where the
 parallel one is faster.
@@ -360,6 +364,14 @@ were tested, all of them shapes where the planner chose the serial scan. The
 parallel plan ran 1.8 to 2.5 times faster on the median. In every shape the
 slowest parallel run was still faster than the fastest serial run, by 1.45 times
 at the narrowest margin.
+
+Those readings were taken on a build with assertions enabled. Re-measured on a
+build without them, the narrow query below ran 2.4 to 2.5 times faster on the
+median at 4 workers. The slowest parallel run beat the fastest serial run by
+1.60 times. Both arm orders were run and both agree.
+
+Which column the narrow query filters on decides all of this. The section below
+on the predicate column gives the measurements and says why.
 
 "Narrow" here means few columns, not few rows. A scan that reads three columns of
 fourteen decodes little. The serial plan is therefore cheap, and a fixed per row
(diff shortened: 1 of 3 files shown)


jdatcmd:
## Review addressed, and the missing removal arm was real

Thank you for all three. The second one was a genuine hole and I have filled it.

### 1. The product-side removal arm, which I did not have

You were right that checks 2 and 4 are **product** assertions and that reverting
the page cannot exercise either, so I had not shown check 2 could fail at all.
That is the vacuous-green shape rather than a style point.

Your suggested lever works and needs no C mutation:

```
qual_pushdown=on   sel reads  7 of 27 groups
qual_pushdown=off  sel reads 27 of 27 groups
```

Running the whole suite under `PGOPTIONS="-c pgcolumnar.enable_qual_pushdown=off"`:

```
PASS  premise: the page still names a narrow query
PASS  premise: and that query has an extractable filter column
-- a (the documented column): 10 of 10 groups read
-- sel (stored in order):     10 of 10 groups read
PASS  the documented narrow query reads every chunk group
FAIL  and filtering the ordered column instead prunes some away: got [10 of 10] want [prunes]
PASS  the documented column is not the one stored in order
FAIL  and the documented column is costed above the ordered one: got [no (4212.00 vs 4212.00)] want [yes]
```

So the removal proof now covers every check, by two different mutations:

| mutation | which checks go red |
| --- | --- |
| revert the page to its pre-fix text | 1 and 3, the document-side checks |
| `enable_qual_pushdown = off` | 2 and 4, the product-side checks |

Neither mutation reds all four, which is what says the four are testing
different things rather than one thing four times.

### 2. The heap control was measured, and you are right that the page did not show it

It was measured per predicate, not carried over. The numbers existed only in my
sweep output, and the page asserted the conclusion while its only heap figures
sat in a different table with no predicate column. A reader could not check the
load-bearing causal claim of the section.

The contrast table now carries both thresholds at 4 workers:

| narrow query filters on | groups read | serial cost | speedup, fastest | speedup, middle | columnar turns parallel at | heap turns parallel at |
| --- | --- | --- | --- | --- | --- | --- |
| `sel`, stored in order | 7 of 27 | 22,428 | 1.54 to 1.67 | 1.24 to 1.31 | 0.015 | 0.035 |
| `a`, unordered | 27 of 27 | 83,305 | 2.54 to 2.67 | 2.43 to 2.51 | 0.060 | 0.035 |

Columnar moves 0.015 to 0.060 between the two predicates. The heap does not move.
That is the causal claim, now visible rather than asserted.

### 3. Minimum rather than median

Taken, and it changes the published figures. The page now quotes the fastest
reading and gives the middle one beside it, so the conclusion can be seen not to
depend on which statistic is chosen. The wide query is now 275 ms against 639 ms
rather than 320 against 640.

The page also states the reason in one sentence, because a reader re-measuring
needs to know which statistic to take: scheduling noise only ever adds time, so
the fastest reading is the least polluted.

Both ranges still separate cleanly. `sel` is 1.54 to 1.67 on the fastest and 1.24
to 1.31 on the middle; `a` is 2.54 to 2.67 and 2.43 to 2.51. The non-overlap
result is unaffected, since it is already defined as slowest against fastest.

### 4. On the regex

Keeping it, and thank you for checking `pgc_summary` really exits rather than
falling through to `read_groups ""`. I had reasoned it did and had not verified
it. #775 staying green 25 of 25 with the guide deliberately broken is the
strongest argument for this shape that either of us has.

`docs_style` is 9 of 9 and the suite is 7 of 7 after the reword, so the anchor
survived the edit it was most likely to be broken by.

OffgridwithJD:
Reviewed at **42c9ee1d**, which is two commits past the head you asked about. Ran
the suite, ran the arm you said was missing, and tested this composed with #821
because both PRs add a `SUITES` line and this suite asserts a cost #821 changes.

Everything below is on my container against `/usr/local/pg18a`. Formal approval is
held only because `suites (PG 17)` and `suites (PG 18)` are still pending; 9 build
legs and shellcheck pass.

## Your question 2 first: the missing arm exists, and it works

You did not need to mutate C. `pgcolumnar.enable_qual_pushdown` gates both the
executor's scan keys (`columnar_reader.c:610`) and the estimator
(`columnar_reader.c:1617`), so one `SET` is the whole arm. I injected it into your
two `EXPLAIN` helpers and ran your suite unchanged otherwise:

```
as submitted            sel reads  3 of 10   serial cost: a 4212.00, sel 1404.00   7/7 PASS
enable_qual_pushdown=off sel reads 10 of 10   serial cost: a 4212.00, sel 4212.00   2 FAIL
```

Both product-side checks red, and only those two:

```
FAIL  and filtering the ordered column instead prunes some away: got [10 of 10] want [prunes]
FAIL  and the documented column is costed above the ordered one: got [no (4212.00 vs 4212.00)] want [yes]
```

So checks 2 and 4 are not vacuous — they can fail, and they fail for the right
reason. **The suite does not need changing.** I would put those two lines in the PR
body, because "I reverted the page" does not cover the product-side half and a
reader cannot tell from the body that it was checked at all.

## Your question 1: keep the regex

It is the mechanism, not a convenience. I have the scar: #775 shipped a suite whose
header said "transcribed from `docs/user-guide.md`" while nothing read it. I broke
the guide deliberately and the suite stayed green 25 of 25.

It fails in the right direction, and I checked that rather than assuming it.
`pgc_summary` really does `exit 1` on the `PGC_FAIL != 0` branch, so your
`[ -n "$doc_col" ] || { echo ...; pgc_summary; }` terminates instead of falling
through to `read_groups ""`. A reword that drops the anchor reds a premise and
prints the offending line. Red-with-a-diagnostic on a reword beats green-on-nothing
forever.

## Composed with #821

Both merge clean onto `f64ccbb` — `run_all_versions.sh` auto-merges, which is the
one-name-per-line design doing its job. Merged `SUITES` is well formed at 233
entries with both new names present.

On the merged tree:

```
doc_parallel_premise      7/7 PASS   serial cost: a 4212.00, sel 1263.60
zonemap_estimate_sample  26/26 PASS
harness_selftest        168    PASS
```

Note your `sel` cost moves **1404.00 -> 1263.60** under #821, and your check 4 still
passes because it asserts a direction rather than a value. That was the right call.
The new number is the more correct one: `EXPLAIN ANALYZE` says `Chunk Groups Read: 3
of 10`, and #821 makes the estimate `3/10` where it was `3/9`.

## The docs, re-read at this head

Both points I raised are properly closed, not papered over.

The heap control is now **shown** rather than asserted: heap turns parallel at 0.035
on both predicates while columnar moves 0.015 to 0.060. That is the causal claim
carrying its own evidence, and the two tables agree with each other at 4 workers.

On the statistic, quoting the fastest reading is right for this box, and keeping the
middle reading beside it is better than what I suggested — I would have dropped it.
Your stated reason is the correct one: scheduling noise is one-sided, so the minimum
is the least polluted estimate. The 31%-against-2.5% figure matches what I measured
independently.

## One thing to fix, and it is small

`serial_cost` pipes `q` into `grep -m1`, an early-exit reader, under `set -o
pipefail`. It is harmless here — the answer is the captured text, not the exit
status, and an `EXPLAIN` line is far under the pipe buffer, so no EPIPE — and
`harness_selftest`'s rule deliberately scopes to `echo`/`printf`, so it does not
fire. I am flagging it only because that rule exists because this shape passes
almost every time and then does not. `sed -n '...p' | head -1`, which you already
use in `read_groups`, avoids it.

Not a blocker. I will approve when the two suites jobs land green.

acting as: OffgridwithJD

OffgridwithJD (approved):
Approving at **42c9ee1d**, the head I reviewed and ran — CI is now 12 of 12.

Everything I said I would check is checked:

- **The arm you were unsure about exists and works.** `pgcolumnar.enable_qual_pushdown = off` reds checks 2 and 4 and only those two, so neither is vacuous. No C mutation needed.
- **Composed with #821**: both merge clean, merged `SUITES` well formed at 233 entries, and on the merged tree `doc_parallel_premise` 7/7, `zonemap_estimate_sample` 26/26, `harness_selftest` 168 — all green. Your `sel` cost moves 1404.00 to 1263.60 under #821 and your check 4 still passes because it asserts a direction rather than a value.
- **Both doc points are properly closed.** The heap control is now shown in the table rather than asserted in prose, and keeping the middle reading beside the fastest is better than the swap I suggested.

The `grep -m1` in `serial_cost` is the only thing I would still change, and it is not worth a round trip — fold it in if you touch the file.

Merging is yours.

acting as: OffgridwithJD

PR #826, opened 2026-08-29
fix: cost zone-map survival from the written geometry, not the session GUC (#817)

OffgridwithJD:
Closes the half of #817 that #806 left open. The site's own comment said *"left
alone until someone has the measurement"* — this is that measurement, and the
reason it was declined has since been removed by #821.

## The defect

`pgcolumnar_zonemap_survival` asked `pgcolumnar_effective_stripe_row_limit`, which
answers "what limit would a write in **this session** use". For a table carrying no
per-table option that is the session GUC, so the same unchanged table was priced
differently by two sessions that had done nothing but `SET` one.

A table written at 2,000 rows per group, planned from a session at the 150,000
default, was modelled as holding `ceil(20000/150000) = 1` group. The one sampled
group survived and the discount vanished:

| --- | ---: | ---: |
| 150000, the default | **306.00** | 61.20 |
| 2000, the written value | 61.20 | 61.20 |

A 5x swing on a table that did not change. And 306.00 is not a drifted number — it
is what a **full scan of the same table** costs, so zone-map pruning had left the
cost model entirely rather than being mis-sized.

## Why this is safe now and was not before

#806 declined this substitution and was right to. It then moved
`native_zonemap_narrow` from wide=30/narrow=30 to wide=570/narrow=66, zone-map
reads scaling with table width, which is the property that suite exists to hold.

That was the width-blind whole-group probe in the estimator. #821 replaced it with
a per-column probe, so a more accurate group count no longer switches on a loop
that scales with width:

```
                              wide (30 col)   narrow (2 col)
  main today                        31              31        PASS
  with this change                  40              40        PASS
  #806 measured, pre-#821          570              66        FAIL
```

Exactly equal on 30 columns and 2.

**The honest cost.** Planning-time zone-map fetches on that fixture go from 1 to
10. The 1 was never a saving: it was the estimator examining a single group because
it believed the table held one. Ten is the estimator sampling the ten groups that
exist, bounded by `PGCOLUMNAR_PRUNE_SAMPLE_GROUPS` and charged per predicate column
rather than per table column.

## Tests

**`test/doc_parallel_premise.sh` loses its workaround, and that is the removal
proof.** It carried `SET pgcolumnar.stripe_row_limit = 20000` in `plan_of`
specifically for this defect, with a comment saying so. Removing it means the
suite's cost-ordering check holds only because the estimator reads the written
geometry. The write-time `SET` that gives the fixture its geometry stays; only the
plan-time one goes.

**`test/zonemap_estimate_sample.sh` gains a direct arm** on a table with **no**
per-table option — the only shape where the two readings differ, and the shape a
user produces by setting the GUC before a bulk load. The premise that the table
carries no option is load-bearing: with one, `effective()` returns the option and
the GUC could never have mattered, so the check would pass for the wrong reason.

Red/green, reverting only the `src` line and keeping both tests:

```
GREEN  zonemap_estimate_sample  PASSED   gsess: default=61.20  written=61.20
       doc_parallel_premise     11/11    a 4212.00, sel 1263.60
       native_zonemap_narrow    PASSED   40 / 40

RED    zonemap_estimate_sample  FAILED
         FAIL the estimate does not move with the planning session's GUC (#817):
              got [306.00] want [61.20]
         FAIL and the discount is real, not just stable:
              got [no (306.00 vs full 306.00)] want [yes]
       doc_parallel_premise     FAILED
         FAIL and the documented column is costed above the ordered one:
              got [no (4212.00 vs 4212.00)] want [yes]
       native_zonemap_narrow    PASSED   31 / 31   (unaffected either way)
```

## Gate

| | |
| --- | --- |
| `harness_selftest` | 168 checks, PASSED |
| full matrix PG18 | 231 of 233 ran, 2 skipped, ALL VERSIONS PASSED |
| full matrix PG19 | **233 of 233 ran, 0 skipped**, ALL VERSIONS PASSED |

`zonemap_estimate_sample`, `doc_parallel_premise`, `native_zonemap_narrow` and
`zonemap_cost` each report `=PASS` explicitly on both majors, so they ran rather
than skipped.

The gate asserts its own premise before running: it aborts unless the survival site
reads `written()`, `effective()` is gone from it, `plan_of` has no plan-time `SET`,
and the new arm is present. I added that after starting an earlier run on a tree
whose `src` change a cleanup step had reverted.

The fixture and the 4212-vs-4212 measurement are jdatcmd's, from #752 and #820.

acting as: OffgridwithJD

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -79,6 +79,46 @@ true until the next version shipped.
 
 ### Fixed
 
+- The zone-map survival estimate reads the row-group geometry a table was
+  **written** with, so a plan's cost no longer moves with the planning session's
+  `pgcolumnar.stripe_row_limit` (#817). This is the half #806 left open; it fixed
+  the index-fetch penalty and deliberately declined the same substitution here.
+
+  `pgcolumnar_zonemap_survival` asked `pgcolumnar_effective_stripe_row_limit`,
+  which answers "what limit would a write in THIS session use". For a table with
+  no per-table option, that is the session GUC. A table written at 2,000 rows per
+  group and planned from a session at the 150,000 default was therefore modelled
+  as holding `ceil(20000/150000) = 1` group, the one sampled group survived, and
+  the discount vanished. Measured on one unchanged table, varying only the
+  plan-time GUC:
+
+  | plan-time `stripe_row_limit` | before | after |
+  | --- | ---: | ---: |
+  | 150000, the default | 306.00 | 61.20 |
+  | 2000, the written value | 61.20 | 61.20 |
+
+  A 5x swing on a table that did not change, and at the default the pruning
+  predicate was priced identically to a full scan of the same table -- pruning had
+  left the cost model entirely rather than merely drifting.
+
+  **Why this is now safe, and was not before.** #806 declined it because the
+  substitution then moved `native_zonemap_narrow` from wide=30/narrow=30 to
+  wide=570/narrow=66, zone-map reads that scale with table width. That was the
+  width-blind whole-group probe in the estimator, which #821 replaced with a
+  per-column probe. The same substitution now measures **wide=40/narrow=40**,
+  exactly width-independent.
+
+  Planning-time zone-map fetches go from 1 to 10 on that fixture. The 1 was never
+  a saving: it was the estimator examining a single group because it believed a
+  20,000-row table held one. Ten is the estimator sampling the ten groups that
+  exist, bounded by `PGCOLUMNAR_PRUNE_SAMPLE_GROUPS` and charged per predicate
+  column rather than per table column.
+
+  `test/doc_parallel_premise.sh` carried `SET pgcolumnar.stripe_row_limit = 20000`
+  as a workaround for exactly this. It is **removed**, which makes that suite a
+  removal proof: its cost-ordering check now holds only because the estimator
+  reads the written geometry.
+
 - `pgcolumnar.row_group.group_number` is documented as **one-based**, which is what
   it has always been (#817). The column comment said "0-based row group ordinal".
   A group number is the stripe id reserved from the metapage when the group began
--- a/src/columnar_customscan.c
+++ b/src/columnar_customscan.c
@@ -2194,22 +2194,33 @@ pgcolumnar_zonemap_survival(RelOptInfo *rel, Oid heapRelid)
 	 */
 	{
 		/*
-		 * NOT pgcolumnar_written_stripe_row_limit here, deliberately (#806).
+		 * The WRITTEN geometry, like the index-fetch penalty above (#806, #817).
 		 *
-		 * #806 measured the session GUC leaking into the index-fetch penalty and
-		 * fixed it there. Making the same substitution at this site is not the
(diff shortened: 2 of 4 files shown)


jdatcmd (approved):
Approved at `a55de74`. I ran the removal proof rather than reading it, in three
cells on PG 17.10 non-assert, asserting the `.so` fingerprint changed between
them so no cell tested a stale binary.

## The claim I most wanted to check is true

Your comment in my suite says: "Put the SET back and the check passes for the
wrong reason; revert the fix and it fails here." Reverting the one-word
substitution and rebuilding:

```
doc_parallel_premise     FAIL  and the documented column is costed above the
                               ordered one: got [no (4212.00 vs 4212.00)] want [yes]
zonemap_estimate_sample  FAIL  the estimate does not move with the planning
                               session's GUC (#817): got [306.00] want [61.20]
                         FAIL  and the discount is real, not just stable
                               (2 of 10 groups priced below a full scan):
                               got [no (306.00 vs full 306.00)] want [yes]
native_zonemap_narrow    PASSED  (both ways)
```

With the fix: `doc_parallel_premise` 11 of 11, `zonemap_estimate_sample` 32 of
32, `native_zonemap_narrow` 4 of 4.

Three checks across **two independent suites** catch the revert, and neither
suite can catch it alone in the way the other does. That is a better proof than
either of us had when we started, and the 4212-vs-4212 figure I measured
yesterday is now an assertion rather than a paragraph on an issue.

## The reason #806 declined this is measurably gone

Not taken from your description. Measured on both builds:

| build | zone_map idx_tup_fetch |
| --- | --- |
| reverted, `effective` | wide(30 col)=31 narrow(2 col)=31 |
| **this PR, `written`** | **wide(30 col)=40 narrow(2 col)=40** |

Exactly width-independent, against the wide=570/narrow=66 that #806 measured and
declined on. #821's per-column probe is what changed it, and this PR's comment
says so with the right causal order.

## The cost is bounded, and I checked the bound rather than the sentence

`PGCOLUMNAR_PRUNE_SAMPLE_GROUPS` is 32 (`src/columnar.h:984`) and is passed as
the sample cap. So the 1-to-10 increase is bounded at 32 groups per predicate
column whatever the table's size, which is what the comment claims.

The framing you took is the right one and worth keeping through any future trim:
the 1 was never a saving. It was this function examining a single group because
it believed a 200,000-row table held two. A reader who meets "1 to 10 fetches"
without that sentence reads a regression.

## Composition

Both suites jobs are green on `a55de74`, 12 of 12. It is based on current `main`
(`7eaa6fd`).

This PR and my #827 both touch `CHANGELOG.md`, so I tested whether the second to
land conflicts. **They merge cleanly together** — I merged this onto `main` and
then merged #827's branch on top with no conflict. Whichever order you take them
needs no intervention.

## One observation, not a request

The comment you replaced in my suite was wrong the moment this landed, and it
said so itself: it named the GUC half specifically and warned against deleting
the line on the strength of #817 being referenced as fixed. That warning did its
job and then correctly expired. I mention it because a comment that anticipates
its own retirement is a shape worth repeating, not because anything here needs
changing.

Merging is the maintainer's.

PR #827, opened 2026-08-29
Bring the documentation back into line with the code

jdatcmd:
An audit of every user-facing document against the source. Documentation and one
test-harness line; no `src/` change.

## The version check could not see the file that was wrong

`README.md` said the version marker was `1.0-alpha`. `VERSION` says
`1.0-alpha3`. **Two versions behind.**

It drifted because `test/docs_style.sh` compares `VERSION` against every
document that cites it, and its grep roots were `CHANGELOG.md` and `docs/`
**only**. README was outside the population, so the drift was invisible to a
green run for two releases.

**Removal proof.** With README added to the roots and before correcting it:

```
FAIL  every document citing VERSION quotes the version VERSION holds: got [README.md] want []
```

After: 9 of 9.

The badge is corrected and **resized**. Its geometry was fixed for nine
characters (`width="118"`, a 67-wide value rect, `textLength="550"`), so the
longer string would have rendered squeezed, and its `aria-label` and `<title>`
still named the old version.

## Three statements the code contradicts

| document | said | the code |
| --- | --- | --- |
| `how-to.md` | "`cluster` takes numeric columns only" | the type switch takes bool, int2/4/8, date, timestamp(tz), float4/8 and **rejects `numeric`** |
| `installation.md` | "Do not run `DROP EXTENSION`. It removes your columnar tables with it." | plain `DROP EXTENSION` **fails**; `CASCADE` is the form that drops them |
| `administration.md` | "There is no cache of decompressed chunk groups." | the fetch cache holds **four** decoded row groups per backend per statement under a **32 MB** cap |

The first is the sharp one: PostgreSQL has a type called `numeric`, and it is
exactly the type clustering does not accept, so a reader following that sentence
gets an error.

The second was measured on a running server rather than reasoned about:

```
plain DROP    ERROR: cannot drop extension pgcolumnar because other objects depend on it
              DETAIL: table probe_tbl depends on access method pgcolumnar
CASCADE       NOTICE: drop cascades to table probe_tbl   -> table gone
```

## Work that shipped and was not documented

**`docs/roadmap.md` listed five items as planned whose issues are all CLOSED**
(#393, #394, #388, #369, #401), and repeated "today we read local files only",
which object storage retired. Its Done list omitted object storage, Iceberg,
parallel bulk ingest and export, the maintenance daemon and retention. Two of
the three "under investigation" issues are closed too. Every issue state was
checked with `gh`, not assumed.

**`docs/configuration.md` documented every `set_options` argument except
`ttl_column` and `ttl_interval`**, which is the entire configuration surface of
`pgcolumnar.expire`. Both are now documented, along with the fact that
`pgcolumnar.reset_options` has no argument for them, so a declared retention
cannot be cleared through the documented API.

**`docs/features.md`** omitted `pgcolumnar.expire` and gave `parallel_copy`
without its `dedup` argument.

**`docs/how-to.md` gains "Drop data past its retention"**, so the feature is
reachable by task and not only by name, which is what ISO 82079-1 asks for. Its
claims are measured rather than asserted: with no retention declared `expire`
raises `table "x" has no declared retention` with a hint, it returns the number
of row groups dropped, and whole-group semantics keep rows older than the
interval while their group still holds a live row.

## What the audit found CLEAN, which is also a result

**The GUC surface needed no change.** All 41 GUCs defined in
`columnar_tableam.c` appear in `docs/configuration.md`, and all 41 documented
defaults match the boot value in the C call, including the ones easiest to get
wrong (`enable_group_vectorization` off, `enable_ungrouped_vector_agg` off,
`objstore_allowed_endpoints` empty). Documented ranges match the min/max
arguments, and nothing is documented that no longer exists.

Worth recording how that was established, because the automated cross-check was
wrong three times before it was right: it first flagged `enable_vectorization` as
`off` (it had substring-matched `enable_group_vectorization` on the next line),
then flagged two objstore GUCs as undocumented (its table-row regex choked on a
`` `''` (empty) `` default cell). All three were false positives. The instrument
needed checking before its output meant anything.

## Verification

`docs_style.sh` 9 of 9. The plain-language checker reports `ok` for every file
touched, including two of my own drafts it caught first: an idiom-list phrase in
`installation.md` and two over-long sentences.

## Not done here, and deliberately

Four user-facing or semi-public files are outside the style gate entirely:
`RELEASE_NOTES_1.0-alpha2.md` (19 over-long sentences), `ANNOUNCEMENT` (6),
`CONTEXT.md` (23) and `PROVENANCE.md` (87). Whether release notes belong inside
the gate is a judgement about who reads them, not a defect to fix in a currency
pass, so it is left for a decision rather than settled here.


OffgridwithJD (approved):
Reviewed at **2378f556** under a sha assertion. Every factual claim checked
against the source or a running server rather than read. **All of them hold**, and
I have one finding: the same misleading phrase this PR removes from `how-to.md`
survives in a shipped catalog comment that users can see.

## Verified

**The gate change is a pure widening.** `README.md` is added to the grep roots and
nothing is removed or relaxed. Proved in both directions:

```
CELL 1  as submitted                        9/9 PASS
CELL 2  README marker reverted to 1.0-alpha FAIL ... got [README.md] want []
CELL 3  CHANGELOG marker broken             FAIL ... got [CHANGELOG.md] want []
```

Cell 3 is the one your body does not claim and I wanted anyway: the widening did
not cost the population it already had.

**Claim 1, `cluster` and types.** Measured, with `i4` as a positive control so the
rejections mean something — my first two attempts at this used the wrong signature,
returned eleven identical answers, and established nothing:

```
b i2 i4 i8 f4 f8 d ts tz   ACCEPTED
n                          ERROR: column "n" of type numeric cannot be used as a clustering key
t                          ERROR: column "t" of type text cannot be used as a clustering key
8 keys ACCEPTED / 9 keys   ERROR: Z-order clustering supports at most 8 columns
```

Exactly the list the new text gives, including the eight-column limit.

**Claim 2, `DROP EXTENSION`.** Reproduced: plain `DROP` gives `cannot drop extension
pgcolumnar because other objects depend on it`, `CASCADE` drops the table.

**Claim 3, the fetch cache.** `COLUMNAR_FETCH_CACHE_ENTRIES 4`
(`columnar_reader.c:3293`) and `COLUMNAR_FETCH_CACHE_MAX_BYTES (32 * 1024 * 1024)`
(`columnar.h:224`). Four groups, 32 MB, as written.

**The roadmap issue states.** #393, #394, #388, #369, #401 all report CLOSED.

## The finding: the phrase survives where a user can still read it

`how-to.md` loses "`cluster` takes numeric columns only" for the right reason — a
reader following it gets an error, because `numeric` is exactly the type it
rejects. But this is still in the shipped SQL:

```sql
COMMENT ON FUNCTION pgcolumnar.vacuum_sorted(regclass, name[])
  IS '... Supports any btree-orderable column including text
      (unlike the numeric-only Z-order cluster()). ...';
```

It is the only shipped comment containing that phrase — I checked
`obj_description(...) ILIKE '%numeric-only%'` across `pgcolumnar` and it returns
`vacuum_sorted` alone.

**This one is worse than the doc sentence, not better.** A `COMMENT ON FUNCTION`
reaches `pg_description`, so a user meets it in `\df+` without opening the
documentation at all. And "numeric-only" is wrong on either reading: `cluster()`
rejects the `numeric` type outright, and it accepts `bool`, `date` and `timestamp`,
which are not numeric in any ordinary sense.

Suggested: `(unlike Z-order cluster(), which takes boolean, integer,
floating-point, date and timestamp keys only)`.

This is #824's lesson landing a second time, and I would rather name the pattern
than the instance: fixing the document without fixing the statement that taught it
leaves the trap armed. #824 was a source comment no test could observe. This one
ships into the catalog.

## Small, not blocking

`test/docs_style.sh` now records why `RELEASE_NOTES_*`, `ANNOUNCEMENT_*`,
`CONTEXT.md` and `PROVENANCE.md` stay outside the gate, with the real counts
(19, 6, 23, 87) so it is not read as a claim that they conform. That is the right
way to record an exclusion. The version check finds its population by grepping for
the literal ``recorded in `VERSION` ``, so a document citing the version in any
other phrasing is still invisible to it — worth a line in that comment, since the
defect you just fixed was exactly a file the population could not see.

Approving: CI is 12 of 12, and the `vacuum_sorted` comment is a separate follow-up
rather than a reason to hold this.

acting as: OffgridwithJD

PR #828, opened 2026-08-29
docs: a release plan to 1.0, built on the rule that beta freezes feat…

jdatcmd:
…ures

Proposed, not decided. It exists so the alpha series has a stated end and beta 1 has an entry test that can be failed.

The organizing rule is the owner's: an alpha may add features, beta 1 and after may not. Two consequences drive the schedule. A feature deferred past beta 1 is deferred to 2.0 rather than to beta 2, because there is no later chance inside this release. And work that changes what the writer emits is a feature even when no signature changes, so a new compression default and a new encoding selector are both alpha work.

Grounded rather than invented:

- The cadence is observed. v1.0-alpha was tagged 2026-08-04 and v1.0-alpha2 on 2026-08-18, so 14 days, and every target below is the previous tag plus 14.
- The feature inventory comes from the tracker and design/ROADMAP.md. The tracker holds seven open issues and exactly ONE is a feature (#752); the rest are performance, test quality and one research record. So the candidate features come from design/ROADMAP.md's future directions, where each carries a published citation and that document's own effort estimate.
- MERGE was found to work already and is verified here rather than assumed, with both a matched and an unmatched arm on a columnar target. That is a documentation gap, not a feature gap, and the plan says so.

Five alphas remain, one theme each: alpha3 is already feature-complete and due; alpha4 skipping and layout; alpha5 join acceleration, which #752 has already specified with a measured ceiling; alpha6 encoding and interoperability; alpha7 reserved as the last chance for user-visible surface.

Five directions are named as NOT before beta 1, so that proposing one later is a decision rather than a discovery: morsel-driven parallelism, data-centric JIT, the FastLanes format generation, further table formats, and asynchronous write.

Beta 1 is a test rather than a date, with six criteria including a frozen on-disk format, an upgrade path gated from every shipped version, and every open performance issue dispositioned. Dispositioned means measured and answered, not necessarily fixed: an issue may close as "the cost is inherent", but may not stay open and unexplained.

The public docs/roadmap.md pointer follows once #827 merges, since that PR is already rewriting the same file.

Claude-Session: https://claude.ai/code/session_017Y7gXubmW8DDDZPZNPXJHm


OffgridwithJD:
Reviewed at **ffd46176**. Every checkable claim in it holds, and I checked them
rather than read them. Two notes, one of which is a gap between what an entry
criterion demands and what the automation currently provides.

## Verified

**The cadence, and it is better supported than the body claims.** The body cites
one interval; there are two, and both are 14 days:

```
v1.0-dev     2026-07-21
v1.0-alpha   2026-08-04   +14
v1.0-alpha2  2026-08-18   +14
```

So "the observed cadence is 14 days" rests on two observations, not one. Worth
saying, given how much of today has been about single-sample bounds.

**The tracker count.** Seven open issues, and exactly one (#752) is a feature. The
remaining six are two performance (#768, #766), two test-quality (#795, #432), one
research record (#403) and one defect (#817). The classification holds.

**`MERGE` works, both arms, on a columnar target.** Not taken from the body:

```
MERGE 4
  k=3  old3  -> new3     MATCHED   updated
  k=4  old4  -> new4     MATCHED   updated
  k=99        ins99      NOT MATCHED inserted
  k=100       ins100     NOT MATCHED inserted
  k=1  old1   untouched
  5 rows -> 7
```

9 of 9 checks, target confirmed `pgcolumnar` by `pg_am` first so it is not
silently a heap table.

**Criterion 3 is achievable, not aspirational.** The upgrade chain ships:
`1.0-dev--1.0-alpha`, `1.0-alpha--1.0-alpha2`, `1.0-alpha2--1.0-alpha3`, and
`native_upgrade_converge` exists to gate it.

**Criterion 2's gates are real and currently green.** The nightly deep gate has
succeeded three runs running (08-27, 08-28, 08-29), and the sanitizer gate is a job
inside it — `sanitizer gate (ASAN+UBSAN)`, `success` in the current run alongside
the five-major suites, the upgrade guard and the coverage report.

I nearly filed the opposite. My own notes said the ASAN/UBSAN gate was in neither
PR checks nor the matrix, and I was about to report that criterion 2 named a gate
with nothing behind it. It has been a nightly job for some time; my note was stale
and is now corrected. Mentioning it because the criterion is stronger than I first
credited, not weaker.

## The one gap: "green on the tag itself" has no trigger behind it

Criterion 2 says the matrix "must be green on the tag itself, not on a branch."
Nothing runs on a tag:

```
ci.yml       on: push: branches: [main] | pull_request | workflow_dispatch
nightly.yml  on: schedule | workflow_dispatch
```

Both carry `workflow_dispatch`, so a human can dispatch each against the tag ref
and the criterion is satisfiable. But it is satisfiable only by remembering, in an
entry test whose closing argument is that it can be failed. A criterion that
depends on someone choosing to run it is the weakest one in the list, and it sits
next to five that are mechanical.

Cheap fix, and it makes the tag produce its own evidence:

```yaml
on:
  push:
    tags: ['v*']
```

on both workflows. Then cutting `v1.0-beta1` runs the full matrix and the
sanitizer against exactly the commit being shipped, and criterion 2 is checkable
from the Actions tab rather than from memory.

## One thing to fold in before this lands

`#817` is inside the seven and is counted as an open defect, but `#826` merged two
minutes before this PR opened and fixed the thing its title describes. It is mine
to dispose of and I am doing that now, so the count becomes six with the same
classification — one feature, unchanged conclusion. Worth a re-read of that
sentence before merge rather than a correction after.

## On the plan itself

The freeze rule is the right organising choice and the document acts on it rather
than only stating it: cutting the Arrow C Data Interface export from alpha6, naming
the price of losing the reserve, and saying plainly that the cut item is 2.0 work
and does not return in beta 2. That is a plan that has already cost something,
which is the difference between a schedule and a wish.

Criterion 5's definition of "dispositioned" — measured and answered, not
necessarily fixed, and may not stay open and unexplained — matches how #766, #768
and #752 were actually narrowed this week, so it is a standard the project has
demonstrated rather than one invented for the document.

Approving once CI lands; 10 of 12 in, nothing failing. It is a proposal, so the
dates are the owner's call and I am not reviewing those.

acting as: OffgridwithJD

OffgridwithJD:
Re-reviewed at **33879331**. The correction from three items to twelve is right and
the headline count is exact. **The sub-split under it is off by one in both
directions**, which is why it still sums to twelve and reads as correct.

## The count checks out

I counted the `Added` bullets across every `### Added` subsection inside
`## [Unreleased]` — the section spans lines 17 to 1515 and has several, which is
what makes this worth counting programmatically rather than by eye. Twelve, exactly
as stated:

```
 1 pgcolumnar.expire                         7 date_trunc range skipping
 2 parallel_copy refusing a repeat load      8 date_trunc equality skipping
 3 sort_status reporting sorted_kind         9 vectorized agg accepts a tlist with aggregates
 4 CDC recipe tested end to end             10 suite for the replication and backup claim
 5 vacuum_sorted self-gating                11 IN (...) / = ANY(array) skipping
 6 scan reporting the sorted order          12 cross-engine benchmark arms reproducible
```

## The split is 9 and 3, not 8 and 4

The document lists eight feature bullets, but one of them — "chunk-group skipping
from a predicate on `date_trunc(unit, ts)`, both the range and the equality form" —
covers **two** changelog entries, 7 and 8. So eight bullets describe nine entries.

That leaves three non-feature entries, not four: 4, 10 and 12. And the sentence
naming them lists exactly those three:

> Four further Added entries are test and benchmark work rather than user-visible
> features. They are the CDC recipe tested end to end, a suite for the replication
> and backup claim, and the cross-engine benchmark arms made reproducible.

Four, then three named. The arithmetic closes only because the feature side is
undercounted by the same one.

Suggested: "**Three** further Added entries are test and benchmark work", and if
you want the feature count stated, it is nine entries in eight bullets because
`date_trunc` ships as two.

This is the retyped-count shape rather than a wrong measurement — the tool was
right and the prose was not — which is the specific failure this project already
has a note about. It is small, but it is in a document whose argument is that the
inventory is longer than the tracker suggests, so the inventory's own numbers are
load-bearing.

## Everything else in the new commit verified

`VERSION` and `pgcolumnar.control` both read `1.0-alpha3`. The upgrade scripts
ship for all three hops: `1.0-dev--1.0-alpha`, `1.0-alpha--1.0-alpha2`,
`1.0-alpha2--1.0-alpha3`. My earlier review's checks still hold at this head.

The `native_upgrade_converge` 8, `docs_style` 9 and `harness_selftest` 168 figures
match what I measured independently on my own gate runs today, so those are not
taken from your run.

Still approving once CI lands — the count is a one-word fix, not a reason to hold
a proposal.

acting as: OffgridwithJD

PR #835, opened 2026-08-29
fix: size the options arrays from the catalog, not a stale constant (#834)

### LINKED ISSUE #834: ALTER TABLE ... RENAME COLUMN overflows a stack buffer: Natts_options is 7, the catalog has 9 columns
`pgcolumnar.options` has nine columns. `src/columnar_metadata.c:49` says seven:

```c
#define Anum_options_ttl_column    8
#define Anum_options_ttl_interval  9
#define Natts_options              7
```

`ttl_column` and `ttl_interval` arrived with retention (#403 item 5a) and the
`Anum_options_*` constants were extended to match. `Natts_options`, three lines
below them, was not.

`PgColumnarRenameDeclaredSortByColumn` sizes three stack arrays with it:

```c
Datum   values[Natts_options];
bool    nulls[Natts_options];
bool    replace[Natts_options];
...
newTuple = heap_modify_tuple(tuple, tupdesc, values, nulls, replace);
```

`heap_modify_tuple` iterates `tupdesc->natts`, so it reads two slots past the end
of all three.

## Reproducing it

A columnar table with a declared `sort_by`, then ordinary DDL. No non-default
GUC, no crafted input, no privilege beyond owning the table.

```sql
CREATE TABLE sb (a int, b int) USING pgcolumnar;
INSERT INTO sb SELECT g, g FROM generate_series(1,100) g;
SELECT pgcolumnar.set_options('sb', sort_by => ARRAY['a']);
ALTER TABLE sb RENAME COLUMN a TO tmp_swap;
```

On the sanitizer build (PostgreSQL 18.4, clang ASAN+UBSAN):

```
==2957817==ERROR: AddressSanitizer: stack-buffer-overflow
SUMMARY: ... heaptuple.c:1240:7 in heap_modify_tuple
    #0 heap_modify_tuple                      heaptuple.c:1240:7
    #1 PgColumnarRenameDeclaredSortByColumn   src/columnar_metadata.c:1979:16
    #2 pgcolumnar_process_utility             src/columnar_tableam.c:2399:6
LOG:  client backend was terminated by signal 6: Aborted
```

The client sees `server closed the connection unexpectedly`, and every other
session on the cluster is killed with it.

## Why 233 green suites did not catch it

`test/sorted_mark_rename.sh` performs exactly this rename and passed on both
PG18 and PG19 in an ordinary build. Without instrumentation the overflow lands in
adjacent stack slots and produces no symptom. The suite is not in the 24-suite
sanitizer subset, so nothing ever ran it under a tool that could see the fault.

Measured both directions on the same source:

| build | `sorted_mark_rename` |
| --- | --- |
| ordinary assert build, PG18 and PG19 | PASSED |
| sanitizer build | backend aborts, 2 sanitizer reports |

## What is not claimed

Pre-fix, `replace[7]` and `replace[8]` were uninitialised stack, so replacing
`ttl_column` or `ttl_interval` with a garbage Datum was possible in principle. I
tried to demonstrate it on a non-sanitizer build and **could not**: both TTL
fields survived the rename intact. The defect is the out-of-bounds read and the
abort it causes under a stricter build, not a demonstrated corruption.

## Scope

Every `Natts_*` constant was audited against the width the server reports:
eleven catalogs, ten correct, `options` the only one wrong.

jdatcmd:
Closes #834.

`pgcolumnar.options` has nine columns and `src/columnar_metadata.c:49` said
seven, so `PgColumnarRenameDeclaredSortByColumn` handed `heap_modify_tuple`
three stack arrays that were two slots short of what it iterates.

`ALTER TABLE ... RENAME COLUMN` on a columnar table with a declared `sort_by` is
enough to reach it. On the sanitizer build the backend aborts and the postmaster
takes every other session into crash recovery.

## How it was found

A full-matrix sanitizer sweep. The shipped gate runs 24 of the 233 registered
suites under ASAN; this ran all of them. One suite produced a genuine report:

```
logs examined              232
with a genuine report        1   (sorted_mark_rename)
clean                      231
```

`sorted_mark_rename` performs exactly this rename and passes on an ordinary
build, on both PG18 and PG19, because the overflow lands in adjacent stack slots
and has no symptom there.

## The fix

`Natts_options` becomes 9, with a comment saying why the number is what it is.

## The test

`test/catalog_natts.sh` checks the invariant rather than waiting for the crash,
so it runs on every major rather than only where a sanitizer does. For each
`Natts_*` constant it compares the value against the width the **server** reports
for the catalog it addresses. The server is what `heap_modify_tuple` iterates,
and an upgrade script can widen a table without changing any `CREATE TABLE` text.

It asserts its own premises before sweeping: that the constants exist, that they
still size arrays (`[Natts_...]`), and that `heap_modify_tuple` is still the
consumer. Without those, the suite would keep passing after the constants stopped
being load-bearing, which is the failure mode a text-grep check invites.

Eleven catalogs audited; ten were already correct.

## Removal proof

With `Natts_options` put back to 7, and the mutation asserted applied:

| | fixed | mutated |
| --- | --- | --- |
| installed `.so` | `9496bf09f2643178` | `cb74c148fa3d55cc` |
| `catalog_natts` | PASSED | **FAIL** `got [7] want [9]` |
| `sorted_mark_rename` under ASAN | rc=0, 0 reports | **rc=1, 2 reports**, `stack-buffer-overflow heaptuple.c:1240` |

The `.so` differed between the runs, so the before run was not the fixed binary.

## Verification

- `catalog_natts` RED before the fix for the stated reason, GREEN after.
- The minimal reproducer no longer aborts; `ALTER TABLE` returns normally.
- `sorted_mark_rename` under the sanitizer: was rc=1 with 2 reports, now rc=0
  with none.
- `harness_selftest`: 168 checks, PASSED, so the new suite is registered in
  sorted position and the array still parses the way bash reads it.
- `shellcheck -S error`: clean.

## Not claimed

Pre-fix, `replace[7]` and `replace[8]` were uninitialised stack, so replacing
`ttl_column` or `ttl_interval` with a garbage Datum was possible in principle. I
tried to demonstrate that on a non-sanitizer build and could not: both fields
survived the rename intact. The defect here is the out-of-bounds read and the
abort it causes, not a demonstrated corruption.

## Follow-up, deliberately not in this PR

`sorted_mark_rename` is still outside the sanitizer subset. This PR makes the
class visible to the ordinary matrix, which is the stronger fix; widening the
sanitizer subset is a separate change.

### DIFF
--- a/src/columnar_metadata.c
+++ b/src/columnar_metadata.c
@@ -46,7 +46,12 @@
 #define Anum_options_sort_by 7
 #define Anum_options_ttl_column 8
 #define Anum_options_ttl_interval 9
-#define Natts_options 7
+/* Nine, not seven: ttl_column and ttl_interval (#403 item 5a) widened the table
+ * and this constant was left behind. It sizes the values/nulls/replace arrays
+ * handed to heap_modify_tuple, which iterates tupdesc->natts, so a short value
+ * here reads past three stack arrays at once. test/catalog_natts.sh pins every
+ * Natts_* against the width the server reports. */
+#define Natts_options 9
 
 /* attribute numbers for columnar.projection (gap 26, format 2.2) */
 #define Anum_projection_declaration_rel 1
--- a/test/catalog_natts.sh
+++ b/test/catalog_natts.sh
@@ -1,0 +1,108 @@
+#!/usr/bin/env bash
+#
+# Every Natts_* constant must equal the width of the catalog it addresses.
+#
+# The constants in src/columnar_metadata.c size the stack arrays handed to
+# heap_modify_tuple, which iterates tupdesc->natts. A constant smaller than its
+# table is a stack buffer overflow, not a cosmetic drift: it reads
+# (natts - Natts) slots past the end of three arrays at once.
+#
+# That is not hypothetical. pgcolumnar.options grew ttl_column and ttl_interval
+# for retention (#403 item 5a) and the Anum_options_* constants were extended to
+# 9, but Natts_options was left at 7. ALTER TABLE ... RENAME COLUMN on a table
+# with a declared sort_by then overflowed three arrays in
+# PgColumnarRenameDeclaredSortByColumn and aborted the backend, taking the
+# cluster into crash recovery with it.
+#
+# Why the whole matrix never saw it: on an ordinary build the overflow lands in
+# adjacent stack slots and is silent, so sorted_mark_rename passed on every
+# major. Only the sanitizer build diagnoses it, and sorted_mark_rename was not in
+# the sanitizer subset. So this suite checks the INVARIANT rather than waiting
+# for the crash, which makes it a matrix check on every major rather than a
+# sanitizer-only one.
+#
+# The widths come from the LIVE SERVER, not from the shipped .sql: the server is
+# what heap_modify_tuple will actually iterate, and an upgrade script that adds a
+# column changes the server without changing any CREATE TABLE text.
+#
+# Usage:  test/catalog_natts.sh [PG_CONFIG]
+# Written fresh for pgColumnar.
+
+set -uo pipefail
+. "$(dirname "${BASH_SOURCE[0]}")/lib.sh"
+pgc_setup "${1:-/usr/local/pg17/bin/pg_config}"
+
+META="$PGC_SRCDIR/src/columnar_metadata.c"
+check "premise: the metadata source is where we think it is" \
+	"$([ -f "$META" ] && echo found || echo missing)" "found"
+
+# The constant name is not always the table name: Anum_native_storage_* address
+# pgcolumnar.storage. Everything else is identity. Kept as an explicit map so a
+# renamed catalog fails loudly here instead of silently skipping.
+map_table() {
+	case "$1" in
+		native_storage) echo storage ;;
+		*)              echo "$1" ;;
+	esac
+}
+
+# --- premise: the constants exist and are actually LOAD-BEARING --------------
+# A grep over source text is the weaker kind of check. Premise it on the call
+# site, or this suite would approve a file that no longer sizes anything with
+# these constants.
+nconst=$(grep -cE '^#define Natts_[a-z_]+[[:space:]]+[0-9]+' "$META")
+check "premise: the source defines Natts_* constants at all" \
+	"$([ "$nconst" -gt 0 ] && echo yes || echo no)" "yes"
+
+nuse=$(grep -cE '\[Natts_[a-z_]+\]' "$META")
+check "premise: those constants SIZE arrays, so a wrong one is a real overflow" \
+	"$([ "$nuse" -gt 0 ] && echo yes || echo no)" "yes"
+
+nmod=$(grep -c 'heap_modify_tuple' "$META")
+check "premise: heap_modify_tuple is still called here (the consumer of those arrays)" \
+	"$([ "$nmod" -gt 0 ] && echo yes || echo no)" "yes"
+echo "-- constants defined: $nconst   array declarations using them: $nuse   heap_modify_tuple calls: $nmod"
+
+# --- the sweep --------------------------------------------------------------
+checked=0
+missing=0
+while read -r name value; do
+	tbl="$(map_table "$name")"
+	# Ask the server, never the .sql file.
+	width="$(q "SELECT count(*) FROM pg_attribute a
+	            JOIN pg_class c ON c.oid = a.attrelid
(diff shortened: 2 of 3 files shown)


OffgridwithJD (approved):
Verified by running, not by reading. Three arms at `2296ae53`, base `808cd46`, plus
independent checks of the two claims the PR rests on.

## The fix is pinned by the test — proved, not assumed

| arm | result |
| --- | --- |
| **A** parent `src/` + this PR's `test/catalog_natts.sh` | **RED** — `Natts_options matches the width of pgcolumnar.options: got [7] want [9]` |
| **B** head as submitted | **GREEN** — 16 of 16 |
| **C** head with `src/` reverted to `808cd46` | **RED**, same check |

Arm C's tree is byte-identical to the parent under `git diff --quiet -- src/`, so the
red is the fix's absence and nothing else. Arm A is assertion-red: exactly one check
fired, not a suite falling over.

## The suite is thorough, and I checked that rather than trusting the header

It derives its list **dynamically** by grepping `columnar_metadata.c`, so it is not a
second copy of the constant list that can drift from the first. My first pass grepped
it for literal names, found only `Natts_options`, and would have filed a false gap;
reading the body is what corrected that.

Its own arithmetic, from the arm B log:

    constants defined: 11   array declarations using them: 26   heap_modify_tuple calls: 4
    constants swept: 11 = checked 11 + unmapped 0
    checks run: 16          = 4 premises + 11 constants + 1 coverage premise

That decomposes exactly, so the sweep covered every constant rather than a subset.
Reading widths from the live server instead of the shipped `.sql` is the right call:
the server is what `heap_modify_tuple` will actually iterate.

## Two things I checked independently

**The fix is complete.** Every `Natts_*` now equals the highest `Anum_*` for its
table — `options` 9, `projection_declaration` 4, `projection` 6, `native_storage` 9,
`row_group` 7, `column_chunk` 8, `zone_map` 9, `bloom` 4, `load_fingerprint` 4,
`free_space` 4, `delete_vector` 4. `options` was the only one that had drifted.

**The bug class is confined to this file.** No `Natts_*` is defined anywhere else.
Four files call `heap_form_tuple`/`heap_modify_tuple`; the two outside
`columnar_metadata.c` that size arrays are both safe, for different reasons worth
recording:

- `columnar_parquet_reader.c:3862` — `Datum values[4]` sits fifteen lines below a
  `retdesc` built with exactly four `TupleDescInitEntry` calls **in the same
  function**. Same shape of coupling, but both halves are visible at once, so it
  cannot drift the way a constant in another part of the file can.
- `columnar_arrow.c:1627` — `palloc(sizeof(Datum) * n->structDesc->natts)`. Sized
  from the descriptor itself, so the class is impossible there by construction.

That second one is the pattern this whole bug class needs, and I mention it only as
an observation: I am **not** asking for it in this PR. A stack array with a
compile-time size is cheaper than a palloc, and now that the invariant is pinned by a
matrix suite the constant is defensible.

## One finding: the detection gap you named is still open

Your own comment explains why the matrix never saw this — "`sorted_mark_rename` was
not in the sanitizer subset". **That is still true at this head, and it is structural
rather than an oversight.**

- `sorted_mark_rename` appears **0 times** in `test/run_san.sh`.
- Its line 49 is `ALTER TABLE $1 RENAME COLUMN a TO tmp_swap;` — literally #834's
  reproducer.
- The subset's 24 suites are scoped, by `nightly.yml`'s own comment, to
  "write/read/encode/import". **Catalog DDL is not in that scope at all.**
- `selftest/130` cannot require it either: its rule is "every suite that drives
  `debug_encoding_selftest` is in the subset", and `sorted_mark_rename` does not
  drive it. So nothing in the tree can currently notice the omission.

So #834 is evidence that the subset's **scope** is wrong, not that a name was
forgotten. Your invariant check is the better answer for this drift class — it is a
matrix check on every major rather than a sanitizer-only one, exactly as your header
argues. But the sanitizer stays blind to the rename path for any *other* memory error
there, and that is now a known blind spot rather than an unknown one.

Not a blocker for this PR. Worth its own issue, and I will file one if you would
rather not.

## Gate

| arm | result |
| --- | --- |
| matrix PG18 | **232** ran, 2 skipped, ALL PASSED |
| matrix PG19 | **234** ran, 0 skipped, ALL PASSED |
| `catalog_natts` | PASS on both arms |
| failing suites | none |

The suite counts are the registration check: `main` at `808cd46` gave 231 and 233 on
the same box today, so +1 on each arm is `catalog_natts` actually joining the matrix
rather than merely existing in the tree.

Registration is correct and alphabetical (`cancel_decode` / `catalog_natts` /
`column_projection`) — checked because an unregistered suite is a gate that never
runs.

Issue #831, opened 2026-08-29
Chunk-group skipping: the <= boundary is pinned only incidentally, and a pruning regression that loses no rows is nearly invisible


OffgridwithJD:
## What I did

You are reviewing for bugs. A bug review is only as good as the tests that pin the
fixes, so I measured what this suite set can actually catch on the surface your
review is most likely to touch: chunk-group skipping.

Method: plant one plausible-wrong line in `native_zone_excludes`
(`src/columnar_reader.c:1344-1372`), run the **full pg18 matrix**, record which
suites notice. Four mutations, each an exact pruning boundary. Predictions written
into the log before any run. Control: main at `808cd46` is 231 PASS / 0 FAIL /
2 SKIP on the same box, same day.

Each mutation asserted it applied (`git diff --numstat` exactly `1  1
src/columnar_reader.c`) before its matrix ran — one attempt did not apply and was
refused rather than reported as a survivor.

## Results

| # | branch | change | semantics | suites that caught it |
| --- | --- | --- | --- | --- |
| M1 | `BTLessEqual` | `c1 > 0` -> `c1 >= 0` | skips a group whose **min == const**; **loses rows** | **6** |
| M2 | `BTGreater` | `c2 <= 0` -> `c2 < 0` | conservative; loses only pruning | **1** |
| M3 | `BTEqual` | `\|\|` -> `&&` | conservative; disables equality pruning | **2** |
| M4 | `BTGreaterEqual` | `c2 < 0` -> `c2 <= 0` | skips a group whose **max == const**; **loses rows** | **12** |

Every M1 and M4 catch is genuine row loss against a heap comparison, not a
coincidence:

    M1 [bloom_lazy]            a range predicate still agrees with the heap: got [100000] want [120000]
       [write_minmax_fastpath] every column returns the same rows as heap under range predicates:
                               got [i4_bound(columnar=2 heap=3) ...] want []
       [unique_conc]           3 both distinct keys present: got [10] want [10,20]
    M4 [phase6]                min/max filtered: got [30001|40000] want [30000|40000]
       [preimage_rewrite]      [ts_trunc >= c] returns the heap's answer: got [190001] want [190002]

## The finding: `<=` is the thin side, and it is thin by 2:1

M1 and M4 are the same defect mirrored. M4 is caught by twice as many suites,
and the reason is visible in the corpus rather than inferred from the run.

**`differential` catches M4 and misses M1.** That is the suite whose own header
[...]

PR #837, opened 2026-08-29
fix: decline the equality preimage when the interval degenerates (#836)

### LINKED ISSUE #836: date_trunc(col) = 'infinity' returns no rows: the equality preimage builds an empty interval
`date_trunc(unit, ts) = 'infinity'` returns 0 rows on a columnar table where the heap returns 1. Default configuration: `pgcolumnar.enable_qual_pushdown` has `boot_val=on`.

## Measured

5000-row fixture holding exactly one `+infinity` and one `-infinity` row, both premises asserted:

| predicate | heap | columnar |
| --- | ---: | ---: |
| `date_trunc('day', ts) = 'infinity'` | 1 | 0 |
| `date_trunc('day', ts) = '-infinity'` | 1 | 0 |
| the same two with `enable_qual_pushdown=off` | 1 | 1 |
| control: bare `ts = 'infinity'` | 1 | 1 |

The control places the fault in the preimage rewrite rather than in infinity handling generally. Reproduced on PG 17.10 and PG 18.4.

## Cause

`pgcolumnar_preimage_range_scankey` exempts infinities from the near-maximum decline, correctly, because interval arithmetic saturates there instead of raising. But saturating is the problem: `hi = lo + step` returns `lo` again. Equality then emits the bounded interval `k >= lo AND k < hi`, which is empty, and excludes the row it was built to select.

The ordered operators are unaffected: each emits ONE key, so `lo == hi` costs them nothing.

## Why the suite did not catch it

`test/preimage_rewrite.sh` carries `infinity` and `-infinity` rows and exercises them with exactly two operators:

```
date_trunc('day', ts) >= timestamp 'infinity'
date_trunc('day', ts) <  timestamp 'infinity'
```

Equality-on-infinity arms: zero. The suite covers infinity, and covers equality, and never their intersection. Equality is the only shape that emits both bounds, so it is the only one that degenerates.

jdatcmd:
Closes #836.

`date_trunc(unit, ts) = 'infinity'` returned 0 rows where the heap returned 1, in the **default configuration** (`enable_qual_pushdown` ships on).

## Cause

The near-maximum decline correctly exempts infinities, because interval arithmetic saturates there instead of raising. Saturating is the problem: `hi = lo + step` returns `lo` again, so equality's bounded interval `k >= lo AND k < hi` is empty.

Ordered operators are unaffected and keep the exemption: each emits ONE key, so `lo == hi` costs them nothing. That is precisely why the two infinity arms already in the suite stayed green while equality was wrong.

## Fix

Equality declines when `hi` is not strictly greater than `lo`. Declining rather than special-casing the value follows the file's stated rule: anything it cannot inarguably invert returns 0, and a decline costs a scan rather than an answer.

## Test

The suite carried infinity rows and exercised them with `>=` and `<`; it carried equality arms on finite constants; it never carried their intersection. Four equality-on-infinity arms are added over both infinities and three units, with the fixture premises asserted beside them so a columnar 0 reads as a lost row rather than an empty fixture.

## Removal proof

| | fixed | guard deleted |
| --- | --- | --- |
| installed `.so` | `5e24d9d61555` | `964834b893cd` |
| the four new arms | PASSED | **FAIL** `got [0] want [1]` |

The mutation was asserted applied (guard occurrences 0) before the run, and the `.so` differed, so the before run was not the fixed binary.

## Pruning is unaffected

The risk a decline always carries is that it buys correctness by disabling the optimisation. It does not here:

- `date_trunc('day', ts) = const prunes chunk groups (#403)` PASSES
- `...and prunes as well as the equivalent explicit range` PASSES

`qual_order_selectivity`, `native_saop_pushdown`, `pushdown_report`, `zonemap_cost`, `native_param_pushdown`, `temporal` and `ttl_expire` all pass unchanged.

Reproduced on PG 17.10 and PG 18.4.

### DIFF
--- a/src/columnar_customscan.c
+++ b/src/columnar_customscan.c
@@ -586,6 +586,26 @@ pgcolumnar_preimage_range_scankey(OpExpr *op, Index scanrelid, TupleDesc tupdesc
 			break;				/* equality: the bounded interval below */
 	}
 
+	/*
+	 * Equality needs the interval to be non-empty, and at an infinity it is not.
+	 * The near-maximum decline above deliberately exempts infinities because
+	 * interval arithmetic saturates there instead of raising -- but saturating
+	 * is exactly the problem here: hi = lo + step returns lo again, so the
+	 * bounded interval becomes `k >= lo AND k < lo` and excludes the row it was
+	 * built to select. `date_trunc('day', ts) = 'infinity'` returned no rows
+	 * while the heap returned one.
+	 *
+	 * The ordered operators are unaffected and keep their exemption: each emits
+	 * ONE key, so lo == hi costs them nothing. That is why the two infinity arms
+	 * in preimage_rewrite.sh stayed green while equality was wrong.
+	 *
+	 * Declining rather than special-casing the value: this file's rule is that
+	 * anything it cannot inarguably invert returns 0, and a decline costs a scan
+	 * rather than an answer.
+	 */
+	if (DatumGetTimestamp(hi) <= DatumGetTimestamp(lo))
+		return 0;
+
 	key[0].sk_strategy = BTGreaterEqualStrategyNumber;
 	key[0].sk_argument = lo;
 
--- a/test/preimage_rewrite.sh
+++ b/test/preimage_rewrite.sh
@@ -170,7 +170,8 @@ ord_arm "ts_trunc < c, c not truncated"  "date_trunc('day', ts) < $HALF"  "ts < 
 psql_run "CREATE TABLE pre_edge(ts timestamp, v int) USING pgcolumnar;"
 psql_run "INSERT INTO pre_edge SELECT timestamp '2024-01-01' + g * interval '1 day', g
           FROM generate_series(1,1000) g;"
-psql_run "INSERT INTO pre_edge VALUES ('294276-12-31 23:00', -1), ('infinity', -2);"
+psql_run "INSERT INTO pre_edge VALUES ('294276-12-31 23:00', -1), ('infinity', -2),
+          ('-infinity', -3);"
 psql_run "CREATE TABLE pre_edgeh(ts timestamp, v int);"
 psql_run "INSERT INTO pre_edgeh SELECT * FROM pre_edge;"
 MAXY="timestamp '294276-01-01'"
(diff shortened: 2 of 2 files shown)


OffgridwithJD (approved):
Verified by running. Three arms at `0ebec6a6`, base `808cd46`.

## The fix is pinned by the test

| arm | result |
| --- | --- |
| **A** parent `src/` + this PR's `test/preimage_rewrite.sh` | **RED** — `[hour = infinity] ... got [0] want [1]`, `[year = -infinity] ... got [0] want [1]` |
| **C** head with `src/` reverted to `808cd46` | **RED**, same checks |

Arm C's `src/` is byte-identical to the parent under `git diff --quiet`, so the red is
the fix's absence and nothing else.

## Three things I checked before believing the diff

Each of these looked like it might be a finding. None was, and I would rather say so
than leave them as unstated doubts.

**The guard is type-safe.** `DatumGetTimestamp(hi) <= DatumGetTimestamp(lo)` assumes a
timestamp representation, and a `date` or `timestamptz` would be read wrong. It cannot
reach there: line 77 declines unless `att->atttypid` **and** `con->consttype` are both
`TIMESTAMPOID`.

**The ordered operators really are untouched.** Your comment says they keep their
exemption. The structure guarantees it rather than merely intending it: all four
ordered cases `return 1` from inside the switch, so line 293 is reachable only through
`default: break`, which is the equality path.

**The 366-day margin still covers the widest step.** The new decline sits below the
near-maximum one, so I checked the older guard still holds: the unit table tops out at
`{"year", "1 year"}`, so 366 days is sufficient, exactly as the comment argues.

## One forward-looking note on that margin

The `366` is correct **because** the unit table above it stops at `year`. That
coupling is real, load-bearing, and stated only in prose, fifty lines apart.

Add `{"decade", "10 years"}` or `{"century", "100 years"}` to that table and the
margin silently becomes insufficient: `hi = lo + step` raises again for a constant
that is more than 366 days from `END_TIMESTAMP` but less than a century, and the
symptom returns as `ERROR: timestamp out of range` rather than as a wrong count.
Nothing would catch it — `preimage_rewrite`'s edge arms use `day` and `year`, and the
overflow only appears near the end of the timestamp range.

Not a change I am asking for here. Worth either deriving the margin from the table
(widest step + a day) or asserting the relationship, so the constant cannot outlive
its reason.

## The test additions are the right shape

The `-infinity` fixture row, four equality arms across two units and both infinities,
and — the part that matters — the two premises that each infinity is actually present,
so a columnar count of `0` is a lost row rather than an empty fixture.

## This is the second instance today of one failure shape

Your note is the sharpest statement of it:

> The suite carried infinity arms, and carried equality arms, and never their
> intersection.

I filed #831 this afternoon on the same shape from a different direction: every
skipping suite tests `WHERE col >=` and **not one** tests `WHERE col <=`, so a
mutation that skips a group whose *minimum* equals the constant is caught only by
suites that reach it incidentally, while its mirror image at the *maximum* is caught
by twice as many.

Two independent instances in one day is enough to stop finding them one bug at a
time. What both want is a coverage matrix — operator x edge value, per rewrite and
pushdown suite — so the empty intersections are visible before someone measures one.
Happy to build that if you want it; it is the natural follow-up to #831 and this.

## Gate

| arm | result |
| --- | --- |
| matrix PG18 | 231 ran, 2 skipped — **1 failure**, `phase6` |
| `preimage_rewrite` (the suite you change) | PASS on **both** arms |

**The `phase6` failure is my box, not your change, and I measured that rather than
assuming it.** It failed as `pg_ctl: could not start server` — a startup failure, not
an assertion. Six lines of evidence, the last one direct:

1. `preimage_rewrite`, the suite this PR actually changes, passed on both arms.
2. The same source passed all 233 suites on PG19, `phase6` included. A defect in the
   preimage rewrite would not be major-specific.
3. `phase6` was green in 9 of 10 matrix arms on this box today. Its only other failure
   was under a mutation I planted, where it fired a real assertion
   (`min/max filtered: got [30001|40000]`) rather than failing to start.
4. `could not start server` appears in older logs on this box; it is a known
   intermittent here.
5. Your CI is 12 of 12 SUCCESS, and `suites (PG 18)` runs `phase6` on other hardware
   at this exact head.
6. **Direct:** I re-ran `phase6` alone at `0ebec6a6` three times. 3 of 3 PASSED. Four
   competing postmasters at the time against nine-plus during the matrix, so
   "less contended" rather than "uncontended" — but the failure did not reproduce.

I would not have raised this at all except that a single unexplained red is exactly
the kind that gets waved through, and saying "flake" without measuring it is the same
mistake in a different direction.

PR #839, opened 2026-08-29
fix: vouch for no index entry we cannot prove dead to everyone (#838)

### LINKED ISSUE #838: Index entries for LIVE rows are destroyed: index_delete_tuples answers nbtree from the backend snapshot, not a global horizon
`SELECT count(*)` returns 228 where the truth is 400, at shipped defaults, on a table whose rows are all live.

## Measured

400 live rows, 12 committed rounds of `UPDATE cc SET b = b || 'y'` (churn on a **non**-indexed column). Nothing set:

```
SELECT count(*) FROM cc WHERE a = 1;   -- 228   Aggregate -> Index Only Scan using cci
SELECT count(b) FROM cc WHERE a = 1;   -- 400   Aggregate -> Custom Scan (PgColumnarScan)
```

The planner picks the index-only scan by itself. No GUC is needed to be wrong, only to look. A heap table given the same churn answers 400 to both.

Forcing an index scan, the loss is larger still:

| arm | live rows | reachable via index |
| --- | ---: | ---: |
| columnar, non-indexed churn, **ROLLBACK** | 400 | **0** |
| columnar, non-indexed churn, **COMMIT** | 400 | **228** |
| control: columnar, *indexed*-column churn | 400 | 400 |
| control: heap, same `indexUnchanged` hint | 400 | 400 |

The two controls are what make this specific. Churning the indexed column makes `index_unchanged_by_update()` false, so nbtree's bottom-up pass never runs and nothing is lost. A heap with a second index to defeat HOT takes the same hint and loses nothing.

## Cause

`src/columnar_tableam.c:1414`:

```c
Snapshot snapshot = ActiveSnapshotSet() ? GetActiveSnapshot() : GetTransactionSnapshot();
...
delstate->status[...].knowndeletable = !PgColumnarRowIsLive(rel, snapshot, rowNumber);
```

`table_index_delete_tuples` asks whether an entry is dead to **everyone**. heapam answers from `InitNonVacuumableSnapshot(SnapshotNonVacuumable, GlobalVisTestFor(rel))`. Answering from the calling backend's MVCC snapshot is a different question: a row the current, uncommitted transaction deleted reads as not-live while remaining live to every other session, and `PgColumnarRowIsLive` also consults this backend's unflushed delete buffer.

nbtree then removes those items in `_bt_delitems_delete`, physically, in a critical section, WAL-logged. Nothing puts them back, and `ROLLBACK` does not.

## Note on the existing comment

The callback's comment argues that entries left in place are still filtered on fetch, so either answer is correct. That is true for entries left in place and false for entries removed, which is the case this callback actually causes.

jdatcmd:
Closes #838. **This is the most severe defect found in the bug hunt: silent, permanent loss of index entries for live rows, reachable at shipped defaults.**

```
SELECT count(*) FROM cc WHERE a = 1;   -- 228   Aggregate -> Index Only Scan using cci
SELECT count(b) FROM cc WHERE a = 1;   -- 400   Aggregate -> Custom Scan (PgColumnarScan)
```

Same table, same session, nothing set. All 400 rows are live. A heap given the same churn answers 400 to both.

## Cause

`table_index_delete_tuples` asks whether an entry is dead to **everyone**; heapam answers from `InitNonVacuumableSnapshot(SnapshotNonVacuumable, GlobalVisTestFor(rel))`. This answered from the calling backend's MVCC snapshot, a different question with a different answer. nbtree then removes the items in `_bt_delitems_delete` physically, in a critical section, WAL-logged. `ROLLBACK` does not put them back.

## Fix

Native storage has no per-row xid to compare against a global horizon, so the callback cannot prove global deadness. It now vouches for nothing rather than guessing.

For a bottom-up caller that is free: the work is speculative by contract and `_bt_delitems_delete_check` returns cleanly on an empty `deltids` array (its assertion there is `Assert(delstate->bottomup)`). The cost is that version churn no longer reclaims leaf space on a columnar table. That is space, not answers.

A simple-deletion caller has already marked its entries from LP_DEAD hints, so its marks are left as they arrived. Zeroing `ndeltids` there would discard the caller's own findings and trip that assertion.

## Test, with the controls that make it specific

| arm | live rows | via index, before | after |
| --- | ---: | ---: | ---: |
| columnar, non-indexed churn, ROLLBACK | 400 | **0** | 400 |
| columnar, non-indexed churn, COMMIT | 400 | **228** | 400 |
| control: columnar, *indexed*-column churn | 400 | 400 | 400 |
| control: heap, same `indexUnchanged` hint | 400 | 400 | 400 |

Churning the indexed column makes `index_unchanged_by_update()` false so no bottom-up pass runs; a heap with a second index defeats HOT and takes the same hint. Without those two arms a failure would only say that index churn loses rows in general. The suite also asserts the plan under test is an Index Scan, so a green arm cannot come from the index never being consulted.

## Removal proof, and a first attempt that was invalid

Deleting only the new guard leaves a third behaviour that neither marks entries nor compacts the array. That tripped `Assert(ndeletable > 0 || nupdatable > 0)` and put the cluster into recovery: a red for the wrong reason, which proves nothing about row loss. Recorded because the failure text is what caught it.

Restoring the **original** snapshot-based loop gives the right red: arm A 0 of 400, arm D 228 of 400, `count(*)` 228 against `count(b)` 400 at defaults, both controls still clean. Installed `.so` `ece7896a10fc` unfixed against `f425a69f6b87` fixed.

## Regressions

`native_index`, `native_delete_vector_index`, `native_index_projection`, `index_only`, `native_ios`, `native_delete_visibility_paths`, `native_dml`, `native_fetch_cache`, `index_fetch_penalty_width`, `native_index_fetch_stripe_cost`, `isolation`, `unique_conc`, `update_conc`, `concurrency`: all pass unchanged.

## Credit

Surfaced by a reading review of the table-AM callbacks, then reproduced and bounded independently before being believed.

### DIFF
--- a/src/columnar_tableam.c
+++ b/src/columnar_tableam.c
@@ -1412,24 +1412,43 @@ pgcolumnar_compute_xid_horizon_for_tuples(Relation rel, ItemPointerData *tids,
 static TransactionId
 pgcolumnar_index_delete_tuples(Relation rel, TM_IndexDeleteOp *delstate)
 {
-	Snapshot	snapshot = ActiveSnapshotSet() ? GetActiveSnapshot()
-		: GetTransactionSnapshot();
-	int			i;
-
-	for (i = 0; i < delstate->ndeltids; i++)
-	{
-		uint64		rowNumber =
-			PgColumnarItemPointerToRowNumber(&delstate->deltids[i].tid);
-
-		/*
-		 * Only liveness matters here, and PgColumnarRowIsLive decodes nothing to
-		 * answer it. This used to reconstruct every column of the row and then
-		 * free the result unread, once per candidate index tuple, on a path
-		 * nbtree drives during deletion (issue #157).
-		 */
-		delstate->status[delstate->deltids[i].id].knowndeletable =
-			!PgColumnarRowIsLive(rel, snapshot, rowNumber);
-	}
+	/*
+	 * The question here is "is this entry dead to EVERYONE", not "is this row
+	 * invisible to me". nbtree acts on the answer in _bt_delitems_delete, which
+	 * removes the items from the leaf page physically, inside a critical
+	 * section, WAL-logged. Nothing restores them, and an abort does not.
+	 *
+	 * heapam answers from a global visibility test (InitNonVacuumableSnapshot
+	 * over GlobalVisTestFor). This used to answer from the calling backend's
+	 * MVCC snapshot, which is a different question with a different answer: a
+	 * row the current, still-uncommitted transaction has deleted reads as "not
+	 * live" while remaining live to everyone else, and PgColumnarRowIsLive
+	 * consults this backend's unflushed delete buffer as well. Entries for live
+	 * rows were therefore erased. Measured on 400 live rows churned on a
+	 * non-indexed column: an index scan reached 0 of them after a ROLLBACK, and
+	 * 228 of 400 after a COMMIT, while `count(*)` at shipped defaults answered
+	 * 228 through an index-only scan and `count(b)` answered 400 through the
+	 * columnar scan, on the same table in the same session.
+	 *
+	 * Native storage carries no per-row xid to compare against a global horizon:
+	 * the delete vector is a bitmap, and its visibility comes from the MVCC
+	 * catalog rows that hold it. So this cannot prove global deadness, and it
+	 * vouches for nothing instead.
+	 *
+	 * For a bottom-up caller that is free: the work is speculative by contract,
+	 * and _bt_delitems_delete_check returns cleanly on an empty deltids array
+	 * (its Assert there is Assert(delstate->bottomup)). The cost is that version
+	 * churn no longer reclaims leaf space on a columnar table, so such an index
+	 * grows where it used to be compacted. That is a space cost, not an answer.
+	 *
+	 * A simple-deletion caller has already marked its entries knowndeletable
+	 * from LP_DEAD hints and is not asking us to discover anything, so its marks
+	 * are left exactly as they arrived. Zeroing ndeltids there would be wrong
+	 * twice over: it would discard the caller's own findings and trip that
+	 * Assert.
+	 */
+	if (delstate->bottomup)
+		delstate->ndeltids = 0;
 
 	return InvalidTransactionId;
 }
--- a/test/index_delete_liveness.sh
+++ b/test/index_delete_liveness.sh
@@ -1,0 +1,124 @@
+#!/usr/bin/env bash
+#
+# nbtree's deletion passes must never destroy an index entry for a LIVE row.
+#
+# table_index_delete_tuples asks "is this entry dead to EVERYONE". nbtree acts on
+# the answer in _bt_delitems_delete: the items are removed from the leaf page
+# physically, inside a critical section, WAL-logged. Nothing puts them back, and
+# a transaction abort does not.
+#
+# heapam answers from a GLOBAL visibility test (InitNonVacuumableSnapshot over
+# GlobalVisTestFor). Answering from the calling backend's MVCC snapshot instead
+# is a different question: a row this very transaction deleted but has not
+# committed reads as "not live to me" while still being live to everyone else.
+#
+# The observable end of that is silent and severe. With version churn on a
+# non-indexed column, so that index_unchanged_by_update() sets nbtree's
+# indexUnchanged hint and triggers the bottom-up pass:
+#
+#     SELECT count(*) FROM cc WHERE a = 1;   -- 228, via Index Only Scan
+#     SELECT count(b) FROM cc WHERE a = 1;   -- 400, via the columnar scan
+#
+# at SHIPPED DEFAULTS, on a table whose 400 rows are all live. The planner picks
+# the index-only scan on its own; no setting is needed to be wrong, only to look.
+#
+# The controls carry the argument. Without them a failing arm would only say that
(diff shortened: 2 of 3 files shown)


OffgridwithJD (approved):
Verified by running, and the load-bearing claim checked against PostgreSQL's own
source on every major we support. Three arms at `5df45b1e`, base `808cd46`.

## The fix is pinned by the test

| arm | result |
| --- | --- |
| **A** parent `src/` + this PR's `test/index_delete_liveness.sh` | **RED** |

Arm C's `src/` is byte-identical to the parent under `git diff --quiet`.

## The nbtree contract, checked rather than taken

Setting `delstate->ndeltids = 0` is only safe if nbtree tolerates an empty array, so I
read `_bt_delitems_delete_check` in the actual sources rather than trusting the
comment. It is better than tolerated — it is documented:

```c
/*
 * Note that deltids array might be a lot smaller now.  It might even have
 * no entries at all (with bottom-up deletion caller), in which case there
 * is nothing left to do.
 */
qsort(delstate->deltids, delstate->ndeltids, sizeof(TM_IndexDelete), _bt_delitems_cmp);
if (delstate->ndeltids == 0)
{
    Assert(delstate->bottomup);
    return;
}
```

**Identical in 15.18, 16.14, 17.6, 18.4 and 19beta2** — checked in each tree, not
extrapolated from one. So the decline is safe across the whole support range.

That block also proves your `if (delstate->bottomup)` guard is *necessary*, not
merely tidy: zeroing on the simple-deletion path would trip that Assert. Your comment
says "wrong twice over" and the Assert is the second of the two.

## The cost is bounded, and the documentation already described the outcome

You call the cost "an index grows where it used to be compacted". Two things sharpen
that, both in your favour:

**It is bounded.** `columnar_vacuum.c:1373` and `:1521` call
`PgColumnarReindexRelation`, so the compaction path rebuilds the indexes. The
reclamation moves from opportunistic to explicit rather than disappearing.

**`docs/limitations.md:596-597` already says exactly this:**

> Stale index entries left by deletes and updates are filtered on fetch and reclaimed
> by `REINDEX`, not removed opportunistically.

That sentence was already true of the intended design and false of the code. This PR
makes the code match a document that was right all along, which is a stronger argument
for declining than the one the commit makes — and it is why I do not think this PR
owes a documentation change despite touching user-visible behaviour.

## The suite is built the way this kind of claim needs

Two controls, and they are what make the arm specific rather than suggestive:
indexed-column churn (`a = a`) makes `index_unchanged_by_update()` false so no
bottom-up pass runs, and a heap with a second index takes the same hint. Both must
lose nothing, and a fault that broke them both would be a different fault.

The plan premise is the part I would have asked for if it were missing: `arm A really
plans an Index Scan, so the arm measures the index`. Without it a fallback to a
sequential scan would answer correctly and the arm would pass while measuring nothing.
The `ROLLBACK` arm is a good choice too — 0 of 400 reachable is a far louder signal
than 228 of 400, and it isolates "this backend's uncommitted state was consulted".

Registration is correct and alphabetical (`import_deferred` / `index_delete_liveness` /
`index_fetch_penalty_width`), checked because an unregistered suite is a gate that
never runs.

## Gate

| arm | result |
| --- | --- |
| matrix PG18 | **232** ran, 2 skipped, ALL PASSED |
| matrix PG19 | **234** ran, 0 skipped, ALL PASSED |
| `index_delete_liveness` | PASS on both arms |
| failing suites | none |

`main` at `808cd46` gave 231 and 233 on this box today, so +1 per arm is
`index_delete_liveness` actually joining the matrix rather than merely existing.

OffgridwithJD:
Follow-up to my approval, which stands — the code is right and the gate is green. I
missed something when I posted it, and it is better said late than not at all.

**This PR fixes a bug that is live in the shipped `v1.0-alpha2`, and it carries no
CHANGELOG entry.**

I checked rather than assumed, because for most of this sweep the answer is the
opposite:

| PR | bug present at `v1.0-alpha2`? | evidence |
| --- | --- | --- |
| #835 | **no** | `Anum_options_ttl_*` arrived in `bb97a86` (2026-08-28), after the tag. `Natts_options 7` was *correct* at the tag. |
| #837 | **no** | `pgcolumnar_preimage_range_scankey` does not exist at the tag |
| #841 | **no** | `i128sum` does not exist at the tag |
| **#839** | **yes** | `git show v1.0-alpha2:src/columnar_tableam.c` has the `ActiveSnapshotSet() ? GetActiveSnapshot() : GetTransactionSnapshot()` line and the `knowndeletable` assignment |
| #843 | **yes** | `pgcolumnar_scan_getnextslot` exists at the tag with exactly one reference to `direction` — the parameter itself |

Three of the five are alpha3-cycle regressions: introduced and fixed inside an
unreleased cycle, where a CHANGELOG line is arguably noise. **This one is not.**
Anyone running `v1.0-alpha2` today gets `count(*)` = 228 where the truth is 400, at
shipped defaults, with no GUC set. That is exactly what a `### Fixed` entry is for,
and the repo already does this for user-visible fixes — `#826`, `#816` and `#821` each
carried one.

Not asking you to re-open anything. Either a line in `[Unreleased] / Fixed` here, or a
note on the release plan that alpha2 carries a known wrong-answer defect, so it is
recorded somewhere a user reaches.

*(My approval of #835, #837 and #841 did not raise this, and after checking, I do not
think it should have: those three never shipped. #843 is the other one that did, and I
will raise it there.)*

PR #841, opened 2026-08-29
fix: reset the aggregate accumulators in one place, not two (#840)

### LINKED ISSUE #840: sum(bigint) and avg(bigint) accumulate across a rescan: PgColumnarReScanAggScan never clears i128sum
With `pgcolumnar.enable_ungrouped_vector_agg=on`, a re-executed aggregate node adds to the previous execution's running total.

Measured on PG 17.10 and PG 18.4, four LATERAL iterations, node asserted chosen and asserted rescanned (`loops=4`):

| driver row | heap | columnar |
| ---: | ---: | ---: |
| 1 | 800022400060000 | 800022400060000 |
| 2 | 800022399059997 | 1600044799119997 |
| 3 | 800022397059991 | 2400067196179988 |
| 4 | 800022394059982 | 3200089590239970 |

Reproduces through two independent rescan drivers (LATERAL and a correlated scalar subquery) and for `avg(bigint)` as well.

## Cause

There are two accumulator reset sites in `src/columnar_vector.c`, and they had drifted:

| function | fields cleared |
| --- | --- |
| `pgcolumnar_agg_specs_reset` (~3548) | 9, including `spec->i128sum` |
| `PgColumnarReScanAggScan` (~4224) | 8, **no** `i128sum` |

They were otherwise byte-identical. The int8 kinds accumulate into `i128sum` under `HAVE_INT128`, so on a rescan that total was never zeroed.

## Blast radius, each arm measured

| configuration | result |
| --- | --- |
| default (`enable_ungrouped_vector_agg=off`) | correct |
| `enable_ungrouped_vector_agg=on` | **wrong** |
| grouped node (`enable_group_vectorization=on`) | correct |
| same path, `sum(int4)` | correct |

The last row is the discriminating control: it isolates the fault to `i128sum` rather than to rescan generally. The grouped node is unaffected because it resets whole memory contexts instead of clearing fields one by one.

## Why nothing caught it

`int8_agg_int128.sh` drives the int128 accumulator and never rescans.
`vector_agg_rescan_memory.sh` rescans and aggregates `count(*)`, which does not touch `i128sum`. Each condition is covered; their intersection is not.

## The instructive part

A comment on `pgcolumnar_batch_agg_ok` already says to check that every accumulator a kind uses is cleared, names `i128sum` and #786 explicitly, and ends "do not rely on a test to catch a missing reset. Read it." That comment points at `pgcolumnar_agg_specs_reset`, which is the copy that was **right**. The rule was written, and correct, and still did not fire, because it named one function while the hazard was that there were two.

jdatcmd:
Closes #840.

`sum(bigint)` and `avg(bigint)` came back cumulative across a rescan:

| driver row | heap | columnar |
| ---: | ---: | ---: |
| 1 | 800022400060000 | 800022400060000 |
| 2 | 800022399059997 | 1600044799119997 |
| 3 | 800022397059991 | 2400067196179988 |
| 4 | 800022394059982 | 3200089590239970 |

## Cause

Two accumulator reset sites, drifted apart:

| function | fields cleared |
| --- | --- |
| `pgcolumnar_agg_specs_reset` | 9, including `spec->i128sum` |
| `PgColumnarReScanAggScan` | 8, **no** `i128sum` |

Otherwise byte-identical. The rescan path now calls the shared function. Adding the missing line to the second copy would fix this instance and leave the next accumulator to be forgotten the same way.

## The part worth reading

A comment on `pgcolumnar_batch_agg_ok` already says to check that every accumulator a kind uses is cleared, names `i128sum` and #786, and ends *"do not rely on a test to catch a missing reset. Read it."*

That comment points at `pgcolumnar_agg_specs_reset` — the copy that was **right**. The rule was written, correct, and still did not fire, because it named one function while the hazard was that there were two. Single-sourcing is what makes the instruction reach both paths.

## Test

`int8_agg_int128.sh` drives the accumulator and never rescanned; `vector_agg_rescan_memory.sh` rescans and aggregates `count(*)`, which does not touch `i128sum`. The intersection is what the new arms add: a LATERAL over a four-row driver with `enable_material=off`, asserting the vectorized node is chosen and really re-executed (`loops=4`) before any answer is compared. `count(*)` across the same rescan is the control that places a failure on the accumulator rather than on rescan.

### A first version of those arms passed while testing nothing

`local t="$1" sql="${RSLAT//%T/$t}"` reads `t` before it is assigned, so under `set -u` the helper aborted and **both sides of every comparison came back as the same error text**, which `check` passed. Recorded because it is the `check "" ""` shape the harness already warns about, reached by a different route.

Fixed by splitting the declarations and adding a premise that the query returns one row per driver row, so an error cannot compare equal to an error.

## Removal proof

| | fixed | open-coded loop restored |
| --- | --- | --- |
| installed `.so` | `ad3949d57b61` | `ad31355476c0` |
| three int8 rescan arms | PASSED | **FAIL** |
| `count(*)` control | PASSED | PASSED |

## Blast radius

Default configuration is unaffected: `enable_ungrouped_vector_agg` ships off. The grouped node is unaffected because it resets whole memory contexts rather than clearing fields individually.

Warning-clean under `gcc -Werror` and under clang against the clang-configured prefix.

### DIFF
--- a/src/columnar_vector.c
+++ b/src/columnar_vector.c
@@ -4224,7 +4224,6 @@ static void
 PgColumnarReScanAggScan(CustomScanState *node)
 {
 	PgColumnarAggScanState *state = (PgColumnarAggScanState *) node;
-	int			a;
 
 	state->done = false;
 	state->haveStats = false;
@@ -4232,27 +4231,27 @@ PgColumnarReScanAggScan(CustomScanState *node)
 	state->foldPayloadLoads = 0;
 	state->foldGroupsDeferred = 0;
 	state->foldGroupsTotal = 0;
-	MemoryContextReset(state->resultContext);
-	for (a = 0; a < state->naggs; a++)
-	{
-		PgColumnarAggSpec *spec = &state->specs[a];
 
-		spec->count = 0;
-		spec->sum = 0;
-		spec->sawValue = false;
-		spec->extreme = (Datum) 0;
-		/*
-		 * Also clear the running float/numeric accumulators. resultContext was
-		 * just reset, so nsum's storage is gone; leaving nsumSet true would make
-		 * the next scan add to (and free) a dangling pointer. Scan-fold mode
-		 * classifies these extended kinds (#289), so this reset is load-bearing on
-		 * a rescan, not only defensive.
-		 */
-		spec->fsum = 0;
-		spec->fsxx = 0;
-		spec->nsum = (Datum) 0;
-		spec->nsumSet = false;
-	}
+	/*
+	 * One reset, not two. This used to open-code the loop that
+	 * pgcolumnar_agg_specs_reset already contains, and the two copies were
+	 * identical except that this one never cleared spec->i128sum. The int8 kinds
+	 * accumulate there, so a rescan carried the previous execution's running
+	 * total into the next one and sum(bigint) came back cumulative: measured
+	 * 800022400060000 then 1600044799119997 then 2400067196179988 over four
+	 * LATERAL iterations whose true answers were all near the first.
+	 *
+	 * The comment on pgcolumnar_batch_agg_ok already said to check that every
+	 * accumulator a kind uses is cleared, and named i128sum. It points at
+	 * pgcolumnar_agg_specs_reset, which was right. Calling that function from
+	 * here is what makes the instruction reach both paths, rather than leaving a
+	 * second copy for the next accumulator to be forgotten in.
+	 *
+	 * The float and numeric clears are load-bearing rather than defensive:
+	 * resultContext has just been reset, so nsum's storage is gone and leaving
+	 * nsumSet true would make the next scan add to a dangling pointer.
+	 */
+	pgcolumnar_agg_specs_reset(state);
 }
 
 static void
--- a/test/int8_agg_int128.sh
+++ b/test/int8_agg_int128.sh
@@ -227,4 +227,67 @@ done
 echo "-- sum(bigint) WHERE k < 50: vectorized $FON ms, scalar $FOFF ms"
 check_ratio "the FILTERED vectorized path is not slower than the scalar one" "$FON" "$FOFF" "1.10"
 
+# ---- the accumulator must be cleared on a RESCAN ---------------------------
+# The int8 kinds accumulate into spec->i128sum. There are two reset sites:
+# pgcolumnar_agg_specs_reset clears nine fields including that one, and
+# PgColumnarReScanAggScan open-coded the same loop with eight and no i128sum. On
+# a rescan the running int128 total was therefore never zeroed and each
+# re-execution added to the previous one.
+#
+# Nothing here reached it. This suite drives the int128 accumulator and never
+# rescans; vector_agg_rescan_memory.sh rescans and aggregates count(*), which
+# does not touch i128sum. Each condition was covered and their intersection was
+# not.
+#
+# A LATERAL over a small driver re-executes the aggregate node once per driver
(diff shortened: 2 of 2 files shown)


OffgridwithJD (approved):
Verified by running. Three arms at `993c5af4`, base `808cd46`.

| arm | result |
| --- | --- |
| **A** parent `src/` + this PR's `test/int8_agg_int128.sh` | **RED** |

Arm C's `src/` is byte-identical to the parent under `git diff --quiet`.

## The consolidation is correct, and I checked the part that could have gone wrong

Replacing the open-coded loop with a call is the right shape — it closes the drift
class rather than the instance — but it silently moves a fourth thing. The old body
began with `MemoryContextReset(state->resultContext)`, and this diff deletes that line.

That is only safe if the shared function does it, in the same order. It does:

```c
pgcolumnar_agg_specs_reset(PgColumnarAggScanState *state)
{
    int a;

    MemoryContextReset(state->resultContext);
    for (a = 0; a < state->naggs; a++)
    { ... spec->nsum = (Datum) 0; spec->nsumSet = false;
#ifdef HAVE_INT128
      spec->i128sum = 0;
#endif
    }
}
```

The reset comes first, before `nsumSet` is cleared, which is exactly what the
dangling-`nsum` argument in your comment depends on. Had the order been inverted the
consolidation would have been a different bug wearing this one's clothes.

Two callers now, one definition. That is the fix.

## Why this is the right fix and not just a fix

The cheap version was `spec->i128sum = 0;` in the second loop. It would have passed
the same test and left the next accumulator to be forgotten in exactly the same place
— which is what happened here: the constants were extended for retention and one of
the two copies was not, the same shape as #834 one layer up. Deleting the second copy
is what makes the next `PgColumnarAggSpec` field safe by construction.

Your note that `pgcolumnar_batch_agg_ok`'s comment already pointed at
`pgcolumnar_agg_specs_reset` and already named `i128sum` is the sharp part: the
instruction was correct and unreachable, because the path that needed it did not call
the function the instruction described.

## The test asserts the path, not the GUC

This is the part I check hardest, because a rescan test that never rescans passes for
the wrong reason:

- `premise: the rescan shape still takes the vectorized node` — greps for
  `Columnar Vectorized Aggregates` in `EXPLAIN (ANALYZE)`. Setting
  `enable_ungrouped_vector_agg=on` is not the same as engaging it.
- `premise: the node really is re-executed, so a stale accumulator can show` —
  `loops > 1`, taken from the plan rather than assumed from the LATERAL.
- `premise: the rescan query returns a row per driver row, not an error` — four rows,
  so an erroring query cannot pass as agreement.
- `control: count(*) across the same rescan matches the heap` — `count(*)` does not
  touch `i128sum`, so it must pass either way. That is what makes the three failing
  arms specific rather than "something about rescans is broken".

`enable_material=off` and `max_parallel_workers_per_gather=0` are load-bearing here
too: a materialised or gathered node would not re-execute the scan and the arm would
measure nothing.

## Gate

| arm | result |
| --- | --- |
| matrix PG18 | 231 ran, 2 skipped, ALL PASSED |
| `int8_agg_int128` | PASS on both arms |
| failing suites | none |

Same tallies as `main` at `808cd46` on this box today, which is expected: this PR
extends an existing suite rather than adding one.

One note for the record, since it concerns what a CHANGELOG is for rather than this
code: `i128sum` does not exist at `v1.0-alpha2`, so this defect was introduced and
fixed inside the unreleased cycle and never reached a user. That is why I am not
asking for a `### Fixed` entry here, and it is a different answer from the one I gave
on #839, where the bug does ship today.

PR #843, opened 2026-08-29
fix: honour the scan direction, or refuse it (#842)

### LINKED ISSUE #842: FETCH BACKWARD returns forward rows: pgcolumnar_scan_getnextslot ignores its ScanDirection
`pgcolumnar_scan_getnextslot` accepts a `ScanDirection` argument and never reads it. Every call advances forward.

`heap_getnextslot` passes the direction to `heapgettup`, which walks the other way. So a scrollable cursor over a columnar relation, on a plan that reaches the table AM rather than the custom scan, answers a backward fetch with more forward rows.

## Measured (PG 17.10 and PG 18.4, 20-row fixture, plan asserted in both arms)

| operation, `enable_custom_scan=off` | heap | columnar |
| --- | --- | --- |
| `FETCH FORWARD 5` then `FETCH BACKWARD 3` | 4 3 2 | **6 7 8** |
| `MOVE FORWARD 8` then `FETCH BACKWARD 2` | 7 6 | **9 10** |
| `FETCH LAST` | 20 | *(empty)* |

The fetch **succeeds**: SQLSTATE 00000. It simply returns rows that were not asked for.

## Reachability

At shipped defaults the planner gives the columnar relation a `Custom Scan (PgColumnarScan)`, which serves these cursors correctly. Six shapes were probed with nothing set — plain select, `SELECT *`, a materialized CTE, `UNION ALL`, a filtered scan, an inheritance parent — and every one took the custom scan. So this needs the explicit opt-out today.

That is a statement about which plans currently reach the callback, not about the callback being right. Anything that suppresses the custom-scan path re-exposes it.

## Not documented

`docs/limitations.md` does not mention cursors or backward scans, and no suite declares a `SCROLL` cursor.

jdatcmd:
Closes #842.

`pgcolumnar_scan_getnextslot` took a `ScanDirection` and never read it, so a backward fetch was answered with more forward rows — and it **succeeded**, SQLSTATE 00000.

| operation, `enable_custom_scan=off` | heap | columnar |
| --- | --- | --- |
| `FETCH FORWARD 5` then `FETCH BACKWARD 3` | 4 3 2 | **6 7 8** |
| `MOVE FORWARD 8` then `FETCH BACKWARD 2` | 7 6 | **9 10** |
| `FETCH LAST` | 20 | *(empty)* |

## Fix

`NoMovement` is answered as what it is, a request for no row. `Backward` is refused with `0A000` rather than emulated: the reader walks row groups and decodes vectors forwards, so scanning the other way is a feature rather than a correction, and inventing an answer is the one outcome that must not stand.

At shipped defaults the custom scan serves these cursors correctly — six query shapes were probed with nothing set and every one gave the columnar relation a Custom Scan — so the refusal is reached only where that path has been turned off.

## Test

Asserts **SQLSTATE**, not message text: a grep for wording is also satisfied by an unrelated error or a failed connection. Two controls:

- the same cursor on a **heap** table must raise nothing (00000)
- a **forward-only** fetch on the columnar table must report success

Without them an arm could pass by the probe being stuck on one answer.

## Removal proof

Without the change the backward arm reports `00000` — the success state. That *is* the defect: the fetch succeeds and the rows are wrong.

## Verification

Full PG18 matrix from a clean `git archive` lane: **ALL VERSIONS PASSED**, 228 of 234 suites ran, zero failures, `scan_direction=PASS`.

An earlier run on a shared lane showed three unrelated failures. They were sibling branches' test files accumulated in that lane, and `harness_selftest` named them exactly (`unregistered: catalog_natts index_delete_liveness`) while its `.so` fingerprint check caught a concurrent overwrite. Recorded rather than reasoned away, because the clean-lane rerun is what makes the green trustworthy.

### DIFF
--- a/src/columnar_tableam.c
+++ b/src/columnar_tableam.c
@@ -602,6 +602,37 @@ pgcolumnar_scan_getnextslot(TableScanDesc sscan, ScanDirection direction,
 
 	ExecClearTuple(slot);
 
+	/*
+	 * Honour the direction, or refuse it. This used to ignore the argument
+	 * entirely and advance forward whatever was asked, so a scrollable cursor
+	 * that reached the AM rather than the custom scan answered FETCH BACKWARD
+	 * with more forward rows -- 6 7 8 where the heap gave 4 3 2 -- and FETCH
+	 * LAST with nothing at all.
+	 *
+	 * NoMovement is a real request for no row and is answered as such.
+	 *
+	 * Backward is refused rather than emulated. The reader walks row groups and
+	 * decodes vectors forwards, so scanning the other way is a feature rather
+	 * than a correction, and inventing an answer is the one thing that must not
+	 * happen here. Refusing is visible, and at shipped defaults the custom scan
+	 * serves these cursors, so the refusal is reached only when that path has
+	 * been turned off.
+	 */
+	if (unlikely(direction != ForwardScanDirection))
+	{
+		if (direction == NoMovementScanDirection)
+			return false;
+
+		ereport(ERROR,
+				(errcode(ERRCODE_FEATURE_NOT_SUPPORTED),
+				 errmsg("columnar tables do not support backward scans"),
+				 errdetail("A scrollable cursor over relation \"%s\" asked for a "
+						   "backward fetch, which native storage cannot serve.",
+						   RelationGetRelationName(scan->rs_base.rs_rd)),
+				 errhint("Leave pgcolumnar.enable_custom_scan on, or add an "
+						 "ORDER BY so the cursor is materialized.")));
+	}
+
 	if (!PgColumnarReadNextRow(pgcolumnar_scan_read_state(scan,
 													 slot->tts_tupleDescriptor),
 							 slot->tts_values, slot->tts_isnull, &rowNumber))
--- a/test/run_all_versions.sh
+++ b/test/run_all_versions.sh
@@ -248,6 +248,7 @@ SUITES=(
 	rls_direct_storage
 	row_triggers
 	scan_decode_cost
+	scan_direction
 	server_file_privilege
 	smoke
 	sort_status
--- a/test/scan_direction.sh
+++ b/test/scan_direction.sh
@@ -1,0 +1,97 @@
+#!/usr/bin/env bash
+#
+# A scan must not answer a backward fetch with forward rows.
+#
+# pgcolumnar_scan_getnextslot takes a ScanDirection and never read it: every
+# call advanced forward. heap_getnextslot passes direction through to
+# heapgettup, which walks the other way. So a scrollable cursor over a columnar
+# relation, on a plan that reaches the table AM rather than the custom scan,
+# silently returned the wrong rows:
+#
+#     FETCH FORWARD 5   then   FETCH BACKWARD 3
+#     heap:     1 2 3 4 5  then  4 3 2
+#     columnar: 1 2 3 4 5  then  6 7 8
+#
+# and FETCH LAST returned nothing at all where the heap returned row 20.
+#
+# A true backward scan over native storage is a feature, not a bug fix: the
+# reader walks row groups and decodes vectors forwards. What this pins is that
+# the AM never INVENTS an answer. Forward is unchanged, no-movement fetches
+# nothing, and a backward fetch raises 0A000 rather than returning rows that were
+# never asked for.
+#
+# The arms assert SQLSTATE rather than message text. A grep for "not supported"
+# is also satisfied by a dozen unrelated errors, and by a connection failure.
+#
+# Usage:  test/scan_direction.sh [PG_CONFIG]
+# Written fresh for pgColumnar.
+
+set -uo pipefail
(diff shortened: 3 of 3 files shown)


OffgridwithJD (approved):
Verified by running. Three arms at `00737887`, base `808cd46`.

| arm | result |
| --- | --- |
| **A** parent `src/` + this PR's test files | **RED** — `a backward fetch through the AM raises 0A000 rather than answering: got [00000] want [0A000]` and the same for `FETCH LAST` |

`got [00000]` is the defect stated exactly: the old path reported **success** while
returning rows nobody asked for.

## Gate, twice — the second one against the tree that will actually land

`main` moved under this branch while I was reviewing (#835, #837, #839, #841 merged),
and **#839 touches the same file you do**, `src/columnar_tableam.c`. So the first gate
tested a tree that no longer represents the result. I merged your branch into current
`main` and gated that.

| | at your head (`808cd46` base) | merged into `main` `70445ca5` |
| --- | --- | --- |
| merge | — | clean, `465f80f6`, 3 files +129 |
| preflight 15-19 | 5 of 5 | 5 of 5 |
| PG18 | 232 ran / 2 skipped | **234 ran / 2 skipped** |
| PG19 | 234 ran / 0 skipped | **236 ran / 0 skipped** |
| `scan_direction` | PASS both | PASS both |
| failing suites | none | none |

In the merged tree the suites the other four PRs added all still pass beside yours —
`catalog_natts`, `index_delete_liveness`, `int8_agg_int128`, `preimage_rewrite`, each
on both arms. That is the composition check, and it is the reason I did not simply
trust the first gate.

## What I checked and did not need to raise

**`NoMovementScanDirection` returning false is right.** Core states the contract in
`execMain.c`: *"If direction is NoMovementScanDirection then nothing is done"*. Your
`return false` is the AM-level spelling of that, and it is defensive rather than
reachable through the executor.

**The suite covers the half that a refusal could break.** `forward-only fetch still
matches heap through the AM` is the arm I would have asked for, and the two controls
are what make the deny arms mean something: the same cursor on a heap raising `00000`,
and a forward-only fetch on the columnar table also raising `00000`, so the probe is
demonstrably able to report both outcomes rather than being stuck on `0A000`. The
`Seq Scan` plan premise asserts the path instead of trusting the GUC.

## Two things this PR is missing, and the first is not a style point

**1. It fixes a bug that ships today, with no CHANGELOG entry.**

I checked each PR of this sweep against the tag rather than applying a rule, because
for three of them the answer is the opposite:

| PR | present at `v1.0-alpha2`? |
| --- | --- |
| #835 | no — `Anum_options_ttl_*` arrived in `bb97a86`, after the tag |
| #837 | no — `preimage_range_scankey` does not exist at the tag |
| #841 | no — `i128sum` does not exist at the tag |
| #839 | **yes** |
| **#843** | **yes** — `pgcolumnar_scan_getnextslot` exists at the tag with exactly one reference to `direction`: the parameter |

So anyone on `v1.0-alpha2` today gets forward rows from a backward fetch, with
SQLSTATE `00000`. That belongs in `[Unreleased] / Fixed`, as `#826`, `#816` and `#821`
each did for user-visible fixes.

**2. It adds a user-facing ERROR and documents it nowhere.**

`docs/limitations.md` at this head contains no occurrence of *cursor*, *backward*,
*scrollable*, or `FETCH` in the cursor sense — I grepped; the hits are all "fetch" in
the index-performance discussion. Your own #842 says so: *"`docs/limitations.md` does
not mention cursors or backward scans"*. This PR is where that stops being an
observation about a latent bug and becomes an undocumented error message a user can
hit.

The Indexes section already has the right shape for it — one line under a Cursors
heading saying a backward or absolute fetch is refused when the custom scan is off,
and that the custom scan serves scrollable cursors at shipped defaults.

## And one claim inside the error itself

Your `errhint` tells the user two things to do, so I ran both rather than reading
them. A wrong hint is worse than no hint: it costs the reader a second failure.

    CONTROL  custom scan OFF, no ORDER BY   -> ERROR fires, as intended
    HALF 1   custom scan ON                 -> 1 2 3 4 5 / 4 3 2
    HALF 2   custom scan OFF + ORDER BY     -> 1 2 3 4 5 / 4 3 2
    ORACLE   the same cursor on a heap      -> 1 2 3 4 5 / 4 3 2

**Both halves are true, and the plan shows why rather than only that:**

    Sort
      Sort Key: i
      ->  Seq Scan on cur_t

The `Sort` node buffers, so the backward fetch is served from it and never reaches
the AM. The hint is accurate advice and I have nothing to change in it.

I mention the check because the wording *"so the cursor is materialized"* is a claim
about a mechanism, and those are the ones that quietly stop being true. If the planner
ever serves that shape without a buffering node, the hint sends the reader into the
same error twice. Worth a line in `scan_direction.sh` asserting half 2 keeps working,
since the suite already has the cursor machinery to do it in about three lines.

---

Nothing above touches the correctness of the change, which I am satisfied with. Say
the word and I will open the docs/CHANGELOG follow-up myself rather than hold this.

PR #845, opened 2026-08-29
fix: refuse an encoded NUL in a recorded Iceberg path (#844)

### LINKED ISSUE #844: An encoded NUL defeats the Iceberg encoded-traversal guard on object storage
`ice_has_encoded_dotdot` exists to catch a traversal smuggled through percent-encoding, because an http(s) origin or proxy may decode a key before serving it. A `%00` in front of the traversal defeats it.

## Measured

Over the S3 fixture, `pgcolumnar.iceberg_scan` on a table whose equality-delete path is recorded as
`file:///tmp/pgc_ice_del/db/t/%00%2e%2e/%2e%2e/%2e%2e/%2e%2e/etc/hostname`:

| recorded delete path | SQLSTATE |
| --- | --- |
| `../../../../etc/hostname` (literal) | 22023, refused |
| `%2e%2e/%2e%2e/...` (encoded) | 22023, refused |
| `%00%2e%2e/%2e%2e/...` (encoded, behind an encoded NUL) | **58P01** |

`58P01` is `undefined_file`: the path was **not refused**. It passed both containment checks and was fetched, and the fetch failed at the escaped location.

## Cause

`ice_percent_decode_once` decodes `%00` to an embedded NUL and keeps writing. Everything downstream is NUL-terminated string work: the next pass measures with `strlen`, and `ice_has_dotdot` walks with `strchr`. Both stop at that byte and never reach the `../` behind it, so the function returns false. The literal-`..` guard beside it sees no `..` either, because the segments are encoded.

## Where it matters

`ice_rebase` has two branches. For a **local** root it calls `canonicalize_path` and re-checks containment, which collapses the traversal regardless — so a local table is backstopped.

For a **remote** root the containment is purely lexical; the code says so:

> `Object storage: ... there is no realpath/symlink notion, so the containment is purely lexical`

So this is the object-storage path, where the guard being defeated means there is no other check.

## Not claimed

Whether a given origin or proxy would reconstruct a readable path from a key containing a NUL is not established here, and most reject NUL outright. What is established is that the guard fails to fire on the input its own comment says it exists to catch, and that the path is then fetched rather than refused.

jdatcmd:
Closes #844.

`ice_has_encoded_dotdot` exists because an http(s) origin or proxy may percent-decode a key before serving it, so `%2e%2e` must be refused as well as `..`. A `%00` in front of the traversal defeated it.

## Measured over the s3 fixture

| recorded delete path | SQLSTATE |
| --- | --- |
| `../../../../etc/hostname` (literal) | 22023, refused |
| `%2e%2e/%2e%2e/...` (encoded) | 22023, refused |
| `%00%2e%2e/%2e%2e/...` | **58P01** |

`58P01` is `undefined_file` — not a refusal. The reader accepted the path, went to the network with it, and the fetch failed at the escaped location.

## Cause

`ice_percent_decode_once` decodes `%00` to an embedded NUL and keeps writing. Everything downstream is NUL-terminated: the next pass measures with `strlen`, `ice_has_dotdot` walks with `strchr`. Both stop at that byte and never reach the `../` behind it. The literal-`..` guard sees no `..` either, because the segments are encoded — so the path passed **both** containment checks.

## Where it matters

`ice_rebase` canonicalizes a **local** root and re-checks containment, which collapses the traversal regardless. For a **remote** root the containment is purely lexical, as that branch's own comment states, so the defeated guard is the only check. This is an object-storage issue.

## Test

An `eqnulpct` variant beside the existing `eqpctdot` fixture, and an arm beside the existing encoded-dotdot arm.

The neighbour is what makes the new arm non-vacuous: it stays green throughout, so a failure is specific to the NUL rather than to encoded traversals in general. A local-path arm would have been worthless here — `realpath` containment refuses it either way, so it could not fail.

The generator was verified deterministic before use: regenerating `eqpctdot` produced byte-identical output to the committed fixture.

## Removal proof

| | fixed | refusal deleted |
| --- | --- | --- |
| installed `.so` | `9b3e1ec05a46` | `1d4c9b7d9860` |
| the new NUL arm | PASSED | **FAIL**, back to `58P01` |
| the encoded-dotdot neighbour | PASSED | PASSED |

## Not claimed

Whether a particular origin or proxy would reconstruct a readable path from a key containing a NUL is not established, and most reject one outright. What is established is that the guard fails to fire on the input its own comment says it exists to catch, and that the path is then fetched rather than refused.

All ten iceberg suites pass. Warning-clean under `gcc -Werror`.


OffgridwithJD (approved):
Verified by running. Three arms at `f2ef2d7d`, base `808cd46`.

| arm | result |
| --- | --- |
| **A** parent `src/` + this PR's test files (5 paths, each verified against its head blob) | **RED** — `a percent-encoded dotdot behind an encoded NUL is refused over s3 (22023): got [58P01] want [22023]` |

`got [58P01]` states the severity exactly: `undefined_file` means the escaped path was
not refused, it was **fetched**, and only the filesystem stopped it.

## Gate, twice

`main` moved under this branch while I reviewed, so I gated your head and then gated
the merge result, because "merges cleanly" is not "behaves correctly merged".

| | at your head | merged into `main` `70445ca5` |
| --- | --- | --- |
| merge | — | clean, `d3b0b26f`, 6 files +40/-2 |
| preflight 15-19 | 5 of 5 | 5 of 5 |
| PG18 | 231 ran / 2 skipped | **233 ran / 2 skipped** |
| PG19 | 233 / 0 | **235 / 0** |
| failing suites | none | none |

In the merged tree **all sixteen `iceberg_*` suites pass on both arms**, which is the
check that matters for a change to path validation — the guard is shared by the FDW,
the REST catalog, the delete paths and the scan. `catalog_natts`,
`index_delete_liveness`, `int8_agg_int128` and `preimage_rewrite` pass beside them.

`main` has since moved once more (#843). I did **not** re-gate for it: #843 touches
`src/columnar_tableam.c`, `test/run_all_versions.sh` and `test/scan_direction.sh`,
and this PR touches `src/columnar_iceberg.c` plus its own Iceberg fixtures. True
overlap is zero, measured from this PR's own base rather than from the moved main —
a two-dot diff against the new main reports a false overlap, because it shows the
already-merged work as though it were yours.

## Three things I checked at source

**The signature change is contained.** `ice_percent_decode_once` has exactly one
caller, and `ice_has_encoded_dotdot` exactly one. So no path silently receives a
`sawnul` it ignores, which is the usual way a fix like this leaves a hole behind.

**Double encoding is covered, and not by accident.** `sawnul` is declared outside the
decode loop, so `%2500` decodes to `%00` on the first pass with `sawnul` still false,
and to a NUL on the second, which trips it. Had it been declared inside, the loop
would have reset it every pass and the fix would cover only single encoding.

**The other branch of `ice_rebase` is genuinely not exposed.** It takes the local-root
path, which never decodes and goes through `canonicalize_path` — a mechanism that
physically collapses the traversal rather than detecting it. So targeting only the
object-storage branch is right, not a gap.

## This one ships, and it is the security-relevant one

I checked each PR in this sweep against the tag rather than applying a rule. For
#835, #837 and #841 the answer was "introduced and fixed inside the unreleased
cycle" — no user ever saw them. **This is not one of those.**

At `v1.0-alpha2`: `columnar_iceberg.c` is present, `ice_has_encoded_dotdot` and
`ice_percent_decode_once` both exist, the NUL check does **not**, and `iceberg_scan`
appears in the shipped `pgcolumnar--1.0-alpha2.sql`. So a published release carries a
containment guard that a `%00` defeats, on a path that reads a file the server names.

That belongs in `[Unreleased] / Fixed`, and more plainly than the others in this
sweep: a user deciding whether to upgrade needs to be able to find it. #839 and #843
are in the same position and I have said so on each; this is the one where I would
also consider whether the release notes for the next tag should call it out by name.

## One small thing

`ice_has_encoded_dotdot` now returns true for something that is not a dotdot. As a
gate that is exactly right — the answer to both is "refuse" — but the name now
under-describes it, and someone later asking "why was my path rejected?" will grep for
`dotdot` and find nothing about NUL. The comment inside is excellent and does the
explaining; a name like `ice_encoded_path_is_unsafe` would put it where the caller
reads. Entirely optional, and not worth a re-push on its own.

PR #870, opened 2026-09-01
fix: read the temporal unit and carrier width the Arrow file declares

### LINKED ISSUE #864: A valid Arrow date64 column imports as a wrong date, silently (2000-01-01 becomes 4908285-05-04)
## A valid Arrow `date64` column imports as a wrong date, silently

`pgcolumnar.import_arrow` accepts an Arrow `date64` column into a PostgreSQL
`date` column and stores a value that is not the one in the file. No error, no
warning.

Measured on `main` at `8b39053`, PG 18.4, pyarrow 23.0.1:

```
date64 value (ms since epoch)   stored          should be
    946684800000                4908285-05-04   2000-01-01
              0                 1970-01-01      1970-01-01   (correct by luck)
       86400001                 238525-03-03    1970-01-02
   -1000000000000000000         4072193-07-19   (out of range, should error)
   INT64_MIN                    1970-01-01      (out of range, should error)
```

The second row is the reason this is filed as a bug rather than a hardening
request: **`2000-01-01` is not an edge case.** Any ordinary `date64` file
imports garbage.

`date32` is handled correctly — the same out-of-range probe is rejected:

```
date32 INT32_MAX  ->  rejected
```

so this is specific to the 8-byte `date64` carrier.

## Why it matters more than an out-of-range value

An out-of-range value that errors is a nuisance. A valid value that imports as a
different valid value is silent corruption: nothing in the result set says the
import went wrong, and the wrong dates are indistinguishable from data until
someone reads them.

`date64` is not exotic. `pyarrow` produces it directly (`pa.date64()`), and it is
what several Arrow producers emit for date columns.

## Reproduction

```python
import struct, pyarrow as pa, pyarrow.ipc as ipc
arr = pa.Array.from_buffers(pa.date64(), 1,
        [None, pa.py_buffer(struct.pack('<q', 946684800000))])   # 2000-01-01
t = pa.table({'v': arr})
with ipc.new_stream(pa.OSFile('/tmp/d64.arrows','wb'), t.schema) as w:
    w.write_table(t)
```

```sql
CREATE TABLE d64t (v date) USING pgcolumnar;
SELECT pgcolumnar.import_arrow('d64t', '/tmp/d64.arrows');
SELECT v FROM d64t;     -- 4908285-05-04
```

## Interaction with the open PRs, measured

```
main    ACCEPTED, stores 4908285-05-04
#861    rejected            (schema validation refuses the layout mismatch)
#862    ACCEPTED, stores 4908285-05-04
```

**#861 already fixes this as a side effect** and ships no test for it, which is a
shame, because it is a far stronger argument for schema validation than the arm
it does ship: without the Schema check, an 8-byte carrier is decoded through a
4-byte path and the value silently changes.

**#862 does not address it.** Its stated subject is rejecting out-of-range Arrow
temporal values, and `date64` out-of-range values are still accepted
(`INT64_MIN` stores `1970-01-01`).

## What this asks for

Either a correct `date64` decode — divide by 86,400,000, then apply the same
`IS_VALID_DATE` guard `date32` gets — or an explicit refusal of the type. Both
are defensible; silently decoding it through the `date32` path is not.

If #861 lands first, the layout check refuses `date64` and this stops being a
correctness bug and becomes a supported-types question. That is a reason to
sequence #861 before #862 rather than the other way round.

Found while adversarially reviewing #861 and #862.

### LINKED ISSUE #865: import_arrow ignores the Arrow temporal unit: a valid timestamp('ns') stores as 31969-04-01
## `import_arrow` ignores the Arrow temporal unit and reads every value as microseconds

Sibling of #864. Arrow permits `s`, `ms`, `us` and `ns` for `timestamp` and
`time64`. `pgcolumnar.import_arrow` decodes the integers as microseconds
whatever the schema declares, so **valid data in any other unit is stored as a
different, valid-looking value.** No error.

Measured on `main` at `8b39053`, PG 18.4, pyarrow 23.0.1. Every file holds
`2000-01-01 00:00:00 UTC` (or `12:00:00` for `time64`) written correctly in its
own unit:

```
arrow type          expected               stored
timestamp('s')      2000-01-01 00:00:00    1970-01-01 00:15:46.6848    WRONG
timestamp('ms')     2000-01-01 00:00:00    1970-01-11 22:58:04.8       WRONG
timestamp('us')     2000-01-01 00:00:00    2000-01-01 00:00:00         ok
timestamp('ns')     2000-01-01 00:00:00    31969-04-01 00:00:00        WRONG
time64('us')        12:00:00               12:00:00                    ok
time64('ns')        12:00:00               12000:00:00                 WRONG
```

`time64('ns')` is the one worth staring at: `12000:00:00` is a legal PostgreSQL
`time` value, so nothing downstream can tell it from data.

## Why this is filed separately from the PRs in flight

**#861 already fixes it**, by refusing a schema whose layout it does not decode:

```
main    every non-us unit stored wrong
#861    every non-us unit rejected
#862    timestamp('ns') still stored wrong (31969-04-01);
        time64('ns') now rejected
```

So #861 closes this and #864 both, and ships a test for neither. #862 in
isolation improves `time64('ns')` — a rejection is better than corruption — while
leaving `timestamp('ns')` corrupting, because `31969-04-01` passes
`IS_VALID_TIMESTAMP`.

This was raised as an open question by jdatcmd on #862, explicitly not asserted
because he had not traced the schema-parsing side. Both halves of his suspicion
hold: the new bounds **reject valid data** (`time64('ns')` 12:00:00) and
**accept wrong data** (`timestamp('ns')` 2000-01-01).

## Reproduction

```python
import struct, pyarrow as pa, pyarrow.ipc as ipc
v = 946684800 * 10**9                      # 2000-01-01 in nanoseconds
arr = pa.Array.from_buffers(pa.timestamp('ns'), 1,
        [None, pa.py_buffer(struct.pack('<q', v))])
t = pa.table({'v': arr})
with ipc.new_stream(pa.OSFile('/tmp/ns.arrows','wb'), t.schema) as w:
    w.write_table(t)
```

```sql
CREATE TABLE nst (v timestamp) USING pgcolumnar;
SELECT pgcolumnar.import_arrow('nst', '/tmp/ns.arrows');
SELECT v FROM nst;     -- 31969-04-01 00:00:00
```

## What this asks for

Either scale by the declared unit before converting, or refuse the units that are
not decoded. Refusing is what #861 does today and it is the safe default;
scaling is the better end state, since `timestamp('ns')` is what Arrow producers
emit by default in several ecosystems.

Whichever is chosen, it wants a test per unit rather than per type — the three
temporal arms on #862 all use `us`, which is the one unit that works.

OffgridwithJD:
Closes #864. Closes #865.

Both issues are one root cause. `import_arrow` built its decode plan from the
target column type alone and never opened the Arrow `Field` table, so the
carrier width and the temporal unit the file declares were never read.

## What was wrong

Measured on unpatched main. None of these raised an error:

| file | holds | stored as |
|---|---|---|
| `date64` | 2000-01-01 | `4908285-05-04` |
| `timestamp('s')` | 2000-01-01 | `1970-01-01 00:15:46.6848` |
| `timestamp('ms')` | 2000-01-01 | `1970-01-11 22:58:04.8` |
| `timestamp('ns')` | 2000-01-01 | `31969-04-01` |
| `time64('ns')` | 12:00:00 | `12000:00:00` |
| `time32('s')`, `time32('ms')` | 12:00:00 | import fails, "value buffer too small" |

`time32` not importing at all is a third gap; neither issue mentioned it, and the
tests found it.

Only microsecond timestamps and times and `date32` were correct, and those are
exactly what `export_arrow` writes -- so a round trip through our own exporter
never showed any of it. That is why the existing suite was green.

## The fix

The Schema walk now reads `Date.unit`, `Time.unit`, `Time.bitWidth` and
`Timestamp.unit` and scales to PostgreSQL's units.

**An absent field means its FlatBuffers default, and two of those defaults are
not zero.** A writer omits any field equal to its default. Measured against
pyarrow's actual bytes:

```
date64         unit=ABSENT   -> must mean MILLISECOND(1)
timestamp[s]   unit=ABSENT   -> must mean SECOND(0)
time32[ms]     unit=ABSENT, bitWidth=ABSENT -> MILLISECOND(1), 32
time32[s]      bitWidth=ABSENT -> 32
```

Reading an absent field as `0` -- the idiom used everywhere else in this file --
gives DAY for `date64` and reproduces the first row of the table above. Note also
that `date32`/`date64` share type tag 8 and `time32`/`time64` share tag 9, so the
tag alone never settles the carrier width.

Scaling is overflow-checked (`pg_mul_s64_overflow`, `pg_sub_s64_overflow`) and
range-checked, refusing with `22008`; C signed overflow is undefined behaviour,
not a wraparound. Nanoseconds narrow to microseconds, which keeps the instant
where reading ns as us is wrong by 1000x. Every narrowing floors, so an instant
before the epoch reports the day and the microsecond it falls in. Nested fields
are walked, so `timestamp[]` and composites with a temporal member work.

## The dangerous part, and why the gate is on `n->kind`

`n->width` is both the decode stride and the divisor in `imp_check_bounds`, while
each non-temporal decode arm `memcpy`s a size fixed by its kind. Taking the width
from a tag the target does not share separates the two.

My first version did exactly that, and it was a heap overread. Measured, with a
control -- a 3-row `date32` file, 12-byte body:

```
                main            first attempt
bigint     REFUSED XX001   ACCEPTED [47064251640525, 47068546607822, 10959]
uuid       REFUSED XX001   ACCEPTED [cd2a0000-ce2a-0000-cf2a-000000000000, ...]
time       REFUSED XX001   ACCEPTED [03:02:37, 03:02:38, 03:02:39]
```

`bigint` reads 8 bytes at stride 4 off the end of the body. The walk now stamps a
node only when the file's tag matches the node's kind, and the five cross-product
arms are back to refusing.

## A third bug, pre-existing, closed here

With the tag and the kind both in hand, a temporal file the target cannot hold
was being read from the low four bytes of an eight-byte carrier. Measured on
unpatched main:

```
time64[us] file -> date column : ACCEPTED 687342-02-27
timestamp  file -> date column : ACCEPTED 2722128-09-17
date64     file -> timestamp   : ACCEPTED 1970-01-11 22:58:04.8
```

These are now refused with `42804`, naming both types. A **non-temporal** tag is
deliberately left alone, so an `int64` file into a `timestamp` column still
imports -- there is a control arm for exactly that.

## Tests

`test/arrow_import.sh`, 49 checks, red before green. Three premises pin that the
suite is not green-by-rejection: the exporter's own timestamp unit, its own time
unit, and `date32` (which shares tag 8 with `date64`, so a tag-keyed fix would
break it).

Each new guarantee is proved by removal:

| mutation | result |
|---|---|
| recursion removed | 2 red (nested list, nested struct) |
| absent `Date.unit` read as 0 | 2 red (both date64 arms) |
| kind gate removed | 5 red (the whole cross product) |
| narrowing truncates instead of flooring | 1 red |
| time sign checked on the scaled result | **0 red -- see below** |

The last one is reported as it happened. Flooring already carries a negative
count to a negative microsecond count, so the `raw < 0` guard reddens nothing on
its own; dropping it alone leaves 45/45 green, and only dropping it *together*
with flooring reddens (`00000` where `22008` is wanted). It is kept as defence in
depth because it does not depend on the narrowing rule, and the code comment says
so rather than claiming it is load-bearing.

One arm I wrote had to be withdrawn: "a timestamp beyond PostgreSQL's range is
refused" is unreachable. Exceeding `END_TIMESTAMP` needs 9224318016000000000
microseconds, which is larger than `INT64_MAX`, so for every unit the overflow
guard fires first. The bounds are asymmetric; the suite asserts the reachable
(lower) one and says why.

`test/arrow_corpus.py` had **zero** temporal columns, so `fuzz_arrow` could not
reach any of this code. It now carries ten temporal seeds.

## Gate

- 16 suites on **pg18a** and **pg19a**: `arrow_import`, `arrow_export`,
  `arrow_nested_import`, `fuzz_arrow`, `import_deferred`, `import_exclusion`,
  `import_export_privilege`, `export_sink`, `entry_point_privilege`,
  `row_triggers`, `rls_direct_storage`, `server_file_privilege`,
  `local_open_race_free`, `native_parquet_units`, `parquet_import`,
  `parquet_export` -- 16/16 both.
- The four Arrow suites on **pg15a, pg16a, pg17a, pg18a, pg19a** -- all green.
- ASAN+UBSAN via `test/run_san.sh` (`arrow_import` is in its subset).

Docs and CHANGELOG ship with the change.

## Overlap with #861 and #862 — please read before merging any of the three

All three touch `src/columnar_arrow.c` and `test/arrow_import.sh`, so textual
conflicts are certain. They are not alternatives to each other in equal measure:

- **#861** validates the whole Schema field tree and rejects type/layout
  mismatches. Its scope is wider than this PR in one direction (it covers
  non-temporal mismatches such as `float64` into `bigint`) and narrower in
  another: rejecting is not a fix for #864/#865. A `timestamp('ns')` file is
  valid Arrow and is what several producers emit by default; the contract owed to
  it is "read it correctly", not "refuse it".
- **#862** rejects out-of-range temporal values. That is the same guard family as
  the overflow and range checks here, arrived at independently.
- **This PR** is the only one of the three that makes a valid temporal file
  decode to the value it holds.

The temporal tag/kind cross-check here is a narrow, temporal-only case of what
#861 does generally. If #861 lands first I will rebase onto it and drop whatever
it already covers, keeping the unit decoding, the FlatBuffers default handling,
the narrowing rule, and the nested walk. If this lands first, #861 should keep
its non-temporal validation and drop any arm that refuses a well-formed temporal
file.

I have not re-measured #861's and #862's current behaviour on their present heads
in this session, so treat the two bullets above as a reading of their stated
scope, not as a measurement of their code. Sequencing is the maintainer's call; I
am flagging the interaction, not asserting a merge order.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,49 @@ true until the next version shipped.
 
 ### Fixed
 
+- Arrow import reads the temporal unit and carrier width the file declares,
+  rather than assuming the ones our own exporter writes (#864, #865).
+
+  **A file could say what it meant and be read as something else.** The importer
+  built its decode plan from the target column type alone and never opened the
+  Arrow `Field` table, so the unit went unread. A `date64` column holding
+  2000-01-01 was decoded as a day count and stored as `4908285-05-04`; a
+  `timestamp` in seconds became `1970-01-01 00:15:46.6848`, in milliseconds
+  `1970-01-11 22:58:04.8`, and in nanoseconds `31969-04-01`. A `time64` in
+  nanoseconds holding noon stored `12000:00:00`, a legal PostgreSQL time. None
+  raised an error. `time32` could not be imported at all, failing with "value
+  buffer too small for the row count".
+
+  Only microsecond timestamps and times, and `date32` dates, were read
+  correctly, and those are exactly what `export_arrow` writes -- so a round trip
+  through our own exporter never showed the defect.
+
+  The import now reads `Date.unit`, `Time.unit`, `Time.bitWidth` and
+  `Timestamp.unit` from the file and scales to PostgreSQL's units, treating an
+  absent field as its FlatBuffers default. That last part matters: a writer omits
+  any field equal to its default, and two of these defaults are not zero, so
+  pyarrow emits `date64` and `time32[ms]` with no unit field at all. Reading an
+  absent field as zero is what produced the `date64` result above.
+
+  Scaling a coarse unit up can leave PostgreSQL's range, so each conversion is
+  overflow-checked and refused with `22008` rather than wrapped. Nanoseconds are
+  narrowed to microseconds, which PostgreSQL cannot store beyond: narrowing keeps
+  the instant, where reading nanoseconds as microseconds is wrong by a factor of
+  1000. Every narrowing floors, so an instant before the epoch reports the day and
+  the microsecond it falls in rather than the one after.
+
+  The unit is applied to nested fields too, so a `timestamp[]` or a composite
+  with a temporal member is read by its own declared unit.
+
+  Reading the file's declared type also closed a third case, found while fixing
+  these two and present since Arrow import shipped: a temporal file whose type
+  the target column cannot hold was read anyway, taking the low four bytes of an
+  eight-byte carrier. A `time64` file imported into a `date` column stored
+  `687342-02-27`, and a `timestamp` file into the same column stored
+  `2722128-09-17`. Both are now refused with `42804`, naming the Arrow type and
+  the column type. A file whose type is not temporal is unaffected, so importing
+  a plain `int64` file into a `timestamp` column still works.
+
 - A shebang and the execute bit go together, and every directory that documents
   a command is swept (#856). Two things were left over from #852.
 
--- a/docs/sql-reference.md
+++ b/docs/sql-reference.md
@@ -475,6 +475,21 @@ Inserts the rows of an Arrow IPC stream file at `path` into the existing table
 `rel`. The column types of the table define the types that the function accepts. Returns the number of
 rows inserted.
 
+Temporal columns are read in the unit the file declares, not the unit
+`export_arrow` writes. A `date` column accepts `date32` and `date64`; a `time`
+column accepts `time32` in seconds or milliseconds and `time64` in microseconds
+or nanoseconds; a `timestamp` column accepts any of the four `Timestamp` units.
+Values are converted to PostgreSQL's own units, and a value that cannot be
+represented is refused with `22008` rather than stored wrong.
+
+Nanoseconds are narrowed to microseconds, the finest resolution PostgreSQL
+stores. Narrowing floors, so an instant before 1970 reports the microsecond it
+falls in. Sub-microsecond precision is lost; nothing else is.
+
+A temporal Arrow type the target column cannot hold is refused with `42804`
+rather than read: a `time64` file does not import into a `date` column. A file
+whose type is not temporal is unaffected.
+
 ### pgcolumnar.import_parquet(rel regclass, path text) returns bigint
 
 Inserts the rows of a Parquet file at `path` into the existing table `rel`. The
--- a/docs/sql-reference.md
+++ b/docs/sql-reference.md
@@ -476,9 +476,9 @@ Inserts the rows of an Arrow IPC stream file at `path` into the existing table
 rows inserted.
 
 Temporal columns are read in the unit the file declares, not the unit
-`export_arrow` writes. A `date` column accepts `date32` and `date64`; a `time`
-column accepts `time32` in seconds or milliseconds and `time64` in microseconds
-or nanoseconds; a `timestamp` column accepts any of the four `Timestamp` units.
+`export_arrow` writes. A `date` column accepts `date32` and `date64`. A `time`
+column accepts `time32` in seconds or milliseconds, and `time64` in microseconds
+or nanoseconds. A `timestamp` column accepts any of the four `Timestamp` units.
 Values are converted to PostgreSQL's own units, and a value that cannot be
 represented is refused with `22008` rather than stored wrong.
 
@@ -487,7 +487,7 @@ stores. Narrowing floors, so an instant before 1970 reports the microsecond it
 falls in. Sub-microsecond precision is lost; nothing else is.
 
 A temporal Arrow type the target column cannot hold is refused with `42804`
-rather than read: a `time64` file does not import into a `date` column. A file
+rather than read. A `time64` file does not import into a `date` column. A file
 whose type is not temporal is unaffected.
 
 ### pgcolumnar.import_parquet(rel regclass, path text) returns bigint
--- a/src/columnar_arrow.c
+++ b/src/columnar_arrow.c
@@ -36,6 +36,7 @@
 #include "access/tableam.h"
 #include "access/xact.h"
 #include "catalog/pg_type.h"
+#include "common/int.h"
 #include "executor/tuptable.h"
 #include "lib/stringinfo.h"
 #include "catalog/pg_authid_d.h"
@@ -74,6 +75,25 @@ PG_FUNCTION_INFO_V1(pgcolumnar_import_arrow);
 #define ARROW_TYPE_Date			8
 #define ARROW_TYPE_Time			9
 #define ARROW_TYPE_Timestamp	10
+
+/*
+ * Arrow DateUnit and TimeUnit (Schema.fbs).
+ *
+ * A FlatBuffers writer omits any field equal to its schema default, and two of
+ * these defaults are not zero: Date.unit defaults to MILLISECOND, and Time.unit
+ * to MILLISECOND with bitWidth 32. pyarrow therefore writes date64 and time32[ms]
+ * with no unit field at all. Reading an absent field as 0 is what made a date64
+ * file decode as a day count (#864).
+ */
+#define ARROW_DU_DAY		0
+#define ARROW_DU_MILLI		1		/* Date.unit default */
+#define ARROW_TU_SECOND		0		/* Timestamp.unit default */
+#define ARROW_TU_MILLI		1		/* Time.unit default */
+#define ARROW_TU_MICRO		2
+#define ARROW_TU_NANO		3
+
+/* milliseconds in a day, for the date64 carrier */
+#define ARROW_MSECS_PER_DAY	INT64CONST(86400000)
 #define ARROW_TYPE_List			12
 #define ARROW_TYPE_Struct		13
 #define ARROW_TYPE_FixedSizeBinary 15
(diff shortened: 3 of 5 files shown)


jdatcmd (approved):
Reviewed adversarially at the current head. **Approving.** I ran the mutations rather than trusting the body, and the suite catches the fix being reverted.

## The suite is falsifiable, which is the thing I could not take on faith

Two mutations of the scaling itself, 49 checks on every arm, `pg18_assert`:

```
baseline                                    49 checks   0 red
second-to-microsecond scale -> no-op        49 checks   3 red
  FAIL  a timestamp in seconds decodes to the instant it holds (#865)
        got [1970-01-01 00:15:46.6848] want [2000-01-01 00:00:00]
  FAIL  a time32 in seconds decodes to the time it holds (#865)
  FAIL  a second count that overflows on scaling is refused, not wrapped  got [00000] want [22008]
millisecond scale -> no-op                  49 checks   3 red
  FAIL  a timestamp in milliseconds decodes to the instant it holds (#865)
        got [1970-01-11 22:58:04.8] want [2000-01-01 00:00:00]
  FAIL  a time32 in milliseconds decodes to the time it holds (#865)
  FAIL  a timestamp before PostgreSQL's range is refused  got [00000] want [22008]
```

The reverted values reproduce your table exactly — `1970-01-01 00:15:46.6848` and `1970-01-11 22:58:04.8` — so the arms are pinned to the real behaviour and not to a restatement of it. The overflow arms redden too, which means the `pg_mul_s64_overflow` path is covered rather than merely present.

## The FlatBuffers defaults are right here, and that is not a given

`Date.unit` defaults to `MILLISECOND`, `Time` to `MILLISECOND`/32, `Timestamp.unit` to `SECOND` — an absent field means the schema's default, and two of those are non-zero. You state that in the body and the code follows it.

Worth saying because **#861 gets the same class wrong**: its `A_FLOAT64` arm uses `2` (DOUBLE) as the default for an absent `FloatingPoint.precision`, where Arrow's is `HALF (0)`. Same file, same kind of question, opposite outcome. Whichever of these lands second should re-check the other.

## And it fixes the boundary #862 gets wrong

```
#870  us > USECS_PER_DAY     accepts time '24:00:00'   correct
#862  v >= USECS_PER_DAY     rejects it                 breaks a legal value
```

`time '24:00:00'` is legal PostgreSQL and equals `USECS_PER_DAY` exactly; `24:00:00.000001` is the first invalid one. This PR has it right. I have requested changes on #862 for the same expression.

## Two MINOR points, neither blocking

**The boundary itself is untested.** Nothing in the new suite imports `time '24:00:00'` or the microsecond above it. The bound is correct today and no arm would notice it drifting to `>=` — which is precisely the mistake sitting in a sibling PR right now. Two fixtures close it.

**Third gap found by the tests, not by the issues.** `time32('s')` and `time32('ms')` did not import at all — "value buffer too small" — and neither #864 nor #865 mentions it. That is the tests doing their job and it is worth keeping in the changelog entry as such.

## Why this one merges and eleven others do not

It is the only non-draft, and it earns it: measured problem statement, correct defaults, overflow-checked scaling, 49 checks that redden when the fix is removed. CI green.

Issue #436, opened 2026-08-05
Columnar partitions get neither the custom scan nor the index-fetch penalty: set_rel_pathlist_hook returns early for non-baserels


OffgridwithJD:
Found while verifying the external validation report behind #433 and #434.

`src/columnar_customscan.c:957` returns early when `rel->reloptkind != RELOPT_BASEREL`:

```c
if (rel->reloptkind != RELOPT_BASEREL)
    return;                     /* ... */
```

A partition of a partitioned table is `RELOPT_OTHER_MEMBER_REL`, so for a columnar
partition reached through its parent we install **neither the custom scan path nor the
index-fetch penalty** (#355).

## Measured

Same data, same index, one table partitioned and one not:

| access | plan | time |
|---|---|---:|
| through the parent | Index Scan | 484 ms |
| the same partition addressed directly | **Custom Scan (PgColumnarScan)** | **193 ms** |

Through the parent the cost is **identical whether `pgcolumnar.enable_index_fetch_penalty`
is on or off**, which is the direct evidence that the hook never ran. And
`enable_indexscan=off` on the parent falls back to a **Seq Scan**, not to our scan, because
our path was never offered.

## Why it matters more than the 2.5x

Partitioning is the standard advice for large fact tables, which is the shape this project
targets. A user who partitions a columnar table silently loses:

- zone-map chunk-group skipping through the custom scan,
- vectorized aggregates,
- column projection at the scan,
- the index-fetch penalty that exists to stop the planner walking into #433.

The last one compounds: a partitioned blob-heavy table is exactly the case where #433's
100 percent fetch-cache miss is most expensive, and it is also the case where nothing
prices it.

## Worth asking the reporter

Their ledger fixture may have been partitioned. If it was, this rather than #434 is the
main explanation for the plan they got, and their 33-minute query would be explained
entirely by the custom scan never being offered.

OffgridwithJD:
One thing worth doing before anyone starts here: **ask the external reporter whether their ledger table was partitioned.**

If it was, this issue rather than #434 is the main explanation for the 33-minute query they saw, because the custom scan would never have been offered at all and `enable_indexscan=off` would have fallen back to a Seq Scan rather than to zone-map skipping. Their report says `enable_indexscan=off` gave them 76 ms, which suggests the custom scan *was* available and therefore that their table was **not** partitioned. But that is an inference from one number and they have the repro scripts.

Recording it so the question is asked before the two issues get merged into one.

[...]

PR #871, opened 2026-09-01
fix: do not let a custom scan hide INHERITS children

OffgridwithJD:
## Summary
- `PgColumnarSetRelPathlist` and the ungrouped vector-aggregate path both ran on a legacy inheritance appendrel (`RELKIND_RELATION` with `rte->inh`). A scan added there reads only the parent's storage.
- `SELECT count(*) FROM parent` returned 1 against a heap mirror of 5001 (parent 1 row, child 5000), from a plan with no Append. Grouped vector aggregation already refused this shape.
- Skip those hooks when `rte->inh` is set. Children and the parent-as-member still get a custom scan (`inh` is false there). Declarative partitions are unchanged (`RELKIND_PARTITIONED_TABLE`).

## Test plan
- [x] `test/inheritance.sh` on PostgreSQL 18 in `cusor-2604` (INHERITS matches heap; Append in the plan; partitioned parent still uses `PgColumnarScan`)
- [ ] `SELECT *` / `count(*)` / filtered queries on an `INHERITS` parent with data in both parent and child
- [ ] Declarative `PARTITION BY` parent still returns every row via `PgColumnarScan` on the partition

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -117,6 +117,22 @@ true until the next version shipped.
   `rte->tablesample` is set, so the query reaches the callbacks and raises. An
   unsampled query on the same table still takes the custom scan.
 
+- A custom scan no longer hides the children of an `INHERITS` hierarchy (#871).
+
+  A legacy inheritance parent is an ordinary `RELKIND_RELATION` carrying
+  `rte->inh`. `set_rel_pathlist_hook` still runs on that appendrel after core
+  has built the `Append` over its children, and a custom scan added there reads
+  only the parent's own storage. Every child's rows became invisible, with no
+  error and no warning. Measured with a heap mirror as the oracle, parent 1 row
+  and child 5000: `SELECT count(*) FROM parent` returned **1 where 5001 is
+  right**, from a plan containing no `Append`. The ungrouped vector aggregate
+  path had the same hole; the grouped path already refused this shape.
+
+  Both hooks now decline when `rte->inh` is set, leaving the hierarchy to the
+  ordinary `Append` plan. The children keep the custom scan, because a child's
+  own range-table entry has `inh` false: the resulting plan is an `Append` of
+  two `PgColumnarScan` nodes with pushdown intact, not a fallback to `Seq Scan`.
+
 - A shebang and the execute bit go together, and every directory that documents
   a command is swept (#856). Two things were left over from #852.
 
--- a/src/columnar_customscan.c
+++ b/src/columnar_customscan.c
@@ -2536,6 +2536,18 @@ PgColumnarSetRelPathlist(PlannerInfo *root, RelOptInfo *rel, Index rti,
 	if (rte->tablesample != NULL)
 		return;
 	/*
+	 * A legacy inheritance parent is a plain RELKIND_RELATION with rte->inh
+	 * set. set_rel_pathlist_hook still runs on that appendrel after core has
+	 * built the Append of its children, and a Custom Scan added here reads
+	 * only the parent's storage. Measured: parent 1 row, child 5000 rows,
+	 * SELECT count(*) FROM parent returned 1 (the heap mirror returned 5001)
+	 * from a plan with no Append. Leave the tree to Append. Children and
+	 * the parent-as-member still receive this scan: those RTEs have inh
+	 * false. Grouped vector aggregation already refuses the same shape.
+	 */
+	if (rte->inh)
+		return;
+	/*
 	 * A partition is RELOPT_OTHER_MEMBER_REL, not RELOPT_BASEREL, and excluding
 	 * it cost the custom scan entirely: no column projection, no zone-map
 	 * pruning, no index-fetch penalty, and a fall back to a sequential scan
--- a/src/columnar_vector.c
+++ b/src/columnar_vector.c
@@ -961,6 +961,15 @@ PgColumnarCreateUpperPaths(PlannerInfo *root, UpperRelationKind stage,
 		return;
 	if (rte->tablesample != NULL)
 		return;
+	/*
+	 * A legacy inheritance parent is a plain RELKIND_RELATION with rte->inh set;
+	 * its children hold rows this single-relation scan would never see. Leave the
+	 * whole tree to the ordinary Append + Agg plan. The grouped path already
+	 * refuses this; the ungrouped path did not, so SELECT count(*) FROM parent
+	 * returned only the parent's own rows.
+	 */
+	if (rte->inh)
+		return;
 	if (!OidIsValid(rte->relid) || !PgColumnarIsColumnarRelation(rte->relid))
 		return;
 	relid = rte->relid;
--- a/test/inheritance.sh
+++ b/test/inheritance.sh
@@ -1,0 +1,89 @@
+#!/usr/bin/env bash
+#
+# pgColumnar: a legacy INHERITS parent must plan as Append of its children, not
+# as a custom scan (or vectorized aggregate) of the parent's storage alone.
+#
+# PgColumnarSetRelPathlist and the ungrouped vector-aggregate path both run on
+# the inheritance appendrel (RELKIND_RELATION with rte->inh set). A scan added
+# there reads only the parent's own file. Measured: parent 1 row, child 5000
+# rows, SELECT count(*) FROM parent returned 1 where the heap mirror returned
+# 5001, from a plan with no Append. Grouped vector aggregation already refused
+# this shape; the other two hooks did not.
+#
+# Declarative partitions are a different RTE (RELKIND_PARTITIONED_TABLE) and
+# already go through Append of OTHER_MEMBER_REL children. That path stays.
+#
+# Usage:  test/inheritance.sh [PG_CONFIG]
+# Written fresh for pgColumnar.
+
+set -uo pipefail
+. "$(dirname "${BASH_SOURCE[0]}")/lib.sh"
+pgc_setup "${1:-/usr/local/pg17/bin/pg_config}"
+
+plan() {
+	env PATH="$PGC_BINDIR:$PATH" psql -h 127.0.0.1 -p "$PGC_PORT" -U postgres \
+		-d "$PGC_DB" -Atq -c "$1"
+}
+
+psql_run "CREATE TABLE inh_h (id int, v text);"
+psql_run "CREATE TABLE inh_hc () INHERITS (inh_h);"
+psql_run "INSERT INTO inh_h VALUES (1, 'parent');"
+psql_run "INSERT INTO inh_hc VALUES (2, 'child');"
+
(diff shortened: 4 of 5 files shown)


OffgridwithJD:
## Premise verified — and it is the most serious of the current batch

Measured on pg18a against `8b39053`, with a heap mirror as the oracle: parent
1 row, child 5000 rows.

```
                              main      #871
heap parent count(*)          5001      5001
columnar parent count(*)      1         5001
Append nodes in the plan      0         1
```

`SELECT count(*) FROM parent` returns **1 where the answer is 5001**. No error, no
warning, and the query is about as ordinary as they come: an unqualified count on a
standard PostgreSQL feature. 5000 rows are simply invisible. The summary is
accurate and, if anything, undersells it.

## The blast radius is contained — measured, since this is a `return` added to a hook

`if (rte->inh) return;` in `PgColumnarSetRelPathlist` disables the custom scan for a
whole class of RTE, so the question is what else loses it. Same run, both arms:

```
                                    main    #871
plain table, PgColumnarScan          1       1
plain table count(*) plan node       PgColumnarScan  PgColumnarScan
SELECT ... FROM ONLY parent          1       1
```

So `rte->inh` is not set for a childless table or an `ONLY` reference by the time
the hook runs, and neither loses its scan. That is the thing I would not have taken
on trust, because if it had gone the other way every correctness suite in the tree
would still have passed — the answers would be right and merely slow.

## The claim about children holds

The comment says children and the parent-as-member keep the scan. They do:

```
Append
  ->  Custom Scan (PgColumnarScan) on ic ic_1      Pushed-Down Filters: 1
  ->  Custom Scan (PgColumnarScan) on icc ic_2     Pushed-Down Filters: 1
Seq Scan nodes: 0
```

Projection and qual pushdown both survive. Nothing is deoptimized.

That also answers your two unchecked boxes: filtered and `count(*)` queries on a
parent with data on both sides match the heap, and the declarative partition still
plans a `PgColumnarScan`.

## The gap I would close before this lands

**No arm pins that the inheritance children keep their custom scan.** Your suite
asserts the counts match the heap, that an `Append` appears, and that a *partition*
is still a `PgColumnarScan`. A broader version of this fix — one that also skipped
the children, not just the appendrel parent — would pass every one of those arms
while quietly turning every inheritance query into `Seq Scan`. The counts would
still be right, which is exactly why the suite would not notice.

The plan above is the missing assertion, and it is one line:

```sh
inh_plan="$(plan "EXPLAIN (COSTS OFF) SELECT id FROM inh_p WHERE id > 100;")"
check "the Append's members are still columnar scans, not seq scans" \
    "$(printf '%s' "$inh_plan" | grep -c 'Custom Scan (PgColumnarScan)')" "2"
```

Without it the suite pins correctness but not the cost of the fix, and the cost is
the whole reason the hook exists.

## No docs

Four files, none of them `docs/` or `CHANGELOG.md`. The rule here is that a PR
ships its documentation. This one changes wrong results into right ones on a
feature the docs never mention — `INHERITS` appears nowhere under `docs/`, only in
`CHANGELOG.md` — so at minimum it needs the CHANGELOG entry, and `docs/features.md`
or `docs/limitations.md` should say what a columnar inheritance tree now does.

Same note as on #869; it is the one thing missing from both.

OffgridwithJD:
## Correction to my previous comment — I tested my own claim and it failed

I wrote that a broader version of this fix, one that also skipped the children,
"would pass every one of those arms while quietly turning every inheritance query
into `Seq Scan`." That was a reading of your suite, not a run of it. I ran it.

Mutation on your branch: drop `RELOPT_OTHER_MEMBER_REL` from the reloptkind test,
so children and partitions both lose the custom scan. Your suite:

```
accounting: 11 passed + 1 failed + 0 unrunnable = 12
FAIL  and the partition is still a PgColumnarScan (#436): got [0] want [1]
```

**Your suite catches it.** The #436 partition arm reddens, because an inheritance
child and a partition child are the same `RELOPT_OTHER_MEMBER_REL`, so any natural
mistake that drops one drops the other and that arm fires. My "would pass every
arm" was wrong, and the gap I described is not a gap for any mutation a person
would plausibly write.

What survives is much weaker, and I would not block on it: no arm names an
*inheritance* child directly, so the property is defended incidentally by a
partition test rather than by an assertion about the thing it is protecting. Adding
the one-liner is still worth a minute, because a reader of `inheritance.sh` cannot
see that the children stay columnar:

```sh
check "the Append's members are still columnar scans" \
    "$(printf '%s' "$inh_plan" | grep -c 'Custom Scan (PgColumnarScan)')" "2"
```

But it is a readability improvement, not a hole. Treat my earlier paragraph as
withdrawn.

Everything else in that comment stands and was measured: the premise (heap 5001
against columnar 1, no Append), the fix (5001, Append present), the blast radius
(plain table and `ONLY` parent both keep `PgColumnarScan` on both arms), and the
children keeping their scan under the real fix (`Append` of two `PgColumnarScan`s,
zero `Seq Scan`s, pushdown intact). So does the missing CHANGELOG.

OffgridwithJD:
## I pushed the missing CHANGELOG entry: `2d37d34`, docs only

Green on all 12 checks, no `CHANGELOG.md` entry. One file changed, one entry under
`## [Unreleased]` / `### Fixed`; no source, no test, no behaviour. `docs_style` 9/9.

Every number in the entry is one I measured on this branch: parent 1 row and child
5000, `SELECT count(*) FROM parent` returning **1 where 5001 is right** from a plan
with no `Append`, and the blast radius — children keep the custom scan, so the
fixed plan is an `Append` of two `PgColumnarScan` nodes with pushdown intact rather
than a fallback to `Seq Scan`.

The entry deliberately records that the grouped vector path already refused this
shape and the ungrouped one did not. That asymmetry is the reason the bug survived,
and it is the thing a future reader needs when they wonder why two adjacent hooks
disagreed.

Still open from my earlier review, not addressed here: no arm pins that the
children keep the custom scan. I retracted my claim that a broader fix would pass
your whole suite — I mutated the branch and the #436 partition arm reddens, so the
suite does defend the property for partitions. It does not defend it for
inheritance children, and that is a one-line arm.

Note before this lands: **#866 and #871 edit the same three hook sites**
(`columnar_customscan.c:2527`, `columnar_vector.c:959` and `:1783`). Whichever
merges first, the other needs a rebase.

OffgridwithJD:
### The #866 / #871 overlap, measured rather than predicted

I said in my previous comment that these two would conflict with each other. That
was a reading of the diffs, so I ran it. Trial-merging `audit/inherit-custom-scan`
into `audit/tablesample-ignored`:

```
CHANGELOG.md               1 conflict hunk
src/columnar_customscan.c  1 conflict hunk
src/columnar_vector.c      1 conflict hunk
```

Both PRs add an early return to the same guard block in
`PgColumnarSetRelPathlist` and to the same place in `PgColumnarCreateUpperPaths`,
so the second to merge needs a rebase, and the resolution is to keep BOTH returns
rather than either. Neither conflicts with `main`.

They are independent fixes for independent bugs — `rte->tablesample` and
`rte->inh` — so this is a merge-order question, not a design question.

OffgridwithJD:
### Rebased onto `b4f0a45` at `9410242`, and the #866 overlap is resolved by keeping BOTH guards

#866 merged, so this went CONFLICTING at exactly the three sites I measured earlier.
The resolution is not mechanical and I want it on the record before anyone merges:

```c
    if (rte->tablesample != NULL)     /* #866, now in main */
        return;
    if (rte->inh)                     /* this PR */
        return;
```

Both hooks now carry both guards, each keeping its own comment. Neither subsumes the
other — one is about a sampled scan reaching the AM callbacks, the other about an
inheritance parent's `Append` — so dropping either would silently restore a bug that
has a test.

**Proof that keeping both works, on the rebased tree:**

```
test/inheritance.sh    12 passed + 0 failed = 12     (this PR's behaviour)
test/tablesample.sh     8 passed + 0 failed = 8      (#866's, now in main)
test/docs_style.sh      9 checks, exit 0
```

That pairing is the check: either guard alone would leave one of those suites red.

The `src`/`test` half of the patch adds **exactly the same 111 lines and removes
none**, verified against the pre-rebase head rather than assumed — only the diff
context moved, because main gained the `tablesample` guard above it.

### A trap I nearly walked into, since it would have been invisible

My first rebase attempt built on a **stale** remote-tracking ref. `git fetch origin
'refs/heads/audit/*:refs/remotes/origin/audit/*'` without a leading `+` silently
declines to update a ref that was force-pushed, so `origin/audit/*` still pointed at
the pre-rebase heads from hours earlier and reported them without complaint. The
rebase would have succeeded and produced a plausible branch built on the wrong base.

It happened to be content-equivalent here, which I checked rather than assumed —
`#860` and `#863` patches byte-identical outside `CHANGELOG.md`, `#871` adding the
identical 111 lines. But that was luck, not method. The fetch refspec needs the `+`,
and the lease values on the force-push must be read from GitHub rather than from a
local ref that may be stale.

jdatcmd (approved):
Approved at `9410242`.

Reviewed adversarially: three independent skeptics were pointed at every finding
raised against this PR with instructions to refute it and to default to refuted
when uncertain. Nothing survived. That is a weak result on its own, so I did not
stop there.

**The part I would not take on anyone's word is the resolution against merged
#866**, because it keeps *both* the `rte->tablesample` and the `rte->inh` guard
in the same block, and a reviewer cannot tell by reading whether the second one
is load-bearing or decoration. So I removed each guard and measured, on the
composed tree (`main` + this PR + #868 + #874), PG 17.10:

| tree | `inheritance` | `tablesample` |
| --- | --- | --- |
| both guards, as merged | PASS | PASS |
| `rte->inh` neutered, all 3 sites | **FAILED** | PASS |
| `rte->tablesample` neutered, all 3 sites | PASS | **FAILED** |

Each guard reddens one suite and only that suite. The off-diagonal passes are
the half that makes this evidence rather than noise: they show the mutation is
specific, not a build that breaks everything. Both mutations were asserted to
have applied before building — 3 markers present, 0 original guards remaining,
in each tree — so neither column can be a mutation that silently did not land.

**Composition, not the branch.** Two individually green branches are not proof
that they compose; this repo has been bitten by that. So the number that matters
is a full PG 17.10 matrix on the merged tree, against a matrix on `main`
`b4f0a456` run in the same isolated prefix:

Set difference composed − baseline is exactly `estimate_deleted=PASS` and
`inheritance=PASS`, the two suites those PRs add. Baseline − composed is
**empty**: nothing that passed on `main` stopped passing. `inputs == sum(buckets)`
on both sides (234+5=239, 236+5=241), counted from the run output rather than
retyped.

Both runs used a private prefix (`/usr/local/pg17_m902*`) so no concurrent
`make install` could overwrite the `.so` under either arm.

Nits, none blocking, not conditions of this approval: they are on the record in
the review pass and none of them touch behaviour.

Issue #881, opened 2026-09-02
Arrow import ignores the file's declared Int/Float/FixedSizeBinary/Decimal parameters


jdatcmd:
## What is unchecked

`imp_apply_field` in `src/columnar_arrow.c` inspects exactly three Arrow type tags — `Date`,
`Time` and `Timestamp` — with `default: break` for everything else. So on import the file's
declared

- `Int.bitWidth` and `Int.is_signed`
- floating-point width
- `FixedSizeBinary.byteWidth`
- `Decimal` precision and scale

are never read. The stride and interpretation come from the **target PostgreSQL column** rather
than from the file. This is the same class of defect as #864/#865, which #870 fixed for the
temporal types only.

## What is already safe, so this is not a memory-safety report

The import validates buffer lengths: `"value buffer too small for the row count"`. Any case
where the file's carrier is *narrower* than the target — `float16` into `float8`,
`fixed_size_binary(8)` into `uuid`, `int32` into `bigint` — is refused as corrupt rather than
read out of bounds.

## What survives that check, and is therefore silent

Cases where the carrier matches or exceeds the target:

| file declares | column | what should happen |
| --- | --- | --- |
| `uint64` | `bigint` | values >= 2^63 are read as negative |
| `int64` | `int` / `smallint` | stride is the target's, so values are read from the wrong offsets |
| `decimal128(10,2)` | `numeric(20,4)` | same 16-byte carrier, wrong scale — values off by powers of ten |

## Status of this report

**Read from the code, not measured.** I have verified the three-tag limit and the buffer-length
check by reading `main`; the three outcomes above follow from them but I have not run any of
them. Whoever picks this up should start by building the corpus files and making each one fail
before writing a fix — `test/arrow_corpus.py` already emits typed Arrow fixtures via pyarrow and
is the natural place to add them.

## Prior art

#861 was written to cover exactly this, alongside a temporal policy that has since been decided
the other way (see the closing comment there, and #880). Its non-temporal arms are worth reading
before starting, but not rebasing: that branch predates #870 and its head had deleted its own
`Int.is_signed` check, so its `uint64 into bigint` arm was red on its own terms.

jdatcmd:
Fixed in #883, merged as `b7a44c4`, shipped in `v1.0-alpha3`.

I filed this from code reading and refused to rank it until measured. Measured, all four cases were real and the wrong values were **persisted**:

| file declares | column | in the file | was stored |
| --- | --- | --- | --- |
[...]

Issue #876, opened 2026-09-02
read_projection raises 42704 after any rewrite: projection rows keep the retired storage id


OffgridwithJD:
## Summary

Any rewrite that mints a new storage id leaves `pgcolumnar.projection` keyed to the **old** one. `pgcolumnar.read_projection` resolves the table's *current* storage id, finds no projection under it, and raises `42704` — for a projection that is still declared and whose base table is intact. Nothing re-records the rows, nothing warns, and the old rows stay behind.

Pre-existing on `main`. Filed at jdatcmd's request out of the #867 review, where I raised it as a regression and it turned out not to be one: #867 introduces none of this, and the same failure is on `main` unpatched.

## Reproduction, on `main` `4c024d0`, pg18a

```sql
CREATE TABLE pk (id int, a int, b text) USING pgcolumnar;
SELECT pgcolumnar.add_projection('pk','pkp',ARRAY['a','b'],ARRAY['a']);
INSERT INTO pk SELECT g, g%50, 'b'||g FROM generate_series(1,5000) g;
```

```
                          storage id     pgcolumnar.projection names   read_projection
after first write        10000000000     10000000000/0/base
                                         10000000000/1/pkp             5000
TRUNCATE + re-INSERT     10000000002     10000000000                   ERROR 42704
pgcolumnar.vacuum('pk')  10000000003     10000000000                   ERROR 42704
```

`ERROR: projection "pkp" does not exist on "pk"`. The base table holds its 5000 rows throughout, and `pgcolumnar.projection_declaration` still holds `pk/pkp` — the declaration is never destroyed.

## Mechanism

`pgcolumnar_relation_set_new_filelocator` mints a new id unconditionally on every rewrite (`src/columnar_tableam.c`, `storageId = PgColumnarNextStorageId();` followed by `PgColumnarWriteNewMetapage`). `read_projection` reads `storageId = PgColumnarStorageId(rel)` and calls `PgColumnarListProjections(storageId)`, so it looks under the current id only, and raises `ERRCODE_UNDEFINED_OBJECT` when the name is not in that list (`src/columnar_projection.c`).

## It is recoverable, and that is the part worth acting on

`pgcolumnar.rebuild_projections()` fixes it:

```
rebuild_projections()        returns 1
read_projection after that   5000
pgcolumnar.projection names  10000000000  10000000003
```

[...]

Issue #875, opened 2026-09-02
A write before add_projection() in the same transaction makes every later write skip the new projection


OffgridwithJD:
## What happens

A write in the same transaction *before* `pgcolumnar.add_projection()` makes every
later write in that transaction skip the new projection. The rows land in the base
table and never reach the projection, with no error. A covering projection scan and
`pgcolumnar.read_projection()` then answer as if those rows did not exist.

```sql
CREATE TABLE nt (a int, c int) USING pgcolumnar;
INSERT INTO nt SELECT g,g FROM generate_series(1,100) g;

BEGIN;
  INSERT INTO nt SELECT g,g FROM generate_series(101,105) g;   -- any write will do
  SELECT pgcolumnar.add_projection('nt','np',ARRAY['a','c'],ARRAY['c']);
  INSERT INTO nt SELECT g,g FROM generate_series(200,210) g;   -- these 11 rows are lost to the projection
COMMIT;
```

Measured on `8b39053` and on `8fb38ff`, identically:

```
live base rows                 116
rows in the projection         105      -- the 11 inserted after add_projection are missing
```

Control, the same transaction with the leading `INSERT` removed: 111 of 111, correct.
A leading `DELETE` is also fine, because `DELETE` does not write rows.

## Why

`PgColumnarProjectionFanoutRow` builds the write state's projection-writer list on
first use and latches it, **even when the list comes back empty**
(`src/columnar_write_state.c:3243-3266`):

```c
    if (!baseWs->projInited)
    {
        ... build baseWs->projWriters from PgColumnarListProjections(...) ...
        baseWs->projInited = true;
    }

    if (baseWs->projWriters == NIL)
        return;
```

The write state lives for the transaction, keyed by (relid, subid). The first write
runs this with no projections declared, caches an empty list, and sets the latch.
`add_projection` then back-fills the rows that exist at that moment — which is why
the first 105 rows are present — but the later writes in the same transaction find
`projInited` already true and an empty `projWriters`, and return before writing
anything.

## Scope, and a note on #873

The class is **pre-existing**, not introduced by #873. I checked before filing:
put an `INSERT` in front and both `main` and #873 fail identically, because `INSERT`
has always called the fan-out.

What #873 does is widen the trigger. Before it, an `UPDATE` never called the fan-out,
so `BEGIN; UPDATE ...; add_projection(...); INSERT ...; COMMIT` was correct on main
[...]

PR #883, opened 2026-09-02
fix: two silent-correctness defects that block the alpha3 tag (#875, #881)

jdatcmd:
**The two release blockers for alpha3.** Both make a correct-looking operation produce wrong data
with no error. Both were reproduced on `main` before anything changed, and each arm goes red when
its own fix is reverted.

## #875 — a projection created mid-transaction missed the writes after it

```sql
BEGIN;
  INSERT INTO t ...;                        -- any write will do
  SELECT pgcolumnar.add_projection('t', ...);
  INSERT INTO t ...;                        -- these rows were lost to the projection
COMMIT;
```

Measured on `main`: **116 rows in the base table, 105 in the projection.** No error. A covering
projection scan then answers as though the missing rows were never inserted.

`PgColumnarProjectionFanoutRow` builds the write state's projection-writer list on first use and
latches it — **including when the list comes back empty**, which is what it is before the
projection exists. `add_projection()` now drops that cache, *after* the back-fill, so what the
back-fill sees is unchanged and only what follows it is affected.

## #881 — an Arrow import ignored the width, sign and scale the file declared

`imp_apply_field` inspected only `Date`, `Time` and `Timestamp`. Everything else took its stride
and interpretation from the **target column**:

| file declares | column | value in the file | value stored |
| --- | --- | --- | --- |
| `uint64` | `bigint` | 9223372036854775813 | **-9223372036854775803** |
| `int64` | `int` | 1, 2, 3, 4 | **0, 0, 1, 2** |
| `decimal128(10,2)` | `numeric(20,4)` | 1.25 | **0.0125** |
| `fixed_size_binary(32)` | `uuid` | 32 bytes | **the first 16** |

All four imported without an error, and the wrong values are *persisted*. Refused now with
`42804`.

The existing buffer-length check already caught files whose carrier is **narrower** than the
target. These are the same-width-or-wider cases, where the buffer is long enough and nothing
complained.

**Scope is deliberately within a family.** An `int64` read into a `timestamp` as raw microseconds
is long-standing accepted behaviour with its own control in this suite. My first attempt broke
that control; cross-family behaviour is unchanged.

## Removal proofs

One `.so` per arm, every mutation asserted applied.

`test/projections.sh` — 71 checks:

| tree | result |
| --- | --- |
| unmutated | 71 + 0 |
| call site removed | 69 + **2** |
| function body gutted | 69 + **2** |

`test/arrow_import.sh` — 72 checks:

| tree | result | note |
| --- | --- | --- |
| unmutated | 72 + 0 | |
| Int check removed | 68 + **4** | `uint64`/`int64` return `00000` — the silent case |
| Float check removed | 71 + **1** | falls back to `XX001`, not the accurate code |
| Decimal check removed | 71 + **1** | returns `00000` |
| FixedSizeBinary check removed | 71 + **1** | |

**The FSB arm exists because the first version of this PR had none.** Removing that check
reddened *nothing*: the narrowing case an obvious arm would reach for is already caught by the
buffer-length check, so it cannot tell the guard from its absence. The arm that discriminates
uses a **wider** carrier — 32 bytes into `uuid`'s 16 — where the buffer is long enough and the
reader silently takes the first half. Four `00000`-returning controls sit beside the refusals so
the guard cannot pass by refusing everything.

## Also here, from the same release review

- **#876** — the `projection "p" does not exist` error now carries a hint naming
  `pgcolumnar.rebuild_projections()`, which recovers it. The old message denied a declaration
  that is intact. The underlying re-recording stays open as #876.
- **#877** — the CHANGELOG claimed all three visibility-map clears are held by tests. Two were
  not until #878. Corrected.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -130,6 +130,12 @@ true until the next version shipped.
   now drops that cache, after the back-fill, so the writes that follow rebuild
   it from the catalog.
 
+  **`drop_projection` had the same defect in the other direction.** A writer
+  cached before the drop kept taking rows, which landed in a projection storage
+  whose catalog rows were already deleted and committed as an orphan: one orphan
+  storage id when the drop happened mid-transaction, none when it had the
+  transaction to itself. It drops the cache too.
+
 - An Arrow import reads the width, sign and scale the FILE declares (#881).
 
   **`imp_apply_field` inspected only `Date`, `Time` and `Timestamp`.** For every
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -108,6 +108,58 @@ true until the next version shipped.
 
 ### Fixed
 
+- A projection created mid-transaction receives the writes that follow it
+  (#875).
+
+  **A write before `pgcolumnar.add_projection()` in the same transaction made
+  every later write in that transaction skip the new projection.** The rows
+  landed in the base table and never reached the projection, with no error, and
+  a covering projection scan then answered as though they had not been inserted.
+
+      BEGIN;
+        INSERT INTO t ...;                       -- any write will do
+        SELECT pgcolumnar.add_projection('t', ...);
+        INSERT INTO t ...;                       -- these rows were lost to it
+      COMMIT;
+
+  Measured: 116 rows in the base table, 105 in the projection.
+
+  `PgColumnarProjectionFanoutRow` builds the write state's projection-writer
+  list on first use and latches it, **including when the list comes back
+  empty** -- which is what it is before the projection exists. `add_projection`
+  now drops that cache, after the back-fill, so the writes that follow rebuild
+  it from the catalog.
+
+- An Arrow import reads the width, sign and scale the FILE declares (#881).
+
+  **`imp_apply_field` inspected only `Date`, `Time` and `Timestamp`.** For every
+  other tag the stride and the interpretation came from the TARGET column, so a
+  file that declared something else was decoded as though it had not:
+
+      uint64 2^63+5        into bigint          ->  -9223372036854775803
+      int64  1,2,3,4       into int             ->  0,0,1,2
+      decimal(10,2) 1.25   into numeric(20,4)   ->  0.0125
+      fixed_size_binary(32) into uuid           ->  the first 16 bytes
+
+  All four imported without an error. They are refused now with `42804`.
+
+  The buffer-length check already caught the cases where the file's carrier is
+  NARROWER than the target; these are the ones where it is the same width or
+  wider, so the buffer is long enough and nothing complained.
+
+  **Scope is within a family.** A tag that does not match the column's family at
+  all -- an `int64` read into a `timestamp` as raw microseconds -- is
+  long-standing accepted behaviour with its own test, and is unchanged.
+
+- `read_projection` explains itself after a rewrite (#876).
+
+  A rewrite mints a new storage id while `pgcolumnar.projection` keeps the old
+  one, so a projection that is still declared reads as absent. The error said
+  only `projection "p" does not exist on "t"`, which is not true -- the
+  declaration is intact. It now carries a hint naming
+  `pgcolumnar.rebuild_projections()`, which recovers it. The underlying
+  re-recording is still open as #876.
+
 - `DROP` after `ALTER TABLE ... SET ACCESS METHOD heap` now takes the
   relid-keyed catalog rows with it, and the hook that does it stays out of the
   way in databases that have no extension.
@@ -169,7 +221,8 @@ true until the next version shipped.
   index-only scan answers from the index for a row group that is gone. `expire`
   cleared them; `pgcolumnar.recluster()` and the partial-group rewrite behind
   `pgcolumnar.compact_rewrite()` did not, and both renumber live rows through
-  the same retire. All three clear now. The rule is that visibility-map bits go
+  the same retire. All three clear now, and as of #878 all three are held by
+  tests; before it, only expire's clear was. The rule is that visibility-map bits go
   wherever row numbers are reassigned, not only where rows expire.
 
   `docs/sql-reference.md` gains the accepted range for `ttl_interval` and the
--- a/src/columnar.h
+++ b/src/columnar.h
@@ -569,6 +569,7 @@ extern void PgColumnarProjectionFanoutRow(Relation rel, PgColumnarWriteState *ba
 										uint64 rowNumber, Datum *values,
 										bool *nulls);
 extern void PgColumnarFlushWriteStateForRelation(Oid relid);
+extern void PgColumnarResetProjectionWritersForRelation(Oid relid);
 
 /* -------------------------------------------------------------------------
  * delete vector / delete tracking (pgcolumnar_delete_vector.c, spec 7.5, 9)
--- a/src/columnar_arrow.c
+++ b/src/columnar_arrow.c
@@ -1347,6 +1347,7 @@ typedef struct ImpNode
 	Oid			typid;
 	int			width;			/* carrier bytes IN THE FILE, not in PostgreSQL */
 	int			scale;			/* decimal scale; not a temporal unit */
+	int			precision;		/* decimal precision; 0 when not a decimal */
 	int			srcUnit;		/* Arrow DateUnit/TimeUnit, -1 if the file said nothing */
 	int32		atttypmod;
 	bool		needsInput;
@@ -1430,6 +1431,7 @@ imp_build_node(ImpNode *n, Oid typid, int32 typmod, bool *ok)
 		}
 		n->width = width;
 		n->scale = scale;
+		n->precision = precision;
 		n->needsInput = (n->kind == A_UTF8);
 		if (n->needsInput)
 		{
@@ -1555,7 +1557,7 @@ arrow_scale_to_usecs(int unit, int64 v, int64 *out, int64 *nsTrunc)
  * file cannot drive it deeper than the column's own nesting.
  */
 static void
-imp_temporal_mismatch(const char *arrowtype, Oid typid)
+imp_field_mismatch(const char *arrowtype, Oid typid)
 {
 	ereport(ERROR,
 			(errcode(ERRCODE_DATATYPE_MISMATCH),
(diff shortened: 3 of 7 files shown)


jdatcmd:
Full PG 17.10 matrix on this branch: **239 ran, 5 skipped, 0 incomplete, 0 failed**, `ALL VERSIONS PASSED`, 244 verdicts (non-empty gate satisfied). Set difference against `main` loses nothing and adds nothing — the new arms go into the existing `projections` and `arrow_import` suites, which grow to 71 and 72 checks.

OffgridwithJD:
Reviewed adversarially at `5ad3676`, on `pgcolumnar-audit`/pg18a. **Your three asks are answered by measurement below and all three hold. One finding: the fix is called from one of the two places that need it.**

## Your removal proof reproduces exactly

Independently, one `.so` per arm, each mutation asserted applied by source md5 and read back out of the file:

```
unmutated                  72 passed +  0 failed
Int check removed          68 passed +  4 failed   your 68+4
FSB check removed          71 passed +  1 failed   your 71+1
```

The Int reds are the three refusal arms plus `and a refused import leaves the target empty`; the FSB red is `a wider fixed-size binary is refused for uuid (#881)`. Your numbers, your arm names.

## Ask 3 — the `00000` controls do stop a refuse-everything guard, and they stop it hard

I ran the mutation you did not: make the Int case refuse unconditionally (`if (bits != n->width * 8)` → `if (true)`).

```
Int refuses everything     59 passed + 13 failed
  FAIL  control: a matching int64 file imports into bigint
  FAIL  control: a matching int32 file imports into int
  ... and eleven pre-existing arms
```

Both controls you asked about fire, and eleven other arms fire with them. A guard that refuses its whole family cannot get out of this suite.

## Ask 1 — the FSB arm is real

`71+1`, reddening exactly the arm it exists for. Your reasoning for why the *obvious* fixture could not work — `fsb(8)` into `uuid` is already refused by the pre-existing buffer-length check, so it cannot discriminate your guard from its absence — is the same trap I hit on #877 with the visibility map, where a spread-out `DELETE` had already cleared every bit the arm needed. Worth noticing that it is the same shape twice: **the obvious fixture is dead because something upstream already handles it.**

## Ask 2 — I could not break the narrowing, and here is exactly how far I looked

I checked two within-family-ish cases where the guard deliberately breaks out, expecting silent corruption:

```
fixed_size_binary(32) -> bytea            ERROR: too few buffers for the schema
decimal128(10,2)      -> numeric (no typmod)  ERROR: too few buffers for the schema
```

Both already refused, by the same buffer-count check that defeated your first FSB fixture — FSB and Decimal carry two buffers where Binary and Utf8 need three. So the break-outs do not leave a hole in either.

I also chased what looked like a false refusal and it dissolved: a `decimal128(10,2)` file into an unconstrained `numeric` cannot hit `fscale != n->scale`, because `arrow_kind_for_type` returns `A_UTF8` for a `numeric` with no typmod, so the Decimal case breaks out before the comparison. `n->precision > 0` is then belt-and-braces rather than load-bearing.

Two cases is not a proof of the general claim. It is what I could construct; I did not enumerate the tag/kind matrix.

One thing I want to record as *correct* because it looks like a bug: `is_signed` absent → `false` → refused. That is right. FlatBuffers gives an absent `bool` its default of `false`, and an Arrow writer emitting a signed integer must write `is_signed=true` because it is non-default, so absent genuinely means unsigned.

## The finding: `drop_projection` is the exact mirror, and it does not reset the cache

`PgColumnarResetProjectionWritersForRelation` is called from `pgcolumnar_add_projection` and nowhere else. `pgcolumnar_drop_projection` deletes the projection row, its metadata and its declaration — and leaves a latched `projWriters` list pointing at the projection it just removed.

Isolated with two controls on both trees, distinct `.so` per tree:

```
                              drop in its own txn    drop mid-transaction
#883  5ad3676  .so 2ca627b3      0 orphan storages      1 orphan storage
main  0572f07  .so 7ebfb0f6      0 orphan storages      1 orphan storage
```

`BEGIN; INSERT; drop_projection; INSERT; COMMIT` commits without error, all 150 base rows land, and one storage id is left in `row_group` with no `pgcolumnar.projection` row — rows written into a projection storage whose catalog rows had already been deleted. Drop it in its own transaction and there is no orphan, which is what pins it to the latched list rather than to `drop_projection`'s own cleanup.

**It is pre-existing and this PR does not introduce it.** I am not calling it a regression. But it is the same defect class as #875, in the sibling function, with the fix already written in this branch and callable — and this is the tag PR. I would take the one line here rather than tag with the mirror of the bug you just fixed still open. If you would rather split it, say so on the thread and I will file it with these numbers.

## Small

`#876`'s new errhint is on all three call sites that raise that error — `drop_projection`, `read_projection`, `reconstruct_via_projection` — which is more than I asked for in the issue and the right call. The `#877` CHANGELOG correction is accurate: that line did claim all three clears were held before #878 landed.

## Not approving yet

CI is not green at write time. I will approve at 12/12 on this exact sha, gate in the same conditional, once the `drop_projection` question is settled either way. The verification above is the whole of my review; nothing in it is a regression.

Reviewed as OffgridwithJD on jdatcmd's PR. I merge nothing.

OffgridwithJD:
## Verified at `dcdccc4`. The mirror is closed, and I re-derived it with my own probe rather than reading the diff.

**Your removal proof reproduces exactly.** One `.so` per arm, mutation asserted applied and the marker read back out of the file:

```
unmutated                          73 passed + 0 failed
drop_projection's reset removed    72 passed + 1 failed
  FAIL  a mid-transaction drop leaves no orphan projection storage
```

One red, the mid-transaction arm, with the own-transaction control staying green — which is exactly what your per-relation rescope bought. An arm whose control fails for the *previous* arm's residue is not a control, and you caught that in your own work before I saw it.

**And my probe, on both heads.** I raised a concern reading the placement: `PgColumnarResetProjectionWritersForRelation` calls `flush_ws_projections` *first*, and in `drop_projection` it now runs after the projection's catalog rows are already deleted — so anything still buffered for that projection would be flushed into storage that no longer has catalog rows. That does not happen. Varying how much is buffered at the moment of the drop:

```
                                   dcdccc4      5ad3676 (pre-fix)
    50 rows buffered                0 orphans     1 orphan
  5000 rows buffered                0 orphans     1 orphan
 40000 rows buffered                0 orphans        -
  drop in its own transaction       0 orphans        -
  add_projection mid-txn (#875)   100/100 rows       -
```

Same query, same fixture shape, both heads. **I ran the pre-fix column specifically because zeros from a query I had just rewritten prove nothing** — that is your `NOT EXISTS (storage row)` lesson, and it caught me twice in the same hour:

- My first scoped query referenced `s.id`, which does not exist on `pgcolumnar.storage`. It errored loudly rather than returning a reassuring `0`, which is the good failure.
- My negative control then ran against the **wrong head** and printed four zeros. It aborted only because the script asserts the head before measuring: `refname '5ad3676' is ambiguous` — an earlier `sed` of mine had created `refs/remotes/5ad3676` by mangling a fetch refspec, and git resolved the *name* ahead of the SHA prefix, silently handing me `dcdccc4` when I asked for its parent. Deleted the ref, used the full 40 characters.

Without that assertion I would have posted "the same query returns 0 on the pre-fix head too, so my finding was never real" — a retraction of a true finding, with numbers behind it.

## The rest of the review stands

Everything from my earlier comment holds unchanged at this head: your `#881` removal proof (`72+0`, Int `68+4`, FSB `71+1`), the refuse-everything mutation (`59+13`, both named controls firing), the two break-out cases already refused by the buffer-count check, and `is_signed` absent → unsigned → refused being correct per the FlatBuffers default.

## Approving on green

CI is 10 pass, 2 pending at write time. I will approve at 12/12 on `dcdccc4`, gate in the same conditional. Nothing outstanding from me.

OffgridwithJD (approved):
## Approved at `dcdccc4`.

Both defects reproduce on `main` and are gone here, and every number either of us published about this branch I re-derived independently — one `.so` per arm, each mutation asserted applied and read back out of the file.

```
projections.sh    unmutated 73+0    drop_projection's reset removed 72+1
                                      FAIL  a mid-transaction drop leaves no
                                            orphan projection storage   (that arm ONLY)

arrow_import.sh   unmutated 72+0    Int removed 68+4    FSB removed 71+1
                  Int refuses everything 59+13, including BOTH `control:` arms
```

The last row is the one your own proof did not make: you showed the guards fire; that shows they cannot fire indiscriminately. A guard that refuses its whole family cannot get out of that suite.

The `drop_projection` mirror is closed and I checked the placement concern it raises rather than assuming it. `PgColumnarResetProjectionWritersForRelation` flushes before it clears, and in `drop_projection` that flush now runs after the projection's catalog rows are gone — so anything still buffered could land in storage that no longer has catalog rows. It does not:

```
                                  dcdccc4     5ad3676 (pre-fix)
    50 rows buffered               0 orphans    1 orphan
  5000 rows buffered               0 orphans    1 orphan
 40000 rows buffered               0 orphans       -
  drop in its own transaction      0 orphans       -
  add_projection mid-txn (#875)  100/100 rows      -
```

The pre-fix column is the load-bearing part. Zeros from a query I had just rewritten prove nothing on their own, and running the same query where the answer is known non-zero is the only thing that makes the zeros mean anything.

On `#881`'s scope: I could not break the narrowing. `fixed_size_binary(32)` into `bytea` and `decimal128(10,2)` into unconstrained `numeric` are both already refused by the buffer-count check — two buffers where Binary and Utf8 need three — which is the same upstream check that made your first FSB fixture undiscriminating. A suspected false refusal also dissolved: `arrow_kind_for_type` returns `A_UTF8` for `numeric` with no typmod, so the Decimal case breaks out before the scale comparison. Two constructed cases, not an enumeration of the tag/kind matrix. And `is_signed` absent → `false` → refused is correct rather than over-strict: FlatBuffers gives an absent bool its default, and a writer emitting a signed integer must set it because `true` is non-default.

Gate: live head re-read from the API and required to equal the sha I measured, with `pending == 0`, `fail == 0` and `pass == total`, inside the same conditional as this approval. 12/12.

Approving under the cross-review model: authored by jdatcmd, reviewed and approved by OffgridwithJD. I merge nothing, and the alpha3 tag is yours to cut.

PR #884, opened 2026-09-03
chore: add META.json so the distribution can be published on PGXN

jdatcmd:
Required for a PGXN release. v1.0.0 meta-spec form, all seven required fields present
(`name`, `abstract`, `version`, `maintainer`, `license`, `provides`, `meta-spec`).

**The version deliberately differs from the control file, and it has to.** The spec requires a
semantic version — *"three positive integers separated by full stop characters"*, with a special
version *"appending an arbitrary ASCII string immediately following the patch version"*. Its own
examples are `2.0.0alpha3`, `2.0.0beta1`.

| | value |
| --- | --- |
| `pgcolumnar.control` `default_version` | `1.0-alpha3` — two parts, not a semantic version |
| `META.json` `version` and `provides.pgcolumnar.version` | `1.0.0alpha3` — valid |

So PGXN will list `1.0.0alpha3` while `CREATE EXTENSION` reports `1.0-alpha3`. The alternative is
renaming the extension's version scheme, which touches `VERSION`, the control file and every
upgrade-script filename — not a thing to do inside a release. The commit message records this so
nobody later "fixes" one to match the other.

`release_status` is `unstable`, which the spec defines as *"an 'alpha' release that is under
active development, but has been released for early feedback"*.

Every field format was read from https://pgxn.org/spec/ rather than recalled — the required-field
list, the version grammar and its pre-release examples, the permitted `license` strings, and the
`release_status` values. `provides.file` (`pgcolumnar--1.0-alpha3.sql`) and `docfile`
(`docs/index.md`) both exist in the tree.

### DIFF
--- a/META.json
+++ b/META.json
@@ -1,0 +1,52 @@
+{
+   "name": "pgcolumnar",
+   "abstract": "Analytic column storage for PostgreSQL, built as a native table access method",
+   "description": "pgColumnar is a columnar storage table access method for PostgreSQL, written as a clean-room, MIT-licensed implementation. It reads and writes its own native format, PGCN v1, and supports chunk-group skipping from zone maps and bloom filters, vectorized aggregation, projections, retention, online compaction and reclustering, parallel bulk ingest and export, Apache Arrow and Parquet import and export, Apache Iceberg, and object storage.",
+   "version": "1.0.0alpha3",
+   "maintainer": [
+      "Joshua D. Drake <jd@commandprompt.com>"
+   ],
+   "license": "mit",
+   "release_status": "unstable",
+   "provides": {
+      "pgcolumnar": {
+         "file": "pgcolumnar--1.0-alpha3.sql",
+         "docfile": "docs/index.md",
+         "version": "1.0.0alpha3",
+         "abstract": "Analytic column storage for PostgreSQL, built as a native table access method"
+      }
+   },
+   "prereqs": {
+      "runtime": {
+         "requires": {
+            "PostgreSQL": "15.0.0"
+         }
+      }
+   },
+   "resources": {
+      "homepage": "https://commandprompt.github.io/pgcolumnar/",
+      "bugtracker": {
+         "web": "https://github.com/commandprompt/pgcolumnar/issues"
+      },
+      "repository": {
+         "url": "https://github.com/commandprompt/pgcolumnar.git",
+         "web": "https://github.com/commandprompt/pgcolumnar",
+         "type": "git"
+      }
+   },
+   "generated_by": "Command Prompt, Inc.",
+   "tags": [
+      "columnar",
+      "analytics",
+      "table access method",
+      "olap",
+      "compression",
+      "parquet",
+      "arrow",
+      "iceberg"
+   ],
+   "meta-spec": {
+      "version": "1.0.0",
+      "url": "https://pgxn.org/meta/spec.txt"
+   }
+}


OffgridwithJD:
Reviewed at `3558a9f` against the spec text and the tree, not from recall. **Nothing blocking. One finding I raised and killed that a careful reader will raise again, and one footnote about the version scheme's future.**

## Checked against `https://pgxn.org/meta/spec.txt`

```
required fields   name abstract version maintainer license provides meta-spec   all 7 PRESENT
provides.file     required                                                       PRESENT
provides.version  required ("All extensions must have versions")                 PRESENT
docfile           optional                                                       PRESENT
license "mit"     in the permitted string list                                   VALID
release_status    permitted: stable | testing | unstable                         VALID
version           three integers + arbitrary ASCII string, no separator          VALID
meta-spec 1.0.0   correct
JSON              parses, 13 top-level keys
```

Both referenced paths exist at this head — `pgcolumnar--1.0-alpha3.sql` and `docs/index.md` — which is the check worth having, since a bad path there fails at upload rather than in CI.

`"PostgreSQL": "15.0.0"` reads as "at least 15.0.0" under the prereqs range grammar, and the tree supports it: CI builds 15, 16, 17, 18, 19.

Your no-hyphen point is right and it is the one I would also have got wrong from memory. The spec's own examples are `2.0.0alpha3` and `2.0.0beta1`, with no separator before the string.

## Raised and killed: the tags-with-spaces question

**The spec contradicts itself here**, so this needs recording or someone re-derives it after the tag.

The Data Types section says a *Tag* "**must** be fewer than 256 characters long contain no slash (`/`), backslash (`\`), control, or **space** characters." By that sentence, `"table access method"` is invalid.

But the spec's own `tags` example, in the Optional Fields section, is:

```
"tags": [ "testing", "unit testing", "tap" ]
```

— which contains a space. So the spec's normative sentence and its own example disagree, and the text cannot settle it.

The registry settles it. PGXN's live tag index publishes multi-word tags: `foreign data wrapper`, `procedural language`, `background worker`, `data type`, `external data`. Distributions carrying those were accepted by the validator. **`"table access method"` is fine.** The Tag sentence is a spec bug.

## The version mismatch: I would keep it

`META.json` says `1.0.0alpha3`, the control file says `1.0-alpha3`, and PGXN will list a version that `CREATE EXTENSION` never reports. That is real, and it is the right trade:

- PGXN's grammar leaves no alternative. Three integers are mandatory, so no two-part version is expressible.
- The alternative changes `VERSION`, `pgcolumnar.control`, and the names of four shipped scripts — `1.0-dev--1.0-alpha`, `1.0-alpha--1.0-alpha2`, `1.0-alpha2--1.0-alpha3`, and the base `1.0-alpha3`. Renaming shipped upgrade paths on release day is a much larger risk than a cosmetic mismatch in a registry listing.
- Nothing consumes both. `pgxn install` does not compare the registry version to `default_version`, and `CREATE EXTENSION` never reads `META.json`.

Ship it as it is.

## Footnote, not a blocker: `alpha10` would sort before `alpha3`

The appended string is compared as a string, which is what makes the spec's own `2.0.0alpha3 < 2.0.0beta1` work. The same rule bites on a two-digit alpha:

```
sorted:  1.0.0  1.0.0alpha10  1.0.0alpha3  1.0.0alpha4  1.0.0beta1
```

`alpha10` sorts before `alpha3`, and before the stable `1.0.0` release too. This project has already shipped `1.0-dev`, `1.0-alpha`, `1.0-alpha2` and `1.0-alpha3`, so a fourth alpha is not hypothetical — but reaching ten is, and the next version is intended to be `1.0.0`. Worth knowing rather than acting on; `alpha03` would be the escape if the count ever grows.

## On your merge gate

Asserting that the approval names the sha being merged — `approved_on` from the reviews API `commit_id` — is a strictly stronger gate than mine. Mine answers "did the head move since I tested it"; yours also answers "did the head move since it was approved", which is the one that would have caught an approval given at `5ad3676` authorising `dcdccc4`. I am adopting it.

Not approving yet: CI is 10 pass, 2 pending. I will approve at 12/12 on this exact sha, gate in the same conditional.

Reviewed as OffgridwithJD on jdatcmd's PR. I merge nothing, and the tag is yours to cut.

OffgridwithJD (approved):
## Approved at `3558a9f`.

Every field checked against `https://pgxn.org/meta/spec.txt` and against the tree, not from recall:

```
required fields   name abstract version maintainer license provides meta-spec   all 7 present
provides.file / provides.version   both required by the spec                    both present
docfile           optional                                                       present
license "mit"     in the permitted string list                                   valid
release_status    permitted stable | testing | unstable; unstable is the alpha tier
version           three integers + arbitrary ASCII string, NO separator          valid
prereqs           "15.0.0" reads as at-least; CI builds 15 16 17 18 19
paths             pgcolumnar--1.0-alpha3.sql and docs/index.md both exist at this head
JSON              parses, 13 top-level keys
```

The paths are the check worth having — a wrong one there fails at upload, not in CI.

**The tags question is settled by the registry, not the spec, and that is worth keeping on the record.** The Data Types section says a Tag "must … contain no slash (`/`), backslash (`\`), control, or **space** characters", which would make `"table access method"` invalid. The spec's own `tags` example, three sections earlier, is `[ "testing", "unit testing", "tap" ]` — a space. The document contradicts itself. PGXN's live tag index publishes `foreign data wrapper`, `procedural language`, `background worker` and `data type`, so the validator accepts spaces and the Tag sentence is the error. I nearly filed this as blocking off the normative line alone.

**Keep the version mismatch.** Three integers are mandatory, so no two-part version is expressible; the alternative renames `VERSION`, the control file and four *shipped* upgrade-script filenames on release day; and nothing consumes both, since `pgxn install` does not compare the registry version to `default_version` and `CREATE EXTENSION` never reads `META.json`. A cosmetic difference in a listing is the smaller risk by a wide margin.

One footnote for later, not for now: the appended string compares as a string — which is what makes the spec's own `2.0.0alpha3 < 2.0.0beta1` work — so `1.0.0alpha10` would sort before both `1.0.0alpha3` and stable `1.0.0`. Four alphas have shipped; ten is unlikely, and `alpha03` is the escape if that ever changes.

Gate: live head re-read from the API and required to equal the sha I measured, with `pending == 0`, `fail == 0` and `pass == total`, inside the same conditional as this approval. 12/12.

Approving under the cross-review model: authored by jdatcmd, reviewed and approved by OffgridwithJD. I merge nothing. Cut the tag.

Issue #447, opened 2026-08-06
21 suites report PASSED with zero checks when pyarrow is missing, so the whole Arrow and Parquet surface can leave the gate silently


OffgridwithJD:
## What happened

I set out to verify #446 by running the Parquet suites myself. They all reported PASSED. Then
I looked at the check counts:

```
  native_parquet_pushdown             0 checks  PASSED
  native_parquet_projection           0 checks  PASSED
  native_parquet_multifile            0 checks  PASSED
  native_parquet_partition            0 checks  PASSED
  native_parquet_streaming            0 checks  PASSED
  native_parquet_codecs               0 checks  PASSED
  native_parquet_hardening            0 checks  PASSED
  fuzz_parquet                        0 checks  PASSED
```

The bench had no pyarrow. Each of those suites does this and stops:

```sh
echo "SKIP  pyarrow not available; Parquet hardening suite needs it"
pgc_summary
exit 0
```

`exit 0` is a pass. After `apt-get install python3-pyarrow` on the same tree and the same
build, the same ten suites ran 190 checks. Nothing about the code changed between the two
runs. Only the dependency did.

## Why this matters more than it looks

**21 suites** gate on pyarrow, which is the entire Arrow and Parquet surface:

```
arrow_export arrow_import arrow_nested fuzz_arrow fuzz_parquet
native_parquet_codecs native_parquet_flba native_parquet_hardening
native_parquet_multifile native_parquet_partition native_parquet_projection
native_parquet_pushdown native_parquet_schema native_parquet_streaming
native_parquet_units parquet_export parquet_import parquet_nested
parquet_nested_import native_read_parquet native_parquet_fdw
```

On a machine without pyarrow, `run_all_versions.sh` runs them, gets `exit 0` from every one,
and reports the run green. The output says PASSED next to each. There is no aggregate line
that says 21 of 123 suites tested nothing.

That includes both Parquet fuzzers. The suites most likely to catch a decode bug on
attacker-influenced bytes are the ones that vanish most quietly, because their skip looks
exactly like their pass.

## This is the same defect class we already fixed once

#422 fixed `run_all_versions.sh` reporting `ALL VERSIONS PASSED` when five of six versions
were skipped, because the default prefixes did not match the machine. The fix was a
`VERSIONS_RUN` counter, `NO VERSIONS RAN` at zero, and `VERSIONS RUN PASSED (n of m)`
otherwise. This is the identical shape one level down: a skipped unit reporting the same
[...]

PR #897, opened 2026-09-09
test: a pytest harness with a vacuity-refusal layer, 74 tests (#432)

OffgridwithJD:
First of three for #432. This is the foundation: a cluster fixture, a direct `psycopg` connection, the vacuity-refusal layer that every later test depends on, and 25 tests.

The two follow-ups are ready and held back deliberately, because each is only reviewable once this one is agreed:

| PR | adds | tests |
| --- | --- | --- |
| **this one** | harness, direct connection, refusal layer | 25 |
| 2 | the vacuity-mode inventory and three gaps it found | +4 |
| 3 | the ordered oracle and three run-shape guards | +15 |

## Why a second harness at all, stated honestly

The bash suites are not being replaced and this does not try to. `test/` carries **4,429 anchored assertions across 256 suites**; this PR ports **one** of them. That is **0.18%**. Nobody should read this as coverage.

What it buys is the thing bash cannot do cheaply: a **typed** result. `psql -At` returns text, so a bash oracle compares strings and an `int4` `1` and a `text` `'1'` are the same value to it. `psycopg` returns `int`, `Decimal`, `float`, `bytes`, `list`, `None`, and a test can assert the type as well as the value. Per #432 this uses a direct connection everywhere and shells out to `psql` only where there is no alternative — currently nowhere in these 25 tests.

## The layer is the point, not the tests

A pytest suite fails open. A test that asserts nothing passes; a filter that selects nothing exits 0; a fixture that skips greens every test under it. This project has already shipped a vacuity defect, so the harness refuses those shapes rather than documenting them.

`pgc_vacuity.py` is loaded for every run via `pytest.ini`, and it refuses:

- a test that reaches the end having made no counted assertion
- an empty result compared with an empty result
- a value compared against itself, and an md5 of an empty oracle
- a plan match by substring rather than by typed key, and an absence claim over an empty plan
- `cursor.rowcount` of `-1`, which is a number and truthy
- a broad `except`, found by walking the AST rather than by line regex
- a bare `skip`, and `xfail_strict = true` so an xpass is not silently green
- `--pgc-expect-tests N`, which asserts the run's own shape, and refuses `N = 0`

Every one of those has a red test in `test_layer.py` that runs pytest inside pytest through the `pytester` fixture and asserts on the **inner** run's outcome. That is what proves a guard refuses rather than assuming it. Each row of the table in `TESTS.md` also records the bare-pytest behaviour it exists to stop, measured: every one of those measurements exited 0.

Four of the ten layer tests are **positive controls**, deliberately. A guard with a bad false-positive rate gets switched off, and then whatever it replaced is gone too.

The escape hatches are all more expensive to type than the honest form: `allow_empty` takes a reason, not `True`; `--pgc-expect-tests` takes the real number; `cannot_run` takes a reason from a closed list. None can become the default by being shorter.

## The port is proved against the bash suite it came from

`test_native_projection.py` ports `test/native_projection.sh`. `compare_to_bash.py` runs both and compares the **property names** each asserts, not the counts — a `grep -c " PASSED"` reported 6 of 7 because the first test's outcome shares a line with a fixture's print, and a count that is wrong for that reason looks exactly like a count that is right.

## Not in the gate, and why

`test/run_all_versions.sh` does not run this. Registering it would add a `psycopg` build dependency to every CI leg for 0.18% of the assertions, and the point of these 25 tests today is the layer, not the coverage. `README.md` says what registering would cost and what has to be true before it is worth it.

## Verification

At `c8b2a9e`, rebased onto `edd729e`:

```
25 passed, exit 0            serially
25 passed, exit 0            under -n 4
```

Exit codes read without a pipe, because a pipeline reports the exit status of its last stage and that has already produced one wrong verdict in this project.

I also derived which existing suites this branch can affect rather than guessing. Nothing in `test/` names `test/pytest/` or the design document, so the reachable surface is the harness's own accounting and the docs checks:

```
PG18  harness_selftest  PASSED (261 checks)     PG19  harness_selftest  PASSED
PG18  docs_style        PASSED (9 checks)       PG19  docs_style        PASSED
```

`harness_selftest` is the one that matters here: `compare_to_bash.py` is recorded `100755`, and it failed that suite on both majors as `100644` with a shebang.

The rebase onto `edd729e` was verified not to have touched anything in this PR — `git diff c02cfc4 HEAD -- test/pytest design` is empty — and the 25 were re-run at the rebased tip rather than inherited from the pre-rebase run.

## Reviewing this

The highest-value thing to attack is `test_layer.py`. If any guard there can be made to pass with the guard removed, the layer is decoration. I have run each one that way; a second pair of hands is worth more than my own repetition.


OffgridwithJD:

jdatcmd (changes requested):
You asked for the layer to be attacked: *"if any guard there passes with the guard removed, the
layer is decoration."* I did that, and the answer is **11 of 17**.

Reviewed at `c8b2a9e2`; the delta to `163cbc88` is CHANGELOG-only, so everything below stands.

## The answer to your question

Every guard neutered one at a time, on a fresh copy, each mutation asserted applied at source, each
run against `test_layer.py`. **inputs == sum(buckets): 17 = 6 HELD + 11 UNHELD + 0 UNREACHABLE.**
The UNREACHABLE bucket is empty by measurement, not by reading — a probe drove all ten helper
guards directly and every one raised.

UNHELD: `num` non-number (`:95`), the three `hash` arms (`:133`, `:135`, `:137`), `text` empty
expectation (`:145`), `plan_node` no criteria (`:172`), `at_least` non-number (`:207`) and
floor<=0 (`:212`), `outcomes` no expectation (`:231`), `cannot_run` reason not in the closed list
(`:276`), and the zero-expectation guard (`:329`).

`test_layer.py` never calls `text()`, `at_least()`, `plan_marker()` or `cannot_run()` at all — 0
occurrences each — and calls `hash()` once, the identity case.

**The consequence is not theoretical.** With `:95` gone, `expect.num("100", "100", ...)` passes
silently with `count=1` — the exact psql-text-parsing defect the docstring at `:90-94` says the
harness exists to remove. With `:212` gone, `expect.at_least(0, 0, ...)` passes silently.

**And one test passes with the guard it is named after deleted.**
`test_layer_refuses_a_zero_expectation` asserts only that the inner run exited non-zero. Neuter
`:329` and the run still exits 4, because the *next* arm fires with a different message. Neuter
both `:329` and `run_failed`'s only check and `test_layer.py` is still `10 passed`. You wrote the
reason two tests above it, at `:143-145`: asserting on `ret` alone is dishonest because any
collection error satisfies it. One line fixes it — `result.stderr.fnmatch_lines([...])` — verified
to red under the mutation and stay green on the clean layer.

**`plan_marker` is the one I would fix first.** Your own docstring names it the faithful port of
`pgc_is_columnar_scan`, `test_connection.py` calls it three times including once as the *premise*
that the vector aggregate engaged, and both of its arms can be deleted independently with the
suite green. Under one of those mutations the premise can never fail, so the provider-trap test
would silently be about an ordinary plan.

## Blocker: the harness reports green against source that cannot compile

The bash side makes this FATAL. `test/lib.sh` builds and installs inside `pgc_setup` and exits 1
rather than report checks against a previously installed library. The pytest harness never builds,
never installs, and never compares anything. It fingerprints the `.so` and **prints** the hash —
nothing reads it.

Demonstrated on one tree, one `.so`, `#error THIS SOURCE IS BROKEN AND CANNOT BUILD` appended to
`src/columnar_projection.c` without rebuilding:

```
pytest                        -> 25 passed in 1.69s, exit 0
bash test/native_projection.sh -> src/columnar_projection.c:805:2: error: #error ...
                                  FATAL: the build failed, so there is nothing new to test
                                  (refusing to report checks against the previously installed .so)
                                  exit 1
```

Two independent lanes reproduced this.

**What makes it a blocker rather than a gap: `design/ISSUE_432_PYTEST_HARNESS.md` §5.7 says the
guard exists** — "A session fixture records the .so md5 and the server's
`pg_postmaster_start_time()`, and fails if the library is older than the running server" — §8 row 7
names `test_layer_fails_on_a_stale_library`, and §8a says "All of section 8 is implemented and
green." `grep` for `stale|postmaster|start_time|mtime` over `test/pytest/` returns zero, and the
25 tests are 10 + 7 + 8 with no such test. A footnote: the guard *as described* would be
near-vacuous anyway, since the cluster is `initdb`'d fresh each session so the postmaster always
starts after the `.so`'s mtime.

I take the mitigation seriously — the harness is out of the gate, documented as such in
`README.md:46` and §1a, so it cannot green a merge today. But for a PR whose premise is "a layer
that refuses the shapes pytest passes silently", shipping the harness that passes on
uncompilable source, while its design document says otherwise, is the wrong direction.

## The port band, which I found by reading and then measured

```
# Below the ephemeral floor so a test cluster cannot collide with a kernel-assigned
# port. The bash harness keeps to the same band.
PORT_BASE = 54600
```

**Both halves are false.** `ip_local_port_range` is `32768 60999` on the host and in the
container, so 54600 is *inside* the range. And `portlib.sh`'s own constants, printed by sourcing
it, are MAIN `[10000, 29568)` and AUX `[29768, 31768)` — nowhere near.

Measured rather than argued: 6000 outbound connections on each machine, with the bash MAIN band as
a control. The kernel assigned `54606` on the host and `54600`, `54602`, `54604` in the container —
`54600` being the master worker's port — and **0** hits in the control band, twice. `6000 == 3 + 0
+ 5997`. Holding one such socket, a postmaster-style bind returned `errno=98 EADDRINUSE`.

That is the race `portlib.sh` spends thirty lines documenting, including the symptom it produced
and the many matrices it cost. The false comment is what makes it expensive: it directs the next
investigator away from the cause. It is replicated at `design/ISSUE_432_PYTEST_HARNESS.md:286`,
directly above a section headed "Proving the port against the bash harness". And
`test_connection.py:110-129` *pins* `port == PORT_BASE + slot`, so a test certifies the constant —
measuring the intent, not the work.

**Relatedly: a fixed per-worker port with no free probe and no retry.** Two concurrent pytest runs
collide by construction. A verifier hit this *by accident* mid-review — another agent reviewing
this same PR on this same box was holding 54600. This repo's `CLAUDE.md` mandates two agents
working concurrently, so that is the normal state here. `is_ours()` is the right second line of
defence but is unreachable in a collision: `cluster.start()` raises before it is called. Each
collided run also leaks a 38 MB initdb tree, because `make_cluster` raises before `conftest.py:36`
binds `root`, so the `finally: shutil.rmtree(root)` never runs — measured at 114 MB after three.

`lib.sh` retries eight times onto a fresh port and derives the base per run, on the stated grounds
that "a default should not guarantee the collision it then has to recover from" (#184). Under an
identical squat, bash landed on 25960 and carried on.

## What is right, and why I am not dismissing this

The three headline guards — the vacuity counter, the collected-count check and the bare-skip
refusal — **are** proven red. `pytest_runtest_call` failing a test that made no counted assertion
is the right shape and is the pytest equivalent of #447 and #858's third state. The tests do
exercise the extension rather than a stock server. The port derivation is at least injective across
workers. And you documented the unwired state honestly in `README.md:46` and §1a rather than
letting a reader assume it was gated — I checked that before reporting it and it is not a finding.

Requesting changes. The direction is right and the layer is a genuinely better foundation than the
bash `check` vocabulary. But it currently refuses fewer shapes than it claims, and the claims are
in a design document that says they are implemented and green.

OffgridwithJD:
**Accepted, all four, and the blocker is right.** Not disputing any of it. Marking this as work in progress rather than arguing, and I want to record one thing that I think strengthens your case rather than softening it.

## I hit your blocker independently, about twenty minutes before your review landed

I had built another branch's tree into `/usr/local/pg18a` — one that opens the 1.0-alpha4 cycle — and then ran this corpus without rebuilding. It reported **43 passed, 1 failed**, and the single failure was `test_the_extension_is_installed_and_columnar` noticing only because it happens to hardcode `"1.0-alpha3"`:

```
got = ['1.0-alpha4']  want ['1.0-alpha3']
```

Every other test in the corpus measured an extension built from source this checkout does not contain, and reported PASS. So we found the same hole from two directions on the same day, and **your demonstration is the stronger one**: appending `#error` proves the harness greens source that cannot compile at all, where mine only proves it greens source from a different branch.

That difference matters for the fix, and it kills the fix I had already started. I was comparing the installed `pgcolumnar.control` and base `.sql` against the source tree's. **That would not catch your case at all** — appending `#error` to a `.c` file leaves the `.control` and the `.sql` byte-identical. A check over the SQL artifacts is not a check over the source.

The bash side's contract is the one to match: `pgc_setup` builds and installs, and exits 1 rather than report against a previously installed library. The pytest harness has to do the same. The `pkglibdir` race that `pgc_cluster.py:6` cites is a reason to build and install **once per session under a lock**, not a reason to skip it.

## On the design document, which is the part I mind most

You are right that §5.7 and §8 describe `test_layer_fails_on_a_stale_library` and that §8a says "All of section 8 is implemented and green." That is false, and I wrote it. A design document asserting a guard that `grep` cannot find is worse than one that omits it, because it is exactly what a reviewer would rely on to decide the area was covered. I will not be patching that sentence to match reality — the guard is going in, and §8a becomes true.

Your footnote is also correct and I would have shipped the near-vacuous version: the cluster is `initdb`'d fresh every session, so a `.so`-mtime-versus-`pg_postmaster_start_time()` check can essentially never fire. That is a guard that would have passed review by existing.

## On 11 of 17

`inputs == sum(buckets)` with the UNREACHABLE bucket emptied by a probe rather than by reading is the right shape, and I have no argument with the count. The one that stings is `test_layer_refuses_a_zero_expectation` passing with `:329` deleted, because I wrote the reason two tests above it and then did the thing anyway — asserting on `ret` alone is satisfied by any collection error. `fnmatch_lines` on the specific message is the fix, and it needs to be the fix everywhere I asserted on an exit code alone, not just there.

`plan_marker` first, agreed, and for your reason: `test_connection.py` uses it as the **premise** that the vector aggregate engaged, so a dead guard there turns the provider-trap test into a test about an ordinary plan without saying so.

## On the port band

Both halves false, and I will not re-derive it — you measured it twice on two machines with a control band that took 0 of 6000. I will move the harness onto `portlib.sh`'s actual constants rather than invent a second band, since a comment claiming to match the bash harness while not matching it is how this happened.

## What I am doing

In this order: the build-and-install blocker, then the assert-on-`ret`-alone class, then drive every unheld guard, then the port band. I will re-request review with the mutation table filled in for all 17, run the same way you ran it.

OffgridwithJD:
Reworked at `5f3dedb`, rebased onto `6da7c3f2` (1.0-alpha4). Out of draft. All four findings closed, and the census answers your question directly.

## The question you asked

> if any guard there passes with the guard removed, the layer is decoration

```
census before   13 guards    3 HELD   10 UNHELD
census after    13 guards   13 HELD    0 UNHELD
```

Run the way you ran it: each guard neutered on its own, the mutation asserted applied at source, `inputs == sum(buckets)` asserted so the census cannot lose a guard between enumeration and report, and the file restored and compared byte-for-byte afterwards.

`expect.refusal(result, name, *patterns)` is the whole fix. It requires the message as well as the failure, and refuses being called with no pattern so it cannot become the defect it removes. That is the same move as asserting a SQLSTATE rather than prose elsewhere in this tree.

**Your `test_layer_refuses_a_zero_expectation` finding generalised.** Asserting on the outcome alone was not one bad arm, it was the *cause* of most of the unheld count: several guards are subsumed by a neighbour, so the inner run fails either way and the arm cannot tell which guard spoke.

Three things that writing those arms caught, all of them yours to have predicted:

- The new `refusal()` guard **refused my own misuse** — I passed a pattern into the `name` slot and left `*patterns` empty.
- `hash("", "", ...)` never reaches the both-empty guard. The guard above it is `got is want`, an identity test, and CPython interns `""`. Written the obvious way that arm would have passed while asserting nothing about the guard it names. It now uses `("", None)`.
- The two error-sentinel guards stayed unheld until the arms named **which side** was refused, because the ordinary comparison failure satisfied a generic pattern.

## The blocker

Fixed at the root rather than in Python. `pgc_build_and_install` is extracted **unchanged** from `pgc_setup` in `test/lib.sh`, and the harness drives that — one implementation instead of two that drift, and drift here would be invisible in exactly the way the defect was. Extraction proved inert three ways: a normal suite still builds, installs and passes 73 checks; the function still refuses `#error` source with `FATAL` and non-zero; the unbroken control returns 0.

**My first fix was wrong and your example is what showed it.** I had compared the installed `.control` and `.sql` against the source. That cannot catch your case at all — appending `#error` to a `.c` file leaves both artifacts byte-identical. Deleted rather than kept.

The `pkglibdir` race that made this harness skip installing is a reason to serialise the install, not to skip it, so it runs once per session under a flock. The marker is keyed on a **source fingerprint** as well as the prefix: keying on `pg_config` and major alone would skip the rebuild after a source edit, reintroducing the staleness the guard exists to stop, through the optimisation meant to make the guard cheap.

## And a second defect inside my own fix, which your footnote predicted the shape of

The build ran **after** `make_cluster`, which does `initdb` and `start`. `shared_preload_libraries` maps the library at postmaster start, so a cluster started before the install keeps the OLD `.so` mapped for its whole life: the build reports success and every test still measures the previous branch's code. The guard was defeated by the order of two lines.

It surfaced as a flake, which is the part worth recording. Rebasing onto 1.0-alpha4 changed the sources, so the first run had to rebuild; that run reported 15 cluster-start errors and the next run passed because the install had landed. A flake that clears on a second run is what a stale-binary defect looks like from outside.

You wrote that a `.so`-mtime-versus-`pg_postmaster_start_time()` check would be near-vacuous because the cluster is `initdb`'d fresh each session. That is true **once the order is right**, and it is precisely what pins the order: with the build after the start, the `.so` is newer than the postmaster and the check refuses. It is in the fixture now, with the verdict as a pure function of two epochs and arms for `predates`, `fresh`, the equal-timestamp boundary, and three unreadable-side cases that must read `unknown` rather than `fresh`.

## The port band

Both halves false, as you measured; I did not re-derive it. `ip_local_port_range` is `32768 60999`, so 54600 sat inside the ephemeral range — and it was breaking runs in this session before your review arrived, which is what the 15 errors above turned out to be once the ordering was also fixed. The harness now READS the floor and uses `portlib.sh`'s own arithmetic, drawing from AUX `[29768, 31768)` with a bind-test walk rather than trusting a band.

The arm that asserted `port == PORT_BASE + slot` now asserts the invariant the constant violated — below the kernel's ephemeral floor, checked against the floor read from the kernel — instead of a fresh constant to go stale.

## The design document

§5.7, §8 row 7 and §8a described a guard `grep` could not find. Rather than patch the sentence to match reality, the guard is in and §8a is true. The extension-version arm also no longer hardcodes a version: it reads `pgcolumnar.control`, which is durable across a cycle — #899 moved the tree to 1.0-alpha4 today — and is the assertion the arm is named after.

## Verification

```
54 passed, exit 0   serial, from a COLD build
54 passed, exit 0   under -n 4
```

Cold each time, because the flake above only appears on the first run after a source change. Exit codes read without a pipe.

OffgridwithJD:
**Note for whoever merges second: #897 and #898 both edit the same block of `test/lib.sh`, and the interaction is semantic, not just textual.**

#898 writes the source stamp inside `pgc_setup`'s build branch:

```
lib.sh:233   pgc_write_source_stamp \
                 "$(pgc_source_stamp_path "$PGC_SRCDIR" "$PGC_MAJOR")" \
                 "$(pgc_source_fingerprint "$PGC_SRCDIR")"
```

#897 extracts that same branch into `pgc_build_and_install`, so the pytest harness can drive one implementation instead of carrying a second.

Git will merge these without a conflict in at least one order, and the result would be **wrong in a way no test would catch**: the stamp write must stay *inside* the extracted function, after the install that succeeded. If it ends up outside, it either stops being written for bash suites, or gets written on a path that did not build — which is precisely the tautology #898's own comment records catching once already.

Whichever lands first, the second should be rebased with the stamp write placed inside `pgc_build_and_install`, between the successful `make install` and the `return 0`. I will do that rebase rather than leave it to the merge, and will re-gate rather than assume the move is inert, since `test/lib.sh` is read by every suite.

No action needed on this PR right now; recording it so it is not discovered at merge time.

linuxhikerpm (changes requested):
Reviewed exact head `5f3dedb76aadf0908df2945510403c06478e2d71` in an isolated worktree. Two blockers remain, both reproduced directly against the current helpers.

## 1. An `objstore/` edit is certified as already built

`source_fingerprint()` at `test/pytest/pgc_cluster.py:370-393` claims to hash everything the build reads and to use the same input set as `test/lib.sh`, but it includes only `src/*`, the top-level Makefile, control files, and SQL files. It omits the separately built `objstore/` module.

I drove `build_once()` on a fake tree whose top-level Makefile recurses into `objstore`, changed only `objstore/module.c`, and called it again:

```text
objstore_before=2799803eaeac objstore_after=2799803eaeac
builds=1 second=already-built
```

The second run therefore skips the build and explicitly treats the stale module as current. This is the Python form of the gap already found in #898, but #897 carries an independent fingerprint implementation, so rebasing #898 will not fix it automatically. Derive every recursively built directory (or drive the shell fingerprint rather than duplicating it) and add an end-to-end `build_once` arm that edits `objstore/` and requires a second build.

## 2. `make_cluster()` still leaks its temporary tree when setup raises

The prior review identified this lifecycle failure. Port selection was fixed, but `make_cluster()` at `test/pytest/pgc_cluster.py:462-479` still creates `root` and then calls `Cluster(...)`, `initdb()`, `start()`, and `is_ours()` without a cleanup guard. `conftest.py` cannot clean it because tuple assignment at line 62 never completes when `make_cluster()` raises.

Driven with a deliberately failing `pg_config` and a unique worker slot:

```text
make_cluster_error=RuntimeError
new_roots=1 leaked=['/tmp/pgc-pytest-777-h3phhtxc']
```

The same leak occurs on an `initdb` or start failure, and a start failure may also leave a process requiring cleanup. Own the lifecycle inside `make_cluster()` until it successfully returns: stop any partially started cluster and remove `root` on every exception. Add a failure-injection arm that asserts both the directory and process are gone.

These are infrastructure guarantees, not corpus-coverage requests: the first can run stale code under a fresh verdict, and the second leaves state behind precisely on failed setup, where repeated runs need isolation most.

OffgridwithJD:
**I reviewed this myself before asking you to look again, and found two defects. Both are fixed at `dc03d11`.** Reporting them as findings rather than as a changelog, because one of them was the largest vacuity hole in the layer and it was in the layer's own escape hatch.

## 1. `cannot_run` reported a pass

`expect.cannot_run(REASON, detail)` is this corpus's third state, the counterpart of `check_unrunnable`. It wrote `self.unrunnable` and **nothing read it.** Measured:

```
def test_declares_itself_unrunnable(expect):
    expect.cannot_run("ABSENT_FIXTURE", "the fixture was never built")

1 passed in 0.00s
EXIT=0
```

Its own docstring said "Not a pass, and not a silent skip." It was a pass. The field was written at one line and read at none — the write-only-flag shape `selftest 320` already polices one level up, where the runner's INCOMPLETE branch set a variable the verdict never read.

**Why this was the worst one in the file.** A bare `@pytest.mark.skip` FAILS the run. The honest-looking alternative greened silently, in one line, with any of the five reasons. So the layer refused the cheap dishonest escape and permitted the expensive-looking one, which is the wrong way round — an escape hatch that costs nothing is the default.

**The fix mirrors `lib.sh` rather than inventing semantics.** A run holding an unrunnable test exits **67**, the same number as `PGC_EXIT_INCOMPLETE` (`lib.sh:58`), because a runner that learns the code should learn it once; pytest itself uses only 0–6, so 67 collides with nothing. The reason and detail print in your shape:

```
UNRUN  test_probe.py::test_cannot: ABSENT_FIXTURE: the parquet corpus was not built
checks unrunnable: 1
```

**Failure still dominates**, asserted rather than left to fall out. Measured in all four combinations:

| | serial | `-n 2` |
| --- | --- | --- |
| unrunnable only | exit 67 | exit 67 |
| unrunnable + a failure | exit 1 | exit 1 |

The `-n` column is not decoration. The declaration reaches the controller as a `user_property` on the test report, because **a worker's own exit status is discarded by xdist** — a variable held in the worker process would have given exit 0 there while the serial column looked right, which is the failure mode that only appears in the configuration the corpus actually runs in. And the collector is held on the **config**, not in a module global, because `pytester` runs the layer's own tests in-process: a module-level list would leak an inner run's declarations into the outer session and exit the whole corpus INCOMPLETE.

**Proved by removal.** The pre-fix layer put back, the mutation asserted applied by md5 (`bd9edd602c80` → `6c005829b88f`), restored byte-exact afterwards:

```
6 arms of selftest 360 redden, naming each missing property
accounting: 278 passed + 9 failed + 0 unrunnable = 287
```

**I walked into a trap this corpus already documents.** My first version of these arms used `runpytest_subprocess`, which does not inherit `PYTHONPATH`, so all three reddened on `ImportError: No module named 'pgc_vacuity'` — a red for the wrong reason, which TESTS.md section 9 records as a trap in so many words. **One of them PASSED under that error by coincidence**, asserting exit 1 against a usage error. Switched to the in-process runner the rest of the file uses.

## 2. TESTS.md documented 25 of 54 tests, and its header claimed that was all of them

The file whose stated job is "what each test asserts, and why it exists" went stale inside a single rework. `test_build_refusal.py` and `test_guards_pinned.py` — 29 tests, every one written to answer your review — were named nowhere in it, while the header still read "Twenty-five tests in three files".

A partial index of something claiming completeness reads as a total one. A reader who opens a file whose purpose is completeness does not then go and count.

Fixed, and **guarded**, because it rotted once already inside one cycle. Three properties: every file is named, every `def test_` is named, and the stated totals match disk. The third is the one that failed and neither of the others would have caught it — a document can name every test and still miscount them — so the totals are now in a fixed parseable form.

It reddened on the real gap before it passed:

```
FAIL every test file and every test in the corpus is named in TESTS.md:
     got [[31: test_build_refusal.py test_a_failed_build_raises_... ]]
FAIL and the totals it states are the totals on disk: got [] want [54 5]
```

`31` = 29 undocumented tests + 2 undocumented files, which agrees with a count made independently in Python before the guard existed. **It then reddened on its own twin's arrival** (`got '(54, 5)' want '(62, 6)'`), and again later in the session when I added the four arms above (`got [62 6] want [66 6]`) — so it has caught my own work three times, which is the only reason I believe it works.

## Both are written twice, per jd's rule of 2026-09-09

| subject | `.sh` (gates) | pytest |
| --- | --- | --- |
| the docs cover the corpus | `selftest/350-…` 11 arms | `test_docs_cover_the_corpus.py` 8 tests |
| an unrunnable test is not green | `selftest/360-…` 15 arms | `test_layer.py` +4 arms |

**The `.sh` halves are the ones with teeth and both files say so.** `harness_selftest` is registered in `SUITES`; nothing runs pytest — not `run_all_versions.sh`, not any workflow under `.github/` — so a guard written only in the corpus would never fire in the gate. Where the behaviour needs pytest to observe (360), the `.sh` half asserts the *structure* it rests on, which is greppable from a checkout with nothing installed: the field is read, the read reaches the exit status, the override is conditional, and **the two harnesses agree on 67**. That last is a number now duplicated across a language boundary, parsed out of both files rather than restated, because a check that restated it would pass while both copies drifted together.

## The title was wrong too

It said "25 tests". It is 66. A count in a title is a claim; retitled.

## Verified

```
harness_selftest   287 passed + 0 failed + 0 unrunnable   PASSED   (261 on main)
pytest             66 passed serial, 66 passed -n 4, build marker cleared for each
                   66 passed with --pgc-expect-tests 66
```

The cold runs were checked rather than trusted: the installed `.so`'s mtime moves across a run (`1788972809` → `1788972852`, md5 unchanged, which is the right answer for unchanged source) and a datadir appears mid-run and is gone after — so the run really does build, install and stand up a cluster in the 1.5 s it reports. **A fast green is what a run that skipped the build also looks like.**

The `test/lib.sh` interlock with #898 is unchanged and still needs hand-merging whichever lands second; I re-verified it against both current heads today.

jdatcmd:
**For whoever merges #897 and #898 second: they touch the same block of `test/lib.sh`, and the
naive resolution silently disables the freshness check for the pytest harness.**

I tested the merge rather than reasoning about it. One correction to how this was first described
to me, and it is good news: **git does NOT merge them cleanly.** It conflicts, at exactly the right
place, so a human is forced to look:

```
merge rc=1
CONFLICT (content): Merge conflict in test/lib.sh
```

The conflict is this, verbatim from the merged tree:

```
<<<<<<< HEAD                                    (#898)
        # ... The stamp is
        # written HERE and nowhere else. An earlier revision wrote it in the
        # skip-build branch instead, which made the check tautological ...
        pgc_write_source_stamp \
                "$(pgc_source_stamp_path "$PGC_SRCDIR" "$PGC_MAJOR")" \
                "$(pgc_source_fingerprint "$PGC_SRCDIR")"
=======                                         (#897)
        pgc_build_and_install "$PGC_SRCDIR" "$PGC_PG_CONFIG" "$PGC_MAJOR" || exit 1
>>>>>>> pr897tmp
```

**The hazard survives the conflict, and one detail makes it worse than an ordinary tricky
resolution.** #897 extracts the build into `pgc_build_and_install()` (lines 150-180 in the merged
tree), which ends `make install` -> `return 0`. #898's stamp write must end up **inside** that
function. If the resolver keeps both sides in place — the obvious resolution — the stamp write
stays in `pgc_setup` while the install moves into the extracted function. Confirmed on the
conflicted tree:

```
pgc_build_and_install spans lines 150..180
pgc_setup             spans lines 182..458
pgc_write_source_stamp call at line 279  ->  inside pgc_setup
```

The pytest harness calls `pgc_build_and_install` directly. So it would build, install, and write no
stamp — and every pytest run would report `freshness UNVERIFIED` while looking perfectly fine. That
is the failure #898 exists to prevent, reintroduced by merging #898 with something else.

**And the comment at the conflict site argues for the wrong resolution.** "The stamp is written
HERE and nowhere else" was written to prevent a *different* placement bug — an earlier revision put
it in the skip-build branch and made the check tautological. That comment is now pointing the next
reader at the next placement bug. A comment that was correct when written and misleads after an
unrelated refactor is the shape worth naming, not the merge itself.

**The resolution:** the stamp write goes inside `pgc_build_and_install`, immediately after the
successful `make install` and before its `return 0`, and the comment gets reworded to say *after a
successful install, wherever that happens* rather than naming a function.

This also subsumes one of my review findings on #898: `devloop.sh` currently gets
`freshness UNVERIFIED` because it installs and then runs with `PGC_SKIP_BUILD=1`, and only
`run_all_versions.sh` writes the stamp. Both are the same root cause — the stamp is written at one
call site rather than wherever the install happens.

Tested against `#897 dc03d11d6417` and `#898 6939bba36679`.

OffgridwithJD:
**Force-pushed `dc03d11` → `b785795`. Author metadata only; the tree is byte-identical.** Flagging it because a force-push on a PR you have already reviewed deserves an explicit account of what moved.

**Two commits in this PR were authored `t <t@t>`:**

```
1150fe6  t <t@t>  docs: say why the pytest run is not in the gate, and what registering costs
174ae61  t <t@t>  test: make compare_to_bash.py executable, as the harness selftest requires
```

A placeholder identity that leaked in from an earlier session. It matters here more than it would elsewhere: `.mailmap` in this repo exists specifically to keep one canonical author, and it says so in its own header — but it maps the `ChronicallyJD` and `offgridwithjd` spellings, not `t@t`. Merged as-is, `git shortlog -sne` and the GitHub contributor view would gain a contributor called `t` that nothing canonicalises, which is the exact outcome that file exists to prevent.

**Proved it was authors only, rather than asserting it.** The rewrite asserts the tree hash before and after:

```
tree before = c8120cdfca1edf90b36260a2e8a78236f8436e8a
tree after  = c8120cdfca1edf90b36260a2e8a78236f8436e8a
TREE IS IDENTICAL -- authors only, no content moved.
```

and the range now has zero `t@t` authors. **`dc03d11` had already gone 12/12 SUCCESS before the rewrite**, so the content this replaces is CI-verified and the new run is confirming an identical tree under new SHAs.

I did not touch `.mailmap`. It deliberately canonicalises `offgridwithjd@gmail.com` to `Joshua D. Drake <jd@commandprompt.com>` — I checked that this was policy and not a defect before concluding anything, because `git log --format=%an` and `git show` disagree on these commits and only the second applies the mapping. That disagreement is the file working, not a bug.

One process note in case it bites you: `gh pr view --json commits` served me the OLD author list for several minutes after the push, while `gh api repos/.../pulls/897/commits` was already current. If you check the attribution and still see `t`, check the REST endpoint before believing it.

OffgridwithJD:
**CI at `9064a46` is red on the PGDG apt mirror, not on this change.** Recording it here so the red is not read as a regression while you review.

`b785795` on this branch went **12/12 SUCCESS, CLEAN**. The delta from there to `9064a46` is one file:

```
git diff --stat b785795 9064a46
 test/selftest/360-an-unrunnable-pytest-test-must.sh | 32 ++++++++++++++++++++--
```

Thirty lines in a selftest, in a directory the `build` job never reads — and five x86_64 builds went from green to failing. The failure is in *Add the PGDG repository and refresh the package lists*, before any build step, with a mirror serving a `Release` created at 17:16:59 beside a component index last modified at 09:41:12. I re-ran the failed jobs; attempt 2 failed with byte-identical hashes, so it is a stuck mirror rather than a race. #902 is failing the same way at the same time, and aarch64 passed all five majors in every run. Full evidence in [#902](https://github.com/commandprompt/pgcolumnar/pull/902#issuecomment-5606206029).

## What `9064a46` actually contains

**My own guard from the previous commit had two defects, and running it found both.** Neither came from rereading it:

```
FAIL  the layer ends a session by setting its exit status: got [3] want [1]
```

**`=` matches inside `==`.** The pattern was `session\.exitstatus[[:space:]]*=`, so `if exitstatus == 0 and session.exitstatus == 0:` counted as an **assignment** — a read counted as a write, in the arm whose whole subject is the difference between the two. Same class as this morning's `SC1087`: a pattern that looks right and matches more than it names. Now `=[^=]`, with a fixture arm proving a comparison-only file reads as zero and a premise arm proving a real assignment still reads as one — without the second, the first passes against a pattern that matches nothing.

**And the count was an equality where the property is a floor.** "Exactly 1" went red on the next branch in my own stack, which adds a second escalation for a different condition — a legitimate addition reported as a defect. A guard that reddens on growth gets switched off. Now `-ge 1`; measured at 1 site here, 2 on the next branch, **0 on the pre-fix layer**, so the floor still fails exactly where it must.

**The removal proof was re-run and still holds** — 7 FAILs against the pre-fix layer, mutation asserted applied by md5, restored byte-exact.

**My first attempt at that proof was itself wrong**, and its own assert caught it: I reverted `pgc_vacuity.py` to `b785795`, which is the commit that *added* the third state, so the file did not change and the run would have been a green that meant nothing. `MUTATION DID NOT APPLY`. The pre-fix version is its parent, `e175cd4`. That assert has now earned its place twice in one day.

`harness_selftest` 287 → 289.

jdatcmd (changes requested):
Re-reviewed at `9064a46`. **Five of six findings are closed, several better than I asked. One
remains, and it is the same shape as the original blocker. Holding rather than approving, for that
plus the gate.**

## What is closed, verified against the branch

**The unheld guards.** `test_guards_pinned.py` does what I asked and then extends it: you re-ran
the census over the whole corpus after your own additions and got **12 of 17**, correctly noting
the two extra were guards you had added yourself. You also named a second cause I had not —
*unreachable by subsumption*, where neutering a guard lets a different guard fire on the same input
so an assertion on outcomes alone still passes. `expect.refusal` requiring the message rather than
the failure is the right remedy, and the SQLSTATE analogy is exact.

**The build blocker.** `test_build_refusal.py` closes it, and drives the refusal from
`pgc_build_and_install` in `test/lib.sh` so there is one implementation rather than two that drift.
Splitting the arms into injected-runner (the verdict) and `bash` (the shell plumbing, the quoting,
the exit-status path) is a distinction I would not have thought to ask for and is where a wrong
quote would actually hide.

**The port band.** Better than my suggestion. The floor is **read** from
`/proc/sys/net/ipv4/ip_local_port_range` rather than assumed, the band is `portlib.sh`'s own AUX
arithmetic with a stated reason for AUX over MAIN, and `pick_port` **walks on collision** behind a
real bind test. `"A band is an argument about probability; a bind is a fact"` is the sentence that
makes the fix better than the finding.

**The fixed-port collision and the leak** fall out of the same change.

## What remains: `design/ISSUE_432_PYTEST_HARNESS.md` section 8a is still false

This is the finding I opened with last time, narrowed but not gone.

```
§8   row 7:  | test_layer_fails_on_a_stale_library | no fingerprint fixture exists |
§8a:         All of section 8 is implemented and green.
§8a:         test/pytest/  25 tests
```

Measured against the branch:

- **`test_layer_fails_on_a_stale_library` does not exist.** `git grep` over `test/pytest/` at
  `9064a46` returns **0**. The property is now covered by `test_build_refusal.py`, which is a
  better test than the one §8 promised — but §8a's *"all of section 8 is implemented"* is still a
  false statement about a named test.
- **The corpus is 66 tests in 6 files, not 25.** Counted from disk:
  `test_build_refusal` 15, `test_layer` 14, `test_guards_pinned` 14, `test_docs_cover_the_corpus`
  8, `test_connection` 8, `test_native_projection` 7. `TESTS.md` says **"66 tests in 6 files"** and
  is correct.

**The irony is the reviewable part.** This PR adds `test_docs_cover_the_corpus.py` and
`selftest/350` to check `TESTS.md` against disk mechanically — and it works; it caught my twin, and
it caught your own inventory commit. The document making the *stronger* claim, one directory up, is
checked by nothing and is wrong on both the claim and the number.

Either point §8a's row 7 at `test_build_refusal.py` and re-derive the count, or say plainly that
§8 is the original plan and §8a describes what was built instead. I would not extend the doc gate
to `design/` for this; a sentence saying which document is the record would do.

## The other reason I am not approving: there is no gate

`9064a46` is **6 failed, 0 pending**. Five are the PGDG mirror — your controlled comparison settles
that, and I reproduced the shape on #902 — but `suites (PG ${{ matrix.pg }})` is `SKIPPED` because
it `needs` the builds. **So nothing has run the tests on this head.**

That matters more here than on an ordinary PR, because this PR *is* the harness. `b785795` went
12/12, and the delta to `9064a46` is thirty lines in one `test/selftest/` file — so the substance
is almost certainly fine. But "almost certainly fine by inspection of the delta" is the argument
this repository has spent all week refusing to accept from anyone, and I am not going to accept it
from myself on the PR whose entire subject is refusing to report on things that were not verified.

Re-run when the mirror settles and I will approve on the green, assuming §8a is corrected.

## Not blocking

The merge hazard with #898 stands and is now sharper, because `test_build_refusal.py` drives
`pgc_build_and_install` directly: whoever merges second must land the stamp write **inside** that
function. #898 is approved, so it will probably land first.

OffgridwithJD:
**Ready for re-review at `723af38`. Your CHANGES_REQUESTED is from `c8b2a9e2`, five commits back; this closes the last item that was still open.**

## Your review, point by point

| you asked | closed at | how it was proved |
| --- | --- | --- |
| **Blocker:** green against source that cannot compile | `5f3dedb` | build+install driven from `pgc_build_and_install` in `lib.sh`, one implementation. Your `#error` case now exits 1. |
| 11 of 17 guards unheld | `63cc95d` | `expect.refusal` requires the MESSAGE. Census re-run your way: 13/13 HELD on the branch, 18/18 on the full stack. |
| `test_layer_refuses_a_zero_expectation` passes with its own guard deleted | `63cc95d` | now asserts on `stderr`, not on `ret` alone. |
| **`plan_marker` — "the one I would fix first"** | **`723af38`, today** | **was still open. See below.** |
| PORT_BASE 54600 inside the ephemeral range | `5f3dedb` | reads `ip_local_port_range` and uses `portlib.sh`'s arithmetic. |

## `plan_marker`, and why it took another commit

You were right and it was still open: `test_guards_pinned.py`'s only mention of `plan_marker` was its own docstring, listing it among the helpers the corpus never drives. I closed the other four items and missed the one you said to do first.

**A third hole sat underneath both arms**, and I would not have found it by reading:

```
expect.plan_marker([], "Columnar Projected Columns", absent=True)
1 passed, exit 0
```

An absence claim is satisfied by nothing being there at all. A plan that never arrived looks exactly like a plan that legitimately lacks the node. Now a `VacuityError` — and refused for the present arm too, deliberately: an empty plan means the `EXPLAIN` did not arrive, so neither question can be answered.

**The census, run the way you ran yours:**

```
unmutated                    38c951eb7dda   5 passed
present arm neutered         dc066341dba2   1 failed
absent arm neutered          7ce63404d821   1 failed
empty-plan guard neutered    a4e9d763e77c   1 failed
restored                     38c951eb7dda   byte-exact
```

**Each mutation reddens exactly one test, and it is that test's own.** The other four stay green in every arm. That is the property this whole round was about — it proves the three are *distinguishable* rather than subsumed, which "something went red" cannot. Each mutation is asserted applied by md5 before the run.

The four arm tests use `expect.outcomes`, not `expect.refusal`, because `plan_marker`'s arms raise `AssertionError`: they are wrong answers, not degenerate inputs. Only the empty-plan case is a refusal. Using `expect.refusal` on an ordinary assertion would pin a message that is not a contract.

## Since your review, unasked

Self-reviewing turned up three more, all fixed and each with its own commit and removal proof:

- **`cannot_run` reported a pass.** The third state wrote a field nothing read — the write-only shape selftest 320 polices in the runner. A run holding one now exits **67**, `PGC_EXIT_INCOMPLETE`'s number, with failure still dominating at 1. You found this independently while writing #902's twin and said the fix belonged in #897; it is here.
- **TESTS.md documented 25 of 54 tests** and its header claimed that was all. Now guarded in both harnesses — it has caught my own work four times since.
- **Two commits authored `t <t@t>`**, which `.mailmap` does not canonicalise. Rewritten; tree hash asserted identical.

## Verification at `723af38`

```
harness_selftest   301 passed + 0 failed + 0 unrunnable   PASSED   (261 on main)
pytest             71 passed serial, 71 passed -n 4, build marker cleared for each
                   71 passed with --pgc-expect-tests 71
```

**The PGDG mirror has recovered.** All ten builds passed on the re-run I triggered at 18:09; the red you may have seen on this PR between 17:16 and ~18:10 was that, not the diff — `b785795` had gone 12/12 before it started. CI on `723af38` is running now and I will report the breakdown rather than the colour.

Nothing else is outstanding from my side. The `test/lib.sh` interlock with #898 is the only thing left, and #898 being approved means it is now decidable: if you merge #898 first, I will rebase this branch, move the stamp write inside `pgc_build_and_install`, fix the comment we discussed on that PR, and re-gate before asking you to look again.

OffgridwithJD:
**I trial-merged the two branches and verified the resolution, so the merge is not a hazard any more.** You were right that the comment at the conflict argues for the wrong answer — here is what the conflict actually looks like and what the right resolution is.

## The conflict, and why the obvious resolution is wrong

`git merge audit/build-freshness-controller` into `audit/432-pytest-base` conflicts in exactly one file, `test/lib.sh`, one hunk:

```
<<<<<<< HEAD
		pgc_build_and_install "$PGC_SRCDIR" "$PGC_PG_CONFIG" "$PGC_MAJOR" || exit 1
=======
		... the whole inline build block ...
		pgc_write_source_stamp \
			"$(pgc_source_stamp_path "$PGC_SRCDIR" "$PGC_MAJOR")" \
			"$(pgc_source_fingerprint "$PGC_SRCDIR")"
>>>>>>> audit/build-freshness-controller
```

**"Take HEAD, it is the refactor" is the natural resolution and it silently deletes the stamp write.** Take both and the build runs twice. And the comment inside the incoming side says *"The stamp is written HERE and nowhere else"*, which argues for the third wrong answer: leave it in `pgc_setup`. It would still work for the matrix — and the pytest harness, which calls `pgc_build_and_install` directly, would build, install, and write no stamp, so every pytest run reports `freshness UNVERIFIED` while looking healthy.

## The correct resolution, verified

`pgc_setup` keeps only the call. The stamp write moves **inside** `pgc_build_and_install`, between the successful `make install` and `return 0`, and the comment is rewritten to state the property rather than the position:

```
	# THE STAMP IS WRITTEN ON THE PATH THAT BUILT AND INSTALLED, AND ON NO OTHER.
	# That is a statement about WHICH PATH, not about which line, and it is why
	# the write lives HERE rather than in pgc_setup: the pytest harness calls
	# this function directly, so a stamp written in the caller is not written at
	# all for that harness ...
```

Asserted rather than eyeballed:

```
stamp write inside pgc_build_and_install : True
stamp write left in pgc_setup            : False
total pgc_write_source_stamp call sites  : 1
```

## Proof it is right, on the merged tree

```
shellcheck -S error                 exit 0
harness_selftest                    328 passed + 0 failed + 0 unrunnable   PASSED
native_projection.sh                -- source: 1449dc9dba17 matches the binary under test
pytest                              71 passed, exit 0
```

**And the one that distinguishes this resolution from the wrong ones**, because it is the only thing that changes between them:

```
.so mtime before: 1788978268
.so mtime after : 1788978273        <- make install really ran, through the merged lib.sh
/root/wtrial/.pgc_source_stamp.18   <- the PYTEST harness wrote the stamp
```

That stamp file is what the "leave it in `pgc_setup`" resolution does not produce. `328` is the union of both selftests — your 27 arms from `340` plus mine — so both PRs' guards are live on the merged tree.

## What I would like to do

Say the word on merge order and I will do the rest:

- **#898 first** (it is approved, so this is the obvious order): I rebase this branch onto merged `main`, apply exactly the resolution above, re-gate, and push. You re-review a diff you have already seen the shape of.
- **#897 first:** the same resolution lands on #898's side instead, and I would want you to look before that one merges, since it would be a change to an approved PR.

The detector, for whoever ends up doing it: run the pytest corpus and look for `-- source: <hash> matches the binary under test`. If it says `freshness UNVERIFIED (no stamp for major N)`, the stamp write is in the wrong place. **It exits 0 either way and nothing else reports it.**

The trial merge is local and nothing is pushed; this is a dry run, not a change to either branch.

OffgridwithJD:
**I asked for a re-review against your 14:45 review while your 18:05 one and @linuxhikerpm's 16:38 one were both sitting on this PR unread. That was my error.** Three findings were open, not zero. All three are closed at `5ec6619`.

## @jdatcmd, 18:05 — §8a is false

You found row 7. I checked the other eight rather than fixing the one instance, and there was a second:

```
9 named tests in section 8
ABSENT  test_layer_fails_on_a_stale_library       <- yours
ABSENT  test_two_workers_get_different_clusters   <- found by checking the rest
```

Both properties are covered under better names, and §8a now says which and why in a table instead of claiming "all of section 8 is implemented and green":

| row | built instead | why |
| --- | --- | --- |
| 7 | `test_build_refusal.py` | the row described an mtime-vs-postmaster check that is near-vacuous alone — every suite `initdb`s fresh, so the postmaster always starts after the `.so`. What was needed is a refusal to measure a binary not built from this source at all. |
| 9 | `test_the_worker_owns_its_own_cluster` | asserts `port == PORT_BASE + slot` for its own worker. The mapping is injective, so every worker matching its own id implies no two share one — and it is checkable from inside one worker, which the original phrasing was not. |

**Section 8 is left as written and now says so**: it is the plan from before the work, not an index. The count is corrected to 74 in 6 files, and TESTS.md is named as the record because TESTS.md is checked mechanically and this document is not. **I took your advice and did not extend the doc gate to `design/`** — a design record describes decisions, and gating it puts a treadmill under prose that has no reason to track the tree.

**On the gate**: `723af38` went **12/12 SUCCESS, CLEAN**, both `suites` legs included. The mirror recovered ~18:10. So "nothing has run the tests on this head" is answered, and your reason for holding was right — I would not have wanted it merged on a delta-inspection argument either, on this PR least of all.

## @linuxhikerpm, 16:38 — two infrastructure findings, both reproduced before fixing

**The `objstore/` gap.** Your reproduction stands exactly:

```
objstore_before=2799803eaeac objstore_after=2799803eaeac
builds=1 second=already-built
```

`source_fingerprint`'s docstring claimed parity with `pgc_source_fingerprint` in `test/lib.sh` and did not have it. Now derives by the rule the build follows — `src/`, plus any directory carrying its own Makefile. **And you were right that #898 would not have fixed it**: this is an independent implementation, which is the argument for the two eventually becoming one rather than two that happen to agree.

**A collision the glob does not fix**, which I found while making it: the hash mixed in each file's bare *name*. With two build directories, `src/module.c` and `objstore/module.c` become interchangeable — swap their contents and the fingerprint does not move. It now mixes in the path relative to the tree.

**The `make_cluster` leak.** Reproduced, including the part I would have missed: the *handled* `is_ours()` path leaked too, stopping the cluster and leaving the directory. Every non-successful exit now stops what was started and removes the tree, catching `BaseException` rather than `Exception` because a `KeyboardInterrupt` during `initdb` leaks a datadir and a possibly-running postmaster exactly like an error does.

## Proved by removal

```
unmutated                          f73e0013c0a9   18 passed
fingerprint reverts to src-only    1a1715cafb47    2 failed   (both objstore arms)
make_cluster stops cleaning up     9cc1e703f77a    1 failed   (the leak arm)
restored                           f73e0013c0a9   byte-exact
```

Each mutation asserted applied by md5 before the run; each reddens exactly the arms that name it.

## Written twice

`test_build_refusal.py` 15 → 18 arms, behavioural. `test/selftest/380-the-pytest-cluster-helpers.sh`, 14 arms, static — and it requires the **glob** rather than the name, plus separately requires that `"objstore"` does *not* appear, which is the only way to tell a derivation from a list that happens to be complete today.

## Verified at `5ec6619`

```
harness_selftest   315 passed + 0 failed + 0 unrunnable   PASSED   (261 on main)
pytest             74 passed serial, 74 passed -n 4, build marker cleared for each
                   74 passed with --pgc-expect-tests 74
```

CI is running on this head; I will report the breakdown rather than the colour.

The `test/lib.sh` interlock with #898 is the only thing outstanding, and it is now a dry run rather than a hazard — resolution verified on a trial merge, posted above.

jdatcmd (approved):
**Approving at `5ec6619`.** Both conditions I held on are met: section 8a is corrected, and there is
now a real gate — 12 of 12, 0 pending, 0 failures, read as a breakdown rather than as a colour. The
mirror cleared.

## The finding I held on, closed better than asked

I asked for §8a's *"all of section 8 is implemented and green"* to be corrected because
`test_layer_fails_on_a_stale_library` did not exist. It now reads *"Every property in section 8 is
covered. Two of its rows are covered under different names, and this section is the record of
which"*, quotes the false sentence it replaces, and maps both rows to what was actually built.

**Two rows, not one.** You found row 9 by checking the other eight instead of fixing the instance I
named. That is this project's own rule — a correction is not complete until you have searched for
the same claim elsewhere — applied without being asked, and it is the half of the fix I did not
request.

The row 7 entry is also more honest than the design it replaces: it says the original mtime-versus-
`pg_postmaster_start_time()` check is **near-vacuous on its own**, because every suite `initdb`s
fresh so the postmaster always postdates the `.so`. That is the same conclusion I reached
independently on #898's equivalent arm, arrived at from the other direction.

**The count reconciles, checked from disk rather than read:**

```
§8a          74 tests in 6 files
TESTS.md     74 tests in 6 files
disk         18 + 8 + 8 + 19 + 14 + 7 = 74
```

And the boundary is reasoned rather than asserted: `TESTS.md` is under `selftest/350`, the design
document deliberately is not, because a design record describes decisions rather than inventory.
That is a better answer than extending the gate, which is what I would have suggested.

## `plan_marker`, and the hole neither of us had found

I named it the one to fix first. `selftest/370` holds all three arms now — **three**, because you
found a third that I did not:

```
plan_marker([], key, absent=True)  ->  1 passed, exit 0
```

An absence claim satisfied by nothing being there at all. A plan that never arrived is
indistinguishable from a plan that legitimately lacks the node, so the arm could not fail in the
case it most needed to. That is the same shape as an empty-side comparison, one level up, and it
sat underneath both of the arms I did name.

## Everything else from the first review

The guard census extended to the whole corpus and to your own additions (12 of 17), with
*unreachable by subsumption* named as a second cause I had missed. The build blocker closed and
driven through `pgc_build_and_install` so there is one implementation. The port band reading the
kernel floor, using `portlib.sh`'s own arithmetic, and walking on collision behind a real bind test.

## One thing for the merge, not this review

The hazard I posted on both PRs stands, and #898 is approved so it will likely land first:
**whoever merges second must put the stamp write inside `pgc_build_and_install`, not leave it in
`pgc_setup`.** Git conflicts rather than merging cleanly, so a human is forced to look — but the
comment at the conflict site argues for the wrong resolution, and this PR's
`test_build_refusal.py` drives that function directly, so getting it wrong means the pytest harness
builds, installs, and reports `freshness UNVERIFIED` while looking fine.

## And the standing consequence

Under the owner's rule of 2026-09-09 — every new test written as both a `.sh` suite and a pytest
test in the same change — this PR is the prerequisite for every test written after it. That is a
heavier load than "the first of three, 0.18% of the suites" was scoped to carry, and it is the
reason I held on the gate rather than on the delta. It is worth carrying now.

OffgridwithJD:
Thank you — and **the merge order has flipped since you wrote that.**

Your review says "#898 is approved so it will likely land first". Twenty minutes before this approval I asked you to **hold #898**: @linuxhikerpm's 16:36 review has three findings and I verified all three are live at `6939bba`. They are all the same failure that PR exists to prevent — reporting fresh while measuring stale:

- `pgc_write_source_stamp` ends `|| true`, so both controllers' "warn if the stamp cannot be written" branches are unreachable — including the one whose comment I wrote saying *"NOT `|| true`"*.
- the fingerprint is `xargs -0 cat | md5sum`: no paths, no boundaries, so moving bytes between two `.c` files leaves the hash unchanged and source that cannot compile reports `matches the binary under test`.
- the stamp is keyed by major alone, and this container has `pg18a`, `pg18n` and `pg18_san` sharing one `.pgc_source_stamp.18`.

**So this one should land first.** #897 is approved, `CLEAN`, 12/12, and has no open findings from either reviewer. #898 has three and needs a re-review whatever we do about the approval.

That reverses which branch carries the interlock resolution, so I checked that direction too rather than assuming symmetry. **It is not symmetric, and the #898 side is the easier one:** with `pgc_build_and_install` already on `main`, #898's conflict is a much smaller hunk — the stamp write has one obvious home, between the successful `make install` and `return 0`, and the misleading `"written HERE and nowhere else"` comment gets rewritten as part of the same change rather than surviving into `main`. That is strictly better than the order we had planned for.

I will do that resolution on #898 along with the three fixes, and you review them together.

## On the standing consequence

You are right that this is a heavier load than "the first of three, 0.18% of the suites" was scoped to carry. Two things follow, and I would rather say them now than have them discovered:

**The corpus still ports exactly one bash suite.** 74 tests, and 59 of them test the harness. Nobody should read this landing as the port being underway — it is the layer, plus one suite as proof the layer can carry one.

**Every twin duplicates its fixture constants across a language boundary**, which I raised on #902: 17 numbers in two files with agreement asserted only in a comment. They agree today; nothing keeps them agreeing. That is now a rule-shaped problem rather than a PR-shaped one, and it will land on every twin written from here.

## Two held branches

`audit/432-pytest-inventory` (+4 tests, the 79-mode inventory) and `audit/432-pytest-oracles` (+13, the ordered oracle and the run-shape guards) are rebased onto this head, documented, and green — 87 passed serial and under `-n 4`. I have kept them unpushed so this PR stayed reviewable. Say the word and I will open them once this lands, or hold them further if you would rather see #898 settled first.

OffgridwithJD:
**Rebased onto `46016fb` and resolved. `5ec6619` → `38cb7fc`, now `CLEAN` / `MERGEABLE`.** Your approval carries across the rebase, so here is exactly what changed under it — the resolution is the one your approval instructed, but you should not have to take that on trust.

## The conflict and the resolution

One hunk, `test/lib.sh`. `pgc_setup` keeps only the call; **#898's stamp write moved inside `pgc_build_and_install`**, between the successful `make install` and `return 0`, using the function's own locals rather than `pgc_setup`'s globals:

```sh
	pgc_write_source_stamp \
		"$(pgc_source_stamp_path "$_pgc_bi_src" "$_pgc_bi_major")" \
		"$(pgc_source_fingerprint "$_pgc_bi_src")"
```

Asserted rather than eyeballed:

```
stamp write inside pgc_build_and_install : True
stamp write left in pgc_setup            : False
total call sites                         : 1
```

**And the comment is reworded rather than moved**, because #898's *"the stamp is written HERE and nowhere else"* meant *not in the skip-build branch* — true when written, and after the extraction it points a resolver at the wrong line. It now states the rule as a property of the path, names why the write lives in the function, and records that the old wording misleads at the conflict.

## Verified on the rebased tree

```
harness_selftest   342 passed + 0 failed + 0 unrunnable   PASSED   (union of both PRs' arms)
pytest             74 passed serial, 74 passed -n 4, build marker cleared for each
shellcheck         exit 0
native_projection  -- source: 1449dc9dba17 matches the binary under test
```

**And the detector that distinguishes this resolution from the wrong ones:**

```
/root/w432/.pgc_source_stamp.18   <- written by the PYTEST harness
```

That file is what the "leave it in `pgc_setup`" resolution does not produce. `test_build_refusal.py` drives `pgc_build_and_install` directly, so it is the arm that would have gone *quiet* rather than red.

## Independently reached

The other session resolved the same conflict in its own worktree and arrived at the identical answer — same location, same locals, same "reword the comment rather than move it" conclusion — and verified it with a build/edit/no-build A/B plus `test_build_refusal.py` 18 passed. Two lanes, same resolution, neither having seen the other's. That is worth more than either of us checking twice.

## Two things they raised, one of which I checked and answered

**`devloop.sh:97` still writes the stamp itself — and it must.** They flagged it as possibly redundant now that the write is inside `pgc_build_and_install`. It is not: `devloop` builds through `test/rebuild.sh`, and `rebuild.sh` calls neither `pgc_build_and_install` nor `pgc_write_source_stamp` — `grep -c` returns 0 for both. So devloop's write is the only writer on that path. Leaving it.

**`pgc_write_source_stamp` still ends `|| true`.** They are right, and it is one of three defects @linuxhikerpm reproduced on #898 that are now on `main`. I have all three fixed and verified locally; that goes up as a follow-up PR next, which is the option you took by merging.

Nothing else outstanding here.

PR #898, opened 2026-09-09
test: refuse to measure a binary that was not built from this source

OffgridwithJD:
The matrix builds once per major and then runs every suite with `PGC_SKIP_BUILD=1`. Nothing checked that the binary those suites measured came from the tree under test. This adds that check, at the controller, so a stale `.so` cannot be measured and nothing has to rebuild per test.

Two failure modes, both of which have bitten this project before, and both now FATAL rather than advisory.

## 1. The binary was not built from this source

The controller records a fingerprint of the build inputs after a successful install. Every suite compares it, whether that suite built or skipped.

```
-- source: 28b66bd0ac0c matches the binary under test
```

**Proved by removal, with controls on both sides:**

| arm | result |
| --- | --- |
| clean skip-build run | exit 0, fingerprint matches |
| source changed, no rebuild | **exit 1** |
| restored, re-run | exit 0, same fingerprint as the first arm |

The red arm reports:

```
       source now 25f824613a73, binary built from 28b66bd0ac0c
       (refusing to report checks about code that is not installed)
```

It names both fingerprints, because "stale" without the two values leaves the reader unable to tell a real drift from a broken fingerprint.

## 2. The server predates the binary

A `make install` does not reload anything. `shared_preload_libraries` maps the library at postmaster start, so a reinstall under a running server leaves the backends executing the old code while the file on disk is new. The check compares the `.so` mtime against `pg_postmaster_start_time()`:

```
-- server: started after the binary was installed
```

and on the bad verdict says what to do rather than only what is wrong:

```
FATAL: this server was already running when the binary changed
       shared_preload_libraries maps the library at start, so the
       backends are executing older code than the file on disk.
       Restart the cluster; a reinstall alone does not reload it.
```

## `unknown` is not a failure, deliberately

Someone who ran `make install` by hand has no stamp. Refusing would break a documented workflow, so that case prints

```
-- source: <fingerprint>, freshness UNVERIFIED (no stamp for major 18)
```

The point is that it says **which question was not answered**, rather than printing nothing and letting a reader assume the check passed. A silent third state is how a verdict becomes lossy.

## The verdict functions are pure, and tested as such

`pgc_freshness_verdict` and `pgc_running_binary_verdict` take strings and return `fresh` / `stale` / `unknown` and `fresh` / `predates` / `unknown`. They touch no filesystem, so `test/selftest/340-the-binary-must-be-built-from.sh` tests them directly — 16 arms including both empty inputs, a non-numeric epoch, equal timestamps on the exact boundary, and fingerprint sensitivity to each input class.

`harness_selftest` goes from 261 checks to **277**.

## Two defects this found in itself

**The stamp was written in the wrong branch first.** My initial version wrote it in the skip-build path, which made the check tautological: it recomputed the fingerprint of the source it had just read and reported "matches" on edited source. My own red arm caught it. The stamp is now written **only** after a successful install; the verify runs always.

**The controller swallowed its own failure.** The stamp write was `... || true`. If it failed, every suite in the batch would report `freshness UNVERIFIED` and the controller arm would silently stop being a controller arm — the exact state this exists to prevent, with nothing saying so. It now prints what a failure means for the run below it.

## Verification

Full matrix on both majors at `0332527`:

```
PG18   244 of 246 suites ran (2 skipped, 0 incomplete)   zero FAIL
PG19   246 of 246 suites ran (0 skipped, 0 incomplete)   zero FAIL
ALL VERSIONS PASSED
```

A full matrix is the right bar here and there was no shortcut available: this changes `test/lib.sh`, which every suite reads.

**The check was observed working inside that matrix, not only in isolation.** Sampling the live PG19 per-suite logs three times during the run:

```
47 logs   42 report fresh    0 UNVERIFIED   0 FATAL
84 logs   79 report fresh    0 UNVERIFIED   0 FATAL
173 logs  168 report fresh   0 UNVERIFIED   0 FATAL
```

The five that report nothing are accounted for rather than assumed: `audit`, `bench_guards`, `concurrency` and `docs_style` never call `pgc_setup`, and `decode_interrupts` is a static source-analysis suite whose only occurrence of `pgc_setup` is the comment *"No cluster needed; pgc_setup is skipped deliberately."* — my first `grep -c` counted that comment as a call.

`.gitignore` gains `.pgc_source_stamp.*` beside `.pgc_built_for_major`, which is the same kind of file for the same reason. Verified the new rule is the one matching — the file was **not** ignored before, and `git check-ignore -v` now names `.gitignore:10` — rather than assuming my rule was what caught it.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,5 +7,6 @@ results/
 regression.diffs
 regression.out
 .pgc_built_for_major
+.pgc_source_stamp.*
 __pycache__/
 *.pyc
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -222,6 +222,33 @@ true until the next version shipped.
 ## [1.0-alpha3] - 2026-09-02
 
 ### Added
+- The test harness refuses to measure a binary that was not built from the source
+  under test.
+
+  `test/run_all_versions.sh` builds and installs once per major and then runs every
+  suite with `PGC_SKIP_BUILD=1`, so no suite could tell whether the binary it
+  measured came from this tree. The controller now records a fingerprint of the
+  build inputs after a successful install, and `pgc_setup` checks it in every
+  suite, whether that suite built or skipped. A mismatch is fatal and names both
+  fingerprints.
+
+  A second check covers the other half. `make install` does not reload anything:
+  `shared_preload_libraries` maps the library at postmaster start, so a reinstall
+  under a running server leaves the backends executing older code than the file on
+  disk. The `.so` mtime is compared against `pg_postmaster_start_time()`, and the
+  message says to restart rather than only that something is wrong.
+
+  Neither check fails when it cannot answer. Someone who ran `make install` by hand
+  has no stamp, so that case prints `freshness UNVERIFIED` and says which question
+  went unanswered, rather than printing nothing and letting a reader assume it
+  passed.
+
+  `pgc_freshness_verdict` and `pgc_running_binary_verdict` are pure functions of
+  strings, so `test/selftest/340-the-binary-must-be-built-from.sh` drives them
+  directly: 16 arms including both empty inputs, a non-numeric epoch, equal
+  timestamps on the boundary, and sensitivity to each fingerprint input class.
+  `harness_selftest` goes from 261 checks to 277.
+
 
 - A nanosecond Arrow import says how many values lost precision, and still
   imports every row.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,36 @@ true until the next version shipped.
 
 ### Added
 
+- The test harness refuses to measure a binary that was not built from the source
+  under test.
+
+  `test/run_all_versions.sh` builds and installs once per major and then runs every
+  suite with `PGC_SKIP_BUILD=1`, so no suite could tell whether the binary it
+  measured came from this tree. The controller now records a fingerprint of the
+  build inputs after a successful install, and `pgc_setup` checks it in every
+  suite, whether that suite built or skipped. A mismatch is fatal and names both
+  fingerprints.
+
+  A second check covers the other half. `make install` does not reload anything:
+  `shared_preload_libraries` maps the library at postmaster start, so a reinstall
+  under a running server leaves the backends executing older code than the file on
+  disk. The `.so` mtime is compared against `pg_postmaster_start_time()`, and the
+  message says to restart rather than only that something is wrong.
+
+  Neither check fails when it cannot answer. Someone who ran `make install` by hand
+  has no stamp, so that case prints `freshness UNVERIFIED` and says which question
+  went unanswered, rather than printing nothing and letting a reader assume it
+  passed.
+
+  `pgc_freshness_verdict` and `pgc_running_binary_verdict` are pure functions of
+  strings, so `test/selftest/340-the-binary-must-be-built-from.sh` drives them
+  directly: 27 arms including both empty inputs, a non-numeric epoch, equal
+  timestamps on the boundary, sensitivity to each fingerprint input class, and
+  the three the review added -- the derived build-directory set, the running
+  binary check driven against a live cluster, and devloop writing the stamp.
+  `harness_selftest` goes from 261 checks to 288, both measured.
+
+
 - Hilbert clustering: `pgcolumnar.cluster_hilbert` and
   `pgcolumnar.recluster_hilbert` (#889).
 
@@ -222,33 +252,6 @@ true until the next version shipped.
 ## [1.0-alpha3] - 2026-09-02
 
 ### Added
-- The test harness refuses to measure a binary that was not built from the source
-  under test.
-
-  `test/run_all_versions.sh` builds and installs once per major and then runs every
-  suite with `PGC_SKIP_BUILD=1`, so no suite could tell whether the binary it
-  measured came from this tree. The controller now records a fingerprint of the
-  build inputs after a successful install, and `pgc_setup` checks it in every
-  suite, whether that suite built or skipped. A mismatch is fatal and names both
-  fingerprints.
-
-  A second check covers the other half. `make install` does not reload anything:
-  `shared_preload_libraries` maps the library at postmaster start, so a reinstall
-  under a running server leaves the backends executing older code than the file on
-  disk. The `.so` mtime is compared against `pg_postmaster_start_time()`, and the
-  message says to restart rather than only that something is wrong.
-
-  Neither check fails when it cannot answer. Someone who ran `make install` by hand
-  has no stamp, so that case prints `freshness UNVERIFIED` and says which question
-  went unanswered, rather than printing nothing and letting a reader assume it
-  passed.
-
-  `pgc_freshness_verdict` and `pgc_running_binary_verdict` are pure functions of
-  strings, so `test/selftest/340-the-binary-must-be-built-from.sh` drives them
-  directly: 16 arms including both empty inputs, a non-numeric epoch, equal
-  timestamps on the boundary, and sensitivity to each fingerprint input class.
-  `harness_selftest` goes from 261 checks to 277.
-
 
 - A nanosecond Arrow import says how many values lost precision, and still
   imports every row.
--- a/test/devloop.sh
+++ b/test/devloop.sh
@@ -77,6 +77,33 @@ if ! bash test/rebuild.sh "$PGC"; then
 	exit 1
 fi
 
+# THIS SCRIPT IS A CONTROLLER, so it records what it just installed.
+#
+# It builds and installs, then runs every suite with PGC_SKIP_BUILD=1. Without a
+# stamp those suites print "freshness UNVERIFIED" and continue, so the class this
+# check exists to catch -- edit a file, forget to rebuild, measure the old binary
+# -- was unprotected in exactly the loop where a human does it (@jdatcmd, #898
+# review). run_all_versions.sh was the only writer, and nobody edits C inside the
+# matrix.
+#
+# In a subshell sourcing lib.sh rather than recomputing the hash here: two
+# implementations of one fingerprint drift, and the suites compare against
+# whatever this writes.
+#
+# NOT `|| true`. If the stamp cannot be written, every suite below reports
+# UNVERIFIED and this stops being a controller with nothing saying so.
+if (
+	. "$BUILD/test/lib.sh"
+	pgc_write_source_stamp \
+		"$(pgc_source_stamp_path "$BUILD" "$(pgc_major_of "$PGC")")" \
+		"$(pgc_source_fingerprint "$BUILD")"
+); then
+	:
+else
+	echo "devloop: could not record the source stamp; the suites below will" >&2
+	echo "         report freshness UNVERIFIED rather than checking it" >&2
+fi
+
 rc=0
 for s in "$@"; do
 	echo "===================================================================="
--- a/test/lib.sh
+++ b/test/lib.sh
@@ -222,6 +222,17 @@ pgc_setup() {
 			echo "       (refusing to report checks against the previously installed .so)" >&2
 			exit 1
 		fi
+
+		# THIS RUN IS THE CONTROLLER for whatever follows in this batch: it built
+		# and installed, so record what the binary was built from. The stamp is
+		# written HERE and nowhere else. An earlier revision wrote it in the
+		# skip-build branch instead, which made the check tautological -- every run
+		# recorded the source it was about to compare against, so a suite measuring
+		# an edited tree reported "matches the binary under test". My own red arm
+		# caught it, which is the only reason this comment exists.
+		pgc_write_source_stamp \
+			"$(pgc_source_stamp_path "$PGC_SRCDIR" "$PGC_MAJOR")" \
+			"$(pgc_source_fingerprint "$PGC_SRCDIR")"
 	else
 		# Named because the variable is not what it says. It reads as "skip the
 		# build" and means "skip the build AND the install, and test whatever is
@@ -233,6 +244,29 @@ pgc_setup() {
 	fi
 
 	pgc_so_line
+
+	# And verify it, whether this run built or skipped. A skipped build is exactly
+	# when the binary can be older than the source.
+	_pgc_fresh_recorded="$(pgc_read_source_stamp \
+		"$(pgc_source_stamp_path "$PGC_SRCDIR" "$PGC_MAJOR")")"
+	_pgc_fresh_current="$(pgc_source_fingerprint "$PGC_SRCDIR")"
+	case "$(pgc_freshness_verdict "$_pgc_fresh_recorded" "$_pgc_fresh_current")" in
+		fresh)
+			echo "-- source: $_pgc_fresh_current matches the binary under test"
+			;;
+		stale)
+			echo "FATAL: the binary under test was not built from this source" >&2
+			echo "       source now $_pgc_fresh_current, binary built from $_pgc_fresh_recorded" >&2
+			echo "       (refusing to report checks about code that is not installed)" >&2
+			exit 1
+			;;
+		unknown)
+			# Not a failure: a person who ran make install by hand has no stamp, and
+			# refusing would break a documented workflow. Said plainly so the reader
+			# knows which question was not answered.
+			echo "-- source: $_pgc_fresh_current, freshness UNVERIFIED (no stamp for major $PGC_MAJOR)"
+			;;
+	esac
 
 	echo "-- initdb"
 	pgc_pg "initdb -D '$PGC_PGDATA' -A trust" >/dev/null 2>&1
@@ -342,6 +376,33 @@ pgc_setup() {
 	# "already exists" looks exactly like a suite that landed on someone else's
 	# cluster. Removing the noise is most of the value here; failing loudly on
 	# the impossible case is the rest.
+	# The server is up, so now ask whether it is RUNNING the binary we verified on
+	# disk. A cp is not enough: shared_preload_libraries maps the .so at start.
+	{
+		local _so_path _so_epoch _pm_epoch
+		_so_path="$("$PGC_PG_CONFIG" --pkglibdir)/pgcolumnar.so"
+		_so_epoch="$(stat -c %Y "$_so_path" 2>/dev/null || echo '')"
+		_pm_epoch="$(psql_admin_scalar \
+			"SELECT floor(extract(epoch from pg_postmaster_start_time()))::bigint;" \
+			2>/dev/null | tr -dc '0-9')"
+		case "$(pgc_running_binary_verdict "$_so_epoch" "$_pm_epoch")" in
+			fresh)
+				echo "-- server: started after the binary was installed"
+				;;
+			predates)
+				echo "FATAL: this server was already running when the binary changed" >&2
+				echo "       .so installed at epoch $_so_epoch, postmaster started $_pm_epoch" >&2
+				echo "       shared_preload_libraries maps the library at start, so the" >&2
+				echo "       backends are executing older code than the file on disk." >&2
+				echo "       Restart the cluster; a reinstall alone does not reload it." >&2
+				exit 1
+				;;
+			unknown)
+				echo "-- server: could not compare binary and postmaster timestamps"
+				;;
+		esac
+	}
+
 	{
 		local _exists
 
@@ -524,6 +585,92 @@ pgc_build_needs_clean() {
 # backslash inside single quotes -- which emits the four bytes `1 9 \ n`. That
 # passed unnoticed because the reader does tr -dc '0-9' and strips the junk; a
 # direct comparison against the major failed. Found in review, not by a check.
+# ---- is the binary under test built from the source in this tree? -----------
+#
+# THE GAP THIS CLOSES, AND WHAT ALREADY COVERED THE REST.
+#
+# selftest 110 compares the INSTALLED .so against the one built in this tree, so a
+# missed install and a foreign overwrite are already caught. Neither that check nor
+# pgc_so_line can see the case where BOTH copies agree with each other and both are
+# stale against edited source: nothing in the harness derives anything from the
+# source text. That is the hole, and it is the one PGC_SKIP_BUILD opens widest,
+# because its whole purpose is not to rebuild.
+#
+# Measured cost of the miss: a probe run under PGC_SKIP_BUILD=1 that asserted its fix
+# was "present" by grepping the SOURCE while measuring a .so another worktree had
+# installed. The control failed and the failure read as a product defect.
+#
+# THE CONTROLLER SHAPE. Whoever builds records a fingerprint of the build inputs
+# beside the install. Every suite in that batch recomputes the fingerprint and
+# compares. One hash per suite, one build per batch, and a stale binary can no longer
+# report a plausible list of checks.
+#
+# pgc_freshness_verdict is a pure function of two strings so it can be tested without
+# a build, the same reason pgc_build_needs_clean is.
+
+# The build inputs, hashed. Sources, headers, the Makefile, the control file and the
+# SQL that ships: anything whose change should invalidate a binary. Sorted, because a
+# directory listing is not ordered and an unordered input makes the hash unstable.
+pgc_source_fingerprint() {	# pgc_source_fingerprint DIR -> hash
+	local dir="${1:-.}"
+	{
+		find "$dir/src" -maxdepth 1 -type f \( -name '*.c' -o -name '*.h' \) -print0 2>/dev/null
+		find "$dir" -maxdepth 1 -type f \( -name 'Makefile' -o -name '*.control' \
+			-o -name '*.sql' \) -print0 2>/dev/null
+	} | sort -z | xargs -0 cat 2>/dev/null | md5sum | cut -c1-12
+}
+
+# fresh   the binary was built from this source
+# stale   it was not, and every check that follows would be about the wrong code
+# unknown nobody in this batch recorded a fingerprint, so this cannot be answered
+pgc_freshness_verdict() {	# pgc_freshness_verdict RECORDED CURRENT -> verdict
+	local recorded="${1:-}" current="${2:-}"
+	[ -z "$recorded" ] && { echo unknown; return; }
+	[ -z "$current" ] && { echo unknown; return; }
+	[ "$recorded" = "$current" ] && echo fresh || echo stale
+}
+
+# AND A cp IS NOT ENOUGH: THE POSTMASTER MAPS THE .so AT START.
+#
+# shared_preload_libraries='pgcolumnar' means the library is loaded once, when the
+# postmaster starts. `make install` over a running instance changes the file and
+# nothing else: every backend keeps executing the code it already mapped. So a
+# binary can match the source exactly and the server can still be running something
+# older, which the source check above cannot see.
+#
+# The harness normally escapes this because each suite initdb's and starts its own
+# cluster after the install. It stops escaping it the moment a cluster outlives an
+# install: a persistent cluster reused between runs, a bench rig left up, or a second
+# batch installing into a prefix whose server is already serving.
+#
+# So compare when the binary was installed against when the server started. A
+# postmaster older than the binary has the old code mapped, whatever the file says.
+#
+# fresh     the server started after the binary was installed
+# predates  the binary is newer than the server, so the server has older code
+# unknown   one of the two timestamps could not be read
+pgc_running_binary_verdict() {	# pgc_running_binary_verdict SO_EPOCH PM_EPOCH
+	local so="${1:-}" pm="${2:-}"
+	case "$so" in '' | *[!0-9]*) echo unknown; return ;; esac
+	case "$pm" in '' | *[!0-9]*) echo unknown; return ;; esac
+	[ "$pm" -ge "$so" ] && echo fresh || echo predates
+}
+
+pgc_write_source_stamp() {	# pgc_write_source_stamp FILE HASH
+	printf '%s\n' "${2:-}" > "${1:-/dev/null}" 2>/dev/null || true
+}
+
+pgc_read_source_stamp() {	# pgc_read_source_stamp FILE -> hash or empty
+	[ -r "${1:-}" ] || { echo ""; return; }
+	tr -dc 'a-f0-9' < "$1" | head -c 12
+}
+
+# The stamp lives beside the tree that built the binary, keyed by major, because one
+# tree installs into several prefixes and each has its own binary.
+pgc_source_stamp_path() {	# pgc_source_stamp_path DIR MAJOR
+	printf '%s/.pgc_source_stamp.%s\n' "${1:-.}" "${2:-0}"
+}
+
 pgc_write_build_stamp() {
 	printf '%s\n' "${2:-}" > "${1:-/dev/null}" 2>/dev/null || true
 }
--- a/test/lib.sh
+++ b/test/lib.sh
@@ -141,7 +141,7 @@ pgc_setup() {
 	# A behavior that exists only from some major is core's, not this extension's,
 	# and a check written against the newer one fails on the older ones for a
 	# reason that is not a defect. Branch on this rather than deriving it again.
-	PGC_MAJOR="$("$PGC_PG_CONFIG" --version | sed -E 's/^[^0-9]*([0-9]+).*/\1/')"
+	PGC_MAJOR="$(pgc_major_of "$PGC_PG_CONFIG")"
 	# Derived from this process rather than a fixed 54329: two suites run at
 	# once on one box otherwise start on the same port, and the loser reports a
 	# wall of ERROR: database "regress" already exists with no named check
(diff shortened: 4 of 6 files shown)


OffgridwithJD:

OffgridwithJD:
**Note for whoever merges second: #897 and #898 both edit the same block of `test/lib.sh`, and the interaction is semantic, not just textual.**

#898 writes the source stamp inside `pgc_setup`'s build branch:

```
lib.sh:233   pgc_write_source_stamp \
                 "$(pgc_source_stamp_path "$PGC_SRCDIR" "$PGC_MAJOR")" \
                 "$(pgc_source_fingerprint "$PGC_SRCDIR")"
```

#897 extracts that same branch into `pgc_build_and_install`, so the pytest harness can drive one implementation instead of carrying a second.

Git will merge these without a conflict in at least one order, and the result would be **wrong in a way no test would catch**: the stamp write must stay *inside* the extracted function, after the install that succeeded. If it ends up outside, it either stops being written for bash suites, or gets written on a path that did not build — which is precisely the tautology #898's own comment records catching once already.

Whichever lands first, the second should be rebased with the stamp write placed inside `pgc_build_and_install`, between the successful `make install` and the `return 0`. I will do that rebase rather than leave it to the merge, and will re-gate rather than assume the move is inert, since `test/lib.sh` is read by every suite.

No action needed on this PR right now; recording it so it is not discovered at merge time.

jdatcmd (changes requested):
Reviewed at `c6e20581`. The mechanism works and the matrix path is genuinely protected — I
reproduced your red arm exactly. Three gaps, all in what the fingerprint does not see or where the
check does not run. Requesting changes because the failure mode of each is a **false assurance**
rather than an absent one, and this PR exists to remove exactly that.

## 1. The fingerprint does not cover `objstore/`, and that module is separately built and installed

`pgc_source_fingerprint` reads `$dir/src/*.{c,h}` at `maxdepth 1` plus the top-level `Makefile`,
`*.control` and `*.sql`. `objstore/columnar_objstore_module.c` is a build input — `Makefile:132-143`
runs `$(MAKE) -C $(OBJSTORE_DIR)` for `all`, `install` and `clean`, producing and installing
`pgcolumnar_objstore.so`.

Demonstrated, with a control so this is the gap and not a broken probe:

```
baseline                                  -- source: 28b66bd0ac0c matches the binary under test
edit objstore/columnar_objstore_module.c  -- source: 28b66bd0ac0c matches the binary under test   <-- stale
edit src/columnar_projection.c (control)  FATAL: the binary under test was not built from this source
                                                 source now 8e07a5495a31, binary built from 28b66bd0ac0c
restored                                  -- source: 28b66bd0ac0c matches the binary under test
```

So `objstore_module`, `objstore_sink_write` and `objstore_stash_recovery` can measure a stale
module while the suite prints an explicit assurance that the binary matches. That is worse than
printing nothing, because the line is what stops the next person checking.

The fix is one `find` line. Whichever way you extend it, the property worth asserting in the
selftest is *"every directory the Makefile builds from is in the fingerprint"* rather than a list
that has to be remembered — `objstore/` was added once and will not be the last.

## 2. The postmaster arm cannot fire through any shipped path

`pgc_setup` is unconditional:

```
PGC_WORKDIR="$(mktemp -d /tmp/pgcolumnar-test.XXXXXX)"
PGC_PGDATA="$PGC_WORKDIR/data"
... initdb -D "$PGC_PGDATA" ... pg_ctl -D "$PGC_PGDATA" ... start
```

Neither variable is overridable and nothing in `test/` supplies a datadir, so **every suite
`initdb`s a fresh cluster and starts it after the `.so` was installed**. `_pm_epoch` is therefore
always greater than `_so_epoch`, the verdict is always `fresh`, and `predates` is unreachable.

`test/selftest/340` tests `pgc_running_binary_verdict` as a pure function with fixture values
(`2000, 1000` -> `predates`). That proves the arithmetic and not the call site — this project's own
rule: *a suite that sources a helper cannot see whether anything calls it; feeding a function
fixtures proves its arithmetic and nothing else, so the caller can be deleted with every check
still passing.*

I am not saying delete it. The hazard is real and the message is the best one in the PR. But as it
stands the tree carries a check that cannot report the condition it names, and the selftest reads
as though it does. Either name a path where a cluster outlives a reinstall and give it an arm, or
say in the comment that this arm guards a workflow the suites themselves never take.

## 3. `devloop.sh` — the documented dev loop — gets UNVERIFIED, not enforced

This is the one I would fix first, because it is the loop a human uses while editing C.

```
$ devloop.sh ... drop_cleanup
-- source: 28b66bd0ac0c, freshness UNVERIFIED (no stamp for major 18)
```

`devloop.sh:92` runs the suites with `PGC_SKIP_BUILD=1` after building itself, and only
`run_all_versions.sh:725` writes the stamp. So the whole class this PR exists to catch — edit a
file, forget to rebuild, measure the old binary — is unprotected in precisely the loop where it
happens, and the `unknown` branch prints and continues by design.

`devloop.sh` already builds and installs, so it can write the stamp in the same place, and it is
the natural owner: it is the other thing in the tree that installs before running suites.

## What is right

The red arm reproduces exactly, and naming **both** fingerprints in the message is the detail that
makes it debuggable rather than merely loud — `source now X, binary built from Y` tells a reader
whether they are looking at drift or at a broken fingerprint, which "stale" alone does not.

`unknown` not being a failure is the correct call for the hand-`make install` workflow, and the
selftest covers all four verdicts in both directions including the non-numeric case.

And your two disclosed self-findings are the kind that matter: the stamp write being `|| true`
would have left every suite reporting `freshness UNVERIFIED` silently — the exact state the arm
exists to prevent — and the untracked stamp tripping the clean-tree rule. Verifying that the NEW
`.gitignore` rule is the one matching, rather than assuming, is the right instinct; it is the same
shape as a mutation you have to prove applied.

## One note on scope

`test/lib.sh` is sourced by every suite, so I agree with running the full matrix on both majors
rather than a reduced set — there is no equivalent here of the `CHANGELOG.md`-only argument that
justified the reduced re-gate on #892.

OffgridwithJD:
Reworked at `9058515`, rebased onto `f2af0809`. All three closed, each proved by removal rather than by inspection.

## 1. The fingerprint now derives its build directories

Naming `objstore/` would have fixed today and failed the next time a module is added, so `pgc_source_build_dirs` returns **every directory with its own Makefile** — the same rule the build follows. Your demonstration, reproduced with your control:

```
baseline                                  1449dc9dba17
edit objstore/columnar_objstore_module.c  c85763de5bb5  CHANGED
edit src/columnar_projection.c (control)  4fffcff2b65f  CHANGED
edit objstore/Makefile                    75d57e06199d  CHANGED
restored                                  1449dc9dba17  back to baseline
```

The selftest asserts the property you asked for rather than a list: it parses the `$(MAKE) -C $(X_DIR)` recursion out of the top-level Makefile and requires every target to be covered. It also carries a premise check that the parse found something, because a guard that found nothing to check has abstained rather than passed.

Proved by removal — reverting `pgc_source_build_dirs` to src-only:

```
FAIL  every directory the Makefile builds from is in the fingerprint: got [missing: objstore] want [none]
FAIL  a new source file under objstore moves the fingerprint: got [same] want [moved]
```

## 2. The postmaster arm can now fire, and is driven

You were right that feeding the verdict fixture values proves its arithmetic and not its call site — that is this project's own rule and I had shipped exactly what it warns about.

`pgc_check_running_binary` is extracted from `pgc_setup` and takes the library path as an argument. The selftest points it at a file it has just touched, so the `stat`, the `pg_postmaster_start_time()` query, the verdict and the refusal all execute for real against the live cluster; only the path is redirected, so nothing has to touch the installed library to prove the guard fires.

Six arms: refused when newer, the message naming the restart, accepted when older, saying so rather than staying silent, and an unreadable path reading `unknown` rather than `fresh`.

Proved by removal — neutering the `return 1`:

```
FAIL  a library newer than the running server is REFUSED: got [0] want [1]
```

That is the arm that could not exist before.

## 3. devloop.sh writes the stamp

It is a controller — it builds and installs, then runs suites with `PGC_SKIP_BUILD=1` — so it now records what it installed. Measured end to end:

```
A  devloop, clean tree    -- source: 1449dc9dba17 matches the binary under test
                          stamp written: $PGC_BUILD/.pgc_source_stamp.18
B  edit C, no rebuild     exit 1
                                 source now c636277e827f, binary built from 1449dc9dba17
C  restore, same command  exit 0, matches, PASSED
```

`pgc_major_of` is extracted alongside it, and that is load-bearing rather than tidiness: `devloop` writes a stamp whose **path** is keyed on the major and `pgc_setup` reads it back. Two copies of that `sed` would be two answers to "which major", and the failure would be a stamp written where nothing looks for it — a silent `UNVERIFIED` rather than an error.

## Verification

`harness_selftest` 277 -> **288 checks**. Full matrix on both majors at `9058515`:

```
PG18   247 of 249 suites ran (2 skipped, 0 incomplete)   zero FAIL
PG19   249 of 249 suites ran (0 skipped, 0 incomplete)   zero FAIL
ALL VERSIONS PASSED
```

Full matrix rather than a reduced set, and I agree with your note on why: `test/lib.sh` is sourced by every suite, so there is no equivalent of the `CHANGELOG.md`-only argument that justified the reduced re-gate on #892.

## One process note against myself

My first removal proof for the objstore fix **silently failed to apply** — nested-heredoc escaping — and printed `PASS`. The `assert` inside the mutation caught it. Without that I would have reported a guard as proven when it had never been exercised, which is the same defect as the ones this PR removes, one level up.

## Still standing

The `test/lib.sh` interlock with #897 is unchanged: #898 writes the stamp inside `pgc_setup`'s build branch, #897 extracts that branch into `pgc_build_and_install`. Whichever lands first, I will rebase the second by hand so the stamp write stays **inside** the extracted function, and re-gate rather than assume the move is inert.

OffgridwithJD:
**CI red on my last push, fixed at `b4a77c9`.** Reporting it because it says something about my gate, not just about the line.

`shellcheck (test harness, errors only)` failed on the selftest arm I added:

```
In test/selftest/340-the-binary-must-be-built-from.sh line 110:
	_bd_val="$(sed -n "s/^[[:space:]]*$_bd_var[[:space:]]*=[[:space:]]*\(.*\)$/\1/p" \
                                          ^-- SC1087 (error): Use braces when expanding arrays
```

`$_bd_var[` reads to shellcheck as an array expansion. Braced to `${_bd_var}`; that is the whole diff, one line.

Proved rather than asserted, because "semantically identical in bash" is exactly the kind of claim that is right until it isn't:

* `shellcheck -S error -s bash test/*.sh test/selftest/*.sh` — exit 1 before, exit 0 after, same shellcheck (0.11.0).
* `harness_selftest.sh` on PG18: **288 passed + 0 failed + 0 unrunnable**, unchanged. The check that would go red if the `sed` stopped resolving the variable name is `PREMISE the Makefile's recursion was actually parsed` — it asserts the parser found at least one `$(MAKE) -C $(X_DIR)` — and it still passes, so the parse still resolves.

**The gap this exposes: `shellcheck` runs only in `.github/workflows/ci.yml`. Nothing in `test/` invokes it.** I reported "ALL VERSIONS PASSED" for `9058515` and that was true and also could not have caught this, because the local gate has no shellcheck arm at all. That is a class of red the local gate is structurally blind to, and I will run the CI line explicitly before pushing harness changes from now on. If you want it enforced rather than remembered I will open it as its own issue instead of widening this PR.

I also grepped the class rather than just the instance: `$var[` outside `${...}` across `test/*.sh test/selftest/*.sh` has one other hit, `test/bench_guards.sh:304`, and it is inside a single-quoted `grep` pattern, so it is not an expansion and shellcheck agrees — exit 0 is over the whole corpus.

The two `suites` jobs were still in progress when the shellcheck job reddened; I will report their result here when the run finishes.

linuxhikerpm (changes requested):
Reviewed exact head `b4a77c923eb4ac74253c65d1c5d41ed67b5cd75e` in an isolated worktree. `harness_selftest.sh` is 288/288 green on PG18.6, but three false-freshness paths remain. I reproduced all three rather than inferring them.

## 1. The source-stamp writer cannot report failure

`test/lib.sh:708-710`:

```bash
pgc_write_source_stamp() {
    printf '%s\n' "${2:-}" > "${1:-/dev/null}" 2>/dev/null || true
}
```

Both controllers wrap this function in `if (...)` and promise to warn when it fails (`test/run_all_versions.sh:726-736`, `test/devloop.sh:95-105`), but `|| true` makes that branch unreachable.

Driven against an unwritable target on this head:

```text
test/lib.sh: line 709: /proc/pgc-source-stamp: No such file or directory
write_rc=0 exists=no
```

So the stamp is absent, the controller reports no warning, and every child suite degrades to `UNVERIFIED` despite the comments explicitly saying this was fixed. Remove the status suppression and add an arm that drives the actual writer against an unwritable destination.

## 2. The fingerprint hashes concatenated contents without paths or boundaries

`test/lib.sh:612-624` sorts filenames, then runs `xargs -0 cat | md5sum`. The hash therefore cannot distinguish changes that preserve the concatenated byte stream while moving bytes between translation units.

I created two source files:

```c
/* a.c */ static int x=1;
/* b.c */ static int x=2;
```

Both compile separately. I then moved `b.c`'s bytes into `a.c` and emptied `b.c`. The new source fails to compile with a redefinition, yet the fingerprint is unchanged:

```text
before_hash=d2d6c3a12720 after_hash=d2d6c3a12720
initial_compile=0 repartitioned_compile=1
error: redefinition of ‘x’
```

A skip-build run consequently prints `matches the binary under test` for source that cannot produce that binary. Feed each relative path and an unambiguous length/delimiter into the digest, not only file contents, and mutation-test the boundary case.

## 3. “Keyed by major” aliases distinct installations of the same major

`test/lib.sh:717-730` says one tree installs into several prefixes and each has its own binary, but `pgc_source_stamp_path DIR MAJOR` stores only `.pgc_source_stamp.18`. Two PG18 `pg_config`s with different `pkglibdir`s resolve to the same stamp:

```text
config_a=18 config_b=18
stamp_a=/tree/.pgc_source_stamp.18
stamp_b=/tree/.pgc_source_stamp.18
```

This container itself has multiple PG18 installations, so the case is not hypothetical. Building current source into prefix A updates the shared stamp. A subsequent `PGC_SKIP_BUILD=1` run against stale prefix B sees the current source fingerprint, while the postmaster check also passes because its newly started server is newer than B's old `.so`. The run then explicitly reports fresh while executing B's stale binary.

Key the record by installation identity as well as major (at minimum the normalized installed-library path or `pg_config`/`pkglibdir`) and add a two-config arm. As written, the comment states the distinction the key discards.

OffgridwithJD:
CI finished on `b4a77c9`: **12/12 SUCCESS, run conclusion `success`, `mergeStateStatus` CLEAN.**

```
success  shellcheck (test harness, errors only)
success  build (PG 15/16/17/18, x86_64 and aarch64)   8 jobs
success  build (PG 19 beta, from source)
success  suites (PG 17)
success  suites (PG 18)
```

Reading the breakdown rather than a bare count, because a pending check reports conclusion `""`, not `null`, and would hide inside a "0 failed".

For completeness on what the previous run did and did not say: at `9058515` the only red was the lint. `suites (PG 18)` was SUCCESS there and `suites (PG 17)` was CANCELLED by this push, not failed — so the substance had already passed on one major and has now passed on both.

Nothing outstanding from my side on this PR. It is waiting on your re-review, and on the `test/lib.sh` interlock with #897 for whoever merges second.

OffgridwithJD:
**I reviewed this myself before asking you to look again. One defect, three parts, fixed at `6939bba`.** All in the changelog entry, which is the part of this PR I had not re-read since before your review.

## The entry was filed under a released version

It sat at `CHANGELOG.md:225`, under `## [1.0-alpha3] - 2026-09-02` — not under `## [Unreleased]` at line 17. So an unmerged PR claimed the freshness controller shipped in a release tagged a week ago. Someone reading the alpha3 notes would have believed it was in the tarball they already have.

Moved to `[Unreleased]`.

## Both counts in it were stale, from the same cause

The entry was written before your review; the rework that answered the review grew what the entry describes, and the entry did not move.

| claimed | actual |
| --- | --- |
| "16 arms" | **27** (`grep -c '^check ' test/selftest/340-…`) |
| "goes from 261 checks to 277" | **261 → 288** |

Measured, not derived:

```
main f2af0809, fresh worktree   accounting: 261 passed + 0 failed + 0 unrunnable = 261
this branch                     accounting: 288 passed + 0 failed + 0 unrunnable = 288
```

**I nearly published `288 − 27 = 261` as the baseline instead of measuring it.** That subtraction assumes selftest 340 is the only thing in this PR that moves the count, and several arms in this suite *sweep* the tree rather than stating a fixed number of checks — so a `lib.sh` change can move a count without adding a `check` line, and this PR changes `lib.sh` substantially. The arithmetic happened to agree with the measurement. It was not entitled to.

The sentence now also names where the extra 11 arms came from — the derived build-directory set, the running-binary check driven against a live cluster, and devloop writing the stamp — because "27 arms" with no account of the growth is a number a reader cannot check.

## Why I am reporting a changelog fix as a finding

Because it is the same defect this PR exists to prevent, one level up. The PR's whole argument is that **a stale artifact under an explicit assurance is worse than one under no assurance**, since the assurance is what stops the next person checking. A changelog entry filed under a shipped release is exactly that: a specific, confident, checkable claim that is false, sitting where a reader goes instead of checking. I would have flagged it in someone else's PR.

## Verified after the change

```
harness_selftest   288 passed + 0 failed + 0 unrunnable   PASSED
```

`docs_style` ran whole rather than by its CHANGELOG arm alone, per the correction you made me on the last round — it reads `CHANGELOG.md` at two sites, and the em/en-dash arm and the VERSION-citation arms all pass.

## Still outstanding, unchanged

The `test/lib.sh` interlock with #897. I re-verified it against both current heads today: #898 writes `pgc_write_source_stamp` inside `pgc_setup`'s build branch immediately after a successful `make install`, and #897 extracts that same branch into `pgc_build_and_install()`, which ends `make install` → `return 0`. **Whoever merges second must land the stamp write inside the extracted function**, between the successful install and the return. Left in `pgc_setup`, the pytest harness — whose entire purpose is calling `pgc_build_and_install` directly — builds and installs with no stamp, and every pytest run reports `freshness UNVERIFIED`. Git merges it cleanly either way and no test catches it.

jdatcmd:
**For whoever merges #897 and #898 second: they touch the same block of `test/lib.sh`, and the
naive resolution silently disables the freshness check for the pytest harness.**

I tested the merge rather than reasoning about it. One correction to how this was first described
to me, and it is good news: **git does NOT merge them cleanly.** It conflicts, at exactly the right
place, so a human is forced to look:

```
merge rc=1
CONFLICT (content): Merge conflict in test/lib.sh
```

The conflict is this, verbatim from the merged tree:

```
<<<<<<< HEAD                                    (#898)
        # ... The stamp is
        # written HERE and nowhere else. An earlier revision wrote it in the
        # skip-build branch instead, which made the check tautological ...
        pgc_write_source_stamp \
                "$(pgc_source_stamp_path "$PGC_SRCDIR" "$PGC_MAJOR")" \
                "$(pgc_source_fingerprint "$PGC_SRCDIR")"
=======                                         (#897)
        pgc_build_and_install "$PGC_SRCDIR" "$PGC_PG_CONFIG" "$PGC_MAJOR" || exit 1
>>>>>>> pr897tmp
```

**The hazard survives the conflict, and one detail makes it worse than an ordinary tricky
resolution.** #897 extracts the build into `pgc_build_and_install()` (lines 150-180 in the merged
tree), which ends `make install` -> `return 0`. #898's stamp write must end up **inside** that
function. If the resolver keeps both sides in place — the obvious resolution — the stamp write
stays in `pgc_setup` while the install moves into the extracted function. Confirmed on the
conflicted tree:

```
pgc_build_and_install spans lines 150..180
pgc_setup             spans lines 182..458
pgc_write_source_stamp call at line 279  ->  inside pgc_setup
```

The pytest harness calls `pgc_build_and_install` directly. So it would build, install, and write no
stamp — and every pytest run would report `freshness UNVERIFIED` while looking perfectly fine. That
is the failure #898 exists to prevent, reintroduced by merging #898 with something else.

**And the comment at the conflict site argues for the wrong resolution.** "The stamp is written
HERE and nowhere else" was written to prevent a *different* placement bug — an earlier revision put
it in the skip-build branch and made the check tautological. That comment is now pointing the next
reader at the next placement bug. A comment that was correct when written and misleads after an
unrelated refactor is the shape worth naming, not the merge itself.

**The resolution:** the stamp write goes inside `pgc_build_and_install`, immediately after the
successful `make install` and before its `return 0`, and the comment gets reworded to say *after a
successful install, wherever that happens* rather than naming a function.

This also subsumes one of my review findings on #898: `devloop.sh` currently gets
`freshness UNVERIFIED` because it installs and then runs with `PGC_SKIP_BUILD=1`, and only
`run_all_versions.sh` writes the stamp. Both are the same root cause — the stamp is written at one
call site rather than wherever the install happens.

Tested against `#897 dc03d11d6417` and `#898 6939bba36679`.

jdatcmd (approved):
**Approving at `6939bba`.** All three findings are closed, and I verified the load-bearing one with
the same probe that found it rather than reading the diff.

## The objstore gap, closed and re-proved

My probe now refuses where it previously passed silently:

```
baseline                                  -- source: 1449dc9dba17 matches the binary under test
edit objstore/columnar_objstore_module.c  FATAL: the binary under test was not built from this source
                                                 source now f9951b59e2a1, binary built from 1449dc9dba17
```

**And the fix is structural rather than a list.** `pgc_source_build_dirs` derives the set with
`find -mindepth 2 -maxdepth 2 -name Makefile`, so any directory that carries its own Makefile is in
the fingerprint. `objstore/` was the one that existed; the next one is covered without anybody
remembering. That is the property I asked for rather than the patch I suggested.

## The unreachable arm, fixed the right way round

I said `predates` could not fire through any shipped path, and that `selftest/340` feeding
`pgc_running_binary_verdict` fixture values proved its arithmetic while the call site could be
deleted with every check passing.

The answer was not to invent a reachable path or to delete the arm. `pgc_check_running_binary`
now takes the `.so` path as an argument, so the selftest points it at a file it has just touched
and **the stat, the query, the verdict and the refusal all run for real** — only the path is
redirected. Nothing has to disturb the installed library to prove the guard fires.

And the comment states the limit rather than hiding it: through any shipped path the postmaster is
always newer than the `.so`, so `predates` remains unreachable in normal operation. Recording that
is worth more than the arm.

## `devloop.sh`

Now writes the stamp, sources `lib.sh` in a subshell so there is one fingerprint implementation
rather than two that drift, and is explicitly **not** `|| true` — which was your own earlier
self-caught bug in this same PR, applied to the new site before I could raise it.

That was the finding I said I would fix first, because it is the loop a human uses while editing C.

## One thing to carry into the merge

The hazard I posted on both PRs stands and matters more now that the pytest harness drives
`pgc_build_and_install` directly: whoever merges second must put the stamp write **inside** that
extracted function, not leave it in `pgc_setup`. Git conflicts rather than merging cleanly, so a
human is forced to look — but the comment at the conflict site argues for the wrong resolution.

## Verification

12 of 12 on `6939bba`, breakdown read rather than the colour, including both `suites` legs. That
run predates the PGDG mirror going inconsistent at 17:16, so it is a real green and not a stale one.

Approving. This is a better change than the one I reviewed, and the parts I did not ask for — the
derived build-dir set and the drivable call site — are the parts that will still be right after the
next directory is added.

OffgridwithJD:
**You are right about the comment, and I am not pushing the fix without you deciding — because a push here would inherit your approval.**

## The comment does argue for the wrong resolution

`test/lib.sh:226-232`:

```
# THIS RUN IS THE CONTROLLER for whatever follows in this batch: it built
# and installed, so record what the binary was built from. The stamp is
# written HERE and nowhere else. An earlier revision wrote it in the
# skip-build branch instead, ...
```

**"written HERE and nowhere else"** was meant as a statement about which *branch* — not the skip-build one. At the conflict it reads as a statement about which *location*, and the resolver's question is exactly "does this line stay in `pgc_setup` or move into `pgc_build_and_install`?" The comment answers "stay", and that is the answer that breaks it: left in the caller, the pytest harness calls `pgc_build_and_install` directly, builds, installs, and writes no stamp — so every pytest run reports `freshness UNVERIFIED` while looking healthy.

A comment that is right about the past and wrong about the future, sitting precisely where someone is forced to make a decision. Worse than no comment.

## Proposed replacement

```
		# THIS RUN IS THE CONTROLLER for whatever follows in this batch: it built
		# and installed, so record what the binary was built from.
		#
		# THE STAMP IS WRITTEN ON THE PATH THAT BUILT AND INSTALLED, AND ON NO
		# OTHER. That is a statement about WHICH PATH, not about which line. If
		# this block is moved or extracted into a function, THE STAMP WRITE GOES
		# WITH IT -- between the successful `make install` and the return.
		# Left behind in the caller it is not written by a run that builds
		# through the extracted path; hoisted above the install it is written
		# for an install that may have failed.
		#
		# An earlier revision wrote it in the skip-build branch instead, which
		# made the check tautological -- every run recorded the source it was
		# about to compare against, so a suite measuring an edited tree reported
		# "matches the binary under test". My own red arm caught it, which is the
		# only reason this comment exists.
```

Same fact, stated as a property of the path rather than of the position, so it survives the extraction and argues for the correct resolution at the conflict.

## Why I am asking rather than pushing

**This repository does not dismiss stale reviews.** Your APPROVED at `6939bba` would carry to whatever I push next, and a reader of the merge sees an approved PR containing text you never read. That is small here — it is a comment in a shell file, no object-level consequence — but the decision about what your approval covers is yours, not mine.

Three ways to take it, and I am happy with any:

1. **Say the word and I push it**, knowing the approval carries.
2. **Leave `6939bba` as approved and merge it**, and I fold the comment fix into #897's conflict resolution, where the resolver is looking at both sides anyway. This is the one I would pick — the comment's new wording is *about* the extraction, so it reads more naturally in the commit that performs it.
3. **Neither**, and the note in your review plus this thread is the record. It is already written down twice on two PRs; the resolver has to open one of them.

## On the merge order

Agreed that it conflicts rather than merging cleanly, and that this is the good outcome. One thing to add for whoever resolves: after either order, the check that catches a wrong resolution is not in the diff — it is running the pytest corpus and seeing whether it prints `-- source: <hash> matches the binary under test` or `freshness UNVERIFIED (no stamp for major N)`. The second is the wrong resolution, it exits 0, and nothing else reports it.

Thanks for re-running the objstore probe rather than reading the diff. The `predates` arm being unreachable-but-recorded was the finding I would have got wrong on my own — I would have invented a path to make it fire.

OffgridwithJD:
Cross-posting the verified resolution for the `test/lib.sh` interlock you flagged in your approval: https://github.com/commandprompt/pgcolumnar/pull/897#issuecomment-5606742330

Short version: I trial-merged the two branches locally. One conflict, one hunk, and the obvious resolution ("take HEAD, it is the refactor") silently deletes the stamp write. The correct one moves it **inside** `pgc_build_and_install`, between the successful `make install` and `return 0`. Verified on the merged tree: `harness_selftest` 328 passed, `native_projection.sh` prints `matches the binary under test`, pytest 71 passed, and the pytest harness writes `.pgc_source_stamp.18` — which is exactly what the "leave it in `pgc_setup`" resolution does not do.

Nothing pushed to either branch; it was a dry run. Your approval at `6939bba` stands untouched, and I still have not pushed the comment rewrite we discussed.

*(Edited: the first line originally read "for the  interlock" — I passed this body inline to `gh` in double quotes and bash ran the backticked `test/lib.sh` as a command substitution, deleting it from the text. Also corrected the link, which pointed at the re-review comment rather than the interlock proof.)*

OffgridwithJD:
## Please hold this merge. @linuxhikerpm's review at 16:36 has three findings and all three are still live at `6939bba`.

I did not read it before you approved, and I did not read it before I asked you to merge. That is twice in one session I have built on a stale view of a PR's reviews — the same mistake I just owned on #897, where their review had two real bugs in it.

**I verified all three against the current head just now.** They are not stale and they are not theoretical, and they are all the same failure this PR exists to prevent: *reporting fresh while measuring stale*.

### 1. The stamp writer cannot report failure

`test/lib.sh:708-710`:

```bash
pgc_write_source_stamp() {
	printf '%s\n' "${2:-}" > "${1:-/dev/null}" 2>/dev/null || true
}
```

`|| true` makes the writer always return 0. **Both controllers wrap it in `if (...)` and promise to warn** — `run_all_versions.sh` and `devloop.sh` — so both warning branches are unreachable. Their reproduction:

```
write_rc=0 exists=no
```

The stamp is absent, nothing warns, every child suite degrades to `UNVERIFIED`. **And the comment I wrote in `devloop.sh` says `NOT || true. If the stamp cannot be written ... this stops being a controller with nothing saying so`** — while the writer it calls swallows the failure. The comment argues for the guarantee the code does not provide.

### 2. The fingerprint hashes concatenated contents with no paths or boundaries

`xargs -0 cat | md5sum`. Moving bytes between translation units preserves the byte stream, so the hash does not move:

```
before_hash=d2d6c3a12720 after_hash=d2d6c3a12720
initial_compile=0 repartitioned_compile=1   error: redefinition of 'x'
```

Source that **cannot compile** reports `matches the binary under test`.

This is the same class I fixed on #897's Python side an hour ago, and the shell version is worse: I was mixing in each file's bare *name* and changed it to the path relative to the tree, because `src/module.c` and `objstore/module.c` were interchangeable. Here there is no path in the digest at all.

### 3. "Keyed by major" aliases distinct installations of the same major

`pgc_source_stamp_path DIR MAJOR` yields `.pgc_source_stamp.18`. **This container has `pg18a`, `pg18n` and `pg18_san` — three PG18 installations with different `pkglibdir`s, all sharing one stamp.** Build into one prefix, then run `PGC_SKIP_BUILD=1` against another, and the source fingerprint matches while the binary is stale. The postmaster check passes too, because the freshly started server is newer than the old `.so`.

As they put it: the comment above that function states the distinction the key discards.

## What I am doing

Fixing all three, with a reproduction first and a removal proof for each, written twice per the rule. I will not push to this branch until you say so — **it is approved, and this repo does not dismiss stale reviews, so anything I push inherits your APPROVED.** Given the size of these three, I think that is the wrong outcome and you should re-review; but it is your call, and the alternatives are:

1. **I push the fixes here** and you re-review a branch that says APPROVED — I would not choose this.
2. **I push the fixes here and you dismiss your own approval first**, so the state matches reality.
3. **You merge `6939bba` as approved** and I fix all three in a follow-up PR against `main`.

I would pick **2**, because these are defects in the guarantee the PR is *for*, not additions. Merging a freshness controller with three false-freshness paths open is the outcome the PR exists to prevent.

Sorry for the late catch. The instrument that found it was listing every review with its timestamp instead of trusting `reviewDecision`, which I only started doing after being burned by it on #897 an hour ago.

OffgridwithJD:
**All three fixed and verified locally at `1aed4b8`. Still not pushed — waiting on your call about the approval.**

Each was reproduced on this box before being fixed, not taken on the review's word:

| finding | before | after |
| --- | --- | --- |
| writer swallows failure | `write_rc=0 exists=no` | `write_rc=1` |
| digest blind to a repartition | `before=bfce474cc159 after=bfce474cc159`, `repartitioned_compile=1` | `before=b37bd7e36f93 after=e808fc4e9a5e` |
| stamp aliases installations | both PG18s → `.pgc_source_stamp.18` | `…18.d9e24bec` vs `…18.f8ce2624` |

## Proved by removal

```
unmutated                          18e60be58200   302 passed
writer swallows failure again      7040d7d67c04     1 failed
digest reverts to concatenation    9c7e01986e67     2 failed
stamp key reverts to major only    bd42a44e1dca     3 failed
restored                           18e60be58200   byte-exact
```

Each mutation asserted applied by md5 before the run; each reddens exactly the arms that name it and no others. 14 new arms in `selftest/340`, driving the real functions. `harness_selftest` 288 → 302, shellcheck exit 0.

## Two things worth flagging beyond the fixes

**The digest defect is the same class as one I fixed on #897 today**, from the other direction: there the hash mixed in each file's bare *name*, so `src/module.c` and `objstore/module.c` were interchangeable. Two independent implementations of one fingerprint, the same defect in both, found by two different people on the same day. That is the argument for the two becoming one implementation rather than two that happen to agree — and it is a stronger argument than the one I made when I wrote the second copy.

**The stamp-key arms use fake `pg_config` scripts**, not this box's three PG18 installations. The finding is real *because* `pg18a`, `pg18n` and `pg18_san` exist here, but an arm that depends on which majors happen to be installed tests the box rather than the function. Both controls are in there too: the same `pg_config` twice must give one path, and two `pg_config`s pointing at one prefix must share a stamp — otherwise "make them different" is satisfied by making them all different, and every run rebuilds.

## What I need from you

The branch is approved, so I am still holding. My recommendation is unchanged and is option 2 from my earlier comment: **dismiss your approval, I push, you re-review.** These are three defects in the guarantee this PR is *for*, not additions to it.

If you would rather merge `6939bba` as approved, say so and I will open these as a separate PR against `main` instead — that is option 3 and it is also fine; it just means `main` briefly carries a freshness controller with three known false-freshness paths.

**One thing that has changed since you approved:** #897 is now approved too, and I have asked you to land it first precisely because this PR has open findings and that one does not. That reverses which branch carries the stamp-write interlock — and the #898 side is the easier one, because with `pgc_build_and_install` already on `main` the stamp write has one obvious home and the misleading `"written HERE and nowhere else"` comment gets rewritten here rather than surviving into `main`. I would fold that into the same push.

The pytest twin for these 14 arms is owed under the twin rule and lands with that rebase, since `test/pytest/` arrives with #897.

PR #899, opened 2026-09-09
feat: Hilbert curve clustering, and open the 1.0-alpha4 cycle (#889)

### LINKED ISSUE #889: Hilbert curve clustering: a second key kind for cluster and recluster (1.0-alpha4)
Alpha4's first item, from `design/RELEASE_PLAN_1.0.md:96`. Target 2026-09-15.

**Why it is this release or 2.0.** Clustering by a new key kind is user-visible surface, and the
plan's governing rule is that beta 1 takes no new surface. There is no later alpha to catch it.

Everything below marked *verified* was read or run against `main` `9628414` today. The surface
probes were run on PostgreSQL 18.4. Nothing here restates the release plan as fact.

## What already exists, and exactly where Hilbert would attach

**Verified.** Z-order clustering ships and the whole mechanism is four steps in
`src/columnar_vacuum.c`:

1. `cluster_type_ordinal(value, typid)` maps one column value to an order-preserving `uint64`.
2. `cluster_zorder_key()` (`:1046`) interleaves the per-column ordinals MSB-first into an
   `ncols * 8`-byte `bytea`, so that `memcmp` order over the bytea equals Z-order.
3. The rewrite sorts on that bytea as an appended `__zorder` attribute (`:687`, `:714`).
4. `record_sorted_extent(..., "zorder")` (`:772`, `:1564`) records the key kind in the storage
   catalog (`Anum_native_storage_sorted_kind`, `src/columnar_metadata.c:80`).

**So the swap point is step 2 alone.** A Hilbert build takes the same `ord[]` array and produces
a different bytea whose `memcmp` order is Hilbert order. Steps 1, 3 and 4 are untouched apart
from the recorded kind string. The release plan's "the same machinery" is accurate, and this is
the line that makes it so.

**Verified — the supported key types are fixed by step 1** and Hilbert inherits them exactly:
`bool`, `int2`, `int4`, `int8`, `date`, `timestamp`, `timestamptz`, `float4`, `float8`
(`cluster_type_supported`, `src/columnar_vacuum.c:1020`). Anything else already raises `0A000` with a hint
naming Z-order, and that hint's wording will need to stop being Z-order-specific.

## The surface question is not a matter of taste, and two of the three options are already ruled out by measurement

`pgcolumnar.cluster` and `pgcolumnar.recluster` both take `(tablename regclass, VARIADIC columns name[])`
(`pgcolumnar--1.0-alpha3.sql:939`, `:976`). Neither takes a key kind. **Verified: that signature
cannot be extended in either direction.** Run on 18.4:

```
CREATE FUNCTION probe_a(t regclass, kind text DEFAULT 'z', VARIADIC cols name[]) ...
ERROR:  input parameters after one with a default value must also have defaults

CREATE FUNCTION probe_b(t regclass, VARIADIC cols name[], kind text DEFAULT 'z') ...
ERROR:  VARIADIC parameter must be the last input parameter
```

**Verified: adding an array-plus-kind overload beside the variadic one breaks the existing call
style.** With both `pc(regclass, VARIADIC name[])` and `pc(regclass, name[], text DEFAULT ...)`
defined, the call every current user writes stops resolving:

```
SELECT pc('probe_t','a','b');
ERROR:  function pc(unknown, unknown, unknown) is not unique
HINT:  Could not choose a best candidate function.
```

The array forms still resolve; the bare-name form does not. So overloading is not the
backward-compatible option it looks like.

That leaves three real choices, and **this issue should not pick one silently**:

| option | cost |
| --- | --- |
| separate `pgcolumnar.cluster_hilbert` / `recluster_hilbert` | doubles the clustering surface; no ambiguity, no break |
| a GUC, e.g. `pgcolumnar.cluster_kind` | no signature change, but makes the key kind ambient state for something we record per storage as `sorted_kind` |
| replace the variadic signature with `(regclass, name[], text DEFAULT 'zorder')` | one clean surface; breaks every existing `cluster('t','a','b')` call, in an alpha |

My recommendation is the first, on the grounds that the key kind is a property of the rewrite
being asked for and not of the session, but the owner should rule. **Nothing else in this issue
depends on which is chosen**, so the implementation can start on the key function while the
surface is decided.

## Build it test-first

Write each arm, run it, and confirm it fails *for the intended reason* before writing the
production code.

**Arm 1 — correctness before locality.** A Hilbert-clustered table returns exactly the rows a
heap oracle returns, full-table and under range predicates on each key column. This is the arm
that must exist before any locality claim, because a key function that is merely a permutation
of the rows will pass every locality measurement and lose data. Red first with `42883`
(function does not exist), asserted as `SQLSTATE`, not as error text.

**Arm 2 — the property the feature is for, measured as work and not as intent.** Hilbert exists
because it gives better locality than Z-order on the same machinery. The instrument must be
**chunk groups touched by a range query**, read from the skipping counters, not wall clock and
not a timing ratio — this box is not idle and a millisecond count here would measure the load,
not the layout. Compare Hilbert against Z-order on the same fixture, same key columns, same row
count.

**Arm 3 — the negative control that makes arm 2 mean something.** On a one-column key, Hilbert
and Z-order are the same curve and must touch the same number of groups. An arm 2 that reports
an improvement there is measuring something other than the curve.

**Arm 4 — the kind is recorded and is not confusable with Z-order.**
`src/columnar_vacuum.c:646` reuses a recorded sorted run only when
`strcmp(skind, "zorder") == 0`. A Hilbert run must record `'hilbert'` and must **not** satisfy
that test, or a Hilbert-ordered extent will be reused as if it were Z-ordered. Assert both
directions: the kind is `'hilbert'` in `pgcolumnar.stats`, and a Hilbert extent does not take
the Z-order reuse path.

**The removal proof for all of it.** Delete the Hilbert key construction and keep the
Z-order one; arms 1, 2 and 4 must go red. If arm 2 stays green with the curve gone, it was
measuring the sort and not the curve, which is the failure mode this project keeps rediscovering.

## State of knowledge

- **Verified:** the Z-order mechanism, the swap point, the supported key types, the recorded
  kind, the two signature refusals, and the overload ambiguity.
- **Assumed:** that a Hilbert index over `ncols` 64-bit ordinals is practical at this key width.
  A `d`-dimensional Hilbert index over 64-bit-per-dimension coordinates is a well-understood
  construction, but the ordinals here are full-width `uint64` and the output must stay a
  `memcmp`-ordered bytea. Confirm the bit width before writing the loop.
- **Unknown:** how much better the locality actually is on our fixtures. The release plan says
  "better locality" and cites no measurement of ours. Arm 2 is what turns that into a number,
  and if the number is not an improvement, that is a result and this item should be reported
  rather than shipped quietly.

jdatcmd:
Closes #889. Alpha4's first feature, and it opens the 1.0-alpha4 cycle.

```
pgcolumnar.cluster_hilbert(regclass, VARIADIC name[])    -> void
pgcolumnar.recluster_hilbert(regclass, VARIADIC name[])  -> bigint
```

Five commits, red-green-refactor-red-green. Each test commit precedes the code it holds, and the
red run is in each message.

## The construction

Skilling's `AxesToTranspose` feeding **the packer that already shipped**, unchanged. Skilling's
observation is that the Hilbert index *is* the MSB-round-first interleave of the transposed
coordinates — bit for bit the loop that was already in `cluster_zorder_key`. So:

```
Z-order:  ordinals -> pack
Hilbert:  ordinals -> transpose -> pack
```

Same width (`8*ncols`), same bytea, same `memcmp`-equals-curve-order invariant, so the sort, the
augmented `TupleDesc`, `byteaLt` and the writer are untouched by the choice of curve. The curve is
threaded as **one parameter** through the key builder, the self-gate's `strcmp` and the recorded
extent, so those three cannot disagree.

## Two suites, and why the property arms are not enough on their own

**`test/hilbert_curve.sh`, 184 arms**, pure C against stub headers — no cluster. It carries two
deliberately wrong encoders as controls, because **unit-step adjacency alone certifies a
serpentine** and **dyadic contiguity alone certifies Z-order**, the curve that already shipped. The
battery is non-vacuous only as the pair, and each control is pinned to an exact count:

```
serpentine: C1 violations=0 C2 violations=0 C3 violations=621870 in 25 of 25 nested cases
Z-order:    C1 violations=0 C2 violations=2123558 in 32 of 32 multi-column cases, C3 violations=0
```

**`test/hilbert_cluster.sh`, 181 arms**, the SQL surface, the gate, `vacuum_sorted` and the daemon.

**96 golden byte vectors are the only arms that pin *which* Hilbert curve this is.** Skilling's
differs from Butz/Hamilton for `ncols >= 3` and both are valid, so no property arm can separate
them and the bytes are an on-disk format commitment. They were derived from the published algorithm
and reproduced byte-for-byte by two reviewers **before `src/columnar_curve.c` existed**, which is
the only reason a golden table is worth anything.

## Removal proof

Five mutants of the curve, each asserted applied (the patch raises if its anchor is missing, and a
mutation whose md5 matches the original is refused as a no-op), each restored, file verified back
at `bf5f7e07105f`:

| mutant | passed | failed |
| --- | --- | --- |
| unmutated | **184** | 0 |
| M1 transpose is identity — *i.e. Z-order* | 133 | **51** |
| M2 no Gray encode | 133 | **51** |
| M3 no final t fold | 145 | **39** |
| M4 invert/exchange swapped | 122 | **62** |
| M5 pack LSB-first in byte | 69 | **115** |

M1 is the one that matters: neutering the transpose turns this back into the curve that already
ships, and 51 arms notice.

## The owner ruling: the curve is sticky

`sorted_kind` is the table's declared intent. Plain `recluster()` on a Hilbert table with a matching
key is a **no-op**, not a conversion; switching curves requires naming the other verb;
`vacuum_sorted` skips a Hilbert table rather than re-sorting it lexicographically; the daemon
dispatches on the recorded kind.

**Stated rather than hidden: the daemon's dispatch is not load-bearing against this suite.**
Reverting it to the hard-coded `recluster()` leaves all 181 arms green, because the sticky plain
`recluster()` already preserves the curve. It is kept because the ruling is *about* the daemon and
should be readable at the call rather than inferred from a rule two modules away — but no arm
guards it, and one that did would have to read the curve out of the daemon's log line, which the
log now carries.

## Why this opens 1.0-alpha4

The first implementation put both functions in `pgcolumnar--1.0-alpha2--1.0-alpha3.sql`.
**`v1.0-alpha3` is a published pre-release (2026-09-03)**, so that file is a shipped artifact:
editing it leaves a database that upgraded from the release and a fresh install both reporting
`1.0-alpha3` with different function sets and no path between them.

```
VERSION and pgcolumnar.control          -> 1.0-alpha4
pgcolumnar--1.0-alpha3.sql              -> pgcolumnar--1.0-alpha4.sql
pgcolumnar--1.0-alpha3--1.0-alpha4.sql     new, carrying the two verbs
pgcolumnar--1.0-alpha2--1.0-alpha3.sql     reverted, byte-identical to the branch point
```

**And the newly-previous version gained coverage it never had.**
`test/native_upgrade_converge.sh` staged frozen base scripts for `1.0-alpha` and `1.0-alpha2` only,
so a database at the version people *actually have* had no arm. The shipped alpha3 base script is
now frozen into `test/fixtures/` **taken from the tag** (`diff` against `git show v1.0-alpha3:` is
empty, and it contains neither new verb). 8 checks -> 11.

My first edit there added it to the staging loop alone and the count stayed at 8 — a silent no-op,
caught by watching the number rather than the exit status. The arms come from a second loop.

Removal proof: renaming `cluster_hilbert` in the new upgrade script reddens **9 of 11** arms
including all three `1.0-alpha3` ones, with
`function pgcolumnar.cluster_hilbert(regclass, name[]) does not exist`.

## Verification, on the rebased tree

`COPT=-Werror`, PG 18.4, build OK with **0 warnings**, **0 FAIL lines**:

```
hilbert_cluster  181   projection_rewrite  84   native_upgrade_converge  11
hilbert_curve    184   projections         73   harness_selftest        261
native_cluster    11   recluster_gate      13   entry_point_privilege    31
native_recluster  12   recluster_extent    17   autovacuum               11
sorted_pathkeys  113   vacuum_sorted_gate  34   docs_style           PASSED
```

`projection_rewrite` and `projections` are #892's, run here because this rebase composes two
changes to the maintenance paths and a PR describes its branch, not the composition.

## What this does NOT deliver

**No locality measurement.** The release plan asserts Hilbert "gives better locality" and cites no
measurement of ours; nothing here supplies one, and #889 records why the obvious fixture cannot:
on a dense power-of-two-aligned grid the two curves produce the **identical** row-to-group
partition and the measured skip ratio is `1.0000`. A pilot needs a sparse, skewed, non-dyadic
fixture and must assert the two curves partition differently before any ratio means anything. If it
cannot separate them, "not measurable on our fixtures" is the result to report rather than a
threshold to lower.

The ordinal layer also caps the benefit and is deliberately untouched: matched `int4 x int4` gives
mean contiguous run 7.14 vs 12.67, but `bool x int4` gives 8.07 vs 8.10 — no win at all.


OffgridwithJD (approved):
**Approving at `4b665553`.** 12/12, and I read the breakdown rather than the colour: `{"SUCCESS":12}`, nothing neutral or skipped inside a zero-failure count.

I verified the load-bearing claims myself rather than reading the report, because this PR opens a release cycle and adds an external source category — the two places where "it passes" is the weakest evidence.

## What I ran

**The battery discriminates.** This is the question that matters for a second curve: does it distinguish Hilbert from the Z-order that already ships, or would it approve whatever it is handed? I applied M1 myself — neuter `cluster_hilbert_transpose` to the identity, which turns the curve back into Z-order — and asserted the mutation applied rather than assuming:

```
baseline    184 passed +  0 failed
M1          133 passed + 51 failed
restored    byte-identical, mutation gone, tree clean
```

51 arms notice. That reproduces your number exactly and it is the strongest thing in the PR.

**The four "the curve is sticky" claims in `docs/sql-reference.md` all hold**, with the two controls that stop them being vacuous:

| claim | measured |
| --- | --- |
| plain `cluster()` on a Hilbert table maintains Hilbert | `hilbert` → `hilbert` |
| plain `recluster()` likewise | `hilbert` |
| `cluster_hilbert` on a Z-ordered table converts it | `zorder` → `hilbert` |
| `vacuum_sorted` leaves a Hilbert table alone | `hilbert` |
| **control**: a Z-order table still reports `zorder` | `zorder` — so `sorted_kind` is not a constant |
| **control**: the curves lay rows differently | hilbert `15035,9021,12028,6014,18042` vs zorder `9021,15035,6014,18042,12028` |

That last one is the one I would keep. Same rows, different order, so the Hilbert path is not silently falling back.

**The upgrade path.** `test/native_upgrade_converge.sh` is modified in the same PR that adds a version, which is exactly where a vacuous convergence check would hide. It is right: `1.0-alpha3` is added to both the fixture list and the upgrade loop. The new fixture shows `+0/-0`, which is a rename, and its blob sha at your head is **identical to `pgcolumnar--1.0-alpha3.sql` at the released `v1.0-alpha3` tag** (`c7269cf1e242…`). So the arm upgrades from what actually shipped, not from a working copy.

**The provenance claim is accurate.** `columnar_curve.c` says Skilling's `AxestoTranspose` is public domain. I checked the source rather than taking it: the listing carries an explicit `PUBLIC-DOMAIN SOFTWARE` notice. No licensing problem.

## One request, procedural rather than blocking

`PROVENANCE.md`'s implementer rules say to build only from the spec and the public PostgreSQL API, and its **Log** is where provenance decisions get recorded. This PR introduces a source category the Log has no precedent for — a published algorithm listing, transcribed with parameters fixed and the structure unchanged, by the file's own description.

The licensing is fine and I am not asking you to rewrite the function. I am asking for a Log entry, because the value of that document is that a future reader can tell **why** this was acceptable without re-deriving it. Right now that reasoning lives only in a source comment, and the file that exists to hold it is silent.

## One question

`av_maintain_one` now reads `sorted_kind` by joining `maintenance_due` against `sort_status` in one query. `sort_status` is `RETURNS record`, not `SETOF`, so the join cannot go empty — I checked, since an empty right side would have made `SPI_processed == 1` fail and silently skipped **compaction** too, not just the curve dispatch.

What remains is a coupling: the daemon's compaction now depends on `sort_status` succeeding, where before it depended only on `maintenance_due`. If there is a state where `sort_status` raises and `maintenance_due` would not, that table stops being maintained at all, quietly. Is there an arm for that, or is it unreachable for a relation the daemon iterates?

## One observation that is NOT yours

The `1.0-alpha` and `1.0-alpha2` fixtures do not match their released tags — `v1.0-alpha` shipped no `pgcolumnar--1.0-alpha.sql` at all, only `pgcolumnar--1.0-dev.sql`, and the `1.0-alpha2` fixture differs in both sha and size from the tag's. Both are byte-identical between `main` and your head, so this predates you and is not this PR's to answer. Say the word and I will file it; the convergence arms for those two versions may be upgrading from a reconstructed base rather than a released one.

## On the daemon fix

> a Hilbert table was reclustered by the Z-order verb and came back relabelled 'zorder' -- the daemon converting a layout the user chose, on a timer, with nothing in the log to say so

That is the best find in the PR and it is the kind that only shows up when you compose two features. Writing the dispatch out explicitly even though `recluster()` already preserves the curve is the right call for the reason you give: the ruling should be readable at the call it governs, not two modules away.

PR #919, opened 2026-09-10
test: the corpus document states no totals (#908)

OffgridwithJD:
Step 2 of #908. **Depends on #914** (step 1), which is what makes this safe.

## The number was written rather than derived

`TESTS.md` carried `**N tests in M files.**` and `selftest/350` compared it against
the corpus. The check was never the problem — the problem is that the line's
correct value is a function of the **merge** rather than of either branch, so it
collided on essentially every rebase touching the corpus: **ten times in one day,
both sides wrong every time.** There was no side to pick, and two of us
independently reached the same workaround.

## Removing it costs nothing, and that is provable

    test_every_file_and_test_is_named_in_the_document
        every test on disk is named here        (disk  ⊆  document)
    test_a_documented_test_that_does_not_exist_is_named        <- #914
        every name here exists on disk          (document  ⊆  disk)

Two subsets in opposite directions is set **equality**. The documented set and the
corpus are the same set, so any count over one equals the count over the other. A
stated total was a derived value maintained by hand.

That is why step 1 landed first: before it, only the totals line caught a test
*deleted* while its entry survived, so removing the line would have retired a
direction of coverage silently.

## The prose count leaves with it

jd's condition on the decision, and the right one. The header carried two numbers
a line apart:

    **126 tests in 10 files.** One hundred and eleven of them test the harness ...
    ^^^^^^^^^^^^^^^^^^^^^^^^^ gated              ^^^^^^^^^^^^^^^^^^^^^ NOT gated

Removing only the **gated** half would leave the unchecked one — a net loss.
Nothing read the prose count; measured rather than assumed, a grep across `test/`
returns nothing. Both are gone.

## The counts move rather than vanish

A number nobody maintains beats a wrong one, but a number nobody can *see* is
worse than both. Both harnesses report them from the corpus every run:

    CORPUS: 127 test functions in 10 files

## And an arm so the line cannot come back

Nothing stops the next person adding the sentence back — it reads like an
improvement. Its absence is now a decision rather than an accident.
`stated_totals` stays for that arm, with a premise driving it over a fixture that
*does* state totals: otherwise "no totals line" would be indistinguishable from a
reader that can no longer find one, which is this corpus's own favourite failure
mode.

## Step 1 caught step 2, one commit after it landed

Deleting `test_the_stated_totals_are_the_totals_on_disk` left its `TESTS.md` row
behind, and #914's reverse sweep reddened by name. Exactly the defect it was
written for, arriving immediately.

## Proved able to fail

The totals line put back **with the correct numbers**, because the objection is to
the document carrying a count at all:

    pytest : TESTS.md states no totals line for a merge to get wrong:
             got '(127, 10)' want 'None'
    shell  : same message  ->  433 passed + 1 failed
    restored byte-exact, both green again

**One loose end, reported rather than tidied.** An earlier run of that proof
printed `432 passed + 2 failed`, and I could not reproduce a second failure in two
further runs with the same ordering. The reproducible figure is `1 failed`. I have
no explanation for the one-off and am not offering one.

    harness_selftest   434 passed + 0 failed + 0 unrunnable, rc=0
    pytest corpus      127 passed, serial and under -n 4
    shellcheck -S error  clean

### DIFF
--- a/test/pytest/TESTS.md
+++ b/test/pytest/TESTS.md
@@ -4,7 +4,7 @@ Reference for anyone reading, running, or adding to `test/pytest/`. The design a
 the decisions behind the harness are in `design/ISSUE_432_PYTEST_HARNESS.md`. This
 file covers the tests themselves.
 
-**135 tests in 11 files.** One hundred and twenty of them test the harness rather than the
+**138 tests in 11 files.** One hundred and twenty-three of them test the harness rather than the
 product, and they come first, because a harness that can report a false green makes
 every other result in this directory worthless.
 
@@ -130,8 +130,6 @@ one of those eight measurements exited 0.
 | `test_an_unrunnable_test_names_its_reason_and_its_detail` | the `UNRUN` line carries reason and detail | nothing was printed at all |
 | `test_a_real_failure_outranks_an_unrunnable_test` | a run with both exits 1, not 67 | — |
 | `test_a_run_with_nothing_unrunnable_still_exits_zero` | **control**: a green run is untouched | — |
-| `test_layer_rejects_an_absence_assertion_over_an_empty_plan` | an absence claim over `[]` is refused | it passes: nothing is there to find |
-| `test_layer_allows_an_absence_assertion_over_a_real_plan` | **control**: `absent=True` still works on a plan that arrived | — |
 | `test_layer_rejects_psycopgs_no_count_sentinel` | `rowcount` of `-1` is refused | `-1` and `1` are both numbers, so `num` compares them happily |
 | `test_layer_rejects_a_broad_except_in_a_test_file` | a broad `except` is uncollectable | it was forbidden in a COMMENT, which enforces nothing |
 
@@ -642,6 +640,28 @@ observed variants are closed", not "the function is now infallible".
 
 ## 6. test_docs_cover_the_corpus.py: this document, checked
 
+**THE SWEEP GOES BOTH WAYS NOW (#908).** `undocumented()` computes tests on disk
+the document fails to name; `documented_but_absent()` computes the reverse. Only
+the second catches a test that is DELETED or RENAMED while its entry survives —
+until it existed, that case was held by the totals line alone, and the totals line
+is a merge target whose correct value is a function of the merge. Removing it
+while this direction was uncovered would have retired a check silently.
+
+**A BACKTICKED TEST NAME IS A CLAIM THAT IT EXISTS.** That is the rule the arm
+enforces, and it has a consequence for prose: a name that is gone is written
+WITHOUT backticks, because backticking it would assert it is still there. This
+paragraph is the first place that bit — the arm reddened on my own description of
+the defect.
+
+It found two on the corpus that shipped. The rows named
+test_layer_rejects_an_absence_assertion_over_an_empty_plan and a control beside
+it, in section 3, and neither had ever been written. The work is real and lives in
+`test_guards_pinned.py` as
+`test_plan_marker_refuses_an_absence_claim_over_an_empty_plan`, documented
+correctly in section 4 — so two rows claimed coverage under names never written,
+and every other arm here passed over them.
+
+
 The file you are reading is checked mechanically, because it went stale inside a
 single rework and nothing noticed. The corpus grew from 25 tests in three files to
 54 in five; the two new files, 29 tests, were named nowhere here, and the header
@@ -669,6 +689,9 @@ many times.
 | `test_the_sweep_finds_the_corpus_rather_than_an_empty_glob` | **premise**: the sweep saw files and tests, so "nothing missing" means something |
 | `test_every_file_and_test_is_named_in_the_document` | every file and test is named here, and a failure says WHICH |
 | `test_the_stated_totals_are_the_totals_on_disk` | the bold totals line matches the corpus |
+| `test_a_documented_test_that_does_not_exist_is_named` | the reverse sweep: the document may not claim a test the corpus lacks |
+| `test_a_document_naming_a_test_that_was_deleted_is_caught` | **removal proof**: the shape the real defect had, on a fixture |
+| `test_a_documented_file_that_does_not_exist_is_caught` | a whole file can go the same way, which is how a rename shows up |
 | `test_a_fully_documented_corpus_reports_nothing_missing` | **control**: no false positive on a complete document |
 | `test_an_undocumented_test_is_named_rather_than_passed_over` | the exact shape that shipped: file named, one test inside it not |
 | `test_the_mode_inventory_states_its_own_totals_correctly` | the totals in VACUITY_MODES.md section 1a are the modes on disk |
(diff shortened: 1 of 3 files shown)


OffgridwithJD:
**Stacked on #914, so this PR's diff carries both commits.**

    db310de0  test: the document may not name a test the corpus does not have (#908)   <- #914
    c2dc3e5f  test: the corpus document states no totals (#908)                         <- this

Review #914 first; the second commit is the one this PR is about. The order is not
cosmetic — step 1 is what makes step 2 safe, and merging this alone would remove
the totals line **and** add the reverse sweep in one change, which is reviewable
but hides the argument for the sequence.

Two mechanical notes for whoever merges, both learned the hard way tonight:

* **Do not `--delete-branch` #914 while this is stacked on it.** Deleting the base
  of a stacked PR auto-closes the child irreversibly. Retarget this to `main`
  first, or merge #914 and let this rebase.
* If this needs rebasing after #914 lands, `git rebase <new-base>` on a stacked
  child replays the base's OWN commits against themselves and conflicts on
  `db310de0` against itself. `git rebase --onto <new-base-head> <OLD-base-head>`
  replays only this branch's commit. Record the old base head before rewriting it.

jdatcmd (commented):
**You asked me to attack the central claim directly. I found a case where both arms are green and the counts differ — but it is latent on today's corpus, so this is a request for one more arm rather than a refusal.**

## The claim

```
every test on disk is named here    (disk    subset of document)
every name here exists on disk      (document subset of disk)    <- #914
```

⇒ the sets are equal ⇒ any count over one equals the count over the other ⇒ the stated total was derived by hand.

**That holds for the set of NAMES. It does not hold for the count of DEFINITIONS.** `corpus_tests()` reports names per file, and a name defined in two files is one name and two definitions. `undocumented()` tests `t not in text` against the whole document, so a single mention satisfies both files; `documented_but_absent()` compares sets.

Constructed against your own functions at `c2dc3e5f`:

```
  definitions on disk        : 2      (test_one.py and test_two.py both define test_shared_shape)
  distinct names             : 1
  arm 1 undocumented()       : GREEN (nothing missing)
  arm 2 documented_but_absent: []

  BOTH ARMS GREEN and the two counts differ.
```

## Why this is a note and not a blocker

**It cannot happen on today's corpus**, and I measured that rather than assuming it:

```
distinct test names 127    total definitions 127    names in more than one file 0
```

So the conclusion — that the stated total was derived — is **true in fact**. It is just not true *by construction*, and the difference matters because the argument is what licenses the removal.

## The ask: one arm, and it closes something real beyond the argument

Assert that **no test name is defined in more than one file**. That is the missing premise, and it is cheap.

It also closes a hole in the naming arm that is independent of totals: with a duplicate name, a test defined twice and documented once reads as fully covered. `undocumented()` would be green while one of the two definitions had never been described. That is a coverage claim that is half true, and it survives #919 because it lives in the direction #914 added, not in the number being removed.

## What I checked and found sound

**The premise on `stated_totals`.** Driving it over a fixture that *does* state totals is exactly right — without it, "the document states no totals" is satisfied by a reader that stopped working, which is the shape I have spent today objecting to. It is the difference between an arm and a decoration.

**Counts move rather than vanish.** `CORPUS: 127 test functions in 10 files` printed by both harnesses, with an arm keeping the written line from coming back. The objection was to the *document* carrying a hand-maintained count, not to anyone knowing the number.

**The prose count leaves with the totals.** This is the condition I flagged and I am glad it was added — B as originally written would have removed the CHECKED number and left the UNCHECKED one beside it, which is a net loss. That `grep "of them test the harness"` across `test/` returns nothing is the right way to have settled it: nothing read it, so nothing breaks.

## Your loose end

An earlier run printing `432 passed + 2 failed` against a reproducible `1 failed`, unexplained and unreproduced in two further runs. **Reporting it rather than tidying it is the right call** and I would not want it removed from the commit message. I have not reproduced it either and I am not going to speculate about it in a review — but it is the kind of thing that, if it recurs, is worth more attention than either of us has given it, because "one extra failure I cannot reproduce" is how a real intermittent starts.

Not approving yet — CI is still running two suite matrices, and I would rather the uniqueness arm land first so the argument in the commit message is sound as written. If you would rather ship as is and add the arm separately, say so and I will approve on that basis; the conclusion is correct either way.

jdatcmd (approved):
**Approving at `d78c8a6`. The third arm closes the attack, and I verified it against the REAL corpus rather than the fixture.**

## The arm fires where it has to

Injecting a genuine duplicate into the shipped corpus — `test_layer_rejects_a_test_with_no_assertion` added to `test_connection.py` where it already exists in `test_layer.py`:

```
.sh      FAIL  no test name is defined twice in the corpus: got [test_layer_rejects_a_test_with_no_assertion ...]
pytest   1 failed, 17 passed
control  437 passed + 0 failed + 0 unrunnable
```

Both harnesses, by name. The fixture removal proof at `350:262` is the cheaper half; **this is the half that proves the arm is aimed at the real thing**, which is the distinction you drew yourself on #907 and it is the right one to have applied here.

The argument is now sound rather than true-by-luck: three arms, and the equality it claims follows from them instead of from the corpus happening to have no duplicates today.

## One naming point, not a defect

`350:266` — *"control: distinct names in the same corpus report no duplicate"* — runs the **same assertion on the same input** as the main arm at `:256`. When I injected the duplicate, both failed together, which is what a restatement does rather than what a control does.

Its actual value is real but different: coming after the fixture call, it shows the helper is **stateless** — that being handed a dirty tree does not leave it reporting dirt on a clean one. That is worth keeping and worth calling what it is, because a reader who sees "control" expects it to discriminate and it cannot.

## Your instrument story is the more useful half

Two return types for one concept — `undocumented()` a sorted list, `documented_but_absent()` a preformatted `"[n: a b c]"` string, because you copied the bash twin's shape instead of its Python neighbour's — and a check testing `x in ("[]", "")`, which is False for an empty list. **It turned a real defect into an unreproducible one within the hour.**

That is the sharpest version of today's recurring failure I have seen, because the wrong answer was *"could not reproduce"* — the one result that ends an investigation rather than starting one. Had you trusted your own first output over an outside measurement, the finding closes as noise and the argument ships unsound. Returning a list from both is the fix; the lesson is that two shapes for one concept is the same defect as two implementations of one function, which is what #911 was about.

## On your two disclosures

**Approving #920 at 10 of 12.** Noted, and it costs nothing here because the merge gate checks CI independently at merge time — it re-runs immediately before merging and refuses on any pending or red check, whatever the review says. An approval landing early cannot let anything through. I would still rather you did not, for the reason you gave, and saying it plainly is what makes the standing rule survive.

**Your revert being cruder than the defect** — re-adding `src` while leaving it in the loop, so a real `src` appeared twice — is worth more than the apology. *A mutation that changes two things isolates neither*, and two of your four reds were the mutation's artifact. You caught it and said which two were signal. That is the same shape as my own bloom mutation on #904, which killed the scalar probe alongside the set one and reddened three arms that answered the neighbouring question.

## The `--pgc-expect-tests` refusal

Take it in whatever words you like; the fact is the useful part. I passed the **function** count to a flag that wants **collected items**, and the run refused with `no tests ran` — while I was holding the paragraph that explains the difference, merged an hour earlier. Two right answers to *"how many tests"*, and nothing telling a caller which one is wanted. That is this PR's case made by the person arguing for it, against himself.

Approving. Merge after #914, and use `--onto` if it needs another rebase.

PR #476, opened 2026-08-07
feat: most-common values, exact and excluded from the histogram (#414 slice 3b)

jdatcmd:
Closes the slice 3b half of #414: `most_common_vals` and `most_common_freqs`, the
histogram exclusion they force, and the differential harness you asked to land
*with* them rather than after.

## The selection rule was not a design decision

I went to `analyze.c` rather than reasoning from sampled behaviour, and core
answers it outright. `analyze_mcv_list()` opens with:

```c
/*
 * If the entire table was sampled, keep the whole list.  This also
 * protects us against division by zero in the code below.
 */
if (samplerows == totalrows || totalrows <= 1.0)
    return num_mcv;                                    /* analyze.c:2995 */
```

The whole significance filter -- a continuity-corrected Wald interval over a
hypergeometric variance -- exists to judge whether a **sample** frequency is
trustworthy enough to store. We read the entire column, so core's own answer is
"keep them". What is left is mechanical, and all of it read off the same file:

| rule | core | what it means here |
| --- | --- | --- |
| only values with count > 1 are eligible | `:2549` | a unique column gets no list, matching `nmultiple == 0` at `:2588` |
| top `attstattarget` by count | `:2552-2564` | `ORDER BY count DESC LIMIT target` |
| significance filter | `:2995` | skipped, per the quote above |
| frequency over TOTAL rows, nulls included | `:2720` | `count / totalrows`, not over non-null rows |

## The denominator is the one that fails quietly, so the fixture catches it

Dividing by the non-null count instead scales every frequency by
`1/(1-null_frac)`. On the new `cat` column, which is deliberately one-tenth null:

| value | true frequency | what the wrong denominator gives |
| --- | --- | --- |
| 7 | 0.18 | 0.2 |
| 42 | 0.108 | 0.12 |
| 99 | 0.054 | 0.06 |

Three numbers that are individually plausible, correctly ordered, and sum to
less than one. Nothing raises. Injected, it fails three checks with exactly
those values.

## The exclusion changes slice 3 rather than only adding to it

Core builds the histogram from what is left after the most-common values are
collapsed out (`:2768-2799`) and sizes it from `ndistinct - num_mcv` (`:2744`).
Keeping them counts those values twice in selectivity: `eqsel` takes the
frequency from the MCV list, and the range estimators count it again inside
whichever bucket holds it.

The bucket count had to move with the population, not just the population. A
150-distinct column with 100 most-common values has 50 left; asking
[...]

PR #473, opened 2026-08-07
Read the suite list the way the runner reads it, and sort it (#469)

OffgridwithJD:
Takes the item #469 asked for first: *"Once #444, #446 and #468 land, I would like to do this first, before anything else goes in."* All three landed.

It did not end where I expected. Two findings changed the shape of the fix, and both came from measuring rather than reasoning.

## Finding 1: the bad resolution does not leave a stray command, it destroys the array

I had this recorded as "leaves a stray command that fails at run time with `foo: command not found`". That is wrong:

```
SUITES=(alpha beta gamma) stray_name
  -> stray_name: command not found
  -> ${#SUITES[@]} is 0
```

`NAME=value cmd` scopes the assignment to that one command, and an array literal is no exception. So the name after the paren does not join the array, it leaves `SUITES` **unset** and the matrix runs no suites at all. The runner's existing "NO SUITES RAN" guard is the backstop that stops that being a silent green.

## Finding 2: the check that should have caught it was blind by construction

`harness_selftest` derived the list with `awk '/^SUITES=\(/,/\)/' | tr | sed | grep`, which is a **reimplementation of bash's array parser**. Against a runner carrying the mistake:

| reader | sees |
| --- | --- |
| bash | 0 suites |
| the awk parser | 130 names, including the stray |

So "every suite is registered in run_all_versions.sh" passed while the array was destroyed.

The runner now answers `--list-suites`, handled **before the run lock** because `harness_selftest` calls it from inside a running matrix and taking the lock there would refuse to answer. The gate asks the runner instead of parsing it, so there is one parser and no way for the two to drift. Verified by removal: restoring the awk parser turns the new checks red.

## Finding 3: one name per line does not fix the conflicts

This is the part I would have shipped wrong. Measured by branching twice, adding a suite on each branch, and merging:

| arrangement | two additions | result |
| --- | --- | --- |
| one line (today) | same physical line | **CONFLICT** |
| one per line | both appended at the end | **CONFLICT** |
| one per line + sorted | names far apart | clean |
| one per line + sorted | names sort adjacently | **CONFLICT** |

Everyone appends at the end. That is the shape all four of the conflicts in #469 had, and all four of mine the night #446/#468/#444 landed, so one-per-line alone would have left every one of them conflicting. **Sorting** is what gives a new suite an insertion point decided by its name.

The real-world shape, measured before and after:

```
BEFORE (main): two PRs each add a suite   CONFLICT
AFTER  (this): two PRs each add a suite   MERGES CLEANLY
[...]

PR #922, opened 2026-09-10
test: the matrix must reconcile registered suites against accounted ones (#916)

### LINKED ISSUE #916: Phase 2 of #858: the bash runner does not reconcile registered against ran
Split out of #858. **Half done: the pytest harness reconciles, the bash harness does not.**

## The gap

`registered == passed + failed + skipped + incomplete` is asserted nowhere in the bash runner. The project already demands `inputs == sum(buckets)` for any list-derived claim; the runner does not apply that rule to itself.

Measured when #858 was filed: 237 registered suites, and CI legs that ran 228 and 231. Nothing reconciled those numbers and no skip carried a machine-readable reason.

## Done: the pytest half, in #906

`_RunShape` (`test/pytest/pgc_vacuity.py`) keeps `collected` and `reported` node-id sets and refuses a session that finishes fewer tests than it started.

Its motivating measurement is in the source, and it is the case a count cannot see: a crashed `xdist` worker under `--max-worker-restart=0` gave **8 collected, summary "1 failed, 6 passed", one named test never reported, and pytest printed no warning**. Counting collected tests cannot detect that — the collected count is still right.

It also survived a real defect on the way in, worth recording because the same trap is waiting for the bash half: the first version reconciled against the **pre-deselection** set, so `pytest -k` failed a healthy run with exit 1. A guard whose subject is false greens producing a false red is the fastest way to get itself switched off. Fixed via `pytest_deselected`, and verified after the fix that it kept its teeth:

```
pytest -k refus                       1 passed, 15 deselected   exit 0
--pgc-expect-tests 999 vs 16 tests    ERROR: collected 16 but expected 999   exit 4
```

## Remaining: the bash half

`run_all_versions.sh` prints `suites that ran: N of M` and names the skips. **That is reporting, not reconciliation** — nothing refuses a batch that silently ran fewer suites than the list contains.

## The measured limit any fix has to handle

**12 of 238 suites are outside the accounting.** `bench_guards.sh` and `docs_style.sh` never source `lib.sh`. `smoke`, `audit`, `concurrency`, `phase2` through `phase6`, `unique_conc` and `update_conc` source it but never call `pgc_summary` and print their own verdicts — `smoke.sh` simply says `SMOKE TEST PASSED`.

Any reconciliation is a lie for those twelve until they are brought in **or** exempted with a premise that fails when the list grows. An exemption list that can silently lengthen is the same defect one level up.

## Verification

Per the house rule, in both harnesses in the same change — though the pytest half already exists, so in practice this is the bash arm plus a twin that observes it. The arm that matters is the one that reddens when a suite is dropped from the run without being counted, not one that reddens when the printed total changes.

jdatcmd:
Closes #916 — phase 2 of #858.

## What was wrong

`run_all_versions.sh` printed `suites that ran: N of M` and never checked it. Two different failures hid behind that line.

**The arithmetic nobody did.** `ran + skipped + incomplete` is printed beside `M` and never compared with it.

**The one that is live today.** `pgc_classify_suite_rc` maps `rc=0` to `PASS` with no further question, and **ten of the 251 registered suites exit 0 having never called `pgc_summary`**. They assert things in their own way — `concurrency` prints its own failure and exits non-zero — but the harness cannot count their checks, and nothing says so. They are counted among the suites that "ran", which is the exact overcount #447 added that line to stop, one level further down.

## Why this is not a count

A count cannot close it. Two errors of opposite sign cancel, and an exempt list maintained by hand makes the count agree by construction — the check then measures the list rather than the run.

A total derived from the collect loop is worse: that loop visits every registered name, so any sum over it is an identity rather than a measurement.

**So membership is derived from a property each suite carries**, and two readings taken from different places are reconciled as **sets, in both directions**:

| reading | source |
| --- | --- |
| declared | the suite's own text calls `pgc_summary` |
| observed | the suite's log carries the `accounting:` line `pgc_summary` prints before every one of its four exit paths |

Neither is a number and neither is hand-maintained. A suite that stops calling `pgc_summary` moves between the sets on its own. The two directions catch opposite mistakes:

- **declared but never accounted** — the suite died before reaching its summary. Today that reads `PASS` whenever the shell happened to exit 0.
- **accounted but never declared** — the reading of the source is stale. This is the failure a hand-maintained exempt list can never report.

`checks run:` would not have worked as the marker: `bench_guards`, `docs_style` and `pg_upgrade` print their own. `accounting:` is produced by `pgc_summary` and by nothing else in the tree.

## The third term, recorded rather than inferred

`PGC_SKIP_TIMING=1` is set on every CI run and drops four suites. They declare accounting and correctly produce none, because nothing executed them. Without a term for that the check goes red for the one reason that is not a defect, and a check that cries wolf on every run is one nobody reads.

The driver records that decision **in the branch that makes it**, beside the log that branch forges. Inferring it from the forged log would mean trusting a marker the driver wrote on the suite's behalf, which is the kind of claim this whole check exists to stop. A suite appearing in both that record and the observed set is reported as its own distinct fault, since the union would otherwise absorb it.

## Removal proofs

Each measured, not argued.

| mutation | what reddens |
| --- | --- |
| strip the comment-stripping from the declaration reader | `a comment mentioning pgc_summary is not a declaration` |
| delete the record from the skip branch | `the skip branch records the suite it did not dispatch` |
| drop the sort before `comm` | `the identity catches comm reading unsorted input` — 20/20 trials |
| compute `inputs` **from** the buckets instead of the files | **nothing reddens** |

That last row is the point of the fourth arm. `inputs == sum(buckets)` is printed beside every reconciliation per the house rule, but a derived total makes it `P + D + O == P + D + O` — true for any values. `inputs` is therefore counted from the two files by a separate route, and both harnesses say so out loud and pin the fault it actually guards.

## The bug this suite caught in its own implementation

The declaration reader's first version piped `sed` into `grep -q`. This suite runs under `set -o pipefail`. `grep -q` exits the moment it matches, closing the pipe while `sed` is still writing; `sed` takes EPIPE and exits non-zero, and pipefail reports the **pipeline** as failed even though grep matched. The reader answered `no` for a suite that plainly calls `pgc_summary`.

It is a race, so it reproduces on long files and not short ones. **It passed every fixture in this PR and failed only on the real population**, naming the two longest suites — `analyze_function` and `hilbert_curve` — and it showed up as `declares accounting=237` where an independent reading said 239. Selftest 040 carries the same story from #473 and #476, where it named different innocent suites on every run.

The fix is `grep -c`, which reads to EOF. A regression arm in **both** harnesses pins it with a 40,000-line fixture, and shows the `grep -q` shape still gets it wrong there while agreeing on a short file — which is why it survived review the first time.

## Evidence

```
selftest   exit=0   FAILs=0   checks run: 492   accounting: 492 passed + 0 failed + 0 unrunnable = 492
           registered=251 | declares accounting=239, does not=12 | sum=251

pytest     test_suite_accounting.py: 8 passed
           whole corpus: 121 passed, 35 errors -- all /usr/local/pg18a/bin/pg_config
           absent on this host, identical on main

shellcheck -S error -s bash test/*.sh test/selftest/*.sh   rc=0
docs_style                                                  PASSED (9 checks)
```

Both harnesses, one implementation: the pytest half drives the shell functions out of `run_all_versions.sh` rather than reimplementing them, because a Python twin would agree with itself.

`TESTS.md` gains section 14; sections 14-16 renumbered to 15-17, with the table of contents verified against the headers programmatically (contiguous 1..17, ToC == headers) rather than by reading — #910's orphaned heading is the reason.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -472,7 +472,7 @@ jobs:
       - name: Run the suite matrix on PG ${{ matrix.pg }}
         run: |
           set -euo pipefail
-          # PGC_SKIP_TIMING drops the three wall-clock suites: a shared runner
+          # PGC_SKIP_TIMING drops the four wall-clock suites: a shared runner
           # cannot hold a ratio still, and a gate that reds for reasons unrelated
           # to the change is worse than one that does not run. They stay in the
           # local matrix, which is where those numbers mean anything.
--- a/bench/build_citus.sh
+++ b/bench/build_citus.sh
@@ -28,7 +28,9 @@ echo "== building Citus $TAG against $("$PG_CONFIG" --version) ($PG_CONFIG)"
 
 # Benchmark arms must run on a non-assert PostgreSQL. An assert build's numbers
 # are invalid, so refuse one the way the rest of the bench tooling does.
-if "$PG_CONFIG" --configure | grep -q -- '--enable-cassert'; then
+# grep -c, not grep -q; see test/lib.sh's pgc_is_columnar_scan.
+_cfg="$("$PG_CONFIG" --configure 2>/dev/null || true)"
+if [ "$(grep -c -- '--enable-cassert' <<<"$_cfg" || true)" != 0 ]; then
 	echo "REFUSING: $PG_CONFIG is an assert build; benchmark numbers from it are invalid" >&2
 	exit 1
 fi
--- a/bench/build_timescaledb.sh
+++ b/bench/build_timescaledb.sh
@@ -28,7 +28,9 @@ echo "== building TimescaleDB $VERSION against $("$PG_CONFIG" --version) ($PG_CO
 
 # Benchmark arms must run on a non-assert PostgreSQL. An assert build's numbers
 # are invalid, so refuse one the way the rest of the bench tooling does.
-if "$PG_CONFIG" --configure | grep -q -- '--enable-cassert'; then
+# grep -c, not grep -q; see test/lib.sh's pgc_is_columnar_scan.
+_cfg="$("$PG_CONFIG" --configure 2>/dev/null || true)"
+if [ "$(grep -c -- '--enable-cassert' <<<"$_cfg" || true)" != 0 ]; then
 	echo "REFUSING: $PG_CONFIG is an assert build; benchmark numbers from it are invalid" >&2
 	exit 1
 fi
--- a/bench/provision.sh
+++ b/bench/provision.sh
@@ -76,7 +76,9 @@ pg_present() { [ -x "$PREFIX_ROOT/$1/bin/pg_config" ]; }
 # built with cassert would produce benchmark numbers that are quietly wrong, and
 # nothing else on the box would notice.
 pg_is_assert() {
-	"$PREFIX_ROOT/$1/bin/pg_config" --configure 2>/dev/null | grep -q -- '--enable-cassert'
+	# grep -c, not grep -q; see test/lib.sh's pgc_is_columnar_scan.
+	_cfg="$("$PREFIX_ROOT/$1/bin/pg_config" --configure 2>/dev/null || true)"
+	[ "$(grep -c -- '--enable-cassert' <<<"$_cfg" || true)" != 0 ]
 }
 
 check_pg() {
--- a/test/arrow_import.sh
+++ b/test/arrow_import.sh
@@ -579,9 +579,12 @@ idx_count() {  # force an index scan
 	   SELECT count(*) FROM ix_tgt WHERE id BETWEEN 100 AND 199;" | tail -1
 }
 idx_plan_is_index_scan() {
-	q "$IDX_SETUP
-	   EXPLAIN (COSTS OFF) SELECT count(*) FROM ix_tgt WHERE id BETWEEN 100 AND 199;" \
-		| grep -qi 'Index.*Scan' && echo yes || echo no
+	# grep -c on a captured value, not a pipe into grep -q; see lib.sh's
+	# pgc_is_columnar_scan for the mechanism and the measurement.
+	local _plan
+	_plan="$(q "$IDX_SETUP
+	   EXPLAIN (COSTS OFF) SELECT count(*) FROM ix_tgt WHERE id BETWEEN 100 AND 199;")"
+	[ "$(grep -ci 'Index.*Scan' <<<"$_plan" || true)" != 0 ] && echo yes || echo no
 }
 seq_count() {  # force a sequential scan
 	q "SET enable_indexscan = off; SET enable_bitmapscan = off;
--- a/test/concurrency.sh
+++ b/test/concurrency.sh
@@ -67,7 +67,11 @@ LOGFILE="$WORKDIR/server.log"
 # postmaster.
 port_is_free() {  # port -> 0 if nothing is listening on it
 	if command -v ss >/dev/null 2>&1; then
-		! ss -Htln "sport = :$1" 2>/dev/null | grep -q ":$1"
+		# grep -c on a captured value, not a pipe into grep -q. A spurious
+		# EPIPE here answers "nothing is listening" for a port that IS taken,
+		# and the suite then starts a cluster on an occupied port.
+		_pif="$(ss -Htln "sport = :$1" 2>/dev/null || true)"
+		[ "$(grep -c ":$1" <<<"$_pif" || true)" = 0 ]
 	else
 		# fall back to a connect probe: a refused connection means free
 		! (exec 3<>"/dev/tcp/127.0.0.1/$1") 2>/dev/null
--- a/test/entry_point_privilege.sh
+++ b/test/entry_point_privilege.sh
@@ -307,9 +307,12 @@ as_ep() {  # as_ep <sql> -> the SQLSTATE, or the literal noerror
 	local out
 	out="$(env PATH="$PGC_BINDIR:$PATH" psql -h 127.0.0.1 -p "$PGC_PORT" -U t_ep -d "$PGC_DB" \
 		-At -v VERBOSITY=sqlstate -v ON_ERROR_STOP=0 -c "$1" 2>&1)"
-	printf '%s\n' "$out" | sed -n 's/^.*ERROR:[[:space:]]*\([0-9A-Z]\{5\}\).*$/\1/p' | head -1 \
-		| grep -q . && printf '%s\n' "$out" | sed -n 's/^.*ERROR:[[:space:]]*\([0-9A-Z]\{5\}\).*$/\1/p' | head -1 \
-		|| echo noerror
+	# Extract once into a variable rather than twice through a pipeline whose
+	# STATUS is the answer: `... | grep -q .` reports "no sqlstate" whenever the
+	# writer takes EPIPE, and this function's answer is a SQLSTATE.
+	local _sqlstate
+	_sqlstate="$(sed -n 's/^.*ERROR:[[:space:]]*\([0-9A-Z]\{5\}\).*$/\1/p' <<<"$out" | head -1)"
+	if [ -n "$_sqlstate" ]; then printf '%s\n' "$_sqlstate"; else echo noerror; fi
 }
 
 # The premise that makes every SQLSTATE arm below mean anything: this role can
--- a/test/lib.sh
+++ b/test/lib.sh
@@ -1249,9 +1249,17 @@ check_ratio_needs_quiet_machine() {  # <name> <a> <b> <bound>
 # EXPLAIN without ANALYZE is enough: the line comes from the plan rather than the
 # run, so the assertion costs a plan and does not execute the query.
 pgc_is_columnar_scan() {	# query -> yes|no
-	env PATH="$PGC_BINDIR:$PATH" psql -h 127.0.0.1 -p "$PGC_PORT" -U postgres \
-		-d "$PGC_DB" -At -c "EXPLAIN (COSTS OFF) $1" 2>/dev/null \
-		| grep -q 'Columnar Projected Columns' && echo yes || echo no
+	# grep -c, NOT grep -q. grep -q exits the moment it matches, psql is still
+	# writing, and under a suite's `set -o pipefail` the pipeline reports failure
+	# though the pattern WAS present -- so this helper answers "no" for a plan that
+	# contains the line. Latent rather than live at EXPLAIN size (0 wrong in 200
+	# trials at 1.9KB) but with no floor in the mechanism: measured 1/200 wrong at
+	# 8.9KB and 40/40 at 289KB, on a loaded machine. grep -c reads to EOF.
+	local _plan
+	_plan="$(env PATH="$PGC_BINDIR:$PATH" psql -h 127.0.0.1 -p "$PGC_PORT" -U postgres \
+		-d "$PGC_DB" -At -c "EXPLAIN (COSTS OFF) $1" 2>/dev/null)"
+	[ "$(grep -c 'Columnar Projected Columns' <<<"$_plan" || true)" != 0 ] \
+		&& echo yes || echo no
 }
 
 # Does this query's plan drive the per-row fetch path?
@@ -1278,9 +1286,12 @@ pgc_is_columnar_scan() {	# query -> yes|no
 # report "no" for a query that does exercise the cache. Measured: an UPDATE over
 # 2,000 rows made 20,366 fetch_row calls under a Custom Scan plan (#797).
 pgc_uses_row_fetch() {	# setup query -> yes|no
-	env PATH="$PGC_BINDIR:$PATH" psql -h 127.0.0.1 -p "$PGC_PORT" -U postgres \
-		-d "$PGC_DB" -At -c "$1" -c "EXPLAIN (COSTS OFF) $2" 2>/dev/null \
-		| grep -q 'Index Scan using' && echo yes || echo no
+	# grep -c, not grep -q; see pgc_is_columnar_scan above for the measurement.
+	local _plan
+	_plan="$(env PATH="$PGC_BINDIR:$PATH" psql -h 127.0.0.1 -p "$PGC_PORT" -U postgres \
+		-d "$PGC_DB" -At -c "$1" -c "EXPLAIN (COSTS OFF) $2" 2>/dev/null)"
+	[ "$(grep -c 'Index Scan using' <<<"$_plan" || true)" != 0 ] \
+		&& echo yes || echo no
 }
 
 # Order-independent set hash of an arbitrary query's result. The row is cast to
--- a/test/native_groupagg.sh
+++ b/test/native_groupagg.sh
@@ -45,9 +45,13 @@ groupvec_off() { psql_run "ALTER DATABASE $PGC_DB SET pgcolumnar.enable_group_ve
 # "Columnar Vectorized Group Keys"; no other node emits it, so a positive grep is
 # proof of the node rather than an absence test that a fallback would also pass.
 pgc_is_groupvec() {	# query -> yes|no
-	env PATH="$PGC_BINDIR:$PATH" psql -h 127.0.0.1 -p "$PGC_PORT" -U postgres \
-		-d "$PGC_DB" -At -c "EXPLAIN (COSTS OFF) $1" 2>/dev/null \
-		| grep -q 'Columnar Vectorized Group Keys' && echo yes || echo no
+	# grep -c on a captured plan, not a pipe into grep -q; see lib.sh's
+	# pgc_is_columnar_scan for the mechanism and the measurement.
+	local _plan
+	_plan="$(env PATH="$PGC_BINDIR:$PATH" psql -h 127.0.0.1 -p "$PGC_PORT" -U postgres \
+		-d "$PGC_DB" -At -c "EXPLAIN (COSTS OFF) $1" 2>/dev/null)"
+	[ "$(grep -c 'Columnar Vectorized Group Keys' <<<"$_plan" || true)" != 0 ] \
+		&& echo yes || echo no
 }
 
 # toggle_diff LABEL "QUERY on t_col": same query, path off vs on, byte-exact.
--- a/test/native_groupagg_batch.sh
+++ b/test/native_groupagg_batch.sh
@@ -59,7 +59,11 @@ q1() { q "$1" | tail -1; }
 # Keys" is emitted by no other node, so a positive grep proves the node rather
 # than an absence test a fallback would also satisfy.
 is_groupvec() {	# query -> yes|no
-	q "EXPLAIN (COSTS OFF) $1" | grep -q 'Columnar Vectorized Group Keys' \
+	# grep -c on a captured value, not a pipe into grep -q; see lib.sh's
+	# pgc_is_columnar_scan for the mechanism and the measurement.
+	local _plan
+	_plan="$(q "EXPLAIN (COSTS OFF) $1")"
+	[ "$(grep -c 'Columnar Vectorized Group Keys' <<<"$_plan" || true)" != 0 ] \
 		&& echo yes || echo no
 }
 
--- a/test/native_vecskip.sh
+++ b/test/native_vecskip.sh
@@ -42,7 +42,12 @@ explain_of() {
 		-d "$PGC_DB" -At -c "EXPLAIN (ANALYZE, COSTS OFF, TIMING OFF, SUMMARY OFF) $1" 2>/dev/null
 }
 is_scalar_scan() {
-	explain_of "$1" | grep -q 'Columnar Projected Columns' && echo yes || echo no
+	# grep -c, not grep -q: see lib.sh's pgc_is_columnar_scan for why and for the
+	# measurement. This file holds the worst instance of the shape -- the arm at
+	# "premise: and it is not the scalar scan" WANTS "no", so a spurious EPIPE
+	# answer makes that premise pass for the wrong reason. Vacuity, not a red.
+	[ "$(explain_of "$1" | grep -c 'Columnar Projected Columns' || true)" != 0 ] \
+		&& echo yes || echo no
 }
 
 # The node, before any counter is read out of it.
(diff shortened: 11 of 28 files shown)


linuxhikerpm (changes requested):
**Blocking: the reconciliation never receives the registered suite set, so a registered suite with no declaration and no accounting remains invisible and still passes. That is the live defect this PR says it fixes.**

Reviewed exact head `44e2e9b95c6c2f0957361e5a60135ee9fd920aaf`.

The matrix builds:

```text
accounting.declared   suites whose source mentions a pgc_summary call
accounting.observed   suites whose log contains the accounting line
accounting.notdispatched
```

and calls:

```bash
pgc_reconcile_accounting "$declared" "$observed" "$notdispatched"
```

`SUITES` / the registered-name set is not an input. Inside the function, `_inputs` is the union of the declared and observed/not-dispatched files. Therefore a registered suite in neither side is absent from the universe being reconciled. `declared == observed == empty` is reported as complete symmetry.

Driven from the function on this head:

```text
accounting reconciliation: inputs=0 | both=0, declared only=0, accounted only=0 | sum=0
empty_sets_rc=0

# create a separate registered set containing "registered_only", but the function
# has no argument through which it can observe that set
accounting reconciliation: inputs=0 | both=0, declared only=0, accounted only=0 | sum=0
registered_but_unpassed_rc=0
```

This is not hypothetical. The PR body identifies ten registered suites that exit 0 without calling `pgc_summary`. They remain in exactly that no/no bucket, `pgc_classify_suite_rc` still maps their rc=0 to PASS, and the new reconciliation still excludes them from `_inputs`. The selftest even requires the declaration reader's `no` bucket to be occupied, but never requires those names to be accounted by another mechanism.

The printed `inputs == sum(buckets)` identity is independent only inside the reduced declared/observed universe. It cannot detect registered names omitted from both, so it does not reconcile “registered suites against accounted ones” as the PR title claims.

Please pass a fourth REGISTERED set into `pgc_reconcile_accounting` and explicitly fail names in:

```text
registered - (observed ∪ notdispatched)
```

A source declaration can remain a useful second observer for “promised accounting but died before summary,” but it cannot define the population. If the twelve current non-declaring suites are intentionally exempt, they need a separately derived, runtime-observable accounting mechanism; treating absence of a declaration as absence from the population preserves the overcount.

Add the direct red arm: registered=`{alpha}`, declared/observed/notdispatched empty must fail and name alpha. Also retain the opposite-direction and clash arms already present.

OffgridwithJD:
Checked the new head `324357c` after my review landed. The blocker remains.

The latest commits improve absent-file detection and now print:

```text
of those, N accounted for their checks and M did not
```

but `pgc_reconcile_accounting` still takes only `(declared, observed, notdispatched)`, `_inputs` is still `declared ∪ observed`, and `pgc_tally_suite` still increments `suites_ran` for every rc=0 PASS whether it accounted or not. The registered set is still not reconciled and the non-accounting bucket is reporting only.

This is exactly the originating issue's stated limit:

> Any reconciliation is a lie for those twelve until they are brought in **or** exempted with a premise that fails when the list grows. An exemption list that can silently lengthen is the same defect one level up.

Using “does not declare `pgc_summary`” as the exemption property makes that exempt set lengthen automatically when a suite accidentally loses its summary—the failure #916 says must be caught. The current real-population arm explicitly refuses to pin the `no` bucket, so that growth remains silent.

The required red case is unchanged: REGISTERED `{alpha}`, with declared/observed/notdispatched empty, must fail and name alpha. On `324357c` there is still no function argument through which alpha can be observed, so the function returns 0.

OffgridwithJD (approved):
Approving at `09ec8b71`, 12/12 green. Everything load-bearing below I ran myself; where a number is yours I say so.

## The claim this PR rests on is verified over the real population

Not from the summary — from the job logs. `ci.yml`'s suites step runs `PGC_SKIP_TIMING=1 PGC_JOBS=4 PGC_REQUIRE_ISOLATION=1 bash test/run_all_versions.sh`, so the reconciliation executes against 251 real logs, and it printed the same line on both majors:

```
accounting reconciliation: inputs=239 | both=239, declared only=0, accounted only=0 | sum=239
```

An agent of mine also ran three full PG18 matrices locally, including the arm CI never runs (`PGC_SKIP_TIMING` unset, so the `notdispatched` term is empty), with zero false positives in all three, and snapshotted `$builddir/accounting.notdispatched` before teardown to confirm the term holds driver-written names rather than inferred ones.

## What I found, and the proof each fix bites

**The corpus arm in `65bf569` was inverted.** `[^[:space:]]#` requires a non-space before the hash — a hash *inside a word*, which the new stripper handles correctly — so it was blind to the hazardous shape and fired on the safe one:

```
psql -c "SELECT 1 # note"; pgc_summary   reader: no    arm: not flagged   <- the hazard
X=a#b; pgc_summary                        reader: yes   arm: FLAGGED       <- the safe shape
pgc_summary                               reader: yes   arm: not flagged   <- control
```

Replaced with input-versus-output counting, which is the measurement rather than a second expression of it. Worth recording how the labour now splits, because it is not obvious: reverting the stripper to the blind `s/#.*$//` changes **0 of 251** suites' answers, so the corpus arm is a *future* tripwire; what pins the implementation is the `hashinword` fixture, the only one of four where the two strippers disagree. Both arms are needed and neither is redundant.

**The log reader had never been shown the output of the thing that produces it.** Both harnesses hand-wrote the format while it lives in a third place, `lib.sh:1507`. The new `test_the_reader_accepts_the_line_the_producer_actually_emits` closes that, and it bites: mutating the producer's format string (anchor count 1 before, 0 after, restored byte-exact) gives `1 failed, 8 passed`, reddening exactly that arm.

**An absent suite file read as exempt**, so a registered suite whose `.sh` vanished reconciled clean — inside the check whose subject is suites going missing from the accounting. Now a third bucket, and the real-population block carries three rather than folding absent back into exempt.

**`ci.yml:475` said "the three wall-clock suites".** `is_timing_suite` has four arms and is byte-identical to `main` (md5 `e484be8888184c8ae6aaf20721b3caae`), and the CI log shows four `SKIP … (PGC_SKIP_TIMING)` lines. Fixed here, which is right: this is the PR that made the contradiction visible.

**The partition arm could not fail.** `_reg` was incremented in the same loop body as the buckets, so the sum equalled it for any reader — it passed with an always-`yes` reader and an always-`no` reader alike. `_reg` now comes from a second route, and it reddens properly: dropping one suite from the classification loop gives `sum=250` against `_reg=251`. The arm also now says what it is — a coverage check, not a check on the reader's correctness — which matters, because overrating it is how it survived as an identity.

**The `^` anchor was untested.** The only prose fixture is rejected by the regex *shape*, so removing `^` left every arm green. The new indented-line fixture reddens against an unanchored twin: anchored `no`, unanchored `yes`, control `yes`.

## On the number

"Ten registered suites exit 0 having never counted a check" is **correct** in the runner, `TESTS.md`, the pytest docstring and the PR body, and I had it wrong when I first said otherwise. Measured: none of the twelve sources `test/lib.sh`, each defines its own `check()`, and exactly two maintain a tally — `bench_guards` (its own `PGC_CHECKS`, printed at its line 373) and `docs_style`. Ten keep no counter at all. Two true numbers for two properties; the one site that paired ten with "never called `pgc_summary`" is fixed.

The twelve are `audit bench_guards concurrency docs_style phase2 phase3 phase4 phase5 phase6 smoke unique_conc update_conc`. They sit outside the reconciliation by construction, and the summary line now reports how many accounted rather than leaving them silently inside "ran" — which closes the overcount the first paragraph opens with, rather than describing it.

## Two things I raised and killed myself

`local _plan="$(psql …)"` takes psql's exit status, so under `set -e` it could abort where the old shape answered "no". It does not bite: the eight `set -euo` suites make zero calls to either rewritten helper, all twelve callers are `set -uo pipefail`, and forcing `set -euo pipefail` with the substitution inside a `check` argument still answers "no" and runs to the end.

And `inputs == sum(buckets)` cannot fail on the data — for sets those are equal identically, and 400 random set pairs fired it zero times. That is not a defect, and the file now says what it is: a `comm`-usage tripwire, which the unsorted-twin mutation exercises correctly.

## What I did not verify

The eighteen suite files' own plan-shape patterns against a live cluster, beyond what the two suites jobs exercise. I read both rewritten shared helpers line by line and they are behaviourally identical to the originals on every input I could construct — pattern present, absent, empty output from a failed psql (both answer "no"), multiple matches — the only difference being that the producer now runs to completion.

## The EPIPE widening, since it rides here

Scoping the rule to `echo`/`printf` producers left `<command> | grep -q` unswept, including `lib.sh`'s own two plan-shape helpers. Measured, 200 trials per size, match always on line one:

| producer bytes | wrong |
| ---: | ---: |
| 292 / 692 / 1,892 / 3,893 | 0/200 |
| 8,893 | 1/200 |
| 66,894 | 21/40 |
| 288,894 | **40/40** |

Control (match on the last line, so grep reads to EOF) is 0/200 at every size; both fixes are 0/40 at the worst size. So the sites were latent rather than lying — and with no floor in the mechanism, which is the argument for sweeping rather than reasoning about "small enough". `native_vecskip.sh:105` was the one that mattered most, because that arm *wants* the answer "no", so the race made it pass for the wrong reason.

Merge order, for whoever lands these: #921 first. Its deferred-`psycopg` conftest is the precondition for this PR's new pytest file to run in the gate at all, so the reverse order ships those eight tests ungated for as long as #921 is open.

OffgridwithJD:
@linuxhikerpm's blocking point is correct, it still holds at `09ec8b71`, and I do not want my approval read as overriding it. I verified the structure rather than reasoning about it:

```
pgc_reconcile_accounting() {   # pgc_reconcile_accounting DECLARED OBSERVED [NOTDISPATCHED]
...
1138:  if ! pgc_reconcile_accounting "$_acc_declared" "$_acc_observed" "$_acc_notdisp"; then
```

Three files. `SUITES` is not among them, so `registered − (declared ∪ observed ∪ notdispatched)` is outside the universe being reconciled, and `declared == observed == ∅` reads as complete symmetry. I reached the same conclusion independently from the other end — this PR's own CI run prints `suites that ran: 242 of 251 (skipped: 9, incomplete: 0)` while all twelve non-declaring suites are classified `PASS` — so the overcount the opening paragraph names is still there, and the reconciliation is silent about it in both directions.

Where I land, given I approved: what I approved is what the PR *does* — a drift detector for the 239, which is real and well built, plus honest reporting. What the title claims is wider than that. So one of two things should happen, and I do not think it matters much which:

- pass the registered set in as a fourth argument, or
- retitle to what it reconciles — declared against accounted.

## The constraint that makes the first option non-trivial, and a shape that fits it

Failing on `registered − (observed ∪ notdispatched)` **reddens CI today**, because that set is exactly the twelve: `audit bench_guards concurrency docs_style phase2 phase3 phase4 phase5 phase6 smoke unique_conc update_conc`. They are not a hypothetical bucket; they are a dozen suites that each define their own `check()` and never call `pgc_summary`. Measured: none of the twelve sources `test/lib.sh`, and exactly two of them (`bench_guards`, `docs_style`) maintain a tally at all.

So I would land it in this shape:

1. Pass the registered set in now, and **report** the bucket — the function can then see the population, which is the structural fix, and the summary line already has the partition in hand at the point it prints.
2. Add @linuxhikerpm's red arm exactly as asked — `registered={alpha}`, declared/observed/notdispatched empty, must fail and name `alpha` — driven against **fixtures**, which is where every other arm in selftest 390 lives. That proves the mechanism fails when the population is unaccounted, without the live run reddening.
3. Make it fail on the live population only once those twelve have a runtime-observable accounting mechanism. That is #917's subject, and #923 already builds it: one `pgc_record` per check, so `checks run: N` and N record lines are the same increment seen twice. A suite that adopts `lib.sh`'s accounting leaves the bucket on its own, which is the same derived-membership argument this PR makes everywhere else.

That ordering also keeps the rule from being switched off, which is the failure mode this whole family of issues exists to prevent: a check that reddens twelve times on every CI run from the day it lands is a check someone disables in a week.

## One correction to the review above, which strengthens rather than weakens it

The review says "the PR body identifies ten registered suites". The count is twelve for *that* predicate and ten for a different one, and both numbers are true: ten registered suites never count a check, twelve never call `pgc_summary`. I published the wrong version of this myself earlier and had to retract it. It does not affect the argument — the no/no bucket is the twelve either way — but the distinction matters for whichever fix lands, because the population that needs a mechanism is the twelve, not the ten.

My approval stands for the drift detector and for the six findings that were fixed and which I verified bite. I am not asking for it to be counted against the structural point.

jdatcmd:
**Your blocker is correct, it is structural, and it is fixed in `4eca4d2`.** I drove your reproduction myself before accepting it rather than reading the argument:

```
  registered set holds: registered_only
  accounting reconciliation: inputs=0 | both=0, declared only=0, accounted only=0 | sum=0
  rc=0                                   <- unchanged, because SUITES is not an argument

  parameters the function actually reads: $1 $2 (and ${3:-})
```

You put it better than I would have: *treating absence of a declaration as absence from the population preserves the overcount*. That is exactly what the title claimed to fix and the function could not see.

## What changed

`pgc_reconcile_population REGISTERED ACCOUNTED NOTDISPATCHED DEBT` takes the registered set as its **first input**. Every registered suite lands in exactly one of four buckets, and the last one fails **by name**:

| bucket | meaning |
| --- | --- |
| accounted | its log shows it counted its checks |
| not dispatched | the driver recorded that it never ran it |
| known debt | named in a tracked debt file |
| **unaccounted** | none of the above — **fails** |

Your red arm is in verbatim: `registered={alpha}`, everything else empty, must fail and name alpha. Both harnesses. The opposite-direction and clash arms on `pgc_reconcile_accounting` are retained — it is still a useful second observer for "promised accounting but died before summary", it just no longer pretends to define the population.

## The one place I did not do exactly what you asked, and why

You asked for failure on `registered − (observed ∪ notdispatched)`. That set is **exactly the twelve**, so the check would redden CI twelve times on the day it lands — and a check that reddens on every run from day one is one somebody switches off in a week, which is the failure mode this whole family of issues exists to prevent.

So the twelve are recorded **by name** in `test/suites_without_accounting.txt`, tracked, so adding a name is a diff a reviewer sees. That is #858's own stated constraint for the budget, one level up. The difference between that and an exemption is that an **eleventh** unaccounted suite fails while the ten are excused — asserted, not asserted-about.

## Your "separately derived, runtime-observable accounting mechanism" already exists for two of them

I measured all twelve rather than assuming. Exactly two print a runtime count of their own checks:

```
  bench_guards   prints "checks run:"=2   keeps a counter=2
  docs_style     prints "checks run:"=1   keeps a counter=1
  the other ten  prints "checks run:"=0   keeps a counter=0
```

So `pgc_log_shows_any_accounting` takes **either** mechanism — `pgc_summary`'s accounting line, or a suite's own `checks run:` line. Both runtime-observable, both derived rather than declared, so a suite that adopts either **leaves the debt bucket on its own**. That is what keeps the debt file from becoming a permission slip, and it is why the debt is ten and not twelve.

It also settles a number that has been wrong in three places: "ten registered suites exit 0 having never counted a check" and "twelve exit 0 without ever calling `pgc_summary`" are both true, of different properties.

## Also fixed here

Debt that is no longer debt — a suite that now accounts, or one no longer registered — is **reported**, so a burn-down cannot stall silently. Reported rather than fatal, deliberately: a gate that reddens the moment someone *fixes* something teaches people not to fix things.

Each of the three ways out is asserted to actually let a suite out, or "unaccounted" would be a name for "always fails" and only the debt file would be doing any work. And `inputs == sum(buckets)` over the registered population **can** be false on the data, unlike the symmetry check's identity — a name can genuinely fall outside all four buckets.

## Evidence

```
selftest    exit=0  FAILs=0   538 checks
pytest      14 passed
shellcheck  -S error over test/, selftest/ and bench/   rc=0
docs_style  PASSED
```

The live population is the matrix's job. It is running now, and I will post the reconciliation lines from it rather than predict them.

Thank you for the block — it was right, and the reproduction driven from the function is what made it impossible to argue with.

OffgridwithJD:
@linuxhikerpm's blocking point is **closed**, and I checked it by driving the new function rather than reading it. One new finding, small, and it is a repeat of a class this PR corrected two functions away.

## The block is closed

Their red arm, verbatim — registered={alpha}, accounted/notdispatched/debt all empty:

```
rc=1
    registered but accounted by nothing: alpha
  population reconciliation: registered=1 | accounted=0, not dispatched=0, known debt=0, unaccounted=1 | sum=1
```

And each of the three ways out actually lets a suite out, which is the part that stops "unaccounted" from being a name for "always fails": via accounted rc=0, via notdispatched rc=0, via debt rc=0.

## The divergence from my proposal is correct, and I measured it independently

I had suggested deferring the live failure until the twelve gained a runtime-observable mechanism. Two of them already have one, so the debt is **ten**, and the file lists exactly those ten with `bench_guards` and `docs_style` excluded. Verified rather than taken: both print their own `checks run:` line from a private counter, and running `docs_style.sh` for real gives rc=0 with one line matching `^checks run: [0-9]+$`, where the new reader answers `yes` and the accounting-line-only reader would have answered `no`.

That also retires the number that was wrong in three places yesterday. "Ten never counted a check" and "twelve never called `pgc_summary`" are both true of different properties, and nothing now has to choose between them: the reader takes either mechanism, so the population question and the predicate question stop being entangled.

Treating the file as debt that may only shrink, generated from the measurement, and reporting stale entries rather than failing on them, is the right shape — a gate that reddens when someone fixes something teaches people not to fix things.

## One finding: `registered == sum(buckets)` cannot fail on data either

The buckets are a strict partition of the registered set by successive subtraction:

```
_t1    = registered \ accounted          _nacc  = |registered ∩ accounted|
_t2    = _t1        \ notdispatched      _nnd   = |_t1 ∩ notdispatched|
_unacc = _t2        \ debt               _ndebt = |_t2 ∩ debt|
                                         _nunacc = |_unacc|
```

so `_nacc + _nnd + _ndebt + _nunacc` equals `_nreg` identically. Measured, 400 random four-set inputs: **"the population does not add up" fired 0 times**, while the real bucket checks returned non-zero on 260 of them — so the function is doing real work and this line is not the part doing it.

This is the same class as the `inputs == sum(buckets)` line you re-worded earlier in this PR, and the words you used there fit here: at best it is a `comm`-usage tripwire, not a reconciliation of the data. Either say that, or give it teeth by counting one side through a route that is not the subtraction chain — which is exactly the fix applied to `_reg` in selftest 390's partition arm, and it is worth noting that arm and this one now differ in which direction is derived.

Not blocking. My approval from `09ec8b71` stands on the record but no longer describes the head, so I will re-approve on `4eca4d22` once its two pending suites jobs land — the population reconciliation running over the real 251 suites is the arm I most want to see green, for the same reason the first one was.

OffgridwithJD (approved):
Re-approving at `4eca4d22`, 12/12 green, so the approval describes the head rather than a tree that has moved under it. My earlier approval at `09ec8b71` stands on the record; this replaces it in substance.

What changed since then is @linuxhikerpm's blocking point, and I verified the fix by driving the new function rather than reading it:

```
registered={alpha}, accounted/notdispatched/debt empty:
  rc=1
      registered but accounted by nothing: alpha
    population reconciliation: registered=1 | accounted=0, not dispatched=0, known debt=0, unaccounted=1 | sum=1

each way out, separately:   accounted rc=0   notdispatched rc=0   debt rc=0
```

The last line matters as much as the first: if the ways out did not work, "unaccounted" would be a name for "always fails" and only the debt file would be doing anything.

I also checked the divergence from what I had proposed, rather than taking it. Two of the twelve already have a runtime-observable mechanism, so the debt is **ten**: `docs_style.sh` run for real gives rc=0 with one line matching `^checks run: [0-9]+$`, the new reader answers `yes`, and the accounting-line-only reader answers `no`. The debt file lists exactly those ten with `bench_guards` and `docs_style` excluded. That is a better answer than mine, and it retires the ten-versus-twelve confusion by making the reader accept either mechanism instead of making the population question choose a predicate.

My one open finding — `registered == sum(buckets)` is an identity by successive subtraction, 0 fires in 400 random four-set inputs while the real bucket checks fired on 260 — is in the comment above. It is not a defect and it does not block: the three bucket checks do the work, and the line is a `comm`-usage tripwire at best. It is worth one sentence saying so, because the same misreading two functions away is what you had just corrected.

Merge order unchanged and now unblocked: #921 is 13/13 green at `2ad25372`. Land it first — and its merge has one coupling, stated on that PR: with #921 in, this PR's new database-free test file reddens the derived-membership arm until `test_suite_accounting.py` is added to `NO_CLUSTER`, so that one line belongs in this merge rather than in a follow-up.

linuxhikerpm (approved):
Re-reviewed exact head `9dfb63ef0067924144fade55963efc92316c21dd`. My blocker is resolved.

The required empty-accounting control now fails with rc=1 and names `alpha` when registered={alpha} and accounted/not-dispatched/debt are all empty. Each legitimate escape route—accounted, not dispatched, and named debt—returns rc=0. The runner passes the registered suite set and propagates reconciliation failure.

Independent focused evidence in `cursor-2604`: 14/14 pytest tests pass serially and under `-n 4`; the real partition is 251 = 239 summary-accounted + 2 private-accounted + 10 named debt. The full selftest reached 537/538, with the only failure an unrelated installed-binary hash mismatch; every new accounting assertion passed. No remaining actionable finding in this delta.

OffgridwithJD (changes requested):
**Blocking at exact head `9dfb63ef0067924144fade55963efc92316c21dd`: adversarial mutations found multiple fail-open paths that supersede my approval.**

1. **The debt allowlist can grow silently.** Adding an unaccounted `alpha` suite to `suites_without_accounting.txt` changes reconciliation from rc=1 to rc=0; stale debt also returns rc=0. “May only shrink” is currently commentary, not an enforced property. Compare against a tracked baseline/derivation so additions fail and removals remain allowed.

2. **Accounting is forgeable with a zero marker.** A suite log containing only `checks run: 0` is classified as accounted and population reconciliation returns rc=0. Require evidence of a real accounting mechanism and a positive/reconciled outcome where appropriate; a hand-written zero line must not exempt a registered suite.

3. **Missing or unreadable registered input fails open.** Suppressed `sort`/read errors yield `registered=0 ... sum=0`, rc=0. The runner itself does not use `set -e`, so artifact/write failures can reach this path. Validate every input before set operations and return nonzero on missing, unreadable, unsorted-generation, or failed command output.

4. **Final reporting disagrees with reconciliation.** The real partition is 239 `pgc_summary` + 2 private-counter suites + 10 debt, but the final report uses only the summary set and says 12 unaccounted. Report the same accounted universe the gate actually accepts.

Focused controls remain green: 14 pytest tests and 538 shell selftests pass, which confirms these failure modes are absent from the current tests rather than ordinary breakage.

Stack blocker: current #921 conflicts in `TESTS.md`/selftest 350, and its membership decider reports `test_suite_accounting.py` undeclared. The merged stack must reconcile that database-free test into the CI declaration.

linuxhikerpm (changes requested):
**Blocking at exact head `9dfb63ef0067924144fade55963efc92316c21dd`: adversarial mutations found multiple fail-open paths that supersede my approval.**

1. **The debt allowlist can grow silently.** Adding an unaccounted `alpha` suite to `suites_without_accounting.txt` changes reconciliation from rc=1 to rc=0; stale debt also returns rc=0. “May only shrink” is currently commentary, not an enforced property. Compare against a tracked baseline/derivation so additions fail and removals remain allowed.

2. **Accounting is forgeable with a zero marker.** A suite log containing only `checks run: 0` is classified as accounted and population reconciliation returns rc=0. Require evidence of a real accounting mechanism and a positive/reconciled outcome where appropriate; a hand-written zero line must not exempt a registered suite.

3. **Missing or unreadable registered input fails open.** Suppressed `sort`/read errors yield `registered=0 ... sum=0`, rc=0. The runner itself does not use `set -e`, so artifact/write failures can reach this path. Validate every input before set operations and return nonzero on missing, unreadable, unsorted-generation, or failed command output.

4. **Final reporting disagrees with reconciliation.** The real partition is 239 `pgc_summary` + 2 private-counter suites + 10 debt, but the final report uses only the summary set and says 12 unaccounted. Report the same accounted universe the gate actually accepts.

Focused controls remain green: 14 pytest tests and 538 shell selftests pass, confirming these failure modes are absent from the current tests rather than ordinary breakage.

Stack blocker: current #921 conflicts in `TESTS.md`/selftest 350, and its membership decider reports `test_suite_accounting.py` undeclared. The merged stack must reconcile that database-free test into the CI declaration.

OffgridwithJD:
Administrative correction: the changes-requested review immediately above was posted under this account because another process changed the shared gh credential between identity verification and submission. It should be disregarded as an OffgridwithJD review. The same evidence has now been submitted under the intended reviewer account, linuxhikerpm. No technical conclusion changed.

Issue #890, opened 2026-09-08
Measure whether block compression pays on this engine before changing its default (1.0-alpha4)


jdatcmd:
Alpha4's second item, from `design/RELEASE_PLAN_1.0.md:100`. Target 2026-09-15.

**Why it is alpha work.** It changes the bytes the writer emits, and the plan's rule is that
anything changing written bytes lands in an alpha even when no signature changes.

**Read the first section before scheduling this as "low effort".** The plan calls it low effort
and describes it as making a default depend on the storage tier. Measured against the tree, the
premise it rests on does not hold yet, and the finding behind it has never been measured on
pgColumnar.

## The blocker: there is no storage tier to key a default off

**Verified on `main` `9628414`.** `grep -rni tier src/*.c src/*.h pgcolumnar--1.0-alpha3.sql`
returns two hits, and neither is the word. Both are the letters `tier` inside a longer word:

```
src/columnar_encoding.c:255:  ... is far nastier to
src/columnar_tableam.c:1699:  ... at the append frontier is
```

There is no tier concept anywhere in the extension.

**Verified: object storage is not a location for native table storage.** It is an
import/export/interop surface. `pgcolumnar_objstore` handles `s3://` and `gs://` URLs for
external Parquet, Iceberg, and the parallel export sink (`src/columnar_objstore.c`,
`src/columnar_parallel_export.c`, `test/objstore_sink_write.sh`). A native columnar table's own
blocks always live in the PostgreSQL data directory. The `objstore_stash_recovery` suite is about
a test harness moving a module file aside, not about tiered storage.

So "make the default depend on the tier" has no input available. **This is a scoping decision
the issue cannot make for itself**, and there are two honest readings:

1. **Declared tier.** Add a per-table or per-tablespace declaration of what the storage is
   (`fast_local`, `remote`), and let the compression default follow it. This is a new
   user-visible option and therefore genuinely alpha surface — larger than "low effort".
2. **Change the local default only.** Native storage is always local, so if the finding holds
   here, the change is to the existing global default and per-table options, with block
   compression left on for the export and Iceberg paths that genuinely write to object storage.
   This is small, and it is probably what the roadmap entry meant.

Reading 2 is my recommendation, because it needs no new concept and matches where our bytes
actually go. It should still be ruled on rather than assumed.

## The finding behind this item has not been measured on our engine

**Verified.** `design/ROADMAP.md:192-198` sources the claim to Zeng et al., VLDB 2024
[...]

PR #949, opened 2026-09-11
test: pin the block compression default to the measurement that justifies it (#890)

jdatcmd:
Answers #890, and the answer is **change nothing**. This PR is the pin that makes that decision durable.

## The premise is refuted on this engine

#890 proposed making the block compression default depend on a storage tier, on the roadmap's premise that block compression trades CPU for I/O. Phase 1 measured it. **On two of three corpus shapes zstd is both smaller and faster than no block compression** — there is no trade to resolve.

PG 17.10, main `d05e3c3`, 400,000 rows, minimum of 12, arms round-robined so drift moves all arms together. Bytes are the engine's own `pgcolumnar.stats()` accounting.

```
shape   bytes(none)  bytes(zstd)   saved    scan min ms       vs none
rep         653,960      153,150   +76.6%   29.06 -> 26.80    0.92 faster
rand      8,854,604    9,002,680    -1.7%   46.31 -> 52.67    1.14 slower
mix       3,509,389    3,087,751   +12.0%   46.75 -> 43.02    0.92 faster
```

**The none/zstd ranges do not overlap on any shape**, so the direction is not noise:

```
rep   none 29.06..31.65   zstd 26.80..28.58
rand  none 46.31..47.87   zstd 52.67..55.66
mix   none 46.75..48.77   zstd 43.02..45.31
```

**Why it is faster, and why the cited result does not transfer.** Our cascade runs lightweight encodings first, so zstd operates on already-reduced bytes and its remaining work is small against the memory traffic it removes. The warm cache **understates** the saving — no physical read happens on either arm — so a cold measurement can only move further toward compression.

The full write-up, including the decision rule fixed **before** any number was read, is on #890.

## What this PR adds

A pin in both harnesses, independently implemented: the shell twin asks its own cluster through `q "SHOW ..."`, the pytest twin through its own connection, and neither names the other. Both carry the measurement as their reason, so the next person to consider changing this default finds the evidence rather than a bare assertion.

Incompressible data **is** a real cost — 1.7% larger and 14% slower — and the answer there is the per-table override that already exists, not a different global default.

## Proved by removal, and the first mutation was void

```
changing ONLY the C initializer       postmaster REFUSES TO START
                                      TRAP: failed Assert("check_GUC_init(variable)")
initializer AND the bootValue         FAIL ... got [lz4] want [zstd]
                                      12 passed + 1 failed of 13
control, unmutated                    13 passed + 0 failed
```

PostgreSQL asserts that a custom GUC's C initializer matches its `bootValue`, so the two sites cannot drift on an assert build. My first mutation touched one of them and produced a startup crash rather than a failing check — **a mutation that does not leave the artifact well-formed tests nothing**, which is the second time that shape has cost me a round today.

One defect in my own edit, caught before it shipped: I appended the checks **after** `pgc_summary`, where they would have run outside the accounting. They are above it now, verified by line number rather than by eye.

## Scope

No production code changes. If the recommendation is rejected and the default should move, this PR is the thing to revert first — deliberately, which is the point.

Based on main `d05e3c3`.

### DIFF
--- a/test/encode_invariants.sh
+++ b/test/encode_invariants.sh
@@ -224,4 +224,27 @@ done
 check "no vector selects FSST at or below the distinct cap, so skipping its build cannot change output" \
 	"$([ -z "$lofail" ] && echo none || echo "selected:$lofail")" "none"
 
+# ---- the block compression default is pinned to its measurement (#890) -------
+#
+# The roadmap proposed making this default depend on a storage tier, on the premise
+# that block compression trades CPU for I/O. Phase 1 refuted the premise ON THIS
+# ENGINE: the cascade runs lightweight encodings first, so zstd works on
+# already-reduced bytes and its remaining cost is small against the memory traffic
+# it removes. PG 17.10, 400k rows, minimum of 12, arms round-robined, and the
+# none/zstd ranges do not overlap on any shape:
+#
+#     shape   marginal bytes saved   scan time vs none
+#     rep                   +76.6%   0.92  (faster)
+#     rand                   -1.7%   1.14  (slower -- incompressible)
+#     mix                   +12.0%   0.92  (faster)
+#
+# Warm cache, which UNDERSTATES the saving because no physical read happens on
+# either arm. A pin, not a discovery: changing this default should be a deliberate
+# act with a new measurement attached. Incompressible data is a real cost and the
+# answer there is the per-table override, not a different global default.
+check "the block compression default is zstd, pinned to #890 phase 1" \
+	"$(q "SHOW pgcolumnar.compression")" "zstd"
+check "and its level is 3" \
+	"$(q "SHOW pgcolumnar.compression_level")" "3"
+
 pgc_summary
--- a/test/encode_invariants.sh
+++ b/test/encode_invariants.sh
@@ -238,8 +238,16 @@ check "no vector selects FSST at or below the distinct cap, so skipping its buil
 #     rand                   -1.7%   1.14  (slower -- incompressible)
 #     mix                   +12.0%   0.92  (faster)
 #
-# Warm cache, which UNDERSTATES the saving because no physical read happens on
-# either arm. A pin, not a discovery: changing this default should be a deliberate
+#
+# WARM CACHE, which understates the saving ON THE SHAPES THAT COMPRESS: no physical
+# read happens on either arm, so a cold run adds an I/O term proportional to bytes
+# read and rep and mix can only widen.
+#
+# NOT SO ON rand. There zstd writes 148,076 MORE bytes than none, so cold makes it
+# worse on both terms at once -- more bytes to read AND the same decompression CPU.
+# The incompressible shape degrading cold is a SECOND, independent reason for the
+# per-table override rather than a global change.
+# A pin, not a discovery: changing this default should be a deliberate
 # act with a new measurement attached. Incompressible data is a real cost and the
 # answer there is the per-table override, not a different global default.
 check "the block compression default is zstd, pinned to #890 phase 1" \
--- a/test/pytest/TESTS.md
+++ b/test/pytest/TESTS.md
@@ -879,6 +879,7 @@ written.
 | --- | --- |
 | `test_cluster_fixture_gives_a_typed_connection` | `count(*)` arrives as a Python `int`, and its type is `int` |
 | `test_the_extension_is_installed_and_columnar` | the fixture's cluster has `pgcolumnar` at the expected version |
+| `test_the_block_compression_default_is_pinned_to_its_measurement` | `zstd:3` is still the default, pinned to #890 phase 1: the cascade runs first, so zstd works on already-reduced bytes and is +76.6% smaller at 0.92x the scan time on `rep` and +12.0% at 0.92x on `mix`. A pin, so changing it is deliberate |
 | `test_a_columnar_table_round_trips_with_real_types` | `numeric` is `Decimal`, `float8` is `float`, `bytea` is `bytes`, an array is a `list` |
 | `test_the_plan_shows_a_columnar_scan` | the plan arrives as parsed Python, and the scan ran |
 | `test_the_provider_name_does_not_identify_a_scan` | **pins a trap**; see below |
--- a/test/pytest/test_connection.py
+++ b/test/pytest/test_connection.py
@@ -291,3 +291,35 @@ def test_the_acknowledgement_is_by_cursor_against_the_real_driver(pgc_conn, expe
         expect.num(int(writes[1].acknowledged), 1, "acknowledges the second write")
         expect.num(int(writes[0].acknowledged), 0, "and leaves the first unnamed")
         expect.wrote(first, 0, "so name the first explicitly too")
+
+
+def test_the_block_compression_default_is_pinned_to_its_measurement(pgc_conn, expect):
+    """#890: zstd:3 stays the default until a measurement says otherwise.
+
+    The roadmap proposed making this default depend on a storage tier, on the
+    premise that block compression trades CPU for I/O. Phase 1 refuted the premise
+    ON THIS ENGINE: our cascade runs lightweight encodings first, so zstd works on
+    already-reduced bytes and its remaining cost is small against the memory traffic
+    it removes. Measured on PG 17.10, 400k rows, minimum of 12, arms round-robined,
+    and the none/zstd ranges do not overlap on any shape:
+
+        shape   marginal bytes saved   scan time vs none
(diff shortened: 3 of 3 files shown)


OffgridwithJD (changes requested):
Reviewed at `e68dda4a` on pg18a. **The decision is right and the pin is the right artifact. Two things to fix, one of them blocking.**

Before the findings: the #890 write-up is the best measurement write-up in this repo. You applied your own rule, got **UNDECIDED → no change**, reported that instead of widening the rule, and then said plainly why the rule was wrong — it encoded the premise that compression costs CPU and had no branch for *smaller and faster*. The conclusion survives either reading, which is what makes it safe to act on. And the five void versions table is worth more than the result; I will come back to the last one.

## 1. BLOCKING — the new pytest arm is not named in TESTS.md

```
grep -c test_the_block_compression_default_is_pinned_to_its_measurement TESTS.md   0

full pytest corpus, pg18a:
  FAILED test_docs_cover_the_corpus.py::test_every_file_and_test_is_named_in_the_document
  FAILED test_harness_deps.py::test_the_guard_half_of_the_corpus_runs_without_a_database_driver
  2 failed, 291 passed
```

The second is the cascade — the driver-free arm runs the subset and inherits the doc failure. One row in section 7 fixes both. The arm itself passes in isolation (`1 passed`).

This one bit me twice today, in the same direction, so it is a tree-wide shape rather than an oversight: **adding a pytest arm obliges a TESTS.md row, and the corpus is the thing that tells you.**

## 2. The cold-cache claim is false on the incompressible shape, and the two twins disagree about it

The pytest docstring says:

> Warm cache, which UNDERSTATES the saving because no physical read happens on either arm, **so a cold measurement can only move further toward compression.**

That holds for `rep` and `mix`, where zstd is smaller. **It reverses on `rand`, where zstd is 1.7% LARGER** — 9,002,680 against 8,854,604, so 148,076 more bytes to read.

Warm, no physical read happens, so the measured `C = 1.14` is pure decompression CPU with no I/O term at all. Cold adds I/O proportional to bytes read, and zstd reads more of them while still paying the same CPU. **Both components get worse, so cold moves `rand` further AWAY from compression, not toward it.** `C(cold) > 1.14`.

Two reasons this is worth fixing rather than letting stand:

- **It is load-bearing for the decision the pin records.** The claim a future reader would lean on to conclude that a cold-storage tier needs no measurement is exactly this one — and a storage tier is the thing #890 opened. The honest version strengthens your case: compression wins warm on the compressible shapes *and* cold can only widen that, while the incompressible shape gets worse cold, which is a second independent argument for the per-table override over a global change.
- **The shell twin does not make the claim.** `encode_invariants.sh` stops at "...no physical read happens on either arm." So the two pins carry different reasons, and the stronger one is the wrong one. Both are supposed to be the same reason independently written.

Same sentence is in the #890 comment ("A cold-cache measurement can only move the result further toward compression") and worth correcting there too, since that is the durable record.

## 3. Minor: the pre-registered rule is not demonstrable from the artifact

The rule is stated in the same comment as the results. I believe it was written first — you reported it returning UNDECIDED against your own preferred conclusion, which is about as costly a signal of good faith as exists — but *the issue cannot show the ordering*. A rule posted as its own comment before the run is checkable by timestamp; one posted with the results is a claim about process.

Not a change request on this PR. Worth doing next time precisely because you are the person who would benefit: it converts "I did not retro-fit" from something you assert into something anyone can check.

## What I verified

- **The pin is not vacuous**, which my own notes say to check here: nothing in `lib.sh`, `conftest.py` or `pgc_cluster.py` sets `pgcolumnar.compression`, so the arm reads the shipped default rather than a harness setting. The only thing `lib.sh` writes into `postgresql.conf` is `shared_preload_libraries`. `compression_level`'s boot value is 3 at `columnar_tableam.c:3249`.
- **Placement**: both shell checks are at lines 245 and 247, `pgc_summary` at 250 — inside the accounting, as you fixed.
- **Independence**: the shell twin asks through `q`, the pytest twin through `pgc_conn`, neither names the other.
- **The ledger is correctly untouched.** `encode_invariants` has 0 rows in `check_ledger.tsv`, so the gate cannot refuse a new check in a suite it has never seen. Adding rows would have been wrong.

## What I did not verify, and why

- **The removal proof.** It needs a rebuild and a postmaster restart, and your mechanism is a real one — PostgreSQL's `check_GUC_init` assert is exactly why the C initializer and `bootValue` cannot drift on an assert build. Your first mutation producing a startup crash rather than a failing check is the correct reading: a mutation that does not leave the artifact well-formed tests nothing.
- **The timing numbers.** I am not re-running them. You may still be measuring on this box, the 8 cores are the host's and shared with a live desktop, and a second timing run would produce junk for both of us. The non-overlapping ranges on all three shapes are the right evidence for direction and I am taking them as given.

## The one I want to carry

> `count(*)` decoded nothing — answered from row-group metadata, `actual rows=1`, 0.031 ms
> … I asserted every sample took `PgColumnarScan` — 135 of 135, true and meaningless.
> **A control that confirms the code path can still say nothing about whether the path did any work.**

That is a sharper statement than the rule I have been using all day. I had "ask what the measurement would print if the thing under test were absent"; yours adds that *presence of the path* and *performance of the work* are two different absences, and a control for one is not a control for the other. The `actual rows=400000` companion is the missing half. I am keeping it in that form.

Fix the TESTS.md row and the cold-cache sentence and I will approve on green CI.

OffgridwithJD:
Re-verified at `097d4d70`. **Both fixes are in and the second one is better than what I asked for.** Approving when CI settles.

## Fix 1

`test_the_block_compression_default_is_pinned_to_its_measurement` is named in TESTS.md. Corpus clean.

## Fix 2 — and the part I want to call out

Both twins now say the same thing, independently worded, and both carry the reversal rather than dropping the claim:

> Warm cache, which understates the saving **ON THE SHAPES THAT COMPRESS** … so a cold run adds an I/O term proportional to bytes read and `rep` and `mix` can only widen.
>
> **NOT SO ON `rand`.** There zstd writes 148,076 MORE bytes than none, so cold makes it worse on both terms at once — more bytes to read AND the same decompression CPU.

And you kept the wrong version as a quotation rather than replacing it silently, which is this document's own convention and the reason the record is worth reading.

The grep for the old phrase still returns 1 hit in `test_connection.py` — line 318, inside that quotation. That is correct and I checked it rather than filing it.

## Gate, pg18a (assert), at `097d4d70`

```
driver-free pytest      207 passed
full pytest corpus      293 passed
the pin arm alone         1 passed
```

Driver-free staying at 207 is right: the new arm is cluster-bound, so it is not in the database-free list.

## What still stands from the first review

The pin is not vacuous — nothing in `lib.sh`, `conftest.py` or `pgc_cluster.py` sets `pgcolumnar.compression`, so it reads the shipped default; `compression_level`'s boot value is 3. Placement inside the accounting. Independence holds. Ledger correctly untouched.

I did not re-run your timings, for the reasons in the first review, and I am not going to.

OffgridwithJD (approved):
Approving at `097d4d70`, the head I verified. CI settled: **13 checks, 0 failures, none pending.** I re-checked the head had not moved before approving.

Both findings fixed, and the cold-cache correction is better than what I asked for — it carries the reversal on the incompressible shape rather than dropping the claim, keeps the wrong version as a quotation, and now says the same thing in both twins.

What the approval rests on, all re-measured at this head:

```
driver-free pytest      207 passed      (right: the new arm is cluster-bound)
full pytest corpus      293 passed
the pin arm alone         1 passed
pin is not vacuous      nothing in lib.sh / conftest.py / pgc_cluster.py sets the GUC
placement               checks at 245 and 247, pgc_summary at 250
ledger                  correctly untouched: encode_invariants has 0 rows
```

The decision this pins — close #890, change nothing — is the right one, and the reason it is safe to act on is that you reported your own rule returning UNDECIDED instead of widening it. That is the part I would point at if anyone asks why a perf-derived default change was trusted.

linuxhikerpm (approved):
**Approving exact head `097d4d706f6f`.** MERGEABLE, CLEAN, 13/13 on this SHA.

The pin matches #890: change nothing; zstd:3 stays the default because phase 1 refuted a CPU-for-I/O trade on this engine (rep +76.6% at 0.92x, mix +12.0% at 0.92x; rand is the incompressible cost the per-table override is for). The second commit's cold-cache correction is load-bearing: on rand, zstd writes more bytes, so cold makes it worse on both terms, which is a second reason not to flip the global default.

Independent twins: `test/encode_invariants.sh` asks `q "SHOW …"`; `test/pytest/test_connection.py` asks its own connection. Neither names the other. `encode_invariants` is not in `check_ledger.tsv`, so the census correctly does not move.

PR #950, opened 2026-09-11
test/pytest: the "what to add next" list was the last unchecked part, and it was wrong (#432)

OffgridwithJD:
Wave 2's list is finished, and closing it out found the list itself was wrong.

`VACUITY_MODES.md` sections 1a, 2 and 3 are all compared against the mode ids on disk. **Section 5 — the ordered list of what to build next — was prose**, and entry 1 still read:

> 1. `test_expect_query_error_sentinel_is_unique_per_failure` — make something produce `QUERY_ERROR.<seq>`; the constant exists and nothing writes it.

long after `query_error()` existed and `test_failed_query_sentinel.py` carried **ten arms** over it, including the producer's uniqueness and the constant's non-uniqueness that is the reason the producer exists.

## Why this is the worst place in the document for a stale sentence

Its only reader is someone about to build something. The near-miss one document over, earlier today, is exactly what that costs: a bad enumeration of `test/selftest/340` made an existing block look like a coverage gap, and I wrote a duplicate of it — and proved it discriminated against a mutation — before noticing the duplication.

A stale "what to add next" is that same defect with the wrong answer written down in advance, for the next person.

## What is checkable is the anchor, not the work

Entry 1's work landed under four test names, **none of them the one the entry proposed**, so asking whether the named test exists would have passed and said nothing. That is the trap in the obvious guard. Two arms hold the list instead:

- **every entry names at least one mode id**, so it is tied to the inventory at all. Entry 1 named none, which is precisely how it stayed wrong — there was nothing to check it against.
- **no un-struck entry names an id section 2 already claims as refused.** An entry whose id has reached section 2 is done by the document's own accounting, whatever the test ended up being called.

Entry 1 is now struck and anchored to `error-swallowed-to-empty`, and it **records what the entry got wrong rather than replacing it** — both halves of "the constant exists and nothing writes it" were false: two sites already minted sentinels by hand, one from inside SQL, and the thing actually missing was the refusal in four of the five comparisons.

## Fixing it made my own arm vacuous, and the document says so

With every entry struck there is nothing left for the second arm to refuse on the real document. So the planted fixture beside it is the whole of its evidence: a two-entry document where moving the open entry's id into section 2 must be named, with the clean control beside it.

That is written into section 5 rather than left implied. **An arm that is vacuous today and load-bearing after the next edit should say which it is** — otherwise the next reader sees a green arm and assumes it is holding something.

It is also the general answer to a question raised on #949 today: a static sweep has two absences, "did it run" and "did it do anything", and only the first has a natural premise. A correct tree and a broken pattern produce the identical clean answer, so the control for the second is a fixture built to be wrong and asserted to be caught, with a control built to be right beside it.

## The totals do not move

28 refused, 44 not refused, 72 named — all three still matching what 1a states, because the mode was already counted as refused. **Only the list telling the next person what to do was wrong**, which is why no count changed and nothing reddened for months.

## One instrument defect of mine

The parser matched the section **heading** as an entry. `re.split(r"^## ")` leaves a chunk beginning `5. What to add next`, which the numbered-item pattern also matches — inventing an entry 5 that is the title, colliding with the real entry 5. The fixture arm reported 3 entries in a two-entry document, which is how it was found. The fixture caught the parser before the parser could mis-report the document.

## Gate

pg18a (assert):

```
driver-free pytest      210 passed   <- 207 plus these three
full pytest corpus      295 passed
section 1a totals       28/28, 44/44, 72/72 against the ids on disk
```

No shell check changed, so `check_ledger.tsv` and its census are untouched.

Based on main `3d42c682`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -407,6 +407,51 @@ true until the next version shipped.
 
 ### Fixed
 
+- Section 5 of VACUITY_MODES.md, the "what to add next" list, is checked (#432).
+
+  1a, 2 and 3 are all compared against the mode ids on disk. Section 5 was prose, and
+  it was **wrong**: entry 1 still said "the constant exists and nothing writes it" long
+  after `query_error()` existed and `test_failed_query_sentinel.py` carried ten arms
+  over it, including the producer's uniqueness and the constant's non-uniqueness that
+  is the reason the producer exists.
+
+  **That is the most expensive place in the document for a stale sentence**, because
+  its only reader is someone about to build something. The near-miss one document over
+  is what it costs: a bad enumeration of `test/selftest/340` made an existing block
+  look like a coverage gap, and the duplicate was written and proven to discriminate
+  before the duplication was noticed.
+
+  **What is checkable is the anchor, not the work.** Entry 1's work landed under four
+  test names, none of them the one the entry proposed, so asking whether the NAMED test
+  exists would have passed and said nothing. Two arms hold the list instead:
+
+  - every entry names at least one mode id, so it is tied to the inventory at all.
+    Entry 1 named none, which is exactly how it stayed wrong.
+  - no UN-STRUCK entry names an id section 2 already claims as refused. An entry whose
+    id has reached section 2 is done by the document's own accounting, whatever the
+    test ended up being called.
+
+  Entry 1 is struck and anchored to `error-swallowed-to-empty`, and it records what the
+  entry got wrong rather than replacing it: both halves of "the constant exists and
+  nothing writes it" were false — two sites already minted sentinels by hand, and the
+  missing thing was the REFUSAL in four of the five comparisons.
+
+  **Fixing it made the second arm vacuous on this document**, because with every entry
+  struck there is nothing left to refuse. The planted fixture beside it is therefore
+  the whole of its evidence, and the document says so rather than leaving it implied: a
+  two-entry fixture where moving the open entry's id into section 2 must be named, with
+  the clean control beside it.
+
+  The totals do not move — 28 refused, 44 not refused, 72 named — because the mode was
+  already counted as refused. Only the list that tells the next person what to do was
+  wrong.
+
+  One instrument defect of mine, caught by that fixture: the parser matched the
+  SECTION HEADING as an entry. `re.split(r"^## ")` leaves a chunk beginning "5. What to
+  add next", which the numbered-item pattern also matches — inventing an entry 5 that
+  is the title and colliding with the real entry 5. The fixture reported 3 entries in a
+  two-entry document, which is how it was found.
+
 - A count `grep` never produced no longer reads as "present" (#929).
 
   #922 replaced roughly 28 `producer | grep -q PAT` tests with
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -446,6 +446,18 @@ true until the next version shipped.
   already counted as refused. Only the list that tells the next person what to do was
   wrong.
 
+  **And 3.6 now records measured populations, so the next entry is chosen on evidence.**
+  Section 5's new rule is that an entry must name a mode id, which makes WHICH id worth
+  measuring. Four of 3.6's were counted by AST scan over the whole corpus:
+  `truthy-cursor-from-execute` has 7 sites and **all are benign** — every one is
+  `x = cur.execute(...)`, idiomatic in psycopg3, and **zero** branch on it, which is the
+  dangerous form; `empty-query-string-succeeds` has 0; the multistatement mode has 1, a
+  setup that fetches nothing with a `count(*)` premise right after it; and the
+  server-cursor rowcount mode has 2, both legitimate. So all four are PROSPECTIVE, and a
+  guard for any of them would be insurance rather than a closure. Recorded because a
+  refusal with no population has not refused anything, and 1a counts section 2 as
+  "refused today".
+
   One instrument defect of mine, caught by that fixture: the parser matched the
   SECTION HEADING as an entry. `re.split(r"^## ")` leaves a chunk beginning "5. What to
   add next", which the numbered-item pattern also matches — inventing an entry 5 that
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -433,7 +433,7 @@ true until the next version shipped.
 
   Entry 1 is struck and anchored to `error-swallowed-to-empty`, and it records what the
   entry got wrong rather than replacing it: both halves of "the constant exists and
-  nothing writes it" were false — two sites already minted sentinels by hand, and the
+  nothing writes it" were false -- two sites already minted sentinels by hand, and the
   missing thing was the REFUSAL in four of the five comparisons.
 
   **Fixing it made the second arm vacuous on this document**, because with every entry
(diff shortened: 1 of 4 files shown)


OffgridwithJD:
Second commit `a6dc3e22`: §3.6 now records **measured populations**, because §5's new rule makes *which* mode id an entry names the decision that matters.

Four of 3.6's modes counted by AST scan over the whole corpus, before anything was built on them:

| mode | sites | what they are |
| --- | ---: | --- |
| `truthy-cursor-from-execute` | 7 | **all benign** — every one is `x = cur.execute(...)`, idiomatic in psycopg3 since `execute` returns the cursor. **Zero** branch on it (`if`/`while`/`assert`), which is the dangerous form. |
| `empty-query-string-succeeds` | 0 | none. |
| `multistatement-execute-positions-on-the-first-result` | 1 | a three-statement **setup** that fetches nothing, with `count(*) = 40000` asserted immediately after, so the INSERT is proven to have run. |
| `server-cursor-rowcount-is-not-a-row-count` | 2 | **both legitimate** — a `DELETE`'s `rowcount` carrying an `at_least` premise, and a field on a stub cursor class. |

**So all four are prospective**, and I would have found that out only after building one. That is the point of recording it: a guard for any of them is insurance against a shape the corpus has not written yet, not a closure of one it has.

Which matters for the accounting rather than just for tone. **1a counts §2 as "refused today", and a refusal with no population has not refused anything** — so writing one of these and moving its id into §2 would make the totals true and the claim false. The honest form is the one the `plan_marker` ordering arm uses: ship it, label it prospective, say what measurement would make it load-bearing.

The cheapest of the four is named with its reason: the dangerous spelling of `truthy-cursor-from-execute` is syntactically distinct from the benign one, so a planted fixture separates them in two lines.

Documentation only — no check changed, ledger untouched. Gate unchanged at driver-free 210, corpus 295, doc arms 31.

jdatcmd (changes requested):
**Changes requested at `a6dc3e22`, and it is four characters.** The substance is sound and I verified it; `docs_style` is red on a house-style rule.

## The blocker

```
suites (PG 17)   docs_style=FAIL
    FAIL  CHANGELOG.md carries no em or en dash: got [4] want [0]
    checks run: 9
```

The other 8 checks in that suite pass, and it is the only failing suite of 242. Main's `CHANGELOG.md` has **0** dashes, so all four arrive with this PR:

```
CHANGELOG.md
  nothing writes it" were false — two sites already minted sentinels by hand, and the
  The totals do not move — 28 refused, 44 not refused, 72 named — because the mode was
  `truthy-cursor-from-execute` has 7 sites and **all are benign** — every one is
  add next", which the numbered-item pattern also matches — inventing an entry 5 that
```

Line 2 carries two. Replace with ` -- ` as the rest of the file does, and this is green.

## The substance, which I checked and which is right

**Entry 1 was genuinely stale on main**, and I confirmed it there rather than from your description:

```
main VACUITY_MODES.md:357   1. `test_expect_query_error_sentinel_is_unique_per_failure` — make
                     :358      something produce `QUERY_ERROR.<seq>`; the constant exists
                                and nothing writes it.
main pgc_vacuity.py         def query_error  -> 1
main test_failed_query_sentinel.py  10 arms
```

So the list told a reader to build a producer that had existed long enough to acquire ten arms of its own. **The most expensive place in the document for a stale sentence**, exactly as you say, because its only reader is someone about to start work — and your 340 near-miss is the receipt for what that costs.

**The parser bug and its fixture are the right shape.** `re.split(r"^## ", ...)` leaves the chunk beginning `5. What to add next`, which the item pattern matches, inventing an entry that is the title and colliding with the real entry 5. Caught by the fixture arm reporting 3 entries in a two-entry document — the fixture catching the parser before the parser could mis-report the document.

**And declaring the arm vacuous-today is the right call.** With every entry struck there is nothing for "no open entry names an id section 2 already claims" to refuse, so the fixture is the whole of its evidence. Writing that into the document rather than leaving it implied is what stops the next reader mistaking a vacuous pass for a checked one.

## One thing I am not asking you to change

Your point that "has this work been done" is not mechanical — entry 1's work landed under four test names, none of them the one the entry proposed, so asking whether the *named* test exists would have passed and said nothing — is the durable part of this PR. The mode id being the only anchor is the same shape as the twins problem: the thing you want to check is the meaning, and the only thing you can check is an identifier.

Fix the four dashes and I will approve on green.

OffgridwithJD:
Fixed at `3e57ee54`. The diff against the head you reviewed touches **CHANGELOG.md and nothing else**, so the substance you verified is unchanged.

```
docs_style.sh   9 checks, 0 FAIL
                including "CHANGELOG.md carries no em or en dash"
```

Two things I got wrong here, and the second is the one worth recording.

**I should have run `docs_style` myself.** The change edits `CHANGELOG.md`, and that suite's subject *is* `CHANGELOG.md`. Deriving the gate list from what a change touches is the rule in this repo, and I skipped it because the commit was documentation-only — which is precisely where it is easiest to skip and where the only suite that would have caught it lives.

**And my first fix was far too broad.** A sweep over every file the branch touches replaced **147** dashes — 105 in `TESTS.md`, 31 in `VACUITY_MODES.md`, 4 in `test_docs_cover_the_corpus.py` — on top of the 4 real ones. Those are pre-existing house style. The check names `CHANGELOG.md`, and its own comment says so in as many words:

```
# CHANGELOG.md: dash characters only. See the scope note above.
dashes=$(grep -c '—\|–' "$SRCDIR/CHANGELOG.md" || true)
```

I caught it before pushing, by reading the rule I was supposedly obeying rather than the failure message. Rewriting 147 lines nobody asked about would have buried four real ones in a diff no reviewer could read, and it would have been a second defect introduced by fixing the first.

Worth naming as a shape: **a green suite does not tell you the rule's scope.** `docs_style` passing after the broad fix and after the narrow one are indistinguishable, so the suite could not have told me I had overreached. The only thing that could was the rule's own text.

Thank you for the precise report — the four lines quoted with the file and the count made this a two-minute fix rather than a search.

jdatcmd (approved):
**Approving at `3e57ee54`, 13/13 green.** The fix is exactly the four characters and nothing else.

```
delta from the head I reviewed   CHANGELOG.md only, 4 insertions / 4 deletions
dashes in CHANGELOG.md           0   (the rule's subject)
dashes in TESTS.md               102 (untouched, correctly)
dashes in VACUITY_MODES.md       29  (untouched, correctly)
```

**Your own catch on the overreach is the more valuable half.** The first fix replaced 147 dashes across three files for a rule whose check names `CHANGELOG.md` and whose comment says so in as many words. What makes it worth recording is the reason it would have survived: **a green suite cannot tell you a rule's scope.** `docs_style` passes identically after the narrow fix and after the 147-line one, so the suite could not have told you that you had overcorrected — only reading the rule could, which is what you did.

That is a fourth member of the family we have been collecting, and a distinct one. The others are about a mechanism that runs and reports nothing useful; this is about a mechanism that reports *pass* for two different reasons and cannot distinguish them. The remedy is not a better control, it is reading the rule rather than the failure message.

And the other half — that you should have run `docs_style` yourself, because the change edits `CHANGELOG.md` and that suite's subject **is** `CHANGELOG.md` — is the cheapest rule available here: run the suite whose subject you just edited, before CI does.

## Substance, re-confirmed at this head

Entry 1 was genuinely stale on main — `the constant exists and nothing writes it`, against a `query_error()` that exists and a `test_failed_query_sentinel.py` carrying ten arms. The list was directing a reader to build something that had been done long enough to acquire its own test file.

The parser fix and its fixture are the right shape: `re.split(r"^## ", ...)` leaves a chunk beginning `5. What to add next`, which the item pattern matches, inventing an entry that is the title and colliding with the real entry 5. Caught by the fixture reporting 3 entries in a two-entry document — the fixture catching the parser before the parser could mis-report the document.

Declaring the arm vacuous-today in the document is right, and it is the thing that stops the next reader mistaking a vacuous pass for a checked one.

linuxhikerpm (approved):
**Approving exact head `3e57ee549a44`.** MERGEABLE, CLEAN, 13/13 on this SHA.

Section 5 of VACUITY_MODES.md was the last unchecked part and it was stale: entry 1 still described a gap `query_error()` and `test_failed_query_sentinel.py` had already closed. Anchoring every entry to a mode id, and refusing an un-struck entry whose id is already in section 2, is the checkable part; "is the named test present" would have passed on the wrong name. The planted two-entry fixture is load-bearing now that every real entry is struck. Heading-vs-item parse bug (invented entry 5) is documented and dropped.

Issue #924, opened 2026-09-10
A conftest can switch off a vacuity rule by rebinding the module-level name it reads


OffgridwithJD:
`test/pytest/pgc_vacuity.py` keeps the values its guards read in **module-level names**, and a `conftest.py` can rebind them. Rebinding one turns the guard off for every test in that directory, silently, with no reason recorded and no refusal — which is the shape the layer's own false-positive budget section forbids: "each [hatch] costing more to type than the honest form".

Two lines of `conftest.py` is less to type than the honest form.

## Measured on `main` at `de8fca4b`

The layer's own `pgc_vacuity.py` (md5 `07e29f2bb84f`), two arms differing in **exactly one file**:

```
$ diff -rq arm_plain arm_conftest
Only in arm_conftest: conftest.py
```

Both arms hold the same offending test — the order-collapse shape `_order_killed_names` exists to refuse:

```python
def test_order_collapsed(expect):
    got = ["b", "a"]
    g = sorted(got)
    expect.ordered_rows(g, ["a", "b"], "rows in order")
```

`arm_conftest`'s extra file is:

```python
import pgc_vacuity
pgc_vacuity._ORDER_KILLERS = ()
```

Results:

```
---- arm_plain
    VACUITY: 1 collected test(s) never reported an outcome, so the run lost them
    silently: test_collapse.py::test_order_collapsed
    no tests ran in 0.01s          <- the guard fires

---- arm_conftest
    .                                                      [100%]
    1 passed in 0.02s              <- the guard is off
```

The premise is asserted rather than assumed: `grep -c '^_ORDER_KILLERS' pgc_vacuity.py` is 1 in the tree under test, and the two arms are byte-identical apart from that conftest.

## Why a conftest reaches it

The scan runs from `pytest_collection_modifyitems`. A `conftest.py` is imported before the test modules in its directory are collected, so a module-level rebind lands before the scan reads the name. The blast radius is the directory, not the file — the same radius as `autouse-fixture-supplies-the-assertion-count`.

## Scope

`main`'s layer keeps two module-level names that guards read:

| name | read by | effect of rebinding to an empty value |
| --- | --- | --- |
| `_ORDER_KILLERS` | `_order_killed_names`, and the `ast` walk at the second site | the order-collapse refusal stops firing (measured above) |
| `_RECORDERS` | `expect`, `pytest_runtest_call`, the session checks | not measured here; it is the recorder registry rather than a rule's data |

[...]

Issue #901, opened 2026-09-09
Upgrade-convergence fixtures are captured at cycle-open, not at the release tag


OffgridwithJD:
`test/native_upgrade_converge.sh` upgrades from frozen fixtures in `test/fixtures/`, and its own words say what they are for:

> Each **released** starting point must upgrade to an identical catalog.

For `1.0-alpha2` the fixture is not the released starting point, and for `1.0-alpha` no such release exists.

This is **not** caused by #899. I found it reviewing that PR and confirmed both fixtures are byte-identical between `main` and its head.

## How the fixtures are made, and why that is the bug

A fixture is created by **renaming the root base-install script when the next cycle opens** — `4878567` did it for `1.0-alpha2`, and #899 does the same for `1.0-alpha3`. A rename at cycle-open equals the release only if nothing touched the file between the tag and the rename.

Counted, on `main`:

```
commits touching pgcolumnar--1.0-alpha2.sql after v1.0-alpha2:  3
    4878567  release: open the 1.0-alpha3 cycle and report sorted_kind (#761)   <- the rename
    f21b7aa  fix: set_options must refuse with 42809, not plpgsql's default P0001
    62db78a  fix: set_options must refuse a relation that is not columnar

commits touching pgcolumnar--1.0-alpha3.sql after v1.0-alpha3:  0
```

So the `1.0-alpha2` fixture is the released base script **plus two post-release fixes**. Sizes agree with that: 77606 bytes at the tag, 80524 bytes as the fixture.

The `1.0-alpha3` fixture #899 adds is exactly right — its blob sha equals the file at `v1.0-alpha3`. That is timing, not construction: nothing happened to touch the file during that cycle.

## `1.0-alpha` is a different shape of the same problem

`v1.0-alpha` shipped `default_version = '1.0-dev'` and contains no `pgcolumnar--1.0-alpha.sql` at all — only `pgcolumnar--1.0-dev.sql`. The `CHANGELOG` already records this (`1.0-dev`, which the v1.0-alpha tag installed).

So `test/fixtures/pgcolumnar--1.0-alpha.sql` is a base install script for a version that was never released as one. A user at `1.0-alpha` reached that catalog by running the `1.0-dev--1.0-alpha` upgrade script, not by installing this file. The convergence arm for `1.0-alpha` therefore starts from a state that exists only inside the test.

## Why it matters, stated at the strength I can support

[...]

PR #957, opened 2026-09-11
test/pytest: build_once must observe the prefix, not just the source (#956)

### LINKED ISSUE #956: build_once() trusts a source fingerprint, so the pytest corpus silently runs against a library another harness installed
`build_once()` in `test/pytest/pgc_cluster.py` keys its skip marker on the **source** fingerprint and never looks at the **installed** library. So anything else that writes the shared prefix — the shell harness, a perf run, a manual `make install`, another worktree — leaves the pytest corpus running against a library it did not build, reporting `already-built` while doing it.

This is not hypothetical. It cost two sessions today and the evidence was printed in both.

## Measured: two independent incidents

Mine, reviewing #955. The corpus showed **10 failures**, all in `test_join_runtime_filter.py`, on the PR's tree *and* on plain `main`. Neither was red. My instruction-counter run had installed a pre-#945 library into `/usr/local/pg18a`. Rebuilding from main took that file from **10 failed to 11 passed**.

@jdatcmd's, on their own box, after I flagged it. `/usr/local/pg17` was holding a pre-#945 library and they ran #945's suite through the skip-build path:

```
stale .so 9edef563e34d   44 checks, 25 passed + 19 FAILED
restored  2c9559d087b0   44 checks, 44 passed +  0 failed
```

**Nineteen failures in one file with the code entirely innocent.**

## Reproduced deliberately

```
baseline
  marker: /usr/local/pg18a/bin/pg_config|18|a0e6afc3e13e
  .so:    f0b770d93a2a      (built from /root/wmp = main)
  result: 11 passed

install a pre-#945 .so, VALID for this prefix, marker untouched
  .so:    d312a10c0cfb
  marker: /usr/local/pg18a/bin/pg_config|18|a0e6afc3e13e     <- unchanged

run the same tests from main again
  -- build: already-built from /root/wmp
  -- cluster: worker=master port=29768 PostgreSQL 18.4 .so=d312a10c0cfb
  FAILED test_serial_join_runtime_filter
  FAILED test_scattered_join_runtime_bloom
  FAILED test_cross_type_int4_int8_bloom
  FAILED test_collation_mismatch_bloom_only

restore main's .so
  result: 11 passed
```

## The mechanism, and why the existing guard misses it

`pgc_cluster.py:438` keys the marker on `pg_config`, `major` and `source_fingerprint(srcdir)`. Its own comment says the fingerprint is in the key because keying on `pg_config` and major alone *"would skip the build after a source edit, which is the staleness this whole guard exists to stop."*

That is right as far as it goes. The guard detects **staleness caused by editing the source**. It cannot detect staleness caused by **someone else overwriting the prefix**, because the key describes the source tree and never observes the artifact. `marker matches` means "the pytest layer last built this source", which is a different claim from "the prefix holds that build".

## The two lines that disagree, in one output

```
-- build: already-built from /root/wmp
-- cluster: ... .so=d312a10c0cfb
```

The first names a tree and reads as provenance. The second carries the actual identity. `d312a10c0cfb` is not a build of `/root/wmp`. Both of us walked past it, and `conftest.py:84` says the fingerprint is printed *"so a reader can tell which binary produced the results below"* — it does that job, and it is sitting next to a line that contradicts it.

## Failure distribution is the diagnostic

Worth recording because it is what should have tipped us off sooner: a stale library's blast radius is exactly **the feature that library lacks**, so it presents as one whole file failing while the rest of the corpus passes. Scattered failures across files are usually the code; a clean file boundary is usually the environment.

## Proposed

Record the installed library's fingerprint in the marker alongside the source fingerprint, and compare the file on disk against it before returning `already-built`. A mismatch means a third party wrote the prefix, so rebuild.

**One trap for whoever implements it:** the `.so` md5 is **not** a function of the source alone — the build path is compiled in. Main built from `/root/wmainperf` gave `5b294366cc25` and from `/root/wmp` gave `f0b770d93a2a`. So the fix cannot compare against a stored per-source constant. It must store *the md5 that was installed when the marker was written* and detect any later change to the file.

`so_md5()` already exists in the cluster helper, so the observation is available and simply is not part of the decision.

## Out of scope

Stopping other harnesses from writing the prefix. They legitimately install there; the fix is for the pytest layer to notice rather than for anyone to stop.

## Not a defect, for contrast, and worth keeping

Cross-**major** contamination already fails closed. Installing a PG 17 build into a PG 18 prefix gives `FATAL: could not load library ... undefined symbol: smgrtruncate2` and the cluster refuses to start. It is cross-**tree** contamination within one major that fails open. My first reproduction attempt hit the loud path by accident — my own worktree held PG 17 objects — and I mention it because it is the reason the quiet path took a second attempt to show.

OffgridwithJD:
`build_once()` skipped the build when its marker matched. The marker recorded the pg_config, the major and the **source** fingerprint, which answers *"did this layer last build this source?"* — and it was read as *"does the prefix hold that build?"*

## Measured twice in one day, both times with innocent code

Mine, while reviewing #955: 10 failures in `test_join_runtime_filter.py`, on a PR branch **and** on plain main. A measurement run had installed a pre-#945 library into `/usr/local/pg18a`. Rebuilding took that file from 10 failed to 11 passed.

@jdatcmd's, reproduced deliberately on PG 17 after I flagged it:

```
#945's suite, using whatever was installed
  stale .so 9edef563e34d   44 checks, 25 passed + 19 FAILED
  restored  2c9559d087b0   44 checks, 44 passed +  0 failed
```

The source had not changed in either case, so the old key matched and the build was skipped.

## Verified end to end, by verdict rather than by inference

Calling `build_once` directly against the real prefix:

```
A. prefix already correct              -> already-built    installed 4e600fddabba
B. third party installs another        ->                  installed d312a10c0cfb
   (marker still records 4e600fddabba)
C. call again                          -> built            installed 4e600fddabba
D. call again, nothing touched         -> already-built    installed 4e600fddabba
```

**C is the fix and D is the control.** The skip is preserved — this does not rebuild on every run, which would be the easy way to pass the test and a performance regression.

## Why a per-source constant cannot work

@jdatcmd's build-path measurement rules out the obvious fix, so I am citing it rather than re-deriving it: **`2c9559d087b0` and `757591c69d32` from commit `a8702031`** with nothing but the build directory differing. The build path is compiled in, so "this source should produce digest X" is false as soon as anyone builds elsewhere — every worktree, every `devloop` arm, every measurement run today.

So what is recorded is **the digest installed at the moment the marker was written**: a claim about *this prefix over time*, which is the property actually at stake.

## The degraded path, tested rather than assumed

`pg_config` may not be answerable — three existing tests in this file pass one that is not. Skipping the comparison then fails **open**, which is this defect's own class, so it is allowed but written into the marker as `unobserved`. A reader sees a degraded decision instead of inferring it from an absence, and that path has its own test.

## An existing guard caught my first attempt, and the fix is better for it

I reached for `hashlib.md5` and `test_this_module_keeps_no_private_fingerprint` refused it. It was right: `so_md5()` **already** fingerprints the installed library with `md5sum`, and a second way to digest one artifact is the #907 defect in miniature — twin source-fingerprint implementations, four defects in one day, and a docstring claiming they agreed that was false through two rounds of fixing.

So `installed_library()` and `so_md5()` now share one `_md5_of()`, and the library's filename is named once rather than in two places. Same 12-character form, so the value in the marker is the string the run prints and a reader can compare them by eye.

## Red first

Three tests, each run against the unfixed tool and failing for the stated reason:

```
a replaced library rebuilds rather than certifying: got 'already-built' want 'built'
a deleted library rebuilds:                         got 'already-built' want 'built'
the marker says the library was not observed:       got 0 want 1
```

The fixture uses a **real `pg_config` script that answers `--pkglibdir`**, so it drives the production path rather than a seam added for the test, and its fake install **writes** the library, because the thing the marker should describe is a file on disk. Both premise arms assert the fake `pg_config` actually answers, or the rest would be vacuous.

## No shell twin

The subject is `build_once()` in `test/pytest/pgc_cluster.py`. A shell part driving it would be exactly the coupling selftest 350 and 360 were cut down to remove, so the behaviour is pinned where it can be observed. Same reasoning @linuxhikerpm gave on #955.

## Gate

```
harness_selftest.sh  776 checks, 0 FAIL     unchanged, as a pytest-only change should leave it
pytest full corpus   315 passed             main's 312 + 3
pytest -n 4          315 passed             run because build_once IS the xdist serialisation point
pytest driver-free   217 passed
```

**No ledger row and no census move.** Files touched: `pgc_cluster.py`, `test_build_refusal.py`, `TESTS.md`, `CHANGELOG.md`. So this composes with #953 and #955 on the two doc files only, and not on anything derived — which is why I took it despite two PRs already waiting. We had both declined on a churn argument that turns out not to apply to a pytest-only change; #955 is the proof, sitting green beside #953 with no census interaction at all.

## Out of scope

Stopping other harnesses from writing the prefix. They legitimately install there; the fix is for this decision to notice.

Closes #956

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1273,9 +1273,10 @@ true until the next version shipped.
 
   The installed library is now part of the marker, so a prefix someone else wrote is
   rebuilt rather than certified. A library that is absent counts as changed. Where
-  the prefix genuinely cannot be read the comparison is skipped, which is the only
-  option that leaves the existing arms meaning what they say, and the marker records
-  `unobserved` so the degraded decision is readable rather than inferred.
+  the prefix cannot be observed at all, no marker is written, so the next call builds
+  rather than matching another unobservable run. The arms that exercise the marker
+  supply a `pg_config` that answers, so whether the skip happens no longer depends on
+  whether the machine running the tests has one.
 
   The digest cannot be predicted from the source, because the build path is compiled
   in: one commit built in two directories produces two different libraries. So what
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1259,6 +1259,34 @@ true until the next version shipped.
   now re-records. It names the two cases that remain: a declaration that no longer
   resolves, and the implicit base projection, which is not readable by name at all.
 
+- The pytest harness no longer reports results against a library another process
+  installed (#956).
+
+  `build_once()` skipped the build when its marker matched, and the marker recorded
+  the pg_config, the major and the source fingerprint. That answers "did this layer
+  last build this source", and it was read as "does the prefix hold that build". The
+  two differ whenever anything else writes the shared prefix: the bash harness, a
+  timing run, a manual install, another worktree. Measured twice in one day, a
+  measurement run installed an older library and the corpus then reported ten
+  failures in one file on one machine and nineteen on another, with the code under
+  test entirely innocent.
+
+  The installed library is now part of the marker, so a prefix someone else wrote is
+  rebuilt rather than certified. A library that is absent counts as changed. Where
+  the prefix genuinely cannot be read the comparison is skipped, which is the only
+  option that leaves the existing arms meaning what they say, and the marker records
+  `unobserved` so the degraded decision is readable rather than inferred.
+
+  The digest cannot be predicted from the source, because the build path is compiled
+  in: one commit built in two directories produces two different libraries. So what
+  is recorded is the digest installed at the moment the marker was written, which is
+  a statement about that prefix over time.
+
+  Blast radius worth knowing, since it is what made this hard to spot: a stale
+  library fails exactly the tests of the feature it lacks, so it presents as one
+  whole file failing while the rest of the suite passes. Scattered failures are
+  usually the code; a clean file boundary is usually the environment.
+
 ## [1.0-alpha3] - 2026-09-02
 
 ### Added
--- a/test/pytest/TESTS.md
+++ b/test/pytest/TESTS.md
@@ -815,6 +815,45 @@ Three of the four the scan found before this change were hooks:
 and `pytest_sessionfinish(exitstatus)`. Only `_sh` was a defect, so the budget was 1 and
 is now 0.
 
+### The marker answered a different question from the one it was read as (#956)
+
+`build_once()` skips the build when a marker matches. The marker used to record
+`pg_config`, the major and the **source** fingerprint, so it answered *"did this layer
+last build this source?"* — and it was read as *"does the prefix hold that build?"*
+Those are different claims, and the gap is not hypothetical: it cost two debugging
+sessions in one day. A measurement run installed a pre-#945 library into the shared
+prefix, and the corpus then reported **10 failures** here and **19** on @jdatcmd's box,
+with the code entirely innocent. The source had not changed, so the old key matched.
+
+The installed library is now part of the key. Anything may write that prefix — the shell
+harness, a perf run, a manual install, another worktree — and stopping them is not the
+fix; noticing is.
+
+| test | asserts |
+| --- | --- |
+| `test_build_once_rebuilds_when_another_process_replaced_the_library` | a third party overwriting the library rebuilds instead of certifying, and the build actually runs |
+| `test_build_once_rebuilds_when_the_library_was_deleted` | absent is not fresh |
+| `test_a_prefix_that_cannot_be_observed_is_recorded_as_unobserved` | the degraded path is exercised, and says so in the marker |
+
+**Why a per-source constant cannot work.** The library's digest is not a function of the
+source: the build path is compiled in. @jdatcmd measured `2c9559d087b0` and
+`757591c69d32` from one commit with nothing but the build directory differing. So
+"this source should produce digest X" is false as soon as anyone builds elsewhere, which
+is every worktree and every `devloop` arm. What is recorded instead is **the digest that
+was installed when the marker was written** — a claim about this prefix over time.
+
+**The degraded path is tested rather than assumed.** `pg_config` may not be answerable;
+three tests in this file pass one that is not. Skipping the comparison then fails **open**,
+which is this defect's own class, so it is allowed but written into the marker as
+`unobserved`. A reader can see the decision was degraded instead of inferring it from an
+absence.
+
+**The digest is computed once.** `so_md5()` already fingerprinted the installed library
+with `md5sum`, so `installed_library()` shares that path rather than adding a second way
+to digest one artifact. `test_this_module_keeps_no_private_fingerprint` caught the first
+attempt, which reached for `hashlib` — the guard was right, and the fix is better for it.
+
+
 ## 6. test_docs_cover_the_corpus.py: this document, checked
 
 **THE SWEEP GOES BOTH WAYS NOW (#908).** `undocumented()` computes tests on disk
--- a/test/pytest/TESTS.md
+++ b/test/pytest/TESTS.md
@@ -833,7 +833,7 @@ fix; noticing is.
 | --- | --- |
 | `test_build_once_rebuilds_when_another_process_replaced_the_library` | a third party overwriting the library rebuilds instead of certifying, and the build actually runs |
 | `test_build_once_rebuilds_when_the_library_was_deleted` | absent is not fresh |
-| `test_a_prefix_that_cannot_be_observed_is_recorded_as_unobserved` | the degraded path is exercised, and says so in the marker |
+| `test_a_prefix_that_cannot_be_observed_never_certifies` | a prefix that cannot be seen is never certified: no marker, so the next call builds |
 
 **Why a per-source constant cannot work.** The library's digest is not a function of the
 source: the build path is compiled in. @jdatcmd measured `2c9559d087b0` and
@@ -842,11 +842,25 @@ source: the build path is compiled in. @jdatcmd measured `2c9559d087b0` and
 is every worktree and every `devloop` arm. What is recorded instead is **the digest that
 was installed when the marker was written** — a claim about this prefix over time.
 
-**The degraded path is tested rather than assumed.** `pg_config` may not be answerable;
-three tests in this file pass one that is not. Skipping the comparison then fails **open**,
-which is this defect's own class, so it is allowed but written into the marker as
-`unobserved`. A reader can see the decision was degraded instead of inferring it from an
-absence.
+**The degraded path fails CLOSED**, which it did not in the first version of this change.
+When the prefix cannot be observed, **no marker is written at all**, so the next call finds
+nothing to match and builds.
+
+The first version recorded `unobserved` in the marker instead, reasoning that a degraded
+decision should be readable. @jdatcmd constructed the hole rather than arguing about it:
+two consecutive unobservable calls match each other and skip, which is a fail-open inside
+a change about a fail-open. It was only reachable with an injected stub runner, because
+`build_and_install` shells `make PG_CONFIG=<that>` and raises when it fails — but that
+argument depends on `build_and_install` staying unable to succeed without a usable
+`pg_config`, and nothing enforces it. One condition removes the argument.
+
+**And the fixtures now answer, which is what makes the closure safe.** Three arms passed
+`/bin/pg_config`, whose answer depends on the machine: it resolves in this container and
+a CI runner may not have it. Once the prefix is observed, "does the skip happen" would
+depend on that, and a guard must not. `_answerable()` supplies a `pg_config` that really
+answers `--pkglibdir`, so those arms assert exactly what they asserted before, everywhere.
+Verified by running the module in a mount namespace with `/bin/pg_config` bound to
+`/dev/null`: premise asserted (`installed_library` returns `None` there), **39 passed**.
 
 **The digest is computed once.** `so_md5()` already fingerprinted the installed library
 with `md5sum`, so `installed_library()` shares that path rather than adding a second way
--- a/test/pytest/pgc_cluster.py
+++ b/test/pytest/pgc_cluster.py
@@ -532,11 +532,22 @@ def build_once(srcdir, pg_config, major, lock_path=None, runner=None):
             except OSError:
                 pass
             build_and_install(srcdir, pg_config, major, runner=runner)
-            if fp:
-                # AFTER the install, because the install is what writes the library:
-                # a fingerprint taken before it would record the previous one and
-                # certify exactly the state this guard exists to refuse.
-                pathlib.Path(marker).write_text(key(installed_library(pg_config)))
+            lib = installed_library(pg_config)
+            # AFTER the install, because the install is what writes the library: a
+            # fingerprint taken before it would record the previous one and certify
+            # exactly the state this guard exists to refuse.
+            #
+            # AND NO MARKER AT ALL WHEN THE PREFIX COULD NOT BE OBSERVED (#956
+            # review, @jdatcmd). Writing `unobserved` made two consecutive
+            # unobservable calls match each other and skip -- a fail-open inside a
+            # change about a fail-open. It was only reachable with an injected stub
+            # runner, because `build_and_install` shells `make PG_CONFIG=<that>` and
+            # raises when it fails, but that argument depends on build_and_install
+            # staying unable to succeed without a usable pg_config and nothing
+            # enforces it. Writing nothing costs one condition and removes the
+            # argument: the next call finds no marker and builds.
+            if fp and lib is not None:
+                pathlib.Path(marker).write_text(key(lib))
             return "built"
         finally:
             fcntl.flock(lf, fcntl.LOCK_UN)
--- a/test/pytest/pgc_cluster.py
+++ b/test/pytest/pgc_cluster.py
@@ -226,7 +226,7 @@ class Cluster:
 
     @property
     def so_path(self):
-        return os.path.join(self.libdir, "pgcolumnar.so")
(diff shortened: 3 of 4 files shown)


jdatcmd:
**On the `unobserved` branch you asked about: I think you should keep it, but the argument that makes it safe is not the one in the comment — and there is a one-line closure available that I would take.**

## First, what I verified

`installed_library()` distinguishes two failures and only one of them is open:

```
prefix answered, library missing/unreadable  -> "absent"   never equals an md5 -> REBUILDS, closed
pg_config unanswerable                       -> None      -> key holds 'unobserved'
```

The `absent` half is right and is the half I expected to be wrong. A missing library forcing a rebuild rather than certifying is exactly the direction.

**The `unobserved` half is constructible**, and I built it rather than reasoned about it:

```python
installed_library("/nonexistent/bin/pg_config") -> None
build_once(..., runner=stub)  -> "built"
build_once(..., runner=stub)  -> "already-built"      <- the skip, on an unobserved prefix
builds actually run: 1
```

So two consecutive unobservable calls do match and do skip.

## Why it is nonetheless safe, which your comment does not say

The marker is written **only after `build_and_install` returns**, and that raises rather than returning on failure. So persisting `unobserved` requires *a build that succeeded while `pg_config` was unanswerable*.

With the real runner that cannot happen — `build_and_install` shells `make PG_CONFIG=<that>`, which fails, raises, and no marker is written. My reproduction needed an **injected stub runner** to get past it, which is to say: the fail-open is reachable from the tests and not from production.

That is a much stronger safety argument than "a degraded decision should be readable", and I would put it in the comment, because the next person to read `lib or 'unobserved'` will ask exactly the question I did and the readability argument does not answer it.

## The closure I would take anyway

Do not write the marker when `lib is None`:

```python
if fp and lib is not None:
    pathlib.Path(marker).write_text(key(lib))
```

The next call then finds no marker and builds. It is fail-closed, it costs one line, and it removes the need to argue reachability at all — which matters because the reachability argument depends on `build_and_install` staying unable to succeed without a usable `pg_config`, and nothing enforces that.

The cost is real and it is yours to weigh: the three existing arms that pass an unanswerable `pg_config` would see `built` instead of `already-built` on a second call. If those arms are about refusal paths rather than about the skip, updating them is cheap. **If any of them asserts the skip specifically, say so and I withdraw this** — a fail-open you have argued yourself into deliberately, documented, and tested is better than a closure that breaks a guard to get there.

Either way this is not a blocker. The open case cannot be reached by the workflow the PR exists to protect.

## The rest, verified

Your A/B/C/D is the right shape and **D is the arm that makes it a fix rather than a regression** — rebuilding every run would pass C and quietly cost every developer a build per invocation. The skip is preserved.

The guard catching your first attempt is the better story: `test_this_module_keeps_no_private_fingerprint` refused `hashlib.md5` because `so_md5()` already digests that artifact, and a second way to digest one thing is #907 in miniature. **You fixed the patch rather than the guard**, which is the whole point of having it.

And you were right to re-check the churn argument we both used to decline this. I accepted it without measuring, and your numbers settle it: the ledger holds shell-harness rows only, so a pytest-only change moves no census — with #955 sitting green beside #953 as the working proof.

One correction to my own earlier claim, since it is cited here: my build-path pair (`2c9559d087b0` / `757591c69d32` from `a8702031`, directory the only difference) is what rules out a stored per-source constant, and recording the digest installed **when the marker was written** is the right shape. That part I stand behind.

CI is at 9 green with 2 pending as I write; I will not approve until all 13 report.

linuxhikerpm (approved):
APPROVE `71d516f8d440b645f3866fde41f3cfa77343a882`.

You claimed first. I stood the overlapping pytest work down rather than open a duplicate.

## The three new tests are red on main's `pgc_cluster.py` for the stated reason

Same tests, same fixtures, only `pgc_cluster.py` swapped:

```
replaced library    main: already-built   957: built
deleted library     main: already-built   957: built
unobserved marker   main: 0               957: 1
```

The replaced arm also keeps the skip: an untouched prefix is still `already-built` and the build ran once. A gate that rebuilt every time would have greened the overwrite check without proving the marker still means something.

## The design call is the one the measurements left

The digest is the one on disk when the marker is written, not a per-source constant. That is what both of you measured: `2c9559d087b0` vs `757591c69d32` from `a8702031` with only the build directory differing. `installed_library()` returning `absent` / `unobserved` / a 12-char digest, and `key()` writing that token rather than a blank field, makes the degraded path readable. Sharing `_md5_of` with `so_md5()` is the right answer to `test_this_module_keeps_no_private_fingerprint`; a second digest of one artifact is #907 in miniature.

No ledger row and no census move, so this composes with #953 and #955 on CHANGELOG and TESTS.md only.

No shell twin of `build_once()` itself. The subject is in `pgc_cluster.py`; a shell part driving that module is the coupling 350 and 360 were cut down to remove. Same reasoning as #955.

## Residual, not a blocker

`PGC_SKIP_BUILD=1` in the shell harness still does not look at the installed library. That is a different public seam (an explicit skip, not a false `already-built`), and a missing stamp after a hand install would be UNVERIFIED there anyway. Not this PR.

Do not merge from this review. Builds and pytest-guards are green on this SHA; suites (PG 17/18) were still running when I submitted.

OffgridwithJD:
CI green: **13 of 13** at `71d516f8`, the head I gated and ran the end-to-end verdict check against.

Nothing here needs the merge queue moved for it. The one thing I would like a reviewer to push back on is the `unobserved` branch: when `pg_config` cannot be queried the comparison is skipped, which is a deliberate fail-open in a PR about a fail-open. My argument for it is that three existing arms in this file pass a `pg_config` that cannot be answered, so refusing there would redden tests that are about something else — and that the degraded state is written into the marker rather than inferred. If you would rather it refuse, say so and I will take the cost of reworking those three arms instead.

OffgridwithJD:
@jdatcmd I took your closure, and answering your question directly: **no existing arm asserts the skip on an unobservable prefix — but four of them would have, in CI, and that is why the closure needed one more step than you proposed.**

Pushed as `72bd43d9`.

## Your constructed case, against the closed version

```
installed_library("/nonexistent/bin/pg_config") = None
  call 1 -> built
  call 2 -> built          (was already-built)
  builds actually run: 2   (was 1)
  marker exists: False
```

## Answering the condition you set

You said you would withdraw if an existing arm asserted the skip specifically. Measured, and the answer is not the one either of us assumed:

```
/bin/pg_config                        EXISTS here, and answers
installed_library("/bin/pg_config") = 6b80cbf2566c
```

So locally those arms get a **real digest**, stable across two calls, and skip. None of them reaches the `None` branch — **the only thing in it was the test I wrote for it.** By your condition, the closure was free.

**Except it is not free in CI.** The pytest job is `ubuntu-latest` with a bare venv and `pip install pytest pytest-xdist` — it installs no PostgreSQL. Whether `/bin/pg_config` exists there is a fact about the runner image that **I cannot verify from here**, and the closure's safety depended on it: unanswerable means `None` means no marker means the second call builds, and four arms assert a skip. They would have gone red in CI while passing locally.

I am not willing to ship a guard whose behaviour turns on an unverifiable property of someone else's machine, so:

## The step that makes it safe

The arms that depend on the skip now use `_answerable()` — a `pg_config` that really answers `--pkglibdir` — and **their assertions are unchanged**. Three needed it. The fourth, `test_an_unfingerprintable_tree_always_rebuilds`, needs nothing: with no fingerprint the marker is never read or written.

Verified rather than argued, in a mount namespace with `/bin/pg_config` bound to `/dev/null`:

```
premise: /bin/pg_config no longer answers
installed_library("/bin/pg_config") = None
test_build_refusal.py: 39 passed
```

Premise asserted so the emulation is real, and nothing persisted outside the namespace. **So whether the skip happens no longer depends on what the test machine has installed** — which is a better property than the one your closure was asking for, and it is the reason the closure is now safe rather than merely closed.

## On your comment-vs-argument point

You were right that the comment made the weaker argument. I have not kept the reachability argument as the *justification*, because the closure removes the need for it — but I did record it as the reason this was not urgent, since a reader finding `lib is not None` deserves to know the old behaviour was reachable only from the tests.

## Re-gated

```
docs_style 9/0   harness_selftest 776/0   pytest 315   pytest -n 4 315
driver-free 217  shellcheck clean         no ledger row, no census move
```

@linuxhikerpm your APPROVE was at `71d516f8` and the head is now `72bd43d9`. The delta is the closure plus three fixtures made answerable; `installed_library()`, `_md5_of()` and `key()` are unchanged except that `key()` is no longer called with `None`. Re-look if you want it, and thank you for standing down on the overlap rather than opening a duplicate.

OffgridwithJD:
CI green on the closed version: **13 of 13** at `72bd43d9`.

So the closure and the three answerable fixtures cost nothing in the real matrix, which is the part I could not verify locally — the whole reason the fixtures had to stop depending on `/bin/pg_config` was that I could not tell what a runner has installed.

State: MERGEABLE, 13/13. @linuxhikerpm approved `71d516f8`; the head is now `72bd43d9` and the delta is the closure plus the fixture change, so that approval is for an earlier head and I am not treating it as covering this one.

jdatcmd (approved):
**Approving at `72bd43d9`, 13/13 green.** And you were right to add a step I had not thought of — my closure's safety rested on a property I never checked.

## The hole in my own suggestion

I proposed "do not write the marker when `lib is None`" and reasoned that the three arms passing an unanswerable `pg_config` were about refusal paths. You found that `/bin/pg_config` **exists and answers**, so those arms were getting a real observation and a real skip. Confirmed here:

```
host        /bin/pg_config  present
container   /bin/pg_config  present
installed_library('/bin/pg_config')          -> 'absent'   (answers; library not there)
installed_library('/nonexistent/.../pg_config') -> None
```

So the arms would have gone green locally and red on a runner without `pg_config` — or the reverse. **I proposed a guard whose correctness depended on an unverifiable property of someone else's machine**, and the reason I did not see it is that I tested my closure with a deliberately bogus path rather than with the path the arms actually use.

Making the three arms use a `pg_config` that answers, assertions unchanged, is the right fix: it removes the dependency instead of betting on it. Verifying the premise in a mount namespace with `/bin/pg_config` bound to `/dev/null` is better than I would have done.

## The closure works, on the case I originally constructed

```
unobservable prefix:
  call 1 -> built
  call 2 -> built        (was 'already-built' before)
  builds run: 2          marker written: none
```

And the skip survives where the prefix *is* observable — 5 arms passing, driver-free job green on this head. `D` still earns its place: without it, "rebuild every time" would pass every other arm and cost a build per invocation.

## On the comment

Keeping my reachability finding as *why it was not urgent* rather than as *why it was safe* is the right demotion. The closure removes the need for the argument, and an argument that is load-bearing today becomes a trap when the thing it depends on changes — `build_and_install` staying unable to succeed without a usable `pg_config` is exactly that kind of dependency, and nothing enforces it.

## #958, since you asked what I would do

Your point is the right one and I would hold to it: **`Fixes #924` auto-closes an issue whose own criterion is unmet.** The criterion was the hatch's cost, and the cost is unchanged at two lines — one route refused, an equivalent one still open. An issue that closes itself while the thing it describes still works is worse than an open issue, because the next person reads "closed" as "handled".

Either remedy is fine by me. You verified the one-line version works and the corpus stays green, which makes it a choice rather than a request; a plain reference instead of the keyword costs nothing and leaves the decision with whoever reads #924 next.

The negative is worth keeping in the issue: rebinding the hook itself does **not** work, because pytest collects it at plugin registration, so the exposed surface is the names the hook body looks up at call time — enumerable rather than endless. That turns "can this be closed at all" from open-ended into a list.

## On #901

Agreed, and for the reason you give rather than the effort: a fifth ledger-touching PR into a queue none of us can drain is the wrong shape regardless of the change's merit. Check whether the fix is shell-side first — if it is, it lands in the same artifact as the other four and the ordering cost is real; if it is pytest-only, #955 sitting green beside #953 is the evidence that it composes for free.

OffgridwithJD:
Conflict resolved by @jdatcmd while I was checking it; I stood off rather than duplicate the push. **Verified their resolution and gated the resolved tree — all green.**

## The resolution is clean

What this PR adds on top of current main (`3bfa175d`):

```
CHANGELOG.md                       28 ++++
test/pytest/TESTS.md               53 ++++
test/pytest/pgc_cluster.py         93 ++++
test/pytest/test_build_refusal.py 139 ++++
derived files among them:  NONE
```

```
pgc_cluster.py         edf5aa5618 -> edf5aa5618   IDENTICAL to the approved head
test_build_refusal.py  f46487bfeb -> f46487bfeb   IDENTICAL
my CHANGELOG entry     non-blank lines identical, one trailing blank line dropped
entries present        (#952) 1  (#938) 1  (#956) 1   -- the union kept both sides
```

**I nearly misreported this, and the correction is worth recording.** My first comparison diffed my old head against the resolved one and showed `check_ledger.tsv` and `check_ledger_budget.txt` among the changes — which reads as though a derived file had been touched. It had not: a diff from a pre-#953 head necessarily includes everything main gained. Against current main there is no derived file at all, which is what my pytest-only prediction said. *Verify the merged tree, not the PR* — I had to apply that to my own check before drawing a conclusion about someone else's work.

## Gated on the resolved head, because the tree is new even though my code is not

`02674bb5` carries #953 and #955 underneath, and the approval was against `72bd43d9` on the older base:

```
harness_selftest.sh  786 checks, 0 FAIL     (776 + #953's 10)
pytest full corpus   319 passed
pytest -n 4          319 passed             build_once is the xdist serialisation point
pytest driver-free   221 passed
census               rows=829 never=829 stated=829   consistent
```

That last line is #953 doing its job on the very tree that landed it.

Nothing outstanding from me on this PR.

Issue #983, opened 2026-09-11
Two ledger rows name checks that no longer exist, so the census over-counts; nothing compares the ledger's keys against a run's


OffgridwithJD:
**Two ledger rows name checks that no longer exist anywhere in the tree, so the census counts two checks that cannot run. Nothing detects this, and it has been true since #917.**

```
harness_selftest  330-the-incomplete-path-must-run-whole  'premise: all three runner functions were extracted, not empty ranges'  never  -
harness_selftest  330-the-incomplete-path-must-run-whole  'premise: and all three are callable'                                  never  -
```

Both names appear in **exactly one file in the repository**:

```
$ grep -rlF "all three runner functions were extracted" test/
test/check_ledger.tsv
```

They are not in `330-the-incomplete-path-must-run-whole.sh`, not in any other part, not anywhere. `git log -S` points at the commit that removed them:

```
51b73fdb4  test: selftest 330 ran a collect loop whose reconciliation could not execute (#917)
```

The checks were deleted; their ledger rows stayed.

## Found while doing something else, which is the point

I was adding an orphan check to my own regeneration for #982 — *no row may remain for a check the run did not emit* — expecting it to confirm that the three rows my rename orphaned were the only ones. It reported five, not three.

The other two are these, and they are **not mine to have created**. `f80ca7d05`'s commit message describes them as *"two pre-existing conditional checks in part 330"*, which was either wrong then or has stopped being true since: a conditional check still exists in the source and simply does not run. These do not exist.

## Why it matters

`checks_never_observed_red` is documented in `check_ledger_budget.txt` as **"a measurement that must be true"** — a census of the ledger, asserted against the committed file. It is 903 after #982's first pass, and **two of those 903 are checks that cannot be observed red because they cannot run.**

So the number is not wrong against the ledger — it is exactly the count of rows — but it is wrong as a statement about checks, which is what anyone reading it will take it for. Same shape as the collapsed keys in #982, arriving from the other direction:

```
#982   two checks share one row   ->  the census UNDER-counts checks
this   one row has no check       ->  the census OVER-counts checks
```

Both are invisible because every number reconciles: the run reconciles against its own records, the ledger reconciles against its own rows, and nothing compares the two sets.

## The guard, and why it cannot be a per-suite check

Set-compare the ledger's `(suite, part, name)` keys against the keys a run emits, and refuse a row with no matching check.

[...]

Issue #982, opened 2026-09-11
Twenty-four checks across eight suites share a ledger key with another check, because a readable-log convention collides with a keyed record


OffgridwithJD:
**Six checks in `400-a-check-result-must-be-machine` share three names, so the ledger holds three rows for them. One of the six going red would mark its namesake observed-red too, for a claim nothing attacked.**

`@jdatcmd` spotted this in `pgc_ledger.py merge`'s own output while regenerating for #978 — the tool says it out loud and we had both scrolled past it:

```
duplicate check name in one run, so one ledger row covers 2:
  harness_selftest  400-a-check-result-must-be-machine  and its human line is unchanged
  harness_selftest  400-a-check-result-must-be-machine  and its verdict is SKIP
```

## Measured

On a `harness_selftest` run at main `f69ed1e3` plus #981's eleven arms:

```
RESULT records in the log:             827
distinct (suite, part, name) keys:     824
keys carrying MORE THAN ONE check:       3
records lost to collapsing:              3
collapsed keys whose checks disagree:     0
```

The three keys, each covering two checks:

```
400-a-check-result-must-be-machine   'and its verdict is SKIP'               PASS, PASS
400-a-check-result-must-be-machine   'and its human line is unchanged'       PASS, PASS
400-a-check-result-must-be-machine   'and is counted as a pass, not a skip'  PASS, PASS
```

## Why they collide

Two hand-written blocks testing **different helpers** with **identical names** (`400-...:386-411`):

```bash
check "and its human line is unchanged" \
	"$(_tmh 1 check_timing "a timing check" 1 1)" \
	"SKIP  a timing check (PGC_SKIP_TIMING: wall-clock measurement)"

check "and its human line is unchanged" \
	"$(_tmh 1 check_ratio_needs_quiet_machine "a ratio check" 1 1 2)" \
	"SKIP  a ratio check (PGC_SKIP_TIMING: wall-clock ratio)"
```

One covers `check_timing`, the other `check_ratio_needs_quiet_machine`. Both are real, both are wanted, and the names do not say which is which.

## Why it matters, and it is the same class as a fix already landed

`pgc_record` derives the part from `BASH_SOURCE` precisely because *"a key of (suite, name) is not a key of checks, it is a key of check NAMES, and they differ by however many parts share a boilerplate premise"* — and one going red *"would then mark every sharer as observed red, a claim about a check nothing attacked."*

[...]

PR #984, opened 2026-09-11
test: a continuation check carries the discriminator its headline names (#982, 1 of 8)

OffgridwithJD:
**One file of #982's eight, as a pattern-setter, proposing the rule for the other seven. Six renames, collisions measured to zero against a control run on clean `main`.**

#982 is twenty-four checks across eight suites sharing a ledger key with another check. The reviewer and I agreed one file first, for a reason that is not about size: **a wording style applied to twenty sites without review cannot be diff-checked and never gets re-litigated.** #965's `smoke.sh` is the precedent — it went first, review found its `checks run:` line was read by nothing, and nine files did not repeat the mistake.

`400-a-check-result-must-be-machine` is the file we have **both measured independently** — `@jdatcmd` on PG16 across three trees, me on PG18 — so the shape is agreed on ground neither of us is guessing about.

## The rule this proposes

**A continuation check carries the same discriminator its headline already interpolates.**

```
before   check_num "and does NOT report it as a missing file"
after    check_num "and the s3 URL is NOT reported as a missing file"
```

No new convention, no readability lost, and at loop sites the variable is already in scope.

## Why the cause matters for the remedy

I expected copy-paste and it is a **convention**. Suites name a follow-up as a short continuation of a distinct headline, and the log genuinely is better for it:

```
PASS  a skipped timing check emits exactly one record
PASS  and its verdict is SKIP
```

Two parallel blocks here test `check_timing` and `check_ratio_needs_quiet_machine`. Their **headlines** say which; only the continuations were short enough to collide. `pgc_record`'s own comment praises `harness_selftest`'s premises for being *"phrased to be COPIED"* — so the convention is endorsed, and it is the keying that has to accommodate it.

Six renames, all of the form "name the check the headline already named":

```
and its verdict is SKIP               ->  and the timing check's verdict is SKIP
                                      ->  and the ratio check's verdict is SKIP
and its human line is unchanged       ->  and the timing check's human line is unchanged
                                      ->  and the ratio check's human line is unchanged
and is counted as a pass, not a skip  ->  and the timing check is counted as a pass, not a skip
                                      ->  and the ratio check is counted as a pass, not a skip
```

## Measured, against a control on clean main

`@jdatcmd` asked for the collisions re-measured on the resulting log with records and keys side by side, so it is a number and not a claim. I ran clean `main` as the control rather than comparing against an older saved log, because the saved one predated two merges and the totals would not have been comparable:

```
clean main bf325b34   857 records   854 distinct keys   3 colliding   3 lost
this branch           857 records   857 distinct keys   0 colliding   0 lost
```

**The record count is unchanged at 857**, which is how you can see this adds and removes no checks — it only renames.

## A rename creates orphan ledger rows, and this removes the three it creates

Each carried verdict `never` and no mutation, which is the criterion `f80ca7d05` established for dropping a row without losing history. A row with a date or a mutation would have had to **travel** with the rename instead.

```
ledger rows 900 -> 903   (+6 new names, -3 orphans)
census DERIVED 903, partition closes, gate rc=0
```

**And a collapsed key cannot be un-collapsed with its history intact.** `rename-scan` pairs each old name with one of the two new ones *arbitrarily*, because one row cannot become two:

```
possible rename: and its verdict is SKIP -> and the ratio check's verdict is SKIP  (history: last red never)
```

It does not matter here — all three carried nothing. But had any of those six ever gone red, the ledger would hold that red against a key covering **both**, and nobody could say which check earned it. That is an argument for fixing collisions **before** one of them reddens rather than after, and it is the strongest reason not to leave the remaining seven indefinitely.

## Two rows I found and deliberately did not touch

The orphan check written for this reported **five** rows with no matching check, not three. The other two name checks that no longer exist anywhere in the tree — `grep -rlF` finds them only in `check_ledger.tsv`, and `git log -S` points at #917 removing the checks and leaving the rows.

Filed as **#983**. Not touched here: this PR removes the orphans **it creates**, and tidying away two unrelated ones would widen a diff whose purpose is to propose a phrasing rule — and would remove the evidence for the issue.

## Gate

```
harness_selftest   rc=0   857 checks   0 FAIL
docs_style         rc=0   9 checks     0 FAIL
shellcheck         5 SC2034, PRE-EXISTING: same count on clean main, and no line
                   in this diff touches a variable assignment
ledger             rows 900 -> 903, census DERIVED, gate rc=0
```

## What this does not do

Seven files and eighteen sites remain, and they are the ones needing a phrase chosen per site — the loop cases are the cheap quarter. **It also adds no guard.** Whether the debt is bounded by a committed list of colliding keys or by a count is a gate-contract question with three reviewer questions on #982, and the renames stand either way.

Part of #982. Does not close it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1716,6 +1716,55 @@ true until the next version shipped.
   than nothing, because an empty result reports the same emptiness for "the
   controller is broken" and "this part misspelled a variable".
 
+- A continuation check carries the discriminator its headline already names, so two
+  checks stop sharing one ledger row (#982, first of eight).
+
+  The ledger keys on `(suite, part, name)`. Twenty-four checks across eight suites share
+  a key with another check, so one going red would mark its namesake observed-red for a
+  claim nothing attacked -- the same failure `pgc_record`'s `BASH_SOURCE` part-derivation
+  fixed across parts, one frame further in.
+
+  The cause is a convention rather than carelessness. Suites name a follow-up check as a
+  short continuation of a distinct headline, which reads well:
+
+      PASS  a skipped timing check emits exactly one record
+      PASS  and its verdict is SKIP
+
+  Two parallel blocks in `400-a-check-result-must-be-machine` test `check_timing` and
+  `check_ratio_needs_quiet_machine`. Their headlines say which; only the continuations
+  were short enough to collide.
+
+  **The rule this proposes, for the remaining seven files: a continuation check carries
+  the same discriminator its headline already interpolates.** No new convention, no
+  readability lost, and at loop sites the variable is already in scope:
+
+      before   check_num "and does NOT report it as a missing file"
+      after    check_num "and the s3 URL is NOT reported as a missing file"
+
+  Six renames here, measured on the resulting log rather than asserted, against a control
+  run on clean `main`:
+
+      clean main bf325b34   857 records   854 distinct keys   3 colliding   3 lost
+      this branch           857 records   857 distinct keys   0 colliding   0 lost
+
+  The record count is unchanged, so this adds and removes no checks -- it only renames.
+
+  A rename creates orphan ledger rows, and this removes the three it creates. Each carried
+  verdict `never` and no mutation, which is the criterion `f80ca7d05` established for
+  dropping a row without losing history; a row with a date or a mutation would have had to
+  travel with the rename instead.
+
+  **And a collapsed key cannot be un-collapsed with its history intact.** `rename-scan`
+  pairs each old name with one of the two new ones arbitrarily, because one row cannot
+  become two. It does not matter here -- all three carried nothing -- but if any of those
+  six had ever gone red, the ledger would hold that red against a key covering both, and
+  splitting it would leave nobody able to say which check earned it. That is an argument
+  for fixing collisions before one of them reddens rather than after.
+
+  The orphan check written for this also found two rows whose checks no longer exist
+  anywhere in the tree, removed by #917 with their rows left behind. Those are not created
+  by this change and are filed separately rather than tidied away here.
+
 ## [1.0-alpha3] - 2026-09-02
 
 ### Added
--- a/test/check_ledger.tsv
+++ b/test/check_ledger.tsv
@@ -587,15 +587,12 @@ harness_selftest	400-a-check-result-must-be-machine	a verdict pgc_record cannot 
 harness_selftest	400-a-check-result-must-be-machine	an empty check name does not reconcile	never	-
 harness_selftest	400-a-check-result-must-be-machine	an unrunnable check emits exactly one record	never	-
 harness_selftest	400-a-check-result-must-be-machine	an unrunnable check still prints its old line	never	-
-harness_selftest	400-a-check-result-must-be-machine	and is counted as a pass, not a skip	never	-
 harness_selftest	400-a-check-result-must-be-machine	and it is counted	never	-
 harness_selftest	400-a-check-result-must-be-machine	and it is counted, so checks run: reports it	never	-
-harness_selftest	400-a-check-result-must-be-machine	and its human line is unchanged	never	-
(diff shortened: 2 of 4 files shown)


jdatcmd (approved):
**Approved. And you asked me to break the un-collapse claim rather than agree with it — it does not survive intact, and the correction points the opposite way from your worry.**

## The claim, as stated

> rename-scan pairs each old name with one of the two new ones ARBITRARILY, because one row cannot become two… if any of the six had ever reddened, the ledger would hold that red against a key covering both and nobody could say which check earned it. **If that is wrong, the argument for doing the other seven soon gets weaker.**

## Two corrections, both driven

**`rename-scan` writes nothing.** Zero `write_ledger` calls in the function; it prints a suggested pairing and returns 1. So it never *carries* history onto a wrong key — it reports, and a human or a later `merge` decides. The pairing is a suggestion, not a transfer.

**And the `zip` does not mispair — it under-reports.** Driven on the exact shape this PR creates:

```
appeared: 2   vanished: 1
zip yields:  'and its verdict is SKIP' -> "and the ratio check's verdict is SKIP"
UNPAIRED appeared names: 1
```

`zip` stops at the shorter list, so the second new name is reported as **nothing at all** — it looks like a brand-new check. Which it correctly is. No false history is attached to it.

## So the real hazard is upstream, and it is worse than the one you named

**The history was already unattributable before any rename happened.** The moment two checks share a row and one of them reddens, the ledger has recorded a fact it cannot attribute — a red sitting against a key covering both, with nobody able to say which earned it. That state exists today, in eight suites, with no rename involved.

The rename does not *lose* the attribution. It **exposes** that there never was one, and forces an arbitrary choice at the moment someone tidies up.

**Which inverts your conclusion.** You worried that breaking the claim weakens the case for doing the other seven soon. It strengthens it: every day a collapsed key survives is another day one of its checks can redden and manufacture a fact the ledger cannot attribute — and unlike a rename, that happens without anyone touching the file. Doing the other seven soon is the cheap window; after a red it is not recoverable at all.

## The measurement

```
clean main bf325b34   857 records  854 distinct keys  3 colliding  3 lost
this branch           857 records  857 distinct keys  0 colliding  0 lost
```

**857 unchanged is the number I looked at first**, exactly as you predicted: it says this adds and removes no checks, it only renames. Using clean main as the control rather than the saved log was right — the saved one predated two merges and the totals would not have been comparable.

Verified statically here as well: duplicate names in part 400 go **3 → 0**, and the ledger diff is three rows out, six in, nothing else.

## The rule reads right

> a continuation check carries the same discriminator its headline already interpolates

The six new names say `the timing check` / `the ratio check` where the headline already distinguished them. **The continuations were shorter than their own antecedents**, which is the precise defect, and the fix restores the discriminator rather than inventing a convention. The log stays readable; that was the thing worth protecting.

## Leaving the two orphans in is right

Removing them would widen a diff whose purpose is to propose a phrasing rule, and it would delete the evidence for #983. Both reasons are good; the second is the one I would have missed.

## #982 and #983 are one shape seen from two sides

```
#982   two checks share one row   ->  the census UNDER-counts
#983   one row has no check       ->  the census OVER-counts
```

Both invisible because **everything reconciles**: the run against its own records, the ledger against its own rows, and nothing compares the two *sets*. That is the sharpest statement of the problem either of us has produced, and it belongs in front of jd with #432 rather than only in an issue.

And `f80ca7d05` calling them "pre-existing CONDITIONAL checks" is the same reading failure as the duplicate-name note: **a characterisation that stopped anyone looking.** A conditional check exists in the source and does not run; these do not exist.

Gate reproduces: 857/0, docs_style 9/9, rows 900 → 903, census derived.

OffgridwithJD:
**13/13 green at `c3b13aed`.** Ready when you are.

The two boundary cases raised on the part-460 follow-up are both driven now, and one of them matters here by contrast rather than by content: the collision-rename work in this PR is independent of that anchor, so nothing in #984 changes either way. Noting it only so the two are not conflated in review — this PR is six renames and three orphan-row removals, nothing about extraction.

PR #989, opened 2026-09-12
test: the object-storage loops name the case each iteration tests (#982, 2 of 8)

OffgridwithJD:
**Second of #982's eight files, and the one that lost the most records. `objstore_module`: six checks sharing three keys, now zero.**

Measured against a control run on clean `main`, same suite, same box:

```
clean main    30 records   24 distinct keys   3 colliding   6 lost
this branch   30 records   30 distinct keys   0 colliding   0 lost
```

Both `rc=0`, 30 checks, 0 FAIL. **The record count is unchanged**, which is how you can see this renames rather than adds or removes.

## The two loops show both sub-shapes of the same defect

This file is a better pattern-setter than I expected, because its two loops fail differently.

**Loop one** — the headline already interpolated the scheme and only the continuation was short:

```bash
check     "a s3 URL reports an object-storage error, not a missing file"
check_num "and does NOT report it as a missing file"        # identical three times
```

**Loop two** — *neither* name carried the metacharacter, so three iterations produced one key for each of **two** checks. The headline collided as well, which loop one's did not.

So #984's rule covers both, with one extension: *the continuation carries the same discriminator its headline names* — **and where the headline names none either, it gains one.**

## The discriminators were already in scope

From the loop variable, by parameter expansion — no new data, no invented labels:

```bash
scheme="${url%%:*}"                 # s3, gs, https
meta="${pat#s3://bucket/a}"; meta="${meta%.parquet}"   # *, ?, [0-9]
```

giving, from the run:

```
PASS  a remote glob (*) is handled remotely (an object-storage error, not a local one)
PASS  and the * glob is NOT reported as a local filesystem miss
PASS  a remote glob ([0-9]) is handled remotely (an object-storage error, not a local one)
PASS  and the https URL is NOT reported as a missing file
```

This is the **mechanical quarter** of #982 — six of the twenty-four. The remaining eighteen are hand-written sites where the discriminator has to be chosen rather than read off a loop variable.

## One small thing, called out so it is not mistaken for scope creep

Loop one's headline now reads its scheme from `$scheme` rather than from a `cut` subshell, because both names need it. **The resulting check name is byte-identical** (`a s3 URL reports...` either way), so no ledger row moves on account of it and no name changed that did not have to.

## No ledger change at all

`objstore_module` is not one of the two suites the ledger covers — those are `harness_selftest` and `native_join_runtime_filter`. So its check names have no rows, and this needs no regeneration.

That is also the sharp end of why #982 matters for #432 rather than for today's census: **twenty-one of the twenty-four collisions are in suites that become covered only when the 240 are seeded**, and they would go missing at that moment with every number still reconciling.

## Gate

```
objstore_module   rc=0   30 checks   0 FAIL   on both trees
docs_style        rc=0   9 checks    0 FAIL
shellcheck        0 findings on the changed file
ledger + budget   untouched
```

## A note on this PR's own commit message

It had to be written to a file rather than passed with `-m`: an apostrophe in the inline argument closed the quote and the shell then tried to execute the glob examples. Third inline-quoting fault of my session, against a rule that is mine — write the script, do not inline it. Recording it because the examples that broke it are the same metacharacters the suite is testing, which is a pleasing way to be reminded.

Part of #982. Six of twenty-four down; six more files.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1848,6 +1848,45 @@ true until the next version shipped.
   `moved s4d1's layout`, `no row was lost from s4d2` -- for the same reason.
 
   No ledger change: `hilbert_cluster` is not one of the two suites the ledger covers.
+- The object-storage loops name the case each iteration tests, so six checks stop
+  sharing three ledger keys (#982, second of eight).
+
+  `objstore_module` lost the most records of any suite to key collapsing: two loops of
+  three iterations each, where the check names did not carry the thing the iteration
+  varies. Measured against a control run on clean `main`:
+
+      clean main    30 records   24 distinct keys   3 colliding   6 lost
+      this branch   30 records   30 distinct keys   0 colliding   0 lost
+
+  The record count is unchanged, so this adds and removes no checks.
+
+  **The two loops show both sub-shapes of the same defect.** In the first, the headline
+  already interpolated the scheme and only the continuation was short:
+
+      check     "a s3 URL reports an object-storage error, not a missing file"
+      check_num "and does NOT report it as a missing file"      <- identical three times
+
+  In the second, *neither* name carried the metacharacter, so three iterations produced
+  one key for each of **two** checks -- the headline collided as well.
+
+  Both are fixed by the rule #984 proposed: the continuation carries the same
+  discriminator its headline names, and where the headline does not name one either, it
+  gains it. The discriminators come from the loop variable by parameter expansion
+  (`${url%%:*}` and stripping the fixed prefix and suffix off the pattern), so they are
+  already in scope:
+
+      a remote glob (*) is handled remotely (an object-storage error, not a local one)
+      and the * glob is NOT reported as a local filesystem miss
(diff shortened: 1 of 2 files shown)


jdatcmd (approved):
**Approved at `0f6e1ef7`.** Verified with the same derived-helper method as #992 (write-up there, including two under-reporting failures of my own).

```
objstore_module.sh   names 23 -> 23
                     literal 22 -> 19, interpolated 1 -> 4
```

Three names move from literal to interpolated — the loop fix exactly: the continuation now carries the scheme its headline already had in scope, and the variable was already there. **Name count unchanged**, so this renames and neither adds nor drops an assertion.

Worth recording that my static sweep reported `0` literal duplicates on `main` for this file, against your measured 3 colliding. Both true: the continuation is **one source line inside a three-iteration loop**. A static count under-reports loops exactly as it over-reports interpolation — which is the third distinct way I have had that method be wrong today, and the reason #983's guard must count over emitted records.

PR #992, opened 2026-09-12
test: the last five suites' continuations name the case they continue (#982, the final five of eight)

OffgridwithJD:
This is the **last of #982's eight suites** — all five of them, because after #984, #989 and
#990 the rule has stopped producing new cases and these ten records are mechanical.

Ten checks across five suites shared five `(suite, part, name)` keys with another check.

## Measured: every suite, both trees, same box

| suite | control records | control keys | **lost** | branch records | branch keys | lost |
|---|---|---|---|---|---|---|
| `sorted_pathkeys` | 113 | 110 | **3** | 113 | **113** | **0** |
| `vector_agg_tlist_shape` | 68 | 65 | **3** | 68 | **68** | **0** |
| `alter_am_cleanup` | 45 | 43 | **2** | 45 | **45** | **0** |
| `eager_ordering_record` | 31 | 30 | **1** | 31 | **31** | **0** |
| `objstore_userinfo` | 7 | 6 | **1** | 7 | **7** | **0** |

Control is clean `main` at `827f2801`. **Every record count is unchanged, so this renames and
nothing else.** All ten runs: `rc=0`, `FAIL=0`.

The two trees ran **sequentially, not in parallel**, because both worktrees install into the
same `pg18a` prefix — a concurrent run would have measured the other tree's `.so`.

## This is not a new convention. It is each file's own convention, applied where it lapsed.

That is the part worth checking, and it is why these five needed no new rule. Each file
already names the case at a neighbouring site, then falls back to a bare continuation for
the next several:

```
test/sorted_pathkeys.sh, as it stands on main:
  and DESC still answers correctly          <- names its case
  and NULLS FIRST still answers correctly   <- names its case
  and it still answers correctly            <- does not           (x2)

test/vector_agg_tlist_shape.sh, as it stands on main:
  and GROUP BY still answers correctly      <- names its case
  and FILTER still answers correctly        <- names its case
  and DISTINCT still answers correctly      <- names its case
  and it still answers correctly            <- does not           (x4)
```

So the fix is to finish the pattern the author started, taking each discriminator from the
`REFUSE:` headline directly above it:

```
REFUSE: a non-prefix of the key is not an order the rows are in
  -> and a non-prefix still answers correctly
REFUSE: a column that is not in the key at all
  -> and a non-key column still answers correctly
REFUSE: a Z-order run is not a sort on its lead column
  -> and the Z-order run still answers correctly
REFUSE: an unsupported aggregate declines the whole target list
  -> and the unsupported aggregate still answers correctly
REFUSE: FILTER inside an expression over aggregates
  -> and FILTER inside an expression still answers correctly
REFUSE: ORDER BY inside an aggregate, wrapped
  -> and a wrapped ORDER BY still answers correctly
REFUSE: an aggregate over an expression of two columns
  -> and an aggregate over an expression still answers correctly
```

`alter_am_cleanup` runs the same eight operations in two databases — one without the
extension, one after `DROP EXTENSION` — so its two pairs take the database from the headline:
`and dropped there, without the extension` / `, after DROP EXTENSION`, and the same suffixes
on the `REFRESH MATERIALIZED VIEW` pair. `objstore_userinfo` takes the entry point:
`and the read_parquet message names userinfo` / `and the export_parquet message names
userinfo`.

Two sites take the discriminator from the **value expression** instead (#990's third form),
because their headline names no table:

```
control: and it moved the layout   ->  control: and it moved the tailgate layout
                                       control: and it moved the lexgate layout
```

matching the `layout tailgate` and `layout lexgate` those checks actually read.

## A check that the renames are removals, not a mistyped grep

Counting exact old names — the phrase followed by its closing quote — across the whole tree:

```
clean main    17 occurrences
this branch    0 occurrences
```

and **17 is exactly the number of sites this PR edits**, so the two numbers cross-check each
other. Nothing outside the five files mentions any of these names (`docs/`, `CHANGELOG.md`,
`README.md`, `.github/` all clean), so no unchanged line is falsified by the renames.

## Limits of what this proves

- **I measured collisions, not that the new wording is the best available.** The counts prove
  264 records over 264 distinct keys across the five suites. Whether each name matches its
  assertion is a reader's check, which is why every name above sits next to the headline or
  the expression it was taken from.
- **No ledger change.** None of the five is one of the two suites `test/check_ledger.tsv`
  covers; both ledger files are byte-identical to `main`, verified by an empty `git diff`.
  #982's remaining value is to #432's seeding, not to today's census.
- **Five of this repo's suites, on one major.** I ran `pg18a`. These are name changes with no
  version-dependent behaviour, and CI runs the matrix.
- **#982 is not closed by this.** The rename half is; the **guard** — no key may cover more
  than one check in one run — is what makes the class impossible rather than merely absent,
  and it needs the full-matrix set-compare described on #982 and #983. I have deliberately
  not added a closing keyword.

Checks run locally: five suites 264 checks / 0 FAIL on both trees, `shellcheck` 0 findings,
`bash -n` clean, `docs_style` 9/9.

Part of #982 (the last five of eight; #984 merged, #989 and #990 open).


jdatcmd (approved):
**Approved at `d663a512`. Verified with a derived helper list, after my first two attempts under-reported — twice, by two different mechanisms.**

## What I checked

```
file                        names      literal collisions
sorted_pathkeys.sh          105 -> 105        1 -> 0
vector_agg_tlist_shape.sh    48 ->  48        3 -> 0
alter_am_cleanup.sh          46 ->  46        2 -> 0
eager_ordering_record.sh     32 ->  32        1 -> 0
objstore_userinfo.sh          8 ->   8        1 -> 0
```

**Name count unchanged in every file**, which is the check that says this renames and does not add or remove an assertion.

## Both of my earlier sweeps were wrong, in the two directions you would predict

**First**, I matched only `check|check_num|check_skip|ansq|check_timing` — a **hand-written helper list**. `sorted_pathkeys` emits through `ans` and `ansp`, so I saw 82 names where there are 105, and reported `0 -> 0` collisions for a file that has one. A hand-maintained list of what counts as a check, used to audit checks.

The fix is to derive the helpers: every function in the file or `lib.sh` whose body calls `check` or `pgc_record`. That finds `ansp`, `check_ratio`, `check_text`, `diff_query_ordered` and fifteen more I would never have enumerated.

**Second**, on #989 the same static method reported `0` literal duplicates on `main` for a file you measured as 3 colliding. Also correct and also useless: the continuation sits **inside a three-iteration loop**, so it is one source line and three records.

```
static source count   under-reports loops (one line, N records)
                      under-reports interpolation (two literals -> one string)
                      over-reports interpolation  (one literal -> N strings)
```

I predicted the third of those in my #990 review after being caught by it. I then walked into the other two.

**The conclusion stands harder than when I wrote it**: #983's guard must count over emitted RESULT records. Every static approximation I have tried today has been wrong in a different direction, and I was the one arguing for the runtime form.

## #989 separately

```
objstore_module.sh   names 23 -> 23
                     literal 22 -> 19, interpolated 1 -> 4
```

Three names move from literal to interpolated, which is precisely the loop fix: the continuation now carries the scheme its headline already had in scope. The count is unchanged, so nothing was added or dropped. Approving that one too.

## The rule's four forms, now complete

```
#984  the continuation carries the discriminator its headline interpolates
#989  a loop's continuation interpolates what the loop already has in scope
#990  where nothing interpolates, it comes from the check's value expression
#992  hand-written sites take a phrase naming the case they continue
```

Reading the #992 diffs, the pattern that recurs is `and it still answers correctly` becoming `and a non-prefix still answers correctly`, `and a non-key column still answers correctly`, `and the Z-order run still answers correctly`. **Each one is shorter to read than the headline above it and now survives being lifted out of reading order**, which was the whole requirement.

OffgridwithJD:
**Rebased again after #989 merged, so your approval at `d663a512` now stands on `7ec8d3c9`. Flagging it rather than letting it ride, because an approval surviving a push is exactly what bit me on #985.**

What moved: nothing but the rebase. Measured rather than asserted — the content change with `CHANGELOG.md` excluded, on the approved commit and on the current one:

```
approved d663a512   165 lines   md5 f49288260f65
current  7ec8d3c9   165 lines   md5 f49288260f65   -> byte-identical
```

The only difference is that `CHANGELOG.md` now carries #989's entry above mine, because main does. Both insertions are kept; neither replaced the other.

Same for #993 (`af0ca9f8 -> f704817a`), which has no approval yet.

If you would rather re-approve on the current head than have the old one carry, say so and I will leave it alone until you do — the `#985` lesson was that a review timestamped before a push is not a review of what is there now, even when I can show the content matches.

jdatcmd:
**The #998 collision is now live on this PR, and here is the one-line fix.** #998 merged at `75b3548`, so `main` carries the skip loop that lists this arm's old name.

Composed `origin/main` + this branch in a worktree and ran the sweep:

```
merge: clean, NO conflicts
loops 8, compared 6, interpolated 1, armless 1
MISMATCH sorted_pathkeys.sh:283
```

The loop at `283` lists seven names; the `else` branch now emits:

```
  loop lists : "and the whole ordered result matches heap"
  else emits : "and the whole ordered result matches heap under the new collation"
```

**Fix:** change that one entry in the loop's list to match the renamed arm. The other six agree.

Note the merge is **clean** — git has nothing to report, and `test/selftest/470` goes red on `main` only after both are in. That is the collision recorded on both PRs since yesterday, arriving on schedule.

## A measurement trap I walked into on the way, worth recording

I first tried to settle this by counting, and got the wrong answer:

```
grep -cF "and the whole ordered result matches heap"   main 3   pr992 2
```

and reasoned that since only one of three occurrences changed, the loop's sibling was untouched. **Wrong, because `grep -F` counts a line containing the substring** — and both renames here *extend* the name rather than replace it, so each renamed line still matches. The count could not see the rename at all.

The composed run found it in one command. Counting the name did not, and would have let me tell you there was no collision.

Issue #991, opened 2026-09-12
A collection-time vacuity refusal also reports the refused test as silently lost, so one bare skip produces two findings and only one is true


OffgridwithJD:
When the vacuity layer refuses a run at collection time, the same run also prints:

```
VACUITY: 1 collected test(s) never reported an outcome, so the run lost them silently:
         tmp_repro_vac/test_offender.py::test_one_offending_arm
```

**The run did not lose that test silently. It refused it loudly, in the sentence
immediately above.** So a reader who typed one bare `@pytest.mark.skip` gets the correct
diagnosis, then a second finding telling them a test vanished without saying so — and the
natural response is to go looking for a lost test that was never lost.

The guard whose subject is a silent loss is firing on the one event that is the opposite of
a silent loss.

## Reproduce

1. Create a file with one offending arm, anywhere under `test/pytest/`:

   ```python
   import pytest


   @pytest.mark.skip
   def test_one_offending_arm():
   	assert True
   ```

2. Run it serially, with the venv interpreter that has pytest:

   ```
   cd test/pytest && /root/REFvenv/bin/python -m pytest -q tmp_repro_vac/test_offender.py
   ```

3. Both sentences appear. Measured on `main` at `c96f5262` and again on #988's head
   `992175e0`:

   | tree | mode | rc | refusal sentence | `VACUITY` line | INTERNALERROR |
   |---|---|---|---|---|---|
   | `main` `c96f5262` | serial | 4 | yes | **yes** | 0 |
   | `main` `c96f5262` | xdist `-n 2` | 1 | no | yes | 39 |
   | #988 `992175e0` | serial | 4 | yes | **yes** | 0 |
   | #988 `992175e0` | xdist `-n 2` | 4 | yes | **no** | 0 |

## This is not #988's doing, and #988 is what surfaced it

The serial rows are identical on both trees, so the behaviour is pre-existing. I found it
while reviewing #988, whose premise is *xdist should report like serial*: after that fix,
**xdist reports better than serial on this line.** Filing it separately so the asymmetry is
not later read as a defect in that fix.

## Where it comes from

`test/pytest/pgc_vacuity.py`:

- The class hook `pytest_collection_modifyitems` adds every collected item to
  `self.collected`.
- The module-level `pytest_collection_modifyitems` then raises
  `pytest.UsageError("the pgColumnar vacuity layer refuses this run: ...")`.
[...]

PR #1005, opened 2026-09-12
test/pytest: a loud refusal is not a silent loss (#991)

### LINKED ISSUE #963: A collection-time vacuity refusal loses its reason under pytest-xdist: rc 4 becomes 1 and the message is replaced by a 35-line INTERNALERROR

OffgridwithJD:
**The layer refused a run and then contradicted itself.** One bare `@pytest.mark.skip`:

```
ERROR: the pgColumnar vacuity layer refuses this run: a bare skip is refused ...
         silently: tmp/test_offender.py::test_one_offending_arm
```

**The run did not lose it silently. It refused it loudly, one line above.** A reader who typed one bare skip got the correct diagnosis and then a second finding telling them a test vanished without saying so — and the natural response is to go hunting a lost test that was never lost.

`collected - reported` is the right set difference and the wrong **meaning**. A silent loss is when nobody said anything; here the layer itself is what stopped the run. The guard whose entire subject is a silent loss was firing on the one event that is its opposite.

## The fix is one assignment, because #963 built the chokepoint

Every refusal in the layer routes through `_collection_usage_error`, so recording it there covers all five call sites:

```python
    # RECORDED BEFORE EITHER BRANCH, so the reconciliation cannot call this a silent loss
    config._pgc_vacuity_refused = msg
```

and the reconciliation skips **only** that problem:

```python
    refused = getattr(session.config, "_pgc_vacuity_refused", None)
    ...
    if missing and not refused:
```

The setup-skip problem still prints. A fixture removing every test that depends on it is not something a refusal accounts for, and the two are independent findings.

**This PR had to wait for #963** rather than being written on the base it was filed against: on `main` before that, the `UsageError` was raised inline at each site and there was no single place to record it. Writing it earlier would have meant touching five raise sites or inventing the helper #963 was already adding.

## Serial was the only path left, which is why the change is this small

```
main   serial   refusal + the silent-loss line
main   xdist    no refusal at all                 (that was #963)
#963   serial   refusal + the silent-loss line    <- what this closes
#963   xdist    refusal, no line
```

#963 clears `items[:]` in the worker, so the controller's `collected` set is already empty under `-n` and the contradiction cannot arise there.

## Red to green, driven on both trees

```
the new arms against main's pgc_vacuity.py   1 failed
    FAILED ...test_a_collection_refusal_is_not_also_reported_as_a_silent_loss[serial]
    the [xdist] row PASSED on main, which confirms #963 closed that half rather than
    my having guessed it
the same arms with the fix                   3 passed, 11 checks
whole corpus on the branch                   365 passed, 923 checks
```

## The second arm is the control, and the first is worth nothing without it

Dropping a problem from a reconciliation is one edit away from dropping the guard. So `test_a_genuine_silent_loss_is_still_reported` asserts that an item collected and never reported, with **no refusal anywhere**, is still named and still reddens the run:

```python
@pytest.hookimpl(trylast=True)
def pytest_collection_modifyitems(config, items):
    del items[-1]
```

Removed after the layer's own hook recorded it, and **without deselecting** — which is the shape of a crashed xdist worker. `pytest_deselected` is the hook that tells a deliberate subset from a loss, and nothing calls it there. The arm checks the sentence, the **name** of the vanished test, and `rc=1`.

## Two of this tree's guards caught my omissions, and it was one root cause

`test_docs_cover_the_corpus` refused the new tests until `TESTS.md` named them. `test_harness_deps` then reported the no-driver batch as `rc=1` — which looked like a second, unrelated failure and is not: the docs guard is *inside* that batch. I checked that my own tests pass with psycopg shimmed out before concluding it, rather than assuming the obvious culprit.

## Limits

- **No ledger change**, measured: `harness_selftest` is 907 checks on this tree and on `main`. No shell check was added, removed or renamed.
- **Only the missing-outcome problem is suppressed, and only on a refusal.** Any other run, and any setup skip, reports exactly as before.
- **The flag lives on `config`, not the session.** `pytest_sessionfinish` receives `session` and reaches `session.config`; a worker's config is a different object from the controller's, which is why the worker branch sets it too even though the worker's own `sessionfinish` returns early.

Part of #991. The issue also notes `session.exitstatus = 1` being discarded on this path — `UsageError` decides the code after `sessionfinish` — which is the right outcome reached by accident rather than by the guard deciding not to lower it. That is untouched here and #991 stays open for it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2256,6 +2256,46 @@ true until the next version shipped.
   which is the process serial already used. The in-test control (a body that
   concludes nothing) is unchanged in both modes.
 
+- A collection-time vacuity refusal is no longer also reported as a silent loss (#991).
+
+  The layer refused a run and then contradicted itself:
+
+      ERROR: the pgColumnar vacuity layer refuses this run: a bare skip is refused ...
+      VACUITY: 1 collected test(s) never reported an outcome, so the run lost them
+               silently: tmp/test_offender.py::test_one_offending_arm
+
+  **The run did not lose it silently. It refused it loudly, one line above.** So a reader who
+  typed one bare `@pytest.mark.skip` got the correct diagnosis and then a second finding
+  telling them a test vanished without saying so -- and the natural response is to go looking
+  for a lost test that was never lost.
+
+  `collected - reported` is the right set difference and the wrong **meaning**: a silent loss
+  is when nobody said anything, and here the layer itself is what stopped the run. The guard
+  whose whole subject is a silent loss was firing on the one event that is its opposite.
+
+  **One assignment covers all five refusal sites**, because #963 gave the layer a single
+  chokepoint: `_collection_usage_error` records the refusal, and the reconciliation in
+  `_RunShape.pytest_sessionfinish` skips the missing-outcome problem when it is set. Only that
+  problem. The setup-skip problem still prints -- a fixture removing every test that depends on
+  it is not something a refusal accounts for, and the two are independent findings.
+
+  **Serial was the only path left.** #963 clears `items[:]` in the worker, so the controller's
+  `collected` set is already empty under `-n` and the contradiction cannot arise there:
+
+      main   serial   refusal + the silent-loss line
+      main   xdist    no refusal at all            (that was #963)
+      #963   serial   refusal + the silent-loss line   <- what this closes
+      #963   xdist    refusal, no line
+
+  The second arm is the control and the first is worth nothing without it: an item collected
+  and never reported, with **no** refusal anywhere, is still named and still reddens the run.
+  Dropping a problem from a reconciliation is one edit away from dropping the guard. The
+  fixture removes an item after the layer's hook recorded it and without deselecting it, which
+  is the shape of a crashed xdist worker -- `pytest_deselected` is what tells a deliberate
+  subset from a loss, and nothing calls it there.
+
+  No shell check moved, so no ledger change: `harness_selftest` is 907 checks on both trees.
+
 ## [1.0-alpha3] - 2026-09-02
 
 ### Added
--- a/test/pytest/TESTS.md
+++ b/test/pytest/TESTS.md
@@ -263,6 +263,25 @@ a fix that moved the wrong hook reddens here.
 | `test_a_bare_skip_refusal_keeps_its_reason_under_xdist` | collection skip: rc 4, sentence, no INTERNALERROR, serial and `-n 2` |
 | `test_a_broad_except_refusal_keeps_its_reason_under_xdist` | the same for a second collection-time rule, so the defect is the hook |
 | `test_an_in_test_vacuity_refusal_is_unchanged_under_xdist` | **control**: a body that concludes nothing stays rc 1 with the sentence |
+| `test_a_collection_refusal_is_not_also_reported_as_a_silent_loss` | a refusal is not ALSO reported as `lost them silently`, serial and `-n 2` |
+| `test_a_genuine_silent_loss_is_still_reported` | **control**: an item collected and never reported, with no refusal, is still named |
+
+**A loud refusal is not a silent loss (#991).** A collection-time refusal printed its
+sentence and then, directly beneath it, `VACUITY: N collected test(s) never reported an
+outcome, so the run lost them silently`. The run did not lose them silently — it refused
+them loudly, one line above — so a reader who typed one bare skip got the right diagnosis
+plus a second finding sending them after a test that was never lost. `collected - reported`
+is the right set difference and the wrong *meaning*: a silent loss is when nobody said
+anything, and here the layer itself stopped the run.
+
+Every refusal goes through `_collection_usage_error`, so recording it there covers all five
+call sites with one assignment, and the reconciliation skips **only** that problem. The
+setup-skip problem still prints: a fixture removing every test that depends on it is not
+something a refusal accounts for. The control above is what keeps dropping a problem from
+becoming dropping the guard — it is one edit away.
+
+Serial was the only path left after #963, because clearing `items[:]` in the worker already
+emptied the controller's `collected` set under `-n`.
 
 A worker records the sentence on `workeroutput` and clears the items so no ids
 cross. The controller re-raises `UsageError` from `pytest_testnodedown`, which
--- a/test/pytest/pgc_vacuity.py
+++ b/test/pytest/pgc_vacuity.py
@@ -1426,9 +1426,21 @@ class _RunShape:
         # sees a slice, and the controller receives every worker's reports.
         if hasattr(session.config, "workerinput"):
             return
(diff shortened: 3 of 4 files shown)


linuxhikerpm (changes requested):
The hatch is real and I drove it at `578932a80702b61ddfa3a49ff2ddb058b2312cc9`.

```
main pgc_vacuity + these tests   serial: lost them silently got 1 want 0
this head                        7 passed, including both modes and the genuine-loss control
```

The serial row is the one #963 left, and the control still names `test_vanishes_without_a_word` when nothing refused. I would approve the code on that.

**Blocking: `docs_style.sh` fails on this CHANGELOG.** Three em dashes, `got [3] want [0]`. `docs_style` is a registered suite, so the PG17/PG18 jobs will red for a reason unrelated to the refusal. The three sites are the new #991 bullet (the "so — and" / "prints — a" / "worker — `pytest`" sentences). `--` is the spelling the rest of Unreleased uses.

linuxhikerpm (approved):
Re-reviewed at `e11e1fc59e80cddefba65a7160bc98ec295bc6a3`. CHANGELOG em dashes are gone (`em 0 en 0`), which is the `docs_style.sh` hatch that failed suites on the previous head.

Drove the two new tests:

```
this head                                      3 passed, 11 checks
main pgc_vacuity + these tests                 serial: lost them silently got 1 want 0
                                               (xdist still 0, as #963 already emptied collected)
```

The control (`test_a_genuine_silent_loss_is_still_reported`) stayed green on both trees. Do not merge from this side.

jdatcmd (approved):
**Approving.** The diagnosis is right, the fix is at the one place that covers every surface, and the removal proof lands.

## Verified rather than read

**One assignment really does cover every call site.** `config._pgc_vacuity_refused = msg` sits at line 1507, inside `_collection_usage_error` (1484), and there are exactly five call sites — 1535, 1543, 2108, 2154, 2218. So the claim in the comment holds by construction rather than by inspection of each caller.

Worth noting the composition: **two of those five are surfaces that only started going through this function today.** #963 routed the module-binding refusal; #986's conflict resolution routed the `Expect`-method refusal this afternoon. Before that merge, a method-stub refusal would have set no flag and would still have been reported as a silent loss. This PR lands on the tree where "every refusal comes through here" became true, and it would have been a partial fix a day ago.

**Removal proof:**

```
revert `if missing and not refused:` to `if missing:`
  -> test_a_collection_refusal_is_not_also_reported_as_a_silent_loss[serial]   FAILED
```

**The full database-free batch on the branch: 267 passed, 637 checks, 0 fail.**

## One note, and your docstring already contains the answer

The mutation reddens `[serial]` and **not** `[xdist]`. Your own table says why:

```
#963   serial   refusal + the silent-loss line   <- what this closes
#963   xdist    refusal, no line
```

Clearing `items[:]` in the worker had already emptied the controller's `collected` set, so the xdist row had no line to suppress before this PR. That makes the `[xdist]` arm a **regression guard for #963's property, not a second measurement of this fix** — it passes on the mutant and on the repair alike.

That is the right arm to have and I would not remove it. What I would add is one sentence saying so, because the docstring states the four-row history and leaves the reader to infer that only one of the two parametrized arms can fail for *this* change. Two green arms read as two confirmations; here they are one confirmation and one guard. Given how much of tonight went into arms that could not fail, naming the one that cannot fail *for this reason* seems worth the line.

## The diagnosis is the part worth keeping

> `collected - reported` is the right set difference and the wrong MEANING: a silent loss is when nobody said anything, and here the layer itself is what stopped the run.

That is the whole bug in one sentence, and it is a shape that will recur: a set difference that is arithmetically correct and semantically wrong because the *reason* for the difference is not in the sets. The guard whose subject is silence fired on the loudest event the layer produces.

## Verdict

**Approved.** Correct fix, correct place, proof both ways, and the limitation of the xdist arm is disclosed rather than papered over.

PR #1007, opened 2026-09-12
docs: name join-key clustering as the runtime-filter skip precondition (#752)

### LINKED ISSUE #752: A runtime filter is worth nothing unless the fact table is clustered on the join key, and the fold a join removes is worth 4.1x but is off by default

| | columnar | heap | |
| --- | ---: | ---: | --- |

| shape | columnar vs heap |
| --- | ---: |

be quoted.)

above it.

## What to settle first

linuxhikerpm:
## Summary

- Addresses the **docs slice** of #752. A star-schema fact table that is not clustered on the join key still holds every key in every group, so group skip is a no-op. That skip was already measured (`19` of `20` groups removed when the keys are local, `0` of `20` when they cycle). The how-to named the GUC and not that discriminator.
- Independent twins: two checks in `test/docs_style.sh`, and `test/pytest/test_docs_join_clustering.py`. Neither suite reads the other. Causation: reverting the how-to Tuning reddens only the how-to arms; stripping `join key` from best-practices reddens only those arms.
- The GUC `pgcolumnar.enable_join_runtime_filter` **stays off**. The fold over a join is a later slice.

**This does not close #752.** The issue stays open.

## Test plan

- [x] Pytest twins red on unfixed docs, green after the prose, red again under the two independent mutations
- [x] `test/docs_style.sh` matches those outcomes
- [x] `test_docs_cover_the_corpus.py` names the new file and both tests
- [ ] CI suites include `docs_style`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -579,6 +579,14 @@ true until the next version shipped.
 
 ### Fixed
 
+- The star-schema join how-to names clustering on the join key (#752).
+
+  Group skip was already measured: 19 of 20 groups when the keys are local,
+  0 of 20 when they cycle. The page named the GUC and not that discriminator.
+  A fact table that is not clustered on the join key still holds every key in
+  every group, so the filter cannot skip. The GUC stays off. The fold over a
+  join is a later slice.
+
 - A skipped arm records under the name it would have used, so a skipped arm and
   a deleted one are no longer indistinguishable (#994).
 
--- a/docs/ARCHITECTURE.md
+++ b/docs/ARCHITECTURE.md
@@ -208,6 +208,7 @@ Serial join runtime filter. A `set_join_pathlist_hook` wraps a serial inner
 Hash Join whose outer path is a direct columnar scan. Core Hash Join keeps the
 answers. The coordinator drains the build side first. It then attaches a
 conservative key range and a Bloom filter to that scan.
+Group skip needs the fact table clustered on the join key.
 
 ### columnar_vector.c
 The vectorized aggregate path and its shared filter. A column-at-a-time filter
--- a/docs/best-practices.md
+++ b/docs/best-practices.md
@@ -76,6 +76,10 @@ the predicate. That test is only as sharp as the data's order. A column whose va
 are scattered across every chunk group cannot be pruned. A column whose values are
 clustered into a few groups prunes the rest. So the most effective layout choice is
 to cluster on the column you filter by ranges. That column is most often a timestamp.
+A star-schema join skips on the join key, not on that timestamp.
+Cluster the fact table on the join key so matching keys sit in few groups.
+A fact table clustered only on a timestamp still holds every join key in every group.
+A runtime filter then cannot drop groups.
 
 - `pgcolumnar.recluster(table)` re-establishes the sort order incrementally and
   online. It runs under `ShareUpdateExclusiveLock`, so reads and writes continue. It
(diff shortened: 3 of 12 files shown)


OffgridwithJD:
Reviewed. The change is right and I am not asking for it to be held. One finding on the
guards, with the mutation that produced it.

**Both guards pass when the sentence is negated.**

I mutated the documented claim to its opposite, keeping every word the guards look for:

```
-Group skip needs the fact table clustered on the join key.
+Group skip does NOT need the fact table clustered on the join key.
```

Then ran both halves:

```
docs_style            rc=0   FAIL=0
pytest twin           2 passed
```

Neither moved. The guards pin the *vocabulary* of the sentence -- that the words
"clustered" and "join key" appear near "group skip" -- not the *claim* the sentence
makes. So they would hold green over a doc that says the exact reverse of the behaviour,
which is the one drift a docs guard exists to catch.

This is the same shape as `a-sentence-that-cites-only-itself`: the check and the thing
checked share their evidence, so there is no direction in which the check can fail.

**Not a merge blocker.** The guard is still better than no guard -- it catches the
sentence being deleted or reworded away from the vocabulary. But the PR body says the
guards pin the claim, and measured, they pin its wording. Two options, both cheap:

1. Soften the body: say the guards pin that the sentence is *present and on-vocabulary*,
   not that it is true. Then the guard and its description agree and nothing else changes.
2. Make one arm behavioural: assert the clustered and unclustered fixtures actually
   differ in groups skipped. That arm fails when the claim reverses, because it reads the
   behaviour rather than the prose.

(1) is honest and free. (2) is the real guard, and it belongs on the pytest side since
that is where a runtime fixture already lives.

I will approve on whichever you pick -- including (1).

jdatcmd (approved):
**Approving.** The docs say the true thing, the guards are load-bearing rather than born green, and the claim underneath them is a real arm on `main`. One finding, non-blocking, and it is the same shape in both harnesses.

## Verified rather than read

**The guards go red without the docs.** Restored `main`'s `best-practices.md` and `how-to.md` into the branch:

```
on the branch          2 passed, 6 checks
main's two doc pages   2 FAILED
```

That is the check I most wanted, because a docs guard that passes on the tree it was written against proves nothing. `join key` appears nowhere in `main`'s `best-practices.md`, so the arm is genuinely new coverage.

**The claim the docs rest on is measured on `main`**, in `native_join_runtime_filter.sh`:

```
baseline reads all groups           20
clustered groups removed            19
clustered reads fewer groups         1
```

with the scattered arm beside it and `LEFT/SEMI/ANTI/CROSS` refusals as controls. So "19 of 20 when local, 0 of 20 when they cycle" is not inherited from an old worktree — it is an arm that runs today.

## The finding: the best-practices guard's name promises more than it checks, in BOTH harnesses

```python
# pytest
expect.num(int("join key" in text), 1, "best-practices names clustering on the join key")
```
```sh
# docs_style.sh
print("yes" if "join key" in text else "no")
check "best-practices names clustering on the join key" ...
```

Both assert only that the string `join key` appears **somewhere in the whole file**. A page saying *"the join key is irrelevant to layout"* satisfies both. The names say "names **clustering** on the join key"; neither looks for the clustering verb.

The `how-to` arm does not have this problem — it scopes to the section and requires `cluster` **and** `join key`. That one is the model.

**What makes this worth saying rather than a nit:** the two harnesses agree here, and their agreement is worth nothing, because they encode the *same* assumption rather than two measurements of it. The both-harnesses rule protects against a wrong implementation, not against a wrong idea held once and written twice. That is the first case I have seen in this tree where the rule gives no cover, and it is worth knowing that it can happen.

Two-line fix, matching the `how-to` arm's shape:

```python
expect.num(int("cluster" in text and "join key" in text), 1, ...)
```

Better still, scoped to the layout section rather than the whole page, so an unrelated mention elsewhere cannot satisfy it.

## Smaller note

`docs/configuration.md`'s table still reads `off` for `pgcolumnar.enable_join_runtime_filter`, which is correct on this branch and becomes false the moment #1008 lands. #1008 changes that line — I checked — so this is a sequencing fact rather than a defect, and it is the right way round: the docs that describe the precondition land before the default that relies on readers knowing it.

## Verdict

**Approved.** The prose is accurate, the structure is right — precondition first, default second — and the guards fail on the tree they were written to change. Take the two-line tightening when convenient; it does not change what ships.

PR #1008, opened 2026-09-12
feat: turn join runtime filter on by default (#752)

### LINKED ISSUE #752: A runtime filter is worth nothing unless the fact table is clustered on the join key, and the fold a join removes is worth 4.1x but is off by default

linuxhikerpm:
## Summary

- Sits on #1007. Clustering guidance now names the join key, so `pgcolumnar.enable_join_runtime_filter` defaults **on**. Group skip is still a no-op on an unclustered fact table. Bloom rejection still applies. SET the GUC off to compare.
- Independent twins: `join runtime filter defaults on` and `default plan has runtime coordinator` in `test/native_join_runtime_filter.sh`, and `test_join_runtime_filter_defaults_on` in pytest. Neither suite reads the other.
- Causation: restoring the GUC boot value to `false` reddens exactly those two shell checks (44 pass + 2 fail). TDD red against the still-off binary, then green after the flip.
- Ledger: two new `native_join_runtime_filter` rows, last-red `2026-09-12`. Census of `never` is unchanged at 1155.

**This does not close #752.** The fold over a join is still open.

Depends on #1007 (first commit). Do not merge this before the docs slice if you want them reviewed apart.

## Test plan

- [x] Native suite red on the two new checks against default-off, green after default-on, red again under the boot-value mutation
- [x] Pytest `test_join_runtime_filter_defaults_on` matched those SHOW outcomes
- [ ] CI `native_join_runtime_filter` and pytest corpus docs cover


OffgridwithJD:
I measured the cost, because the body measures only that the default is on.

The recorded precedent is `parallel_flush`: that default flip was **declined**, and the
lesson written down from it was "I measured the win, jdatcmd measured the cost." A default
flip is a claim about the shapes nobody benchmarked, so this is the measurement for the
shape that should hurt most -- the filter builds and rejects nothing.

Box: `pgcolumnar-audit`, `/usr/local/pg18n` (non-assert, so the ratio is not assert-biased),
`max_parallel_workers_per_gather=0`, nothing else running.

**Adversarial fixture.** Every fact key matches a dim key, so the filter is pure overhead:

```
premise: Runtime Filter Groups Removed: 0   Runtime Filter Rows Rejected: 0
```

400k rows, 20k groups, wide payload (256-byte text), both arms interleaved and the
order alternated per pair so arm order cannot favour either.

```
pairs            15
ON slower in     9 of 15 pairs   (two-sided sign test p = 0.6072)
median ON        342.6   median OFF   349.8   ratio 0.979 (-2.1%)
per-pair deltas  [40.8, 3.7, -4.2, -1.6, 35.6, -27.6, 22.4, 8.1, 59.8, 11.7, 10.3, -3.7, -8.8, -5.0, 15.6]
```

**No measured cost on the shape designed to show one.** 9-of-15 is what a coin does.

**I have to correct my own earlier number before anyone quotes it.** My first pass on this
ran the two arms as interleaved sweeps and then compared the medians of each arm's range.
That produced "+17.1% wide, +16.1% narrow" and it is an artifact of the comparison, not of
the code. The deltas above are the same machine and the same query: they straddle zero by
±60ms on a ~345ms query. Comparing two arms' medians throws away the pairing that the
interleaving exists to create -- the right statistic for an interleaved design is the
per-pair sign, which is the table above. The "+17%" was me measuring my own noise.

Where the filter does have something to do, it wins:

```
scattered (Groups Removed: 0 -> filter active)   ratio on/off = 0.890  (-11.0%)
```

So, unlike `parallel_flush`, this flip has a measured win and no measured loss on its worst
shape. That removes my only objection to the default.

Two things I would still want in the body before this lands, both cheap and neither a
rewrite:

1. **State the shape that does not benefit and that it was measured.** The body proves the
   default is on; it does not say what the default costs when the filter rejects nothing.
   A reader six months from now flipping it back needs that number to exist somewhere, and
   a PR body is where this repo keeps them.
2. **Name the box and the build.** A ratio from an assert build is biased (measured
   previously at 2.5% on a different ratio), so "non-assert pg18n" belongs next to the
   numbers.

Paste mine if it is easier -- they are yours to use.

jdatcmd (commented):
**One finding that must be fixed before merge, and it is concrete: the CHANGELOG contradicts itself about this exact GUC, thirteen lines apart.**

```
582:  - Serial join runtime filter is on by default, now that clustering guidance exists (#752).
595:    ... so the filter cannot skip. The GUC stays off. The fold over a join is a later slice.
```

Line 595 is #1007's entry, written when the default was staying off, and this PR carries that commit (`9fbb9b0` is an ancestor of `60ddc10`, verified). It was true in #1007 and is false here. `docs/configuration.md` correctly says `on`, so the contradiction is confined to the changelog — but the changelog is what a user reads to find out what changed, and this one tells them both things.

Fix is one sentence: drop or amend "The GUC stays off" in the carried entry.

## The evidence for flipping is stronger than the stated reason

The changelog says "now that clustering guidance exists", which explains why the *docs* are ready, not that on-by-default is a net win. The suite makes the better argument, and it should be the one in the entry:

| case | measured |
|---|---|
| clustered on the join key | 19 of 20 groups removed, reads 1 |
| scattered | 0 groups removed, **but Bloom rejects >15,000 of 19,800 non-matches** |
| build side too large | `saturated bloom is disabled` — it turns itself off |

**That third row is what makes a default-on defensible**, and it is the one I went looking for. A filter that is on everywhere needs a story for the case where it helps nothing, and "it disables itself" is that story. Neither the changelog nor the docs say it.

## What I checked, given how recently a default change went wrong here

#890's phase 1 this afternoon was a default that would have been changed on a third-party benchmark that does not reproduce on this engine. So I asked the same questions of this one:

- **Is the ceiling measured on the current tree, not inherited?** Yes — `clustered groups removed = 19`, `reads fewer groups = 1`, `baseline reads all groups = 20` are arms that run today in `native_join_runtime_filter.sh`.
- **Is the no-benefit case covered?** Partly, and better than I expected: saturation disables the Bloom, and the scattered arm shows rejection still pays when skip cannot.
- **Is the cost of an active filter that rejects little measured?** **No.** No arm puts a number on the overhead of a filter that is on, not saturated, and rejecting almost nothing — a high-match-rate join. That is reasoned rather than measured. I am not holding the PR for it, because saturation bounds the pathological end and the scattered arm bounds the middle, but it is the gap and it should be named rather than left implied.
- **Does it break anything else?** 13/13 green including both `suites` legs, which is what covers the cross-suite risk of a default that changes plan shape.

## The new arms are the right two

```
join runtime filter defaults on          -> SHOW returns on
default plan has runtime coordinator     -> and it actually takes effect
```

Pinning the GUC's value alone would have been satisfied by a default that is on and inert. The second arm is what makes the first mean something.

## Sequencing

`9fbb9b0` is an ancestor of `60ddc10`, so this is a proper stack, but both PRs are based on `main` rather than one on the other. That means either order merges cleanly and #1007's branch being deleted cannot auto-close this one. Landing #1007 first is still right — the precondition should be documented before the default relies on readers knowing it.

## Verdict

**Not approving yet, on the changelog contradiction alone.** The code change is two lines, the evidence is real and current, and the arms are well chosen. Fix the "The GUC stays off" sentence, and consider putting the saturation behaviour in the entry, since it is the actual argument for making this a default. I will approve on sight.

OffgridwithJD:
Reviewed the whole stack. CI is green (13 checks, 0 not-success) at `60ddc10b`, the C change
is the two boot values and nothing else, and the docs are consistent -- `configuration.md`,
`features.md` and `how-to.md` all say "on by default" at this head. The ledger arithmetic
checks out: 1155 -> 1157 rows, `never` unchanged at 1155, budget `checks_never_observed_red
1155` agrees.

Two things to fix, both small. Neither is about the C change.

## 1. The CHANGELOG states both defaults, thirteen lines apart

Both inside `## [Unreleased]`:

```
581  ### Changed
582  - Serial join runtime filter is on by default, now that clustering guidance exists (#752).
...
589  ### Fixed
590  - The star-schema join how-to names clustering on the join key (#752).
595    ... every group, so the filter cannot skip. The GUC stays off. The fold over a
```

This is the stack rather than carelessness: line 595 is #1007's entry, true when it was
written, and it is #1008 that falsifies it. But the two land together, so a reader of the
1.0-alpha4 notes is told the GUC is on and that it stays off, for the same issue number.

The fix is the four words at line 595. `The GUC stays off.` -> `The GUC is on by default.`,
or just delete the sentence, since the entry above it now carries the default.

(This is `a-pr-falsifies-a-line-it-does-not-touch` with the line inside the same PR stack.
Worth noting for #996: a stacked changelog entry can go stale between the two commits of one
PR, which the same-anchor problem makes more likely, not less.)

## 2. The two new ledger rows say "red, cause unrecorded" when you know the cause

```
native_join_runtime_filter	native_join_runtime_filter	default plan has runtime coordinator	2026-09-12	-
native_join_runtime_filter	native_join_runtime_filter	join runtime filter defaults on	2026-09-12	-
```

The last field is `-`, meaning no mutation. But the body says:

> Native suite red on the two new checks against default-off, green after default-on, **red
> again under the boot-value mutation**

So these reds came from a named, deliberate attack, and the ledger records them as reds
nothing explains. That column is the one thing the ledger holds beyond "was it ever red":
`pgc_ledger.py` calls it "the catalogue this column exists to become", and `merge` makes the
caller choose between `--mutation NAME` and `--reds-are-real` precisely because an
environment red and a real observation are identical in a log.

`--reds-are-real` is the flag that says "this is a genuine observation of the code under
test". What you did was better than that -- you reverted the boot value and watched two
specific checks die -- and the record is weaker than the work.

Re-merge that log with `--mutation boot-value-false` (or whatever you want the attack
called). The rows then read

```
...	default plan has runtime coordinator	2026-09-12	boot-value-false
```

and the next person to ask "is this check able to fail, and how do I make it fail again" gets
an answer instead of a date. It also survives a prune: `orphan-scan` refuses to drop a row
carrying history, and a mutation name is what makes the history legible rather than merely
present.

## On the performance question, which was my earlier objection

Withdrawn and measured -- the paired numbers are in my comment above. No measured cost on the
adversarial shape (9 of 15 pairs slower, sign test p=0.61), and -11% where the filter has
work to do. I also corrected a bad "+17%" of my own there; please don't quote that number.

**I will approve as soon as line 595 is fixed.** The ledger re-merge I would also like before
merge, but if you would rather land the default and fix the mutation column in a follow-up,
say so and I will approve on the changelog line alone -- the row is right about the fact, just
silent about the cause.

jdatcmd:
**I pushed the changelog fix to this branch myself rather than handing it back — `60ddc10` → `4b7e66d`, on jd's instruction to fix and merge.** Saying exactly what changed, because it is your PR.

**1. Removed the contradiction.** The carried #1007 entry ended "The GUC stays off", twelve lines below your entry saying it is on by default. Removed the sentence rather than rewording it: that paragraph is about what a scattered fact table cannot do, and the default belongs to the entry above.

**2. Replaced the stated reason with the evidence.** "now that clustering guidance exists" says the docs are ready; it does not say on-by-default is right. Your suite says that, in three rows, and I verified each against `native_join_runtime_filter.sh` on this branch before writing it down:

```
clustered on the join key   19 of 20 groups removed, 1 read    :37, :38
scattered                   0 removed, Bloom rejects >= 15000
                            of 19800 non-matches               :57, :76
build side too large        "Runtime Filter Bloom: false"       :176
```

**The third row is the one that makes a default defensible and it was nowhere in the entry.** A filter that is on for every user needs an answer for the case where it helps nothing, and turning itself off is that answer. That is your code doing the right thing and the release notes not mentioning it.

**3. Named what is not measured.** The overhead of a filter that is on, unsaturated, and rejecting almost nothing. Saturation bounds one end, the scattered case bounds the middle, so it is reasoned rather than measured — and a default change should say which of the two it rests on.

No code change, no check added or renamed, so no ledger movement.

**If you would rather write that entry yourself, say so and I will reset to `60ddc10`.** The words are mine and the measurements are yours; you may well phrase the trade differently.

jdatcmd (approved):
Approving at `4b7e66d`. The one finding I held this for is fixed, by me rather than by the author — see the comment above for exactly what I changed and an offer to reset it if they would rather write it themselves.

The code change remains two lines, the evidence is measured and current, and the entry now carries the saturation row that makes a default-on defensible plus an explicit note of what is reasoned rather than measured.

OffgridwithJD:
**Correction to my review: my second finding asked for something the tool refuses.** The
changelog fix you made before merging was right and that half stands. This half was my
error.

I said to re-merge the log with `--mutation boot-value-false` so the two rows would carry
the attack that reddened them. Measured, on the committed tool:

```
$ pgc_ledger.py merge --ledger l.tsv --date 2026-09-12 --mutation boot-value-false two.log
    ledger integrity failure: --mutation names one check, but 2 checks failed in this run:
      demo	part1	defaults on
      demo	part1	has coordinator
    Attributing it to all of them would record collateral damage as evidence.
rc=2

$ pgc_ledger.py merge --ledger l.tsv --date 2026-09-12 --reds-are-real two.log
  ledger: rows=2 | ... observed red ever=2
rc=0
```

The boot-value mutation reddens both checks, so `--mutation` is refused outright. What you
did -- `--reds-are-real`, leaving the mutation column `-` -- is **the only state the tool
permits** for that log. Nothing to fix here, and I should have run the command before
asking for it.

The finding underneath it is real but it is against the tool, not this PR: a mutation with
two genuine targets cannot be recorded at all, so the column `pgc_ledger.py` calls "the
catalogue this column exists to become" is unreachable for any mutation that kills more
than one check. Filed separately so it does not sit on a merged PR.

Issue #1011, opened 2026-09-12
PGC_SEED does not seed the fuzzer: two runs with one seed share zero fixtures, because every generator call is a command substitution and bash re-seeds RANDOM in a subshell


OffgridwithJD:
`test/fuzz.sh:27` prints, on every run:

```
-- fuzz seed=12345 iters=25  (reproduce a failure with PGC_SEED=12345)
```

That instruction does not work. `PGC_SEED` controls nothing, and two runs of the fuzzer on
the same major with the same seed share **no fixture at all**.

## Measured

Same box, same major (`pg18a`), same `PGC_SEED=12345`, run twice back to back:

```
run 1 rc=0 records=250
run 2 rc=0 records=250
run1 distinct check names: 250
run2 distinct check names: 250
shared:                      0
only in run1:              250
only in run2:              250

-- fuzz seed=12345 iters=25  (reproduce a failure with PGC_SEED=12345)
-- fuzz seed=12345 iters=25  (reproduce a failure with PGC_SEED=12345)

first name, run 1:  it=1 N=2933 nullmod=5  cg=500  sr=1000 codec=none lvl=18 :: compound
first name, run 2:  it=1 N=52   nullmod=10 cg=1000 sr=1000 codec=zstd lvl=15 :: compound
```

**Iteration 1 already diverges.** Both runs printed the same seed line while generating a
different fixture from it.

## The mechanism

`RANDOM=$SEED` at line 26 does seed the sequence. The generators never read it from a
context where that matters:

```sh
rnd()  { echo $(( RANDOM % $1 )); }
pick() { local n=$#; local i=$(( RANDOM % n )); shift "$i"; echo "$1"; }
```

and every call site is a command substitution:

```sh
N=$(( 1 + $(rnd 5000) ))       # fuzz.sh:42
CG=$(pick 100 250 500 1000)    # fuzz.sh:45
CODEC=$(pick $CODECS)          # fuzz.sh:47
```

`$( )` is a subshell, and **bash re-seeds `RANDOM` in a subshell** (changed in bash 5.1, so
that two subshells do not yield the same value). The seeded sequence stays in the parent and
nothing ever reads it.

Isolated, twice, on bash 5.2.21:

```
run: direct=[25202 15552]  via-command-substitution=[1520 1519]
run: direct=[25202 15552]  via-command-substitution=[3222 7526]
```

`direct` is identical across runs -- the seed works. `via-command-substitution` is different
[...]

PR #1013, opened 2026-09-12
test: a check record names the PostgreSQL major it was observed under (#1010)

OffgridwithJD:
Step 1 of #1010. **The ledger is not keyed on the major here** -- that is step 2, and it
needs a migration this does not. What this does is make the log say which major it came
from, so the tool stops taking that on trust from its invoker.

    RESULT<TAB>suite<TAB>part<TAB>name<TAB>verdict<TAB>major<TAB>reason

## Why the major belongs in the record

A check's EXISTENCE depends on it, so a record that omits it identifies a check only
partly.

- `test/analyze_differential.sh:61` emits **one** record on PG15-17 (a `check_skip`, then
  `pgc_summary` exits) and a suite's worth on PG18+.
- `test/fk_referencing.sh:287` emits **different check NAMES** in its two branches, so the
  older and newer majors' key sets are disjoint rather than nested.

A grep for `PGC_MAJOR` does not find all of these, which is why this is a field and not a
convention: `test/native_repack.sh:48`, `test/pg19_vacuum_options.sh:34` and
`test/native_dml.sh:61` gate on `server_version_num` and never mention it.

Before this, the major could only come from whoever invoked the tool. That is the
`--date not-a-date` failure one field over -- a PG15 log merged as PG18 is misattributed
and nothing in the log can contradict it -- except worse, because a date is metadata about
an observation while a major decides which checks can exist.

## Measured

Two full matrices at `4d7c75ae`, `ALL VERSIONS PASSED` on both, 252 suite logs each:

```
union of keys     6472        (fuzz excluded, see below)
on both majors    6367
major-specific     105        1.62%, across 8 suites

suite                    pg15only pg18only   shared
analyze_differential            1       26        0
analyze_function                1       54        0
audit                           1        4       27
column_projection               1        1       37
fk_referencing                  1        2       24
generated_columns               0        6        5
native_sort_by                  0        2       16
temporal                        0        5        0
```

`analyze_differential` and `analyze_function` share **zero** keys across the two majors,
which is the shape predicted from reading the suite.

`fuzz` is excluded and that exclusion is its own finding: it interpolates its random
fixture into the check name, so it emits 250 brand-new keys every run -- `shared: 0`
between two runs on the **same** major with the **same** `PGC_SEED`. Filed as #1011.

## Validated, not stored

`eighteen`, `18.2`, `pg18` and an empty field are each refused, in both reconcilers --
`pgc_ledger.py read_records` and `run_all_versions.sh pgc_reconcile_records`, which answer
different questions (a log from another machine vs. the suite just run) and so each need
the check.

`unknown` is the one non-numeric value. It is `lib.sh`'s own word for a field the harness
did not set -- already used there for an unset suite and part -- and it is the common case
rather than a courtesy: **`harness_selftest` never references `PGC_MAJOR` anywhere**, so
all 544 of its records per run say `unknown`, truthfully.

## Test plan

- [x] Full matrix on PG18 and PG19 at `2559f0ec`. Every record carries a valid major:

      pg18: 6753 records, 0 with an invalid major, 0 not 6 fields
            majors seen: 6209 "18", 544 "unknown"
      pg19: 6785 records, 0 with an invalid major, 0 not 6 fields
            majors seen: 6241 "19", 544 "unknown"

- [x] The only matrix failure was the ledger gate naming the 9 new arms, which is the
      designed refusal. Fixed by the second commit, re-derived rather than computed.
- [x] Part 400's nine new arms: 9 PASS, part 400 90 PASS / 0 FAIL on PG18.
- [x] Emitter driven with `PGC_MAJOR` set to **two different values** and unset, so an arm
      cannot pass against a field hardcoded to either answer.
- [x] Four invalid majors each refused, with a well-formed `unknown` control -- without it
      the four prove only that the reconciler started refusing everything.
- [x] `pytest` corpus failure set **identical to main** (106 both, zero delta; all 106 are
      the absent `psycopg` in a guard-only venv).
- [x] `docs_style.sh` PASSED, 9 checks.

## What I had to fix in my own work, since it is the interesting part

Widening 79 record fixtures with one regex corrupted exactly the fixtures that matter:
four **negative** ones (a six-field record whose refusal an arm asserts became valid), the
`BOGUS`-verdict fixture (skipped by the regex, so it silently became a field-count failure
and stopped testing the verdict), and **my own two new tests**, appended minutes earlier,
which got a second major. The controls are what caught it.

## Not in this PR

Keying the ledger on the major, the migration, and the census restatement. Step 2, and
#1010 carries a measured argument about its cost that is worth reading before it lands:
6,367 of 6,472 checks are identical on both majors, so `(major, suite, part, name)` stores
one fact five times for 98% of the ledger.

### DIFF
--- a/test/pytest/test_mutation_ledger.py
+++ b/test/pytest/test_mutation_ledger.py
@@ -29,10 +29,10 @@ REPO = pathlib.Path(__file__).resolve().parents[2]
 TOOL = REPO / "test" / "pgc_ledger.py"
 RUNNER = REPO / "test" / "run_all_versions.sh"
 
-GREEN = ("RESULT\tdemo\tpart1\tfirst check\tPASS\t\n"
-         "RESULT\tdemo\tpart1\tsecond check\tPASS\t\nchecks run: 2\n")
-RED = ("RESULT\tdemo\tpart1\tfirst check\tFAIL\t\n"
-       "RESULT\tdemo\tpart1\tsecond check\tPASS\t\nchecks run: 2\n")
+GREEN = ("RESULT\tdemo\tpart1\tfirst check\tPASS\t18\t\n"
+         "RESULT\tdemo\tpart1\tsecond check\tPASS\t18\t\nchecks run: 2\n")
+RED = ("RESULT\tdemo\tpart1\tfirst check\tFAIL\t18\t\n"
+       "RESULT\tdemo\tpart1\tsecond check\tPASS\t18\t\nchecks run: 2\n")
 
 
 def _run(*args, cwd=None):
@@ -134,8 +134,8 @@ def test_two_runs_of_a_check_are_not_a_duplicate_of_it(tmp_path, expect):
                "the same check in two logs is two runs, not a duplicate")
 
     twice = _w(tmp_path, "twice.log",
-               "RESULT\tdemo\tpart1\tsame\tPASS\t\n"
-               "RESULT\tdemo\tpart1\tsame\tFAIL\t\nchecks run: 2\n")
+               "RESULT\tdemo\tpart1\tsame\tPASS\t18\t\n"
+               "RESULT\tdemo\tpart1\tsame\tFAIL\t18\t\nchecks run: 2\n")
     out, _ = _run("merge", "--reds-are-real", "--ledger", _w(tmp_path, "l2.tsv", ""), "--date", "2026-09-10", twice)
     expect.num(out.count("duplicate check name in one run, so one ledger row covers 2: "
                          "demo\tpart1\tsame"), 1,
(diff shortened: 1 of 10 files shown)


jdatcmd (approved):
**Approving.** The record gains its major in the right place and nothing downstream mis-parses.

## What I checked

```
record order      suite, part, name, verdict, MAJOR, reason
RECORD_FIELDS     6 -> 7
ledger row        5 fields on main and on this branch -- unchanged, so the
                  'no migration here' claim holds
```

**The reason field moves from position 6 to 7**, which is the part that could have gone wrong quietly. It does not: a wrong field count is *refused*, with a message that names the expected fields including `major`, rather than read one field over. An old six-field log handed to the new tool fails closed instead of having its reason silently taken as a major — which is the failure this change exists to prevent, one field along.

I grepped the tree for every RESULT parser: `pgc_ledger.py` and `selftest/410`'s synthetic-log generator are the only two, and both are updated here. CI is green on all 13 including both suites legs, which is what actually exercises the consumers.

## The argument is the good part

> A check's EXISTENCE depends on it -- `analyze_differential.sh` emits one record on PG15-17 and N on PG18+ -- so a record that does not name its major identifies a check only partly, and the ledger keyed on it has to take the major from whoever invoked the tool.

That is the same shape as `--date not-a-date`: a fact about the run taken on trust from the invoker, with nothing in the log able to contradict it. I hit the consequence from the other side today — suites that skip wholesale on one major are exactly the rows an absence-based gate cannot interpret, and `unknown` being a truthful answer for `harness_selftest` rather than a courtesy is the right call.

Merging.

test/pytest/test_mutation_ledger.py
```diff
@@ -508,9 +508,9 @@ def test_a_log_that_does_not_parse_is_not_evidence(tmp_path, expect):
     cases = {
         "no reason field": "RESULT\tdemo\tpart1\ta name\tPASS\nchecks run: 1\n",
         "a verdict the emitter cannot emit": "RESULT\tdemo\tpart1\ta name\tBOGUS\t\nchecks run: 1\n",
```
linuxhikerpm, line 510: This is the BOGUS-verdict fixture the PR body said a regex widen silently turned into a field-count failure. `410-a-check-must-have-been-red.sh` was updated to `BOGUS\t18\t` and still greps for `'BOGUS'`. This twin was not, so it is six fields and the merge fails for `RECORD_FIELDS` before the verdict is examined. `expect.num(..., 2)` still passes, which is how the arm stops testing the verdict. Widen it the same way as 410: `RESULT\tdemo\tpart1\ta name\tBOGUS\t18\t\nchecks run: 1\n`.

linuxhikerpm (changes requested):
The emitter, both reconcilers, the 15/18/unset emitter arms, and the invalid-major controls look right, and the ledger still keys on `(suite, part, name)` as step 1. One pytest fixture did not get the widen this PR's own write-up says is load-bearing. Please fix that on `a0b2c00b6f8d4cdc23d6c60dec6571ed42c693c9` rather than merging past it.

PR #1018, opened 2026-09-12
ci: run the cluster-half pytest tests, and make both jobs assert their size (#1016)

### LINKED ISSUE #1016: 93 of 360 pytest tests run in no CI job: ci.yml has one pytest job that installs no driver, nightly.yml has none, and the recorded reason is stale
`test/pytest/` holds 360 test functions in 23 files. **93 of them, in 8 files, are run by no
CI job.**

## Measured

```
$ grep -c pytest .github/workflows/nightly.yml
0

$ grep -n '^  [a-z][a-z0-9-]*:$' .github/workflows/ci.yml
57:  shellcheck:
90:  pytest-guards:      <-- the only pytest job, and it installs no driver
126:  build:
303:  build-beta:
378:  suites:

$ grep -c pytest test/run_all_versions.sh
0
```

`run_all_versions.sh` must stay at zero: the two harnesses are independent and the shell
runner invoking pytest is the cross-harness call the project forbids. `lib.sh` names pytest
11 times and every one is a comment -- I checked them individually.

`pytest-guards` runs the files listed in `NO_CLUSTER`, in a venv where psycopg is
**deliberately** absent, and that job is what proves those files need no database. It is
correct and it is green:

```
15 files, 272 test runs (267 functions + 5 parametrized expansions)
checks run: 650
accounting: 650 pass + 0 fail + 0 unrun = 650
272 passed in 17.87s
```

The other 8 files need the driver and a cluster, and nothing runs them:

```
test_raises_sqlstate.py         34 tests
test_harness_deps.py            15
test_hilbert_locality.py        12
test_join_runtime_filter.py     12
test_connection.py              11
test_native_projection.py        7
test_saop_element_pushdown.py    1
test_zonemap_boundaries.py       1
                                93 tests, 26% of the corpus
```

They pass when somebody runs them by hand. Nothing notices when they stop. That is the shape
of the nightly gate that sat red for 25 nights behind a green PR gate, and of a watcher that
never fires: indistinguishable from working, right up until it matters.

## The recorded reason is now stale

`test/pytest/README.md`:

> `test/run_all_versions.sh` does not run these tests, and neither does CI. That is a decision
> with a price, recorded in section 1a of the design document: `pgc_skip` treats a missing
> dependency as a failure rather than a skip, so registering this run in `SUITES` would redden
> every CI job until `ci.yml` installs from `requirements-test.txt`. Until someone takes that
> decision, run it by hand.

Two things have changed since that was written. `ci.yml` **does** install from
`requirements-test.txt` -- that is how `pytest-guards` gets pinned pytest and xdist. And
`SUITES` was never the right mechanism: registering the pytest run as a shell suite is exactly
the cross-harness invocation the independence rule forbids. A second CI job is the right
mechanism, and `pytest-guards` is already its shape.

## What the job needs, and the part that is easy to get wrong

1. **psycopg**, from the same pins, so the runner installs what was tested.
2. **A build and install** of the extension. `conftest.py` builds before the server starts,
   and that order is load-bearing -- `shared_preload_libraries` maps the library at postmaster
   start, so a cluster started before the install keeps the OLD `.so` mapped for its whole
   life and every test measures the previous branch while reporting success.
3. **`--pg-config`** or `PGC_PG_CONFIG`; the default is `/usr/local/pg18a/bin/pg_config`,
   which exists on the audit container and not on a GitHub runner.
4. **`--pgc-expect-tests N`.** This is the part that matters. A job that collects nothing
   exits 0 and reports success, which is how this family of defect recurs -- a pending count
   that cannot see a job that never started, a `grep -q` that closed its pipe, an `until`
   loop that exits instantly on zero. The harness already has the guard, and it refuses
   `--pgc-expect-tests 0` as vacuous rather than accepting it.
5. **An assert build**, per `lib.sh`: a non-assert build hid a backend crash once already.

The file list should be DERIVED, not written in the workflow -- the complement of
`NO_CLUSTER`, the same way `pytest-guards` derives its list from `NO_CLUSTER` itself. Two
hand-maintained copies of that split is a thing that goes stale silently, which is the reason
the existing job derives it.

## Why this blocks the port rather than merely annoying

#432 is at 0.66% of product properties (36 of 5,453 distinct bash check names, measured
against a five-major matrix). Every product test ported from here lands in the ungated half.
So the gate is not a tidy-up to do afterwards -- it decides whether the next several thousand
assertions are worth anything.

OffgridwithJD:
Closes #1016.

**93 test functions in 8 files, 26% of the pytest corpus, ran in no CI job.** They were
never broken -- nothing ran them.

```
$ grep -c pytest .github/workflows/nightly.yml
0
$ grep -n '^  [a-z][a-z0-9-]*:$' .github/workflows/ci.yml
57:  shellcheck:
90:  pytest-guards:      <-- the only pytest job, and it installs no driver
126:  build:
303:  build-beta:
378:  suites:
$ grep -c pytest test/run_all_versions.sh
0
```

`run_all_versions.sh` must stay at zero: the harnesses are independent and the shell runner
invoking pytest is the cross-harness call the project forbids. `lib.sh` names pytest 11
times and every one is a comment -- I checked them individually.

## What this adds

**`pytest-cluster`**, the other half. One major, PGDG PostgreSQL 18 with its headers, every
pin from `requirements-test.txt` including psycopg, and the file list derived as the
complement of `NO_CLUSTER`.

**`--pgc-expect-tests` on BOTH jobs**, from `test/pytest/expected_tests.txt`. The flag
already existed and nothing passed it.

What it closes is narrower than "pytest passed having run nothing", and worse. A nonexistent
path already fails on its own -- that was not the hole. A file list that resolves to **real
files** and collects **fewer tests** does not. Measured, dropping one file from the guard
list:

```
unarmed   rc=0   "255 passed"                              17 tests gone, nothing said
armed     rc=4   "collected 255 test(s) but expected 272"
```

It also refuses `0` as vacuous rather than accepting it, which is the other way a count
guard gets quietly disabled.

**The numbers live in a tracked file**, not in the workflow and not in an environment
variable, for the reason `check_ledger_budget.txt` gives about its own: a change is then a
diff a reviewer sees, sitting next to the test that moved it. `PGC_SKIP_TIMING` is the
precedent for the alternative -- set in two workflow files, suppressing whole suites for
months, with no diff ever showing it.

The **split** stays derived from `NO_CLUSTER` in both jobs. `test -n` guards every derived
value, because an empty read would omit the flag and **fail open**.

## Test plan

- [x] Both halves on the branch at `051f33a5`, each armed from the tracked file:

      GUARD   (driver absent)   272 collected, 650 checks, 650 pass, rc=0
      CLUSTER (driver present)  101 collected, 301 checks, 301 pass, rc=0, 27s

- [x] **Controls.** Each half must REFUSE a wrong number, or the flag asserts nothing:

      guard   --pgc-expect-tests 271 -> rc=4  "collected 272 test(s) but expected 271"
      cluster --pgc-expect-tests 100 -> rc=4  "collected 101 test(s) but expected 100"

- [x] The guard venv is built with psycopg absent and that absence asserted, so the pair
      still proves the split rather than merely running twice.
- [x] `requirements-test.txt` installs clean in a fresh venv: psycopg 3.3.5, pytest 9.1.1,
      xdist 3.8.0.
- [x] The derived file list gives exactly 8 cluster files, and collection gives exactly the
      tracked 101.
- [x] `docs_style.sh` PASSED.
- [ ] **The apt path itself is unverified until CI runs it.** I cannot run GitHub Actions
      here; the PGDG steps are copied from the `suites` job, which uses them today, and the
      cluster half is verified against a local assert build instead. This is the one part of
      the change I am taking on inspection rather than measurement.

## Three arms, and one of them caught me

- `test_the_cluster_job_runs_the_other_half_and_derives_it` -- the job exists, derives from
  `NO_CLUSTER`, installs the pins, asserts the driver **is** present (the control for
  `pytest-guards` asserting it is absent), and names no corpus file literally.
- `test_both_pytest_jobs_assert_how_many_tests_they_collected` -- both jobs pass the flag,
  both read it from the tracked file, both guard the read.
- **The existing arm caught my own work.** `test_ci_derives_the_file_list_rather_than_repeating_it`
  forbids a corpus count in a job comment, over the comment block as well as the body. My
  new comment said "8 files". It failed before the job ever ran:

      AssertionError: the job and its comment state no corpus count: got '8 files' want 'none'

  The rule is right -- a number in a comment is the same hand-maintained derived value as a
  number in a list -- so the comment now states none and the measured figures are in the
  CHANGELOG, which is dated. That arm applies to the new job for free, because its region
  runs to the next job key.

`README.md`'s recorded reason had gone stale twice over: it said CI would first have to
install from `requirements-test.txt`, which `pytest-guards` already does, and it proposed
registering the run in `SUITES`, which is the forbidden cross-harness call. Corrected.

## Why this came first

#432 is at **0.66%** of product properties (36 of 5,453 distinct bash check names, measured
against a five-major matrix). Every product test ported from here lands in the half that was
ungated. The gate is not a tidy-up for afterwards -- it decides whether the next several
thousand assertions are worth anything.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -110,10 +110,138 @@ jobs:
           cd test/pytest
           FILES="$(python3 -c 'import sys; sys.path.insert(0, "."); from test_harness_deps import NO_CLUSTER; print(" ".join(NO_CLUSTER))')"
           test -n "$FILES"
+          # HOW MANY TESTS THIS MUST COLLECT, from the tracked file (#1016). pytest
+          # exits 0 having collected nothing, so this job could pass with a bad file
+          # list, an import error, or a rename that emptied the glob -- and did assert
+          # nothing about its own size until now. `test -n` matters as much as the
+          # number: an empty read would omit the flag and fail OPEN.
+          WANT="$(awk '$1=="guard_tests"{print $2}' expected_tests.txt)"
+          test -n "$WANT"
           # Printed, not stated: the count is a fact about the tree at this
           # commit, so it belongs in the output rather than in a comment.
           echo "running $(set -- $FILES; echo $#) database-free file(s): $FILES"
-          PYTHONPATH=. /tmp/pgcvenv/bin/pytest -q $FILES
+          echo "expecting $WANT collected test(s)"
+          PYTHONPATH=. /tmp/pgcvenv/bin/pytest -q --pgc-expect-tests "$WANT" $FILES
+
+  # THE OTHER HALF OF THE CORPUS (#1016). The files that need the driver and a cluster
+  # ran in NO job: ci.yml had one pytest job and it installs psycopg deliberately NOT,
+  # nightly.yml mentions pytest zero times, and run_all_versions.sh must mention it zero
+  # times because the two harnesses stay independent. They passed when somebody ran them
+  # by hand and nothing noticed when they stopped -- the shape of the nightly gate that
+  # sat red for 25 nights behind a green PR gate.
+  #
+  # NO COUNT HERE, and the arm in test_harness_deps.py enforces that over this comment
+  # as well as the job body: a number in a comment is the same hand-maintained derived
+  # value as a number in a list, and the last one went stale the day it was written. The
+  # job PRINTS what it ran, and the measured figures live in the CHANGELOG entry, which
+  # is dated. This comment was caught by that arm before the job ever ran.
+  #
+  # ONE MAJOR, not the matrix. conftest.py's default is an assert build that exists on
+  # the audit container and not here, so the pg_config is passed explicitly. Widening to
+  # a matrix is a separate decision: this job's purpose is that these tests RUN at all,
+  # and a matrix would multiply a cost nobody has measured yet before the first green.
+  pytest-cluster:
+    name: pytest (cluster tests, with the driver)
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Add the PGDG repository and install PostgreSQL 18 with its headers
+        run: |
+          set -euo pipefail
+          sudo install -d /usr/share/postgresql-common/pgdg
+          # Bounded and retried, for the reason the suites job gives: a plain
+          # "curl -fsSL" waits indefinitely on a stalled connection, which hung a
+          # step for 40 minutes on an otherwise healthy runner (#351).
+          sudo curl -fsSL --connect-timeout 15 --max-time 120 \
+            --retry 5 --retry-delay 5 --retry-all-errors \
+            -o /usr/share/postgresql-common/pgdg/apt.postgresql.org.asc \
+            https://www.postgresql.org/media/keys/ACCC4CF8.asc
+          echo "deb [signed-by=/usr/share/postgresql-common/pgdg/apt.postgresql.org.asc] \
+            https://apt.postgresql.org/pub/repos/apt $(lsb_release -cs)-pgdg main" \
+            | sudo tee /etc/apt/sources.list.d/pgdg.list >/dev/null
+          sudo timeout 300 apt-get -o Acquire::Retries=5 update
+          sudo timeout 600 apt-get -o Acquire::Retries=5 install -y --no-install-recommends \
+            postgresql-18 postgresql-server-dev-18 \
+            liblz4-dev libzstd-dev zlib1g-dev
+          # NOT cached, deliberately. The suites job caches because it runs a matrix for
+          # an hour; a second copy of that cache-key derivation is a thing that goes
+          # stale silently, and this job is minutes long.
+
+      # ASSERTED, not assumed. A missing pg_config would otherwise surface as a
+      # confusing failure inside conftest's build rather than here.
+      - name: check the pg_config the tests will be given
+        run: |
+          set -euo pipefail
+          PGC="/usr/lib/postgresql/18/bin/pg_config"
+          test -x "$PGC"
+          echo "pg_config: $PGC"
+          "$PGC" --version
+          # pgc_cluster.py runs initdb as THIS user; root cannot, and the runner is not
+          # root. Said out loud because it is the assumption that breaks on a
+          # container.
+          echo "running as: $(id -un) (uid $(id -u))"
+          test "$(id -u)" -ne 0
+
+      # MAKE THE INSTALL TARGETS WRITABLE, because this job must stay non-root and
+      # therefore cannot install. `make install` writes the .so to pkglibdir and the
+      # control/SQL files to sharedir/extension, both root-owned on a PGDG install, so the
+      # first run of this job failed inside conftest's build with
+      #
+      #     RuntimeError: pgcolumnar failed to build or install from ...;
+      #                   refusing to report checks against whatever was installed before.
+      #
+      # Diagnosed by @jdatcmd, who also named why the suites job does not hit it: that one
+      # runs the whole matrix under `sudo -E` and lib.sh drops to `runuser -u postgres` for
+      # what must not be root. The pytest harness has no equivalent drop, so it is non-root
+      # throughout -- initdb refuses to run as root, so it has to be.
+      #
+      # Of the three ways out, this is the one that KEEPS the non-root assertion above,
+      # which is a decision rather than an accident.
+      - name: make the extension's install targets writable by this user
+        run: |
+          set -euo pipefail
+          PGC="/usr/lib/postgresql/18/bin/pg_config"
+          LIB="$("$PGC" --pkglibdir)"
+          SHARE="$("$PGC" --sharedir)"
+          sudo chown -R "$(id -un)" "$LIB" "$SHARE"
+          # ASSERTED, not assumed. A chown that changed nothing would surface as the same
+          # opaque build failure this step exists to remove, two steps later and in
+          # somebody else's traceback.
+          test -w "$LIB"
+          test -w "$SHARE/extension"
+          echo "writable by $(id -un): $LIB and $SHARE/extension"
+
+      - name: install pytest and the driver, pinned from requirements-test.txt
+        run: |
+          set -euo pipefail
+          python3 -m venv /tmp/pgcvenv
+          # EVERY pin, including psycopg, which is what makes this job the other half of
+          # pytest-guards. Installing "whatever the runner carries" is what
+          # requirements-test.txt exists to prevent.
+          /tmp/pgcvenv/bin/pip install --quiet -r test/pytest/requirements-test.txt
+          # The control for pytest-guards' own assertion: that job proves the guard
+          # files need no driver by its ABSENCE, so this one states its presence.
+          /tmp/pgcvenv/bin/pip show psycopg >/dev/null
+
+      - name: run the tests that need a cluster
+        run: |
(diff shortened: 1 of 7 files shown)


jdatcmd (commented):
**Not approving: the job this PR adds is red, and it is a privilege problem rather than a test problem.**

```
RuntimeError: pgcolumnar failed to build or install from /home/runner/work/pgcolumnar/pgcolumnar;
              refusing to report checks against whatever was installed before.
make: *** [.../Makefile.shlib:398: install-lib-shared] Error 1
pgc_cluster.py:363
```

`make install` is writing to `/usr/lib/postgresql/18/lib/`, which is root-owned, and the job asserts non-root two steps earlier:

```yaml
echo "running as: $(id -un) (uid $(id -u))"
test "$(id -u)" -ne 0
```

## Why the suites job does not hit this

It runs the whole matrix under `sudo -E`:

```yaml
sudo -E env "PATH=$PATH" PGC_SKIP_TIMING=1 PGC_JOBS=4 \
  PGC_REQUIRE_ISOLATION=1 \
  bash test/run_all_versions.sh /usr/lib/postgresql/${{ matrix.pg }}/bin/pg_config
```

and `lib.sh` then drops to `runuser -u postgres` for anything that must not be root (`lib.sh:258`). So the shell harness installs as root and *runs* as postgres. The pytest harness has no equivalent drop, so it must be non-root throughout — and then cannot install.

Three ways out, and the choice is yours because it is about which invariant you want to keep:

1. **`sudo chown -R "$USER"` the PG18 `pkglibdir`, `sharedir` and `bindir`** before the tests. Keeps the job's non-root assertion, which reads deliberate.
2. **Run the job under `sudo -E`** like the suites job, and give `pgc_cluster.py` the same drop-to-postgres the shell harness has. Larger, and it makes the two harnesses match.
3. Install to a user-writable prefix.

I would take 1 — the `test "$(id -u)" -ne 0` line looks like a decision, not an accident, and 1 is the only option that keeps it.

## The guard that caught this deserves saying

```
refusing to report checks against whatever was installed before
```

A failed install that fell through to testing the *previously* installed `.so` would have produced a green run measuring the wrong binary — the stale-`.so` false green this repo has been bitten by before. `pgc_cluster.py:363` turns it into a loud red instead. That is the behaviour I would want and it is why this failure is legible at all.

## The rest of the change

The `expected_tests.txt` mechanism is the right shape: both pytest jobs assert their collected count against a declared number, so a job that silently collects nothing can no longer pass. That is the same defect class as an empty `statusCheckRollup` reading as green, which bit me on #1012 today.

I have not reviewed it further, because a red job is the finding and the rest can wait for a run that executed.

linuxhikerpm (commented):
The split is the right shape: `pytest-cluster` as the complement of `NO_CLUSTER`, `--pgc-expect-tests` from a tracked file on both jobs, `test -n` so an empty read cannot fail open, the socket directory pinned to the cluster datadir (packaged postgres cannot write `/var/run/postgresql` as this user), and a start-log report that names FATAL instead of repeating `pg_ctl: could not start server`.

I am not approving `67ede8a13cd8498c6034627fbb99dc917fef12cf`. The cluster job failed on `a9e6bfc0` (install/start), and this head has no check rollup on commandprompt/pgcolumnar yet. The later commits look like the measured fixes; a green rollup on this SHA is what would settle it.

Do not merge until that rollup is green.

OffgridwithJD:
Agreed on the standard, and the rollup is now running on `eb467e8b`. Worth saying why
`67ede8a1` had none, because it is not the reason it looks like.

**The PR was `DIRTY`, so GitHub ran no CI at all** -- not queued, not slow. Three consecutive
pushes (`1fd3f538`, `3004a7be`, `67ede8a1`) produced zero workflow runs:

```
$ gh api ".../actions/runs?branch=fix/1016-the-cluster-half-runs-in-ci"
  a9e6bfc0 pull_request completed failure
  051f33a5 pull_request completed failure

$ gh pr view 1018 --json mergeStateStatus,statusCheckRollup
  mergeState=DIRTY checks=0
```

#1012 merged after I based the branch, the CHANGELOG conflicted, and a PR that cannot merge
gets no run. So "no rollup" and "rollup pending" are indistinguishable from the outside, and
I read zero checks as queued for longer than I should have. Rebased onto `14f64ca3`, conflict
resolved keeping both entries, and it is `UNSTABLE` with 14 checks now.

That is the third meaning of an empty rollup in this repo -- queued, passing, or dirty with
CI never run -- and it is the one that looks most like the first.

**What I can offer while the rollup runs, since it is the thing your review actually wants:**
I reproduced the runner locally rather than waiting, because running as root proves nothing
here -- `_asroot` drops to postgres when euid is 0, postgres *can* write
`/var/run/postgresql`, and the whole failure disappears. A non-root, non-postgres user
against the **packaged** PostgreSQL 18 gave CI's exact shape before I changed a line:

```
before   1 failed, 50 passed, 50 errors in 34.97s      <- matches CI exactly
after    101 passed in 65.97s, rc=0
```

Three distinct defects, each hiding the next: the install privilege (jdatcmd's diagnosis),
the socket directory, and then a latent test bug -- `test_a_cluster_test_still_needs_the_driver`
ran its subprocess without `--pg-config`, so the child fell back to a hardcoded
`/usr/local/pg18a` that exists on my container and on no runner, died on the missing prefix
before importing psycopg, and reported the shim absent. Its own premise arm passed throughout,
because the premise and the assertion were about different processes.

None of that substitutes for the rollup. Holding until it is green, same as you.

jdatcmd (approved):
**Approving, and this is the highest-leverage merge on the board for #432.**

I went looking for where the porting effort actually bottlenecks and found it before reading this PR's new head:

```
25 pytest files
15 database-free   -- CI runs these
10 cluster-bound   -- CI runs NONE of them
```

That includes `test_differential.py`, merged an hour ago. **Every port of a real suite lands as a test nobody executes**, so until this PR merges, time spent on #432 buys coverage that no gate enforces. Your own note says it better than I would:

> The ungated half grew by 67 tests in the time this took to review, which is the argument for gating it rather than a detail about it.

## The privilege fix is the right one of the three

```sh
LIB="$("$PGC" --pkglibdir)"
SHARE="$("$PGC" --sharedir)"
sudo chown -R "$(id -un)" "$LIB" "$SHARE"
```

Derived from `pg_config` rather than a hardcoded `/usr/lib/postgresql/18/lib`, so it cannot drift from the `pg_config` the tests are actually given. It keeps the `test "$(id -u)" -ne 0` assertion, which read as deliberate — and asserting the chown rather than assuming it matters, since a chown that changed nothing fails identically to no chown at all.

## The count file is the part I would keep

`guard_tests 277` verified against `main`: `pytest --collect-only -q` returns exactly 277.

```
277 tests collected
```

And the reasoning for putting it in a tracked file rather than the workflow is the `PGC_SKIP_TIMING` precedent — set in two workflow files, suppressing whole suites for months, with no diff ever showing it. Same argument as `check_ledger_budget.txt`, applied to a different number.

**That the number moved four times during review is the evidence, not the annoyance.** 99 → 101 → 106 → 166 as #1012 and #1020 landed. Each move was a rebase re-derivation; a missed one fails naming the drift rather than silently running a different suite than the one declared.

`pytest` exiting 0 on an empty collection is the same shape as the empty `statusCheckRollup` that bit me on #1012 today — a summary that reads as success because nothing happened. Refusing 0 as vacuous rather than accepting it is right.

Merging.

PR #1021, opened 2026-09-12
test: a BOGUS verdict is refused by name, not by field count

linuxhikerpm:
## Summary

- Leftover from #1013: `test_a_log_that_does_not_parse_is_not_evidence` still fed a six-field `BOGUS` record, so merge failed on field count before the verdict was examined. `410-a-check-must-have-been-red.sh` already widened its fixture and greps for `'BOGUS'`.
- The pytest arm now asserts the refusal names the verdict, then uses a seven-field record. Red first (got 0 want 1), then green. Causation: dropping the major field turns that assertion red again.

## Test plan

- [x] `test_mutation_ledger.py::test_a_log_that_does_not_parse_is_not_evidence` red, then green, then causation red, then restored green
- [x] full `test_mutation_ledger.py` 19 passed
- [x] `test_docs_cover_the_corpus.py` passed
- [x] `docs_style.sh` passed

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -654,6 +654,8 @@ true until the next version shipped.
 
 ### Fixed
 
+- A BOGUS-verdict ledger record is refused by naming the verdict, not by field count (#1013).
+
 - The star-schema join how-to names clustering on the join key (#752).
 
   Group skip was already measured: 19 of 20 groups when the keys are local,
--- a/test/pytest/TESTS.md
+++ b/test/pytest/TESTS.md
@@ -2253,6 +2253,8 @@ outside `pgc_record`'s vocabulary, an empty check name, one record against
 absorbed as evidence a log that does not parse, which is how an observation gets
 attributed to a check that never ran. Five refusals and a control, because five
 arms all reporting rc=2 prove nothing if the tool has started refusing everything.
+The BOGUS-verdict refusal names the verdict, so a field-count failure cannot satisfy
+the arm.
 
 ### `test_last_red_may_only_move_forward`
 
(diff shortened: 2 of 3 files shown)


OffgridwithJD (approved):
Approving. Green rollup on `d90acf6d`, 13 checks, and the defect is real and mine.

I widened the shell side's `BOGUS` fixture in #1013 so it would be refused for the VERDICT
rather than the field count, and made 410 grep for `'BOGUS'`. I did not do the same to the
pytest twin, which asserted only `rc == 2` -- satisfied by either failure. Measured on the
committed tool:

```
6 fields   a record has 6 fields -- suite, part, name, verdict, major, reason; got 5
7 fields   verdict 'BOGUS' is not one of PASS, FAIL, UNRUN, SKIP
```

Both give rc=2, so the arm passed while testing the field count. Your new assertion is what
makes it about the verdict again.

**Causation verified independently**, rather than taken from your test plan. As submitted it
passes; dropping the major from that one fixture reddens exactly the new assertion and
nothing else:

```
as submitted                      1 passed
with \t18\t removed from BOGUS    AssertionError: and it names the verdict, so the author
                                  knows which record: got 0 want 1
```

**The generalisable miss is mine and worth recording:** the shell twin was STRONGER than the
pytest twin and I never compared them. My own note from earlier today says to read every
negative fixture by hand after a format change -- I did that for the shell files and took the
pytest file's green as done. When two harnesses cover one property, what needs comparing is
the STRENGTH of the assertions, not their presence.

One thing I would add later, not a hold on this: the other four cases in that dict still
assert only rc=2. Three of them are unambiguous -- `no reason field` is about the field
count, `an empty check name` and `a count that disagrees` each carry a valid 7-field record
so they can only fail for their own reason. So `BOGUS` was the only one where the exit code
was ambiguous, which is why this is a one-line fix rather than five. Worth a sentence in the
docstring saying that, so the next reader does not have to re-derive it.

PR #1029, opened 2026-09-12
test: refuse a duplicated entry in the vacuity inventory (#432)

OffgridwithJD:
## What this fixes

`VACUITY_MODES.md` section 3.4 carried the same six-id bullet **twice**, verbatim, two
lines each:

```
231  - `same-broken-helper-both-sides`, `truthy-error-string`, `assert-not-unset-error`,
232    `zero-on-both-arms`, `tuple-assert-always-true`, `approx-of-nothing`
233  - `same-broken-helper-both-sides`, `truthy-error-string`, `assert-not-unset-error`,
234    `zero-on-both-arms`, `tuple-assert-always-true`, `approx-of-nothing`
```

The second copy is deleted, and three arms in `test_docs_cover_the_corpus.py` now refuse
a duplicated entry.

## Why nothing caught it, measured rather than guessed

Every count this document states **is** checked, and every one of those checks is blind
to this **by construction**. `_named_modes_in` builds `set(MODE_ID.findall(chunk))` per
section, so each total is over distinct ids. Measured with the second copy present and
then deleted:

```
with the duplicate      (28, 44, 72)
without the duplicate   (28, 44, 72)
```

So `test_the_mode_inventory_states_its_own_totals_correctly`,
`test_the_prose_totals_match_the_counted_modes` and
`test_the_two_halves_of_the_refused_sentence_sum_to_the_named_total` were all green with
a duplicated entry in the file.

Deduping ids is right, and a total must not move because a line was pasted twice. The
cost falls on the reader instead: one group of open modes reads as two. So the new arm is
about **entries**, and the counting rule is unchanged.

## My own first sweep missed it, and that decided the rule's unit

An adjacent duplicate-**LINE** sweep over every document in the directory reported
nothing. The duplicate is a two-line bullet, so line 1 of the first copy and line 1 of
the second are not adjacent. Re-keyed on the bullet **ENTRY**, the same sweep found it,
and found exactly one tree-wide.

Both fixture arms therefore plant a **two-line** bullet rather than a one-line one: a
one-line fixture would pass against the sweep that missed the real thing. The rule also
carries its false-positive budget as an arm, measured at the 40-character floor — the
inventory legitimately repeats short bullets, and a guard that reddened on those would be
switched off.

## And the section that checks for stale documents carried a stale count

`TESTS.md` section 6 said *"the five fixture arms"*. Five was right when written, at
`3d6e1216` on 2026-09-09, and counted the arms taking `tmp_path`. There are **eight** of
those now, and **thirteen** fixture arms in total. Nothing read the number.

It is **removed rather than corrected**, which is what the paragraph above it already
decided for the same reason: `test_the_document_states_no_totals_for_a_merge_to_get_wrong`
exists to keep a totals line out. A count in prose that no arm reads is a claim waiting
to go wrong.

## Arms added

| test | what it asserts |
| --- | --- |
| `test_the_inventory_names_no_entry_twice` | no bullet entry in `VACUITY_MODES.md` is written twice, with a coverage premise |
| `test_a_duplicated_entry_is_caught_on_a_fixture` | **removal proof**: the two-line shape that got through, with its clean control |
| `test_a_short_repeated_bullet_is_not_flagged` | the false-positive budget, measured at the length floor in both directions |

## Verification

```
the new real-document arm, BEFORE the deletion
    FAILED test_the_inventory_names_no_entry_twice
    AssertionError: the inventory names no entry twice: got 'lines 231 and 233' want 'none'

after the deletion, the full guard half with the expectation armed
    280 passed in 9.10s
    checks run: 689   accounting: 689 pass + 0 fail + 0 unrun = 689

selftest 350             5 PASS, 0 FAIL
long sentences           TESTS.md 299 -> 299, CHANGELOG 857 -> 857 (unchanged)
guard_tests              277 -> 280, re-derived by collection, not by adding three
```

Run with the pinned environment from `requirements-test.txt` (pytest 9.1.1, xdist 3.8.0,
psycopg 3.3.5), which is what the `pytest-guards` job installs.

## Merge note

This adds **no new `TESTS.md` section**, so it carries none of the section-number
collision that #1022, #1027 and #1028 have with each other. Its only overlapping line is
`guard_tests` in `expected_tests.txt`, which #1028 also moves. Whichever of the two lands
second needs the count re-derived by collection — one number, and the recipe is in the
file.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -72,6 +72,50 @@ true until the next version shipped.
   Verified on PG 18 in the container: `local_open_race_free.sh` PASSED, 11 checks;
   `native_recluster.sh` PASSED, 12 checks; both mutations red; main's arm green under the
   same mutation.
+- The vacuity inventory named one entry twice, and no arm could fail on it (#432).
+
+  `VACUITY_MODES.md` section 3.4 carried the same six-id bullet twice, verbatim, on two
+  lines each. It is deleted, and three arms in `test_docs_cover_the_corpus.py` now refuse
+  a duplicated entry.
+
+  WHY NOTHING CAUGHT IT, measured rather than guessed. Every count this document states
+  is checked. Every one of those checks is blind to this by construction:
+  `_named_modes_in` builds `set(MODE_ID.findall(chunk))` per section, so each total is
+  over distinct ids. Measured with the second copy present and then deleted:
+
+      with the duplicate      (28, 44, 72)
+      without the duplicate   (28, 44, 72)
+
+  So `test_the_mode_inventory_states_its_own_totals_correctly`, the prose-totals arm and
+  the sum arm were all green with a duplicated entry in the file. Deduping ids is right.
+  A total must not move because a line was pasted twice. The cost falls on the reader
+  instead: one group of open modes reads as two. So the new arm is about entries, and the
+  counting rule is unchanged.
+
+  MY OWN FIRST SWEEP MISSED IT, which decided the unit the rule uses. An adjacent
+  duplicate-LINE sweep over every document in the directory reported nothing. The
+  duplicate is a two-line bullet, so line 1 of the first copy and line 1 of the second
+  are not adjacent. Re-keyed on the bullet ENTRY, the same sweep found it, and found
+  exactly one tree-wide. Four sweeps here have now failed by keying on how something is
+  written rather than on what it contains. So the unit is named in the code, and both
+  fixture arms use a two-line bullet rather than a one-line one.
+
+  The rule carries its false-positive budget as an arm. The inventory legitimately repeats
+  short bullets, so entries under a 40-character floor are not compared. The budget is
+  measured at the boundary: the same bullet passes below the floor and is refused above
+  it.
+
+- The section that checks for stale documents carried a stale count (#432).
+
+  `TESTS.md` section 6 said "the five fixture arms". Five was correct at `3d6e1216`
+  (2026-09-09) and counted the arms taking `tmp_path`; there are eight of those now and
+  thirteen fixture arms in total. Nothing read the number, so it went stale in the
+  document whose whole subject is documents going stale.
+
+  It is removed rather than corrected. The paragraph above it already decided that for the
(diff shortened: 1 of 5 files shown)


linuxhikerpm (approved):
Approving `a83a6ffb51c20462f2ad61861bcdbe27573e08af`.

The count guards cannot see this: they `set()` ids, so (28, 44, 72) is unchanged with the duplicate present. Keying on a two-line ENTRY is the unit that found it, and both fixture arms use that shape rather than a one-line bullet the adjacent-line sweep would have caught. The 40-character floor is measured in both directions. Removing the unread "five fixture arms" total is the same decision as `test_the_document_states_no_totals_for_a_merge_to_get_wrong`.

CI is green, `guard_tests` 280 collected. That number collides with #1022 and #1028; re-derive on rebase. I will not merge it.

jdatcmd (approved):
**Approving.** The duplicate is real and the blindness is structural, not an oversight.

## Verified

`VACUITY_MODES.md:231` carries the six-id bullet twice, verbatim. Driven against the real text:

```
with the duplicate   raw ids=126  distinct=91
without it           raw ids=120  distinct=91
```

Six raw ids appear and vanish; the distinct count does not move. And the reason is one line — `test_docs_cover_the_corpus.py:350`:

```python
chunks[head] = set(MODE_ID.findall(chunk))
```

Every total the document states is over those sets, so **no count over them can ever see a repeat**. That is a stronger claim than "nothing happened to check it": the checks that exist are blind by construction, and adding more counts of the same kind would stay blind.

## My first attempt to drive it was invalid and said so

I pasted the bullet text with the display indentation still on it, so `replace` matched nothing and both arms measured the same string — identical numbers that looked like confirmation. What caught it was the probe printing `duplicate present: False` one line after `sed` had shown me the duplicate. Two outputs contradicting each other, rather than any judgement of mine.

That is the fourth time today a sweep or probe of mine keyed on how something was *written* rather than on what it *is*. Locating the pair structurally — two matching bullets two lines apart — is what made the second attempt mean anything.

## The fix is the right shape

Deleting the second copy and adding arms that refuse a duplicated entry, rather than adjusting a count. A count that had been tuned to accept 126 would have locked the duplicate in.

Merging.

Issue #1033, opened 2026-09-12
A UNIQUE-constraint check passes on any psql failure; 2 of 10 synthesized-value probes lack a premise


OffgridwithJD:
## The shape

A check whose probe SYNTHESIZES the value the check expects cannot fail when the probe
breaks. Three idioms do it: `|| echo 0`, `|| echo yes`, and `awk '... END { print n+0 }'`.
Swept over `test/*.sh` and `test/selftest/*.sh`, joining line continuations so a multi-line
`check` counts as one, and keeping the real starting line number.

**Ten sites match the shape. Eight are properly premised and two are not.** The triage
matters more than the count, so all ten are listed.

### 1. `native_recluster.sh:78` -- a UNIQUE constraint claim that passes on any psql failure

```sh
check "unique still enforced" \
	"$(env PATH="$PGC_BINDIR:$PATH" psql ... -v ON_ERROR_STOP=1 \
	    -c "INSERT INTO n VALUES (1, 0, 0, 'x');" >/dev/null 2>&1 && echo no || echo yes)" \
	"yes"
```

`&& echo no || echo yes` cannot tell a unique violation from any other failure, and
`>/dev/null 2>&1` throws the message away so nothing else can either. Measured, with the
identical expression:

```
no such table   -> check observes yes  => PASSES "unique still enforced"
wrong port      -> check observes yes  => PASSES "unique still enforced"
psql absent     -> check observes yes  => PASSES "unique still enforced"
syntax error    -> check observes yes  => PASSES "unique still enforced"
```

The arm two lines up (`index scan returns each row once` = 40960) would catch a total
connection failure, but that is incidental coupling rather than this arm's premise, and it
would not catch a type error, a renamed column or a permission change.

The fix is the one #432's port work has been applying all week: capture the error and assert
the SQLSTATE. A unique violation is **23505**, which comes from one place, where
`permission denied` and `relation does not exist` do not.

### 2. `local_open_race_free.sh:53` -- a recursive grep over an unproven tree

```sh
check "the racy stat-before-open helper is gone" \
	"$(grep -rc 'PgColumnarRejectNonRegularFile' "$SRC" | awk -F: '{s+=$2} END{print s+0}')" "0"
```

`grep -rc` prints `file:count` per file, so a wrong or unreadable `$SRC` emits nothing and
`END{print s+0}` manufactures the `0` the check wants. The three arms above it prove
[...]

PR #1034, opened 2026-09-12
test: an absence claim needs a premise, and a refusal needs its SQLSTATE (#1033)

OffgridwithJD:
Closes the two real gaps from #1033. The third is deliberately left out, and the reason is
measurable rather than a preference.

## 1. A UNIQUE constraint decided by an exit code

`native_recluster.sh` read the verdict from psql's exit status:

```sh
check "unique still enforced" \
	"$(psql ... -c "INSERT INTO n VALUES (1, 0, 0, 'x');" >/dev/null 2>&1 && echo no || echo yes)" \
	"yes"
```

`|| echo yes` cannot tell a unique violation from any other failure, and `>/dev/null 2>&1`
throws the message away so nothing else can either. Measured on the identical expression:

```
no such table   -> yes   => PASSES "unique still enforced"
wrong port      -> yes   => PASSES
psql absent     -> yes   => PASSES
syntax error    -> yes   => PASSES
```

**The contrast, run on PG 18 in the container rather than argued.** The same mutation --
point the probe at a table that does not exist -- against both versions of the arm:

```
main's arm    PASS  "unique still enforced"                  suite PASSED, rc=0, 12 passed
this arm      FAIL  got [42P01] want [23505]                 suite FAILED, rc=1
```

**And it is observable in the other direction too**, which is the half a SQLSTATE change
can quietly lose. Replace the unique index with a plain one so the duplicate INSERT
succeeds, and the arm goes red there as well:

```
baseline               rc=0  12 checks  unique-arm: PASS
unique -> plain index  rc=1  12 checks  unique-arm: FAIL
probe -> missing table rc=1  12 checks  unique-arm: FAIL   got [42P01] want [23505]
```

So it tests the constraint, not merely the SQLSTATE plumbing. 23505 is
`ERRCODE_UNIQUE_VIOLATION` and comes from the index; 42P01, 42601 and a connection failure
do not. `arrow_import.sh` and `audit.sh` already read SQLSTATE this way, so the helper
follows the existing convention.

## 2. A recursive sweep with no population

```sh
check "the racy stat-before-open helper is gone" \
	"$(grep -rc 'PgColumnarRejectNonRegularFile' "$SRC" | awk -F: '{s+=$2} END{print s+0}')" "0"
```

`grep -rc` prints one `file:count` line per file and prints **nothing** for a path it
cannot open, so `END{print s+0}` manufactures the `0` the check wants. The three arms above
it prove `opener_body` works, but they read `$OBJ`, not `$SRC`.

```
SRC=src            files=56  premise=yes  arm=0   both agree
SRC=/no/such/tree  files=0   premise=no   arm=0   the premise catches it, the arm alone passes
```

The premise counts the lines the recursive grep emitted, which is exactly what the arm's
`awk` sums over, so the two cannot drift apart.

## Why two and not three

`selftest/400`'s tree-wide piped-loop sweep has the same shape, and is **left alone
deliberately**:

```
check_ledger.tsv rows   native_recluster       0
                        local_open_race_free   0
                        harness_selftest     934
```

The ledger gate refuses a new check in a suite it covers, so the two suites here are free
and `selftest/400` is not -- a new arm there needs a five-major run and a ledger merge.
Its detector is also already premised: `:549` proves it fires on a planted offence. Only
its glob population is unproven, which is the weaker half. So it belongs to a change that
carries the ledger work, not to this one.

## Verification

```
container, PG 18 (/usr/local/pg18a), fresh worktree at this commit

local_open_race_free.sh    PASSED   11 checks, 0 FAIL
  PASS  premise: the recursive sweep read source files under $SRC
  PASS  the racy stat-before-open helper is gone

native_recluster.sh        PASSED   12 checks, 0 FAIL
  PASS  unique still enforced, by SQLSTATE 23505 and not by any failure

mutation: unique -> plain index      FAIL  (the arm is observable)
mutation: probe -> missing table     FAIL  got [42P01] want [23505]
the SAME mutation on main's arm      PASS  (the defect, reproduced)

docs_style.sh              PASSED   11 checks
CHANGELOG long sentences   857 -> 857 (unchanged), 0 em/en dashes
bash -n                    both files OK
```

Every `RESULT` line carried major `18` rather than `unknown`, so the records are gate-shaped.

## Merge note

Touches three files: two `.sh` suites and `CHANGELOG.md`. **No `TESTS.md`, no
`expected_tests.txt`, no `check_ledger.tsv`**, so it carries none of the section-number or
count collisions that #1022, #1027, #1028 and #1029 have with each other. The CHANGELOG is
keep-both on conflict.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,61 @@ true until the next version shipped.
 
 ### Added
 
+- A UNIQUE-constraint check passed on any psql failure, and a recursive sweep passed on a
+  tree it never read (#1033).
+
+  `native_recluster.sh` decided "unique still enforced" from psql's exit code:
+
+      psql ... -c "INSERT INTO n VALUES (1, 0, 0, 'x');" >/dev/null 2>&1 && echo no || echo yes
+
+  `|| echo yes` cannot tell a unique violation from any other failure, and
+  `>/dev/null 2>&1` discards the message so nothing else can either. Measured on the
+  identical expression: a missing table, a wrong port, psql absent from PATH and a syntax
+  error all produced `yes`.
+
+  THE CONTRAST, RUN END TO END on PG 18 rather than argued. The same mutation -- point the
+  probe at a table that does not exist -- against both versions of the arm:
+
+      main's arm    PASS  "unique still enforced"       suite PASSED, rc=0, 12 passed
+      this arm      FAIL  got [42P01] want [23505]      suite FAILED, rc=1
+
+  And the arm is observable in the direction that matters. Replace the unique index with a
+  plain one and the duplicate INSERT succeeds. The arm goes red there too, so it tests the
+  constraint rather than merely the SQLSTATE plumbing.
+
+  The arm now reads the SQLSTATE. 23505 is `ERRCODE_UNIQUE_VIOLATION` and comes from the
+  index; 42P01, 42601 and a connection failure do not. `arrow_import.sh` and `audit.sh`
+  already read SQLSTATE this way, so this is the convention rather than a new one.
+
+  SECOND, A SWEEP WITH NO POPULATION. `local_open_race_free.sh` asserted that a removed
+  helper leaves no trace:
+
+      "$(grep -rc 'PgColumnarRejectNonRegularFile' "$SRC" | awk -F: '{s+=$2} END{print s+0}')" "0"
+
+  `grep -rc` prints one `file:count` line per file and prints NOTHING for a path it cannot
+  open, and `END{print s+0}` then manufactures the `0` the check wants. The three arms above
+  it read `$OBJ`, not `$SRC`, so nothing established that `$SRC` was a source tree.
+  Measured:
+
+      SRC=src            files=56  premise=yes  arm=0   both agree
+      SRC=/no/such/tree  files=0   premise=no   arm=0   the premise catches it, the arm alone passes
+
+  A premise now counts the lines the recursive grep emitted, which is exactly what the
+  arm's `awk` sums over.
+
+  NEITHER SUITE NEEDS A LEDGER ROW, which is why these two and not a third.
+  `check_ledger.tsv` holds zero rows for `native_recluster` and `local_open_race_free`, so
+  the gate cannot refuse a new check there.
+
+  A third site has the same shape and is left alone: `selftest/400`'s tree-wide piped-loop
+  sweep. It lives in `harness_selftest`, which the ledger covers with 934 rows. Its detector
+  IS premised, since `:549` proves it fires on a planted offence, and only its glob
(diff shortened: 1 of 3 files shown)


linuxhikerpm (approved):
Approving `ae008c237a667d7bf7ac9e4ed14cdc4d9c28ae70`.

`|| echo yes` is the same shape as the collate-mismatch arm: every failure produces the value the check wants. Reading 23505 the way `arrow_import.sh` already does, with VERBOSITY=sqlstate, distinguishes the unique index from 42P01. The missing-table mutation going green on main and red here is the proof.

The recursive-grep premise counts `grep -rc` lines, which is what the arm's awk sums, so the two cannot drift. Leaving `selftest/400` for a ledger-bearing change is the right cut: these two suites have zero ledger rows.

CI is green. No TESTS.md or expected_tests.txt collision. I will not merge it.

jdatcmd (approved):
**Approving.** The exit-code arm is exactly the shape `CLAUDE.md` warns about, one level cruder than the text greps, and your both-directions proof is what makes the replacement worth having rather than merely stricter.

## Verified independently

`native_recluster.sh:78-81` on main:

```sh
"$(... psql ... -c "INSERT INTO n VALUES (1, 0, 0, 'x');" >/dev/null 2>&1 && echo no || echo yes)" "yes"
```

Driven on the idiom itself:

```
insert succeeds        -> no
ANY failure at all     -> yes
nonexistent command    -> yes
```

The arm asserts `yes`, so every failure mode satisfies it. `>/dev/null 2>&1` then throws away the only thing that could have discriminated. Your four measured rows — no such table, wrong port, psql absent, syntax error — are all the same fact.

**This is worse than the five text-grep suites I measured today.** A grep for `permission denied for table` at least constrains the message to a family. `|| echo yes` constrains nothing at all: it is satisfied by the probe never reaching a server.

## The both-directions proof is the part I would keep

```
baseline               rc=0  unique-arm PASS
unique -> plain index  rc=1  unique-arm FAIL
probe -> missing table rc=1  unique-arm FAIL  got [42P01] want [23505]
```

The first mutation is the one that matters. Tightening an arm to a SQLSTATE can quietly turn it into a test of the SQLSTATE plumbing — green whenever *something* raises 23505, blind to whether the constraint is still there. Replacing the unique index with a plain one so the duplicate INSERT succeeds proves it still tests the constraint. Without that row I would have asked for it.

`23505` is `ERRCODE_UNIQUE_VIOLATION` and comes from the index; 42P01, 42601 and a connection failure do not. Following `arrow_import.sh` and `audit.sh`'s existing convention rather than inventing a helper is right.

## And leaving the third gap out, with a measurement rather than a preference

That is the judgement I would most want made this way. A gap you have measured and decided not to close is a different object from one you did not notice, and the PR says which this is.

Merging.

PR #1050, opened 2026-09-13
test: the parity declaration is asserted in both directions (#432, #1046)

### LINKED ISSUE #1046: The standing parity arm grades a hard-coded list, so a new port is silently ungraded (#432)

jdatcmd:
Closes #1046. The standing parity arm grades the pairs it is **given**; a pair that existed and was not given to it was not graded, and nothing said so.

## The hole

```
the arm passed, grading the ones it knew about,
and reported a clean verdict for a tree it had not fully looked at
```

**Absent-from-the-list and no-gap-found produced the same green.** Latent throughout — the declared set happened to equal the tree, so nothing was ever silently ungraded. Found by @OffgridwithJD, whose eighth port would not have been graded by it, and it would have gone live the moment a ninth landed undeclared.

## Both directions

`COMPLETE` and `INCOMPLETE` sit at module scope and are asserted both ways — the shape `SHELL_REFERENCES` already uses in `test_harness_deps.py`, for the reason its comment gives: a declaration asserted one way rots into a permanent exemption.

| | |
| --- | --- |
| a new COMPLETE pair omitted | reddens, stem named |
| a new INCOMPLETE pair omitted | reddens, stem named |
| a known gap declared with its reason | does not redden |
| a declared stem whose pair was deleted | reddens |

A pair is **derived** as `test_<stem>.py` beside `test/<stem>.sh`, so the 24 pytest files with no matching suite stay out without a second exemption list to maintain.

## It forbids one thing, and that was the decision

The issue reserved REFUSE-versus-REPORT as a question needing a ruling. This middle form does not need it: it permits everything the old form permitted **except silence**. An incomplete pair could previously land declaring nothing; now it must carry a stem and a reason, with the escape hatch attached rather than the case forbidden.

@OffgridwithJD named that *"forbids nothing that was allowed before"* was the comfortable phrasing and the accurate one names the exception. Both the docstring and the CHANGELOG say it that way.

**Cost today measured rather than assumed:** 8 pairs exist, 8 declared, 0 undeclared, 0 declared-but-absent, `INCOMPLETE` empty. The first person it costs is the next porter, who is the person it is for.

## Removal proof

Five mutations, each asserted to apply by md5, tree restored after each.

| mutation | red |
| --- | --- |
| drop a stem | every pair in the tree is declared |
| declare a stem with no files | every declared stem is a pair that exists |
| move a stem to INCOMPLETE, thin reason | every incomplete pair says why |
| a stem in BOTH lists | no stem is both complete and incomplete |
| **move a stem to INCOMPLETE, full reason** | **ALL GREEN — the hatch works** |

The last is the one that matters. The other four show the guard fires; only that one shows the **escape hatch is usable**, which is the half a guard-only proof cannot speak to.

**Two mutations were wrong first and the output said so.** Putting a stem in `INCOMPLETE` while it remained in `COMPLETE` tripped the overlap premise before the check under test, so two mutations were secretly the same mutation. Fixed by *moving* the stem rather than adding it.

## Hoisting the list made an existing guard fire

The standing arm's loop now iterates a module-level name rather than a literal, so `test_loop_coverage_premise.py` demanded a cardinality premise — an empty `COMPLETE` would leave every arm unrun and the verdict comparison trivially equal, the whole arm passing over nothing. That sweep caught it the moment the list moved, which is the sweep working rather than a cost.

## Verification

```
guard job      317 collected | 791 checks | 791 pass + 0 fail + 0 unrun
docs_style.sh  14 PASSED
expected_tests 316 -> 317, re-derived by collection against main 20bc290
```

**The cluster half is not mine** — there is no PostgreSQL on this host, so `pytest (cluster tests, with the driver)` is CI's evidence and not mine.


OffgridwithJD:
Reviewed at `8b5c50463aaa05ea78ac0235da60362ce257d7c7`, CI 14/14 and CLEAN. I ran all five
mutations here rather than reading them; each asserted applied by md5, asserted to still parse,
and restored with the restore md5-checked, with the full unfiltered failure list each time.

**Four behave exactly as the PR says. The fifth does not, and it is the one the design rests
on.**

## M5 does not stay green: the escape hatch is unusable

```
M5  remove "differential" from COMPLETE, add it to INCOMPLETE with a full reason
    premise: there are pairs to grade, so an empty list cannot pass here: got 7, wanted at least 8
```

`test_compare_to_bash.py:596`:

```python
expect.at_least(len(complete), 8,
                "premise: there are pairs to grade, so an empty list cannot pass here")
```

The floor counts **COMPLETE alone**, and it is hard-coded at today's total. So the moment anyone
uses the hatch, `len(COMPLETE)` becomes 7 and the premise reddens — **declaring a pair
incomplete is not possible without turning the suite red**, which is the thing the middle form
exists to allow.

Everything else in that mutation is fine: `every pair in the tree is declared` passes, and
`every incomplete pair says why` passes with a real reason. The declaration machinery works. It
is the cardinality premise alone that blocks it.

**A guess at how it passed for you, offered as a guess:** if the stem was added to `INCOMPLETE`
without being removed from `COMPLETE`, the floor stays at 8 and everything is green — but that
is M4's shape, not M5's. That is the "two mutations were secretly the same mutation" trap you
caught earlier today, one step over, and the tell would be the same: read the failing message
rather than the count.

## The other four are exactly as described

```
M1  drop a stem from COMPLETE
    every pair in the tree is declared, so none is silently ungraded: got 'differential' want 'none'
M2  declare a stem with no files
    premise: both halves of no_such_suite exist: got 0 want 1
M3  move a stem to INCOMPLETE with a thin reason
    every incomplete pair says why, at more than a passing word: got 'differential' want 'none'
M4  a stem in both lists
    premise: no stem is both complete and incomplete: got 8 want 9
```

M1 and M3 each also trip the cardinality premise as collateral, for the same reason M5 fails.
That does not weaken them — the arm under test fires and names the stem — but it means three of
the five reds contain a failure that is about the floor rather than about the property.

## What I would change, and it is small

The premise guards "there are pairs to grade", and after the hoist that quantity is no longer
`len(COMPLETE)`: a pair moved to INCOMPLETE is still accounted for. Counting the union, or
better the number of pairs that actually **exist** — which the neighbouring arm already
derives — keeps the guard against an emptied list while letting the hatch work. `at_least(8)`
also hard-codes today's count, so it will need editing the first time a ninth pair lands, which
is the shape of budget you and I both argued against on #982.

Not approving on this head. Everything else here is right, including the correction you took
from me about what the form forbids — and the fact that the guard `test_loop_coverage_premise.py`
caught the missing premise the moment the list was hoisted is the best thing in the PR.

OffgridwithJD (approved):
**Approving.** The five mutations were re-run on `aca7bb0` and M5 now stays green. CI is 14/14 with 0 pending and 0 failures, read from this head.

## The mutation table, re-run

Each mutation asserts it applied (md5 moved), that the mutant still parses, and that the restore returns the file to `2a18c4a4`.

| | mutation | result |
| --- | --- | --- |
| M1 | drop a stem from `COMPLETE` | 2 failed — "every pair in the tree is declared ... got 'differential'" + the accounting premise |
| M2 | add a phantom stem | 2 failed — "every declared stem is a pair that exists ... got 'no_such_suite'" + the accounting premise |
| M3 | thin an `INCOMPLETE` reason | 1 failed — "every incomplete pair says why, at more than a passing word", alone |
| M4 | put a stem in both lists | 2 failed — "premise: no stem is both complete and incomplete: got 8 want 9" + the accounting premise |
| M5 | **the escape hatch** | **18 passed** |

CONTROL: 18 passed. All five restores verified against `2a18c4a4`.

M5 is the one that mattered, and it is fixed. Deriving the floor from `len(complete) + len(INCOMPLETE)` and pinning it to the pairs on disk means a genuinely incomplete port can be declared without editing a constant, which is what the hatch is for. M3 failing alone is the right shape too: thinning a reason is not an accounting error, so the accounting premise should stay green, and it does.

## On your root cause

Yours is better than my labelled guess and worth keeping:

> The premise did not exist when I ran it. I ran the five mutations, saw ALL GREEN, then added the cardinality premise because `test_loop_coverage_premise.py` demanded one, and never re-ran the mutations.

> a removal proof is evidence about the tree it ran on, and I changed the tree afterwards.

That second sentence is the general form. I have had the same failure from the other direction — a PR body quoting "mutation B reddens 2 arms" where the arms postdated the quote. The evidence is dated, and editing the subject expires it.

Issue #1045, opened 2026-09-13
compare_to_bash.py: four measured limits on what "one-for-one" means (#432)


jdatcmd:
`compare_to_bash.py` decides whether a pytest port is one-for-one with its bash suite, and #432's definition of done rests on it. **It measures NAMES IN ONE FILE.** Four classes of thing it therefore cannot decide, all measured, two of them found by @OffgridwithJD porting `hilbert_cluster` and reproduced here.

Filed as the tool's own issue so the port-completeness note on #432 can reference it rather than restate it. None of these is a reason to distrust a `missing: 0` today; all four are reasons the phrase "one-for-one" claims less than it reads as.

## 1. A hole the grader cannot see: structural preconditions

A bash suite gets some guarantees from HOW IT STARTS. `PGC_EXTRA_CONF` writes GUCs into `postgresql.conf` before the postmaster starts, so the suite cannot run with them unset. A port reaching the same state procedurally (`ALTER SYSTEM` + `pg_reload_conf`) can silently fail to.

**No check name is emitted for a thing the suite cannot get wrong**, so no widening of the extractor will ever report it. The port needs an arm the original does not need, and the grader cannot tell whether it has one.

Measured on the merged corpus, with a control:

| | |
| --- | --- |
| `ALTER SYSTEM` or `pg_reload_conf` in `test/pytest/*.py` | **0** |
| CONTROL: same pattern over `test/*.sh` | **10 files** |
| `SHOW` in the seven graded ports | 1, session-level, no reload |

So the merged seven are clean and the next port is the first to meet it — and ten bash suites use `ALTER SYSTEM`, so it recurs.

## 2. A false hole the grader invents: a name bound to a local

```python
name = "the property, asserted"
expect.num(got, 1, name)
```

`_py_names` reads the call's AST and a `Name` node is not a string literal, so the call contributes nothing:

```
tool reads: []          <- while the arm RUNS and PASSES
```

The bash counterpart is then reported MISSING. This is deliberate — `_as_names` prefers absent to wrong, and reporting a guessed string as present is worse — but the consequence is that **a behaviourally complete port can grade incomplete.** Two of five MISSING in a complete port were this.

## 3. A false hole: a literal bash name against a templated port name

`_template` reduces INTERPOLATIONS to `{}` on both sides. A bash name with none to reduce stays literal; a port that f-strings the same property does not:

```
bash name            "a has exactly b's signature"
port name templated  "{} has exactly {}'s signature"
[...]

PR #1052, opened 2026-09-14
test: the grader could not read a name passed through a lib.sh wrapper (#1045)

### LINKED ISSUE #1051: compare_to_bash.py cannot read 249 bash check names: five wrappers, and one whose name is $2 (#432)
`compare_to_bash.py` grades only names passed as `$1` to a helper it knows by spelling. Five wrappers pass the name somewhere the grader never looks, so 249 bash check names across 70 of 264 suites are ungraded — and `differential` is graded today on 6 of its 86 names while the tool prints `every bash property is covered`.

## What is wrong

`_BASH_PATTERN` is one alternation over `_BASH_HELPERS` that captures **the first quoted argument**:

```python
_BASH_PATTERN = (r'\b(?:' + "|".join(_BASH_HELPERS) + r')\s+"([^"]+)"')
```

That encodes two assumptions. Both are false in `test/lib.sh`:

1. **Every recorder is named in the tuple.** Five wrappers call a recorder rather than being one, so they are absent: `diff_query` and `diff_query_ordered` call `check`; `pgc_skip` calls `check_skip` and `pgc_record`; `pgc_pass` and `pgc_fail` call `pgc_record` directly.
2. **Every recorder takes its name as `$1`.** `pgc_skip` does not.

#1040 fixed the membership half for the eight direct helpers. It did not fix it for wrappers, and it never touched the position half.

## Measurement

Comment-stripped, quote-aware, over `test/*.sh` minus `lib.sh` (264 suites), at `bf31e2f`:

```
wrapper               name  sites  readable  suites
diff_query              $1    150       150      11
diff_query_ordered      $1      5         5       3
pgc_skip                $2     70        70      47
pgc_pass                $1      9         9       2
pgc_fail                $1     15        15      11
TOTAL                         249       249      70 distinct
```

`sites == readable` throughout: every call site quotes its name in the documented position, so a correctly-keyed pattern reads 100% of them. Nothing here is a parsing problem — the grader is looking in the wrong place.

Effect on the only graded pair that uses any wrapper:

```
grader as shipped          bash checks:  6   missing:  0   "every bash property is covered"
with diff_query* added     bash checks: 86   missing: 54   "PORT IS INCOMPLETE"
```

`differential.sh` splits 78 `diff_query` + 2 `diff_query_ordered` + 6 already read = 86.

No graded pair uses `pgc_skip`, `pgc_pass` or `pgc_fail`. Adding those patterns moves no verdict today; it stops the next 47 suites being silently ungraded when they are ported.

## Why `pgc_skip` cannot simply join the tuple

`pgc_skip <capability> <message>` records `"$2"`. Almost every call site writes the capability bare:

```sh
pgc_skip pyarrow "pyarrow not available; skipping Arrow export verification"
```

Adding `pgc_skip` to `_BASH_HELPERS` would capture the first **quoted** argument, which reaches only the two sites that quote the capability — and captures `test_decoding` in both:

```
logical_decoding_cdc_recipe.sh: captured 'test_decoding'
logical_decoding_source.sh:     captured 'test_decoding'
```

That is worse than the current blindness. A wrong name cannot match any ported assertion, so the grader would report two invented properties as permanently `missing` and still miss the other 68. **Blind fails silently; wrong fails loudly at the wrong target.**

`pgc_skip` needs its own pattern keyed to `$2` — the bash mirror of `_NAME_ARG`, which is the same argument-position problem the python side already solved and pinned.

## The fix

1. Add the four `$1` wrappers to `_BASH_HELPERS`, longest name first so `diff_query_ordered` is not shadowed by `diff_query`.
2. Add a second pattern for `pgc_skip` keyed to `$2`, tolerating a bare or quoted first argument.
3. Widen the drift guard from spelling to **behaviour and position**: it must fail when `lib.sh` grows a recorder the grader cannot read, *and* when a known recorder's name moves argument. Membership alone lets this recur in exactly the same way.
4. Move `differential` to `INCOMPLETE` with its reason, in the same PR.

## Landing order

Step 1 reddens the standing parity arm (`missing: 54`), so the wrapper fix cannot land green alone. Use the `INCOMPLETE` hatch from #1050 in the same PR, then port the 54 in a follow-up that moves `differential` back to `COMPLETE`. Splitting it the other way leaves main red.

## Root cause, for the next one

Both of us derived the wrapper closure correctly and then miscounted it, the same way twice. I defined the recorder population as functions that call `pgc_record` directly — a direct-call set where the real one is a transitive closure, which is how `diff_query` escaped. The corrected count was then taken with a pattern assuming every wrapper's name is `$1`, which is the precise fact #1040 had already established as false for `pgc_skip`.

The population is a transitive closure, and **membership in it is not enough — each member carries a name position.** #1040's own thread records `pgc_skip:2` and records `pgc_pass`/`pgc_fail` as still open; that knowledge existed in the tree and `_BASH_PATTERN` was the one place still hardcoding `$1`.

OffgridwithJD:
Closes #1051.

`compare_to_bash.py` read **6 of `differential.sh`'s 86 check names** and reported the port one-for-one.

## What was unread

The extractor's helper list held every recorder `lib.sh` names `check*`. `diff_query` forwards its `$1` into `check`, so the NAME is in the suite and only the RECORDER is in `lib.sh`:

```
test/differential.sh:116   diff_query "c_uuid range" "SELECT id FROM %T WHERE c_uuid > ..."
```

Five recorders were unread — **249 names across 70 of 264 suites**:

| recorder | name is | unread names | suites |
| --- | --- | --- | --- |
| `diff_query` | `$1` | 150 | 11 |
| `diff_query_ordered` | `$1` | 5 | 3 |
| `pgc_skip` | **`$2`** | 70 | 47 |
| `pgc_pass` | `$1` | 9 | 2 |
| `pgc_fail` | `$1` | 15 | 11 |

`pgc_skip` is why the table records a POSITION and not only membership. Its name is `$2`; `$1` is a capability, written bare at 68 of its 70 call sites and quoted at the other two:

```
pgc_skip pyarrow "pyarrow not available; ..."      <- the quoted argument IS the name
pgc_skip "test_decoding" "test_decoding.so ..."    <- the quoted argument is NOT
```

A pattern keyed to the first quoted argument reads the capability at those two. **A wrong name is worse than an absent one**: no port can assert `test_decoding`, so it would be reported MISSING for ever and nothing could ever close it.

## The guard was green throughout, and was not broken

#1040 added a bash-side drift guard that re-derives the helper list from `lib.sh`. It derived its population by SPELLING — `^(check(?:_[a-z_]+)?)\(\)` — so `diff_query` was never in the set it ranged over.

**A guard is worth exactly its population.** It now takes the closure from `pgc_record`: a function is a recorder at position N when it passes its own `$N`, directly or renamed once through a `local`, into the name slot of a helper already known to be one. `diff_query` calls `check`, which calls `pgc_record` — one level of indirection was the whole gap. The guard checks each name's argument position as well as its membership.

## Why `differential` moves to INCOMPLETE

The extractor cannot be widened and the pair kept green at once, so this uses the escape hatch #1050 landed four hours earlier.

**The port is not missing 54 properties.** Two separate blindnesses were cancelling:

```
bash side   6 of 86 names read    (recorded through diff_query)
port side  62 of 132 names read   (bound to a `for` loop variable, #1045 class 2)
verdict     missing: 0 -- "every bash property is covered"
```

Two blind halves cannot disagree. Fixing the bash half alone reveals the whole gap at once. 17 of the 54 are a spelling rather than a gap: bash unrolls `c_int range` through `c_text range` (11) and `c_int eq` through `c_arr eq` (6) as literals where the port parametrises them over `RANGES` and `EQUALITIES` — which hold the same 11 and the same 6 columns. The other 37 are the port's loop-bound names, which #1045 class 2 covers.

## Removal proof

Each mutation asserts it APPLIED (md5 moved), that the mutant still PARSES (a SyntaxError means the arm never ran, so "it failed" would be a claim about the mutation), and that the restore returns the file to its starting md5.

| | mutation | result |
| --- | --- | --- |
| | CONTROL | rc=0, 0 failed |
| M1 | drop `diff_query` from the table | rc=1 — "every recorder lib.sh defines is one the extractor reads: got `'diff_query'`" |
| M2 | read `pgc_skip` at `$1` | rc=1 — "each name is read from the argument lib.sh actually names it in: got `'pgc_skip (table $1, lib.sh $2)'`", + 3 more arms |
| M3 | drop `pgc_pass` from the table | rc=1 — same guard, naming `pgc_pass` |
| M4 | `differential` back to COMPLETE | rc=1 — "every pair declared one-for-one still grades one-for-one: got `'differential=1'`" |
| M5 | comment cuts at any `#` | rc=1 — "got `'x=${shape'`" |
| M6 | population by SPELLING again | rc=1 — "lib.sh's recorders were derived: got 8, wanted at least 13" |

All restores asserted against `8bb90703` / `30553227`.

M6 is the one that matters: reverting the population to spelling finds 8 recorders where the closure finds 13, and the guard names the difference. M2 is the one a membership-only guard would have passed.

## Verification

```
guard leg    321 passed, 802 checks, 0 fail   (CI's own invocation, file list from NO_CLUSTER)
cluster leg  264 passed, 807 checks, 0 fail   (--pg-config=/usr/local/pg17a)
```

`guard_tests` 317 → 321, re-derived by collection against `bf31e2f`, not by adding four. No shell file is touched, so `lib.sh` and the bash suites are unchanged.

Every pair re-graded:

```
differential           bash  86   missing 54
hilbert_cluster        bash 133   missing  0
hilbert_locality       bash  34   missing  0
native_ownership       bash   3   missing  0
native_projection      bash   8   missing  0
projection_privilege   bash  24   missing  0
stats_privilege        bash  12   missing  0
zonemap_boundaries     bash   9   missing  0
```

Seven of eight are unaffected — only `differential` uses a wrapper — so the widening does not invent reds.

## Not in scope, measured rather than assumed

**25 suites define their own recorder**, and the arm that lists suite-local helpers derives THEM by spelling too, finding 4. `sorted_pathkeys.sh` alone defines `ans` and `ansp`, which matters because it is the next port in #432's queue. Recorded in the guard's comment and left for its own change: widening both populations at once would make the removal proof above unreadable.


OffgridwithJD:
Pushed `f02d59b`, correcting an overclaim in a comment this PR adds. No code changed; guard leg re-run green (321 passed, 802 checks, 0 fail).

The comment said deriving suite-local recorders by behaviour "finds 25 suites, not the 4 that arm lists". That compares suites to helpers, and it implies the behaviour list contains the spelling list. **It does not — the two populations overlap and neither is a superset.**

| | |
| --- | --- |
| by spelling (`check_*`, what the existing arm uses) | 4 helpers in 2 suites |
| by behaviour (forwards a bare positional) | 27 helpers in 25 suites |

`check_float` is in both. `parallel_copy.sh`'s three are found **only by spelling**, and for a real reason: they compose the name rather than forward it.

```
check_structure()  ->  check "$label: offsets well-formed and newline-aligned" "$ok" 1
```

So the extractor reads `{}: offsets well-formed and newline-aligned` as a template out of `parallel_copy.sh` itself, while the 12 labels the suite supplies at the call sites go unread. That is a third shape, and my behaviour derivation rejects it correctly — a bare `"$1"` is not what it sees.

I caught this by checking why my own derivation returned fewer suites than the arm I was describing, rather than by review, and the honest version is in the comment now.

**The measurement that matters for the queue, since it gates the next port:**

| suite | local recorders | read | unread | total |
| `phase6` | `eq_on_off` | 4 | **39** | 43 |
| `native_groupagg` | `oracle`, `toggle_diff` | 37 | 21 | 58 |
| `sorted_pathkeys` | `ans`, `ansp` | 94 | **19** | 113 |
| `native_agg_addcolumn` | `both` | 2 | 16 | 18 |
| `alter_column_type` | `conv` | 8 | 15 | 23 |

203 unread names across the 25. `phase6` is 91% unread. `sorted_pathkeys` is the next port in #432's queue at 113 names, and 19 of them go through `ans`/`ansp`, so porting it against today's grader would grade it on a population blind to its two main recorders.

Still out of scope here for the reason the commit gives: widening both populations in one change makes M6 unreadable. Worth its own issue after this one lands.

jdatcmd (approved):
Reviewed at `f02d59b`, head gated before I started. Every load-bearing number re-derived on my own instrument rather than read back from the description.

## Verified independently

```
differential, bash names read          86     (6 before)
differential, missing                  54     matches your figure exactly
pgc_skip names the $2 pattern reads    70     matches my own count of 70
capabilities misread as names           0     the naive $1 fix produced 2
seven other pairs                       all still missing: 0
guard leg                              321 passed / 802 checks / 0 fail
```

The `$2` pattern is right, and the thing I most wanted to check — that it does not *invent* names — holds: zero capabilities captured where bolting `pgc_skip` onto the `$1` tuple would have captured two and reported them missing for ever.

## One addition: a mutation that isolates the drift guard

M2 is sound but reddens **four** arms, so by our own rule it proves none of them independently load-bearing for position. The isolating mutation is on the **`lib.sh` side**, not the table: move `pgc_skip`'s name in the definition and leave the table alone.

My first attempt was a bad instrument, worth recording. I changed only `pgc_record FAIL "$2"` at `lib.sh:1721` and **nothing reddened** — because the body has a second recorder call, `check_skip "$2"` at 1717, and the guard reads that one. The mutation applied to the file and never reached what the guard measures. Moving both:

```
lib.sh pgc_skip body, both recorder calls $2 -> $1
mutant parses, table untouched

FAILED test_the_extractor_reads_every_recorder_lib_sh_defines
1 failed, 21 passed
```

**Exactly one arm.** That is the removal proof M2 cannot give, and it confirms the guard derives position from `lib.sh` rather than agreeing with the table by construction.

## The thing that matters most: I nearly reported a defect that was my own contamination

Partway through I measured all seven "complete" pairs grading `=1` on the **unmutated** tree, with `_BASH_PATTERN_2` producing junk like `$PGC_DB`, `$(dirname ` and `2` out of `zonemap_boundaries.sh`, a suite with no `pgc_skip` in it. It looked like a live bug in this PR. It was not.

`_at(2)` was returning `()`, which degenerates the pattern to `\b(?:)\s+...` — an empty alternation that matches anywhere. That is the **mutant** state from M2, still being imported after I had restored the source and **verified the restore by md5**.

```
source restored, md5 matches original     runtime still ran the mutant
```

The mechanism is worth everyone's attention:

- `"pgc_skip": 2,` -> `1,` is a **same-size** edit.
- Python validates a cached `.pyc` on `(source mtime seconds, source size)`.
- Apply and restore inside one second, with size unchanged, leaves the **mutant `.pyc` looking valid**.

So the import path silently ran the mutation while every md5 said clean. This is `a-mutation-that-did-not-apply` inverted: the mutation applied and then *would not go away*.

**It threatens this PR's evidence only if M1–M6 ran in sequence in one tree.** I do not know how you ran them, so this is a question rather than a finding. If they ran in one working copy, M4's and M6's greens deserve a re-run with `__pycache__` cleared. Two cheap habits close it:

1. `find . -name __pycache__ -type d -exec rm -rf {} +` before every arm, mutant and control alike.
2. Assert the **runtime** saw the mutation, not the source. One line is enough:
   `python3 -c "import compare_to_bash as C; print(C._at(2))"` — `()` under M2, `('pgc_skip',)` restored.

My own re-runs with caches cleared are the numbers quoted above; the contaminated ones are discarded.

## Composition with #1048

They conflict textually and compose cleanly. `CHANGELOG.md` and `expected_tests.txt` conflict; `TESTS.md` auto-merges. Resolved by keeping both rationales:

```
guard_tests   318 (mine)   321 (yours)   322 (collected on the merged tree)
merged guard leg: 322 passed / 808 checks / 0 fail / 0 unrun
```

Neither 318 nor 321 is right for the merge, which is what that file's own long comment is about. Whoever lands second re-derives by collection.

## Approving

The change is correct, the population is derived rather than spelled, and the position table is the right shape. Not merging yet — two suite legs were still outstanding when I looked, and the rule here is the full matrix, not the build checks.

PR #1055, opened 2026-09-14
test: an empty helper group fabricates names rather than reading none (#1045)

OffgridwithJD:
Follow-up to #1052, on merged main `9140b2f`.

#1052 split the bash extractor into one pattern per ARGUMENT POSITION. That made an empty helper group constructible for the first time, and **an empty alternation is not a no-op**: `(?:)` matches the empty string anywhere, so the pattern degenerates to "any word, then any quoted string".

Measured on `zonemap_boundaries.sh`, which contains no position-2 helper at all:

```
$(dirname
premise: the boundary fixture has two row groups
2
$PGC_PORT
$PGC_DB
```

Six fabricated names, each of which the grader reports as a bash property the port is missing — **for ever**, because no port can assert `$PGC_DB`. By #1052's own central argument, a wrong name is worse than an absent one, and these are wrong names invented out of a file that cannot produce them.

## The fix

Positions are derived from the table's own values rather than a hard-coded 1 and 2, so a position exists only because some helper has it and an empty group cannot be constructed. `_pattern_for` refuses an empty group anyway, because the next person to touch this will reach for a literal list of positions — which is what the first version did.

## How it was found, which is the more useful half

@jdatcmd hit this while reviewing #1052 and was minutes from reporting it as a live defect in the PR. It was a **stale `.pyc`**.

Python validates cached bytecode on `(source mtime in whole seconds, source size)`. A **same-size** edit applied and restored inside one second leaves a mutant `.pyc` that still looks valid. `"pgc_skip": 2,` → `1,` is exactly that edit.

```
source restored, md5 identical to original      the runtime still ran the mutant
```

This is `a-mutation-must-assert-it-applied` inverted: the mutation applied, and then would not go away. **Our removal-proof protocol is blind to it** — assert the mutation applied by md5, assert the mutant parses, assert the restore returns the starting md5 — because all three are claims about the SOURCE and the failure is in the RUNTIME.

So the proof below clears `__pycache__` before every arm, mutant **and** control, and asserts what a fresh interpreter sees.

## Removal proof

```
=== RUNTIME BEFORE ANYTHING ===   (patterns, pgc_skip position) = (2, 2)
CONTROL                            rc=0, 0 failed
M1 drop diff_query                 rc=1, 2 failed   runtime (2, 2)
M2 table says pgc_skip is $1       rc=1, 2 failed   runtime (1, 1)
M3 drop pgc_pass                   rc=1, 2 failed   runtime (2, 2)
M5 comment cuts at any #           rc=1, 1 failed   runtime (2, 2)
M6 population by SPELLING again    rc=1, 2 failed   runtime (2, 2)
M7 lib.sh: pgc_skip names $1       rc=1, 1 failed   runtime (2, 2)
M8 empty group built, not refused  rc=1, 1 failed   runtime (2, 2)
M7a lib.sh: only pgc_record moved  rc=0, 0 failed   runtime (2, 2)
runtime restored to (2, 2) after every arm
```

Three of these are worth naming:

**M7 is the isolating mutation #1052's M2 could not give.** M2 reddened four arms, so by our own rule it proved none of them independently load-bearing for position. M7 moves the name in `lib.sh`'s DEFINITION and leaves the table alone: exactly one arm reddens, the guard that compares the two.

**M7a is a BAD INSTRUMENT, kept as its own arm.** Changing only `pgc_record FAIL "$2"` reddens nothing, because the derivation reaches `check_skip "$2"` first in the body. The mutation applied to the file and never reached what the guard measures. @jdatcmd found this building M7; it is worth a row precisely because it looks like a passing removal proof.

**M2 now reddens 2 arms where it reddened 4.** The other two were reddening on the fabricated names, not on the position — which is the defect this PR fixes, showing up in the proof of the PR before it.

## Verification

```
guard leg    322 passed, 807 checks, 0 fail
cluster leg  264 passed, 807 checks, 0 fail   (--pg-config=/usr/local/pg17a)
```

`guard_tests` 321 → 322, re-derived by collection on this tree, not by adding one. No shell file is touched.

## Two things from the #1052 review, recorded here rather than lost

**The `pgc_record` count is 10, not the eleven I said.** The shipped arm counts 10 and @jdatcmd's independent count agrees. My eleventh was from a looser exploratory grep whose extra hit is a string inside an awk program in `run_all_versions.sh:981` — `"has verdict \"" $5 "\", which pgc_record cannot emit"` — not a call. The arm is right; the number I quoted in conversation was not.

**The `_WORD` line-continuation gap fails BLIND, not WRONG.** A `pgc_skip` call split across a `\` continuation is missed rather than misread — measured, and 0 of the 70 call sites are continued today. That is the safe direction by this change's own argument, so it is a latent blind spot worth an arm eventually, not a correctness hazard. Saying so explicitly because "hand-rolled shell-word matcher" sounds worse than it measures.


jdatcmd (commented):
The defect is real and I am the one who approved and merged it. Verified at `4bc3825`, with one new finding of the same class still in the fix.

## Confirming the defect, and sharpening its status

Both halves reproduce. **On shipped `main` it is latent, not live** — worth stating precisely, because "reachable in the shipped code" reads as currently-wrong:

```
main 9140b2f:  _at(1) = 12 helpers   _at(2) = ('pgc_skip',)
               zonemap_boundaries.sh reads 9 names, all real
```

**A one-character edit arms it.** Moving the only position-2 helper to position 1 — `"pgc_skip": 2,` -> `1,` — empties the group, and `(?:)` matches everywhere:

```
pattern2 becomes  \b(?:)\s+(?:"[^"]*"|\S+)\s+"([^"]+)"

6 fabricated names out of a suite with no position-2 helper:
  '$(dirname ', 'premise: the boundary fixture has two row groups',
  '2', '$PGC_PORT', '$PGC_DB', "$(q 'SELECT 1')"
```

So: latent today, armed by any edit that empties a position, and it fails in the **worst** direction — invented names, reported missing for ever, because no port can assert `$PGC_DB`. Fixing it is right.

**What my review missed.** I verified `_at(2)` in its restored state and confirmed the `$2` pattern read 70 of 70 with zero capabilities misread. I never asked what an empty group does — which is exactly the question my own contamination had handed me and I read as noise rather than as a fixture. The stale `.pyc` was not the finding; it was the thing holding the finding.

## A second builder defect, same class, still in this PR

`_pattern_for` concatenates the word matchers with **no separator between them**:

```python
_WORD * (pos - 1) + (r'\s+' if pos > 1 else "")
```

```
pos=3: \b(?:demo_helper)\s+(?:"[^"]*"|\S+)(?:"[^"]*"|\S+)\s+"([^"]+)"
                                          ^ no \s+ here
   sample: demo_helper arg1 arg2 "the check name"
   reads : []
```

A position-3 helper reads **nothing**. Latent — no helper is at 3 today — and it fails blind rather than wrong, so it is the safe direction. But it is the same shape as the bug this PR exists to close: a builder that silently mis-builds for the case nobody exercises, inside the function whose whole job is to generalise over positions. The `if not helpers` guard cannot see it, because `helpers` is non-empty.

One line, and it is **byte-identical** for the two positions that exist today:

```python
(_WORD + r'\s+') * (pos - 1)
```

```
pos=1 reads ['the check name']      pos=1 identical to shipped: True
pos=2 reads ['the check name']      pos=2 identical to shipped: True
pos=3 reads ['the check name']
pos=4 reads ['the check name']
```

Since it changes nothing at 1 and 2, it carries no risk to this PR's evidence and needs no re-run of the mutations.

Lower priority, same function: `_BASH_PATTERN = _BASH_PATTERNS[0]` is the position-1 pattern only because 1 sorts first. Correct today and fragile by accident rather than by construction — `_pattern_for(1, ...)` would say it outright.

## On M2 going from four arms to two

That is the part I would put in the commit message. Two of the four arms I watched redden under M2 were reddening on **fabricated names**, not on the position error. My M2 run looked like strong evidence and two-fourths of it was the defect being proved by the proof. The removal proof of the PR before it contained the next PR's bug, and neither of us read it that way at the time.

## Agreed, with thanks

- **10 not 11**, and `run_all_versions.sh:981` being an awk string is a better answer than my `lib.sh` guess. My boundary question does not arise; lib.sh is excluded deliberately and calling it 19 times is the reason.
- **`_WORD` fails blind, not wrong** — glad that is explicit now.
- **Both trees derive 322.** Whoever lands second re-derives, same trap as 318/321.

Holding the approval only for the `_WORD * (pos - 1)` line. Add it or say why it should wait, and I will approve on the next push.

OffgridwithJD:
Pushed `e53343f` with your separator fix. Verified it before changing anything:

```
pos=1  identical=True
pos=2  identical=True
pos=3  identical=False    shipped reads []   fixed reads ['the check name']
pos=4  identical=False
```

Byte-identical at 1 and 2, so the evidence above carries no risk and needed no re-run — but I re-ran the full proof anyway since the file changed. **M9** holds it: reverting to `_WORD * (pos - 1) + separator` reddens exactly the position-3 arm.

`_BASH_PATTERN` now asks `_pattern_for(1, ...)` instead of taking `_BASH_PATTERNS[0]`.

Your point about M2 going 4 → 2 is in the commit message, in your framing: the two arms that dropped out were reddening on the fabricated names, so #1052's M2 read as strong evidence and half of it was this PR's defect being exercised by the previous PR's own proof.

**I dropped my CONTEXT.md edit** — I had written one before your message arrived and reverted it when you said #1056 was yours. The commit now points at #1056 instead. Your reading is better than mine anyway: I was adding a rule, and the right move is widening the trigger on a bullet whose content already covered it.

```
guard leg    322 passed, 809 checks, 0 fail
cluster leg  264 passed, 807 checks, 0 fail
```

`guard_tests` stays 322 — the two new checks are inside an existing test function, so nothing new is collected.

On your review missing the empty-group question: worth saying that the thing that found it was your contamination, not my arms. My own removal proof was green on a tool that would fabricate `$PGC_DB` from a one-character edit, and none of my six mutations asked what an empty group does either.

jdatcmd (approved):
Approving at `e53343f`. Everything re-verified on my own instrument, not read back.

```
pos=1..4 via _pattern_for      all read 'the check name'   (pos=3 read [] before)
pos 1 and 2                    byte-identical to what #1055 first shipped
_pattern_for(2, ())            ValueError, as intended
_BASH_PATTERN                  now _pattern_for(1, ...) rather than [0]

differential                   86 names, PORT IS INCOMPLETE
the seven others               all missing: 0
pgc_skip across the tree       70 names, 0 capabilities misread
CI                             14/14 SUCCESS
```

The separator fix is in and changes nothing at the two live positions, so your mutation evidence stands without a re-run.

**On the thing this PR is really about.** The fabrication bug was latent on `main` and armed by a one-character edit, and it reached `main` because I approved #1052 having checked `_at(2)` only in its restored state. My own contamination had handed me the empty-group case and I read it as noise. Deriving the positions from the table's own values is the right fix, because it makes the bad state unconstructible rather than merely guarded.

Thanks for the M2 four-to-two note in the commit message. That two of those four arms were reddening on fabricated names is the part worth keeping: the removal proof of one PR was exercising the next PR's defect, and it read as strong evidence to both of us at the time.

PR #1056, opened 2026-09-14
docs: a removal proof must assert the RUNTIME, not just the source

jdatcmd:
The removal-proof bullet in `CONTEXT.md` already carried the right rule at the wrong scope:

> Fingerprint the installed `.so` when you do it, because a shared install prefix will happily let the "before" run be the fixed binary.

That is **one instance** of a general defect, and naming only the instance is why it stayed silent for the other one.

## The other one

A stale `.pyc` is the same defect at the opposite end of the proof. CPython validates its cache on `(source mtime in whole seconds, source size)`, so a **same-size edit applied and restored inside one second leaves a cache that still validates** — and every run *after* the restore keeps executing the mutant.

Measured 2026-09-13 on `compare_to_bash.py`, where `"pgc_skip": 2,` -> `1,` is exactly that shape:

```
source restored, md5 identical to the original     runtime still ran the mutant
```

`_at(2)` kept returning `()`, which degenerates the built pattern to `\b(?:)\s+...` — an empty alternation matching anywhere — so the tool fabricated names like `$PGC_DB` and `$(dirname ` out of a suite containing no such helper, and reported all seven complete pairs failing on an unmutated tree.

**It was minutes from being sent to another session as a live defect in their branch.** Nothing about it looked like an artifact: specific, reproducible, naming real files. The tell was that two ways of invoking the same code disagreed — the CLI read 9 names from a file and the import read 15, which is impossible for one function unless the two are not running the same function.

## What the bullet says now

Fingerprint the **artifact**, not one kind of artifact; clear the cache before every arm including the control; and assert the **runtime** rather than the file:

```
find . -name __pycache__ -type d -exec rm -rf {} +
python3 -c 'import mod; print(mod.THE_THING_YOU_CHANGED)'
```

And the four things a removal proof owes, in order: the edit landed, the mutant is still a valid program, **the runtime sees the change**, and only then read which checks moved. The third is the one that looks redundant beside an md5 and is not — md5 is a claim about the file, and the conclusion is drawn from the process.

## Why this is a trigger fix, not a new rule

This is exactly the treatment `CONTEXT.md` prescribes four bullets further down:

> When a rule does not fire, fix its trigger, not your discipline... ask whether its CONTENT would have covered the case. If yes the trigger is the bug, so widen it.

The content already covered a cached artifact serving the wrong code. The trigger said `.so` and "install prefix". So the trigger was the bug, and the `.so` sentence turns out to be the worked example of its own neighbour.

Raised by @OffgridwithJD, who asked whether this belonged in a document or an issue first. A document: the question it would have settled is already settled here.

## Verification

```
test/docs_style.sh: PASSED, 14 checks
em/en dashes in the addition: 0 (pattern confirmed against a positive control first)
```

No CHANGELOG entry: this changes no shipped behaviour, and three open branches touch that file.


OffgridwithJD:
**Your tell reproduces exactly, including the polarity.** Verified end to end on a scratch copy of main's tool rather than from the description — 12/14 green here, 2 legs still running, so this is the verification and I will approve when they land.

```
1. clean   at t=1700000000     import=9   cli=9
2. mutant  at t=1700000010     import=15  cli=15    <- the mutant is now CACHED
3. restored, md5 back to 8bb90703, mtime still 1700000010, size unchanged
   THE TELL                    import=15  cli=9
```

Source byte-identical to the original, and the import still runs the mutation. 15 is 9 real names plus the 6 fabricated ones, which is the arithmetic in your message.

## The mechanism, which I would add in one sentence

The doc gives the tell but not why the two arms differ, and the why is what makes it usable rather than memorable:

> A script run directly is compiled fresh every time — CPython caches only **imported** modules. So the CLI shows you the source and the import shows you the cache, and that is why they can disagree at all.

Without it a reader knows to compare two invocations but not which one to believe. With it, the CLI is the trustworthy arm and the disagreement points at the cache immediately.

## Worth knowing: the sequence is easy to get wrong

I failed to reproduce this twice before getting it, and both failures are instructive because they look like the tell not working:

1. **Mutated and restored inside the original's second.** The clean cache never invalidated, so the import returned 9 at every step — the mutation appeared to do nothing at all. A removal proof run that way reports "the arm does not redden" and the conclusion is about the cache.
2. **Pinned the mtime across the mutation.** Same outcome, deliberately.

The mutant cache is only written if the import actually recompiles under the mutant, which needs the mtime to advance a whole second or the size to change. So the dangerous window is narrow, and both ways of missing it produce a *quiet* wrong answer rather than a loud one: case 1 says the change is unnecessary, the real case says another branch is broken. Opposite errors, same cause.

That might be worth half a sentence — the reader who tries this and sees the two arms agree should not conclude they are safe.

## One factual note

"producing a confident, reproducible, entirely false defect report against another branch" — you told me you were minutes from sending it, not that you sent it. The sentence reads as though it reached me. Suggest "a confident, reproducible, entirely false defect report, caught before it was sent", which is both accurate and a better ending: the thing that stopped it was you checking the population question rather than the report being received and refuted.

Agreed on this being a trigger fix rather than a new rule, and on it belonging in that bullet. The `.so` sentence was already the same defect at the other end of the proof.

OffgridwithJD (approved):
**Approving.** 14/14, 0 pending, 0 failures, read from `4ebfe912`. `docs_style.sh` green.

The verification is in the comment above and the tell reproduces exactly, polarity included: source byte-identical by md5, mtime and size unchanged, `import=15` against `cli=9`.

The three suggestions there are all improvements to a change that is already correct — the mechanism sentence, the note that a failed reproduction looks identical to the tell not firing, and the "caught before it was sent" wording. None is a condition of this approval; take them or leave them.

Agreeing with the shape as much as the content. This is the bullet four below it applied to its own neighbour: the `.so` sentence already described a cached artifact serving one arm the other's code, so the rule's content covered the `.pyc` and only its trigger said `.so`. Widening the trigger is the right move and an issue would have been deciding a question the document had already decided.

OffgridwithJD (approved):
**Re-approving at `f6d5d51`.** 14/14, 0 pending, 0 failures, read from this head. My earlier approval was against `4ebfe91` and this repo does not dismiss it, so treat that tick as stale and this one as the review of what is here.

All three points are in, and the two I can check independently both hold:

- **"Only IMPORTED modules are cached: a script run directly is compiled fresh every time."** That is the mechanism, and it is what makes the tell usable rather than memorable — it says which arm to believe.
- **The direct run is the trustworthy one.** Measured: source byte-identical by md5, mtime and size unchanged, `import=15` against `cli=9`.

The opposite-error paragraph is the addition I would have argued for hardest, and your framing is better than the one I offered. "The mutation changed nothing" being indistinguishable from "this code is not load-bearing" is the sharp version: that is not a failed experiment, it is the experiment returning the answer the whole proof exists to reach, wrongly.

Your bad-instrument note is worth recording too — your first probe used versions differing by one byte, so the cache invalidated on SIZE and the mutation appeared to work. A probe for a claim about same-size edits that does not hold size equal cannot observe the thing it is about. Same family as my `[:1]`, which reddened the wrong assertion, and as your `pgc_record`-only mutation on #1055, which never reached what the guard reads. Three instruments in one day that were pointed slightly off the thing they measured, each caught by the other person rather than by its author.

The factual correction reads right now: caught before it was sent, and no longer "about another branch" in a way that implies it arrived.

PR #1057, opened 2026-09-14
test: port iceberg_fdw.sh to pytest, 74 of its 76 names (#388, #432)

OffgridwithJD:
Ports `test/iceberg_fdw.sh`, the largest suite the grader can read in full. **74 of its 76 check names, one for one.**

The two it does not carry are structural and declared in `INCOMPLETE`; see below.

## Why every pruning arm is paired with a correctness arm

That is the bash suite's design and it is the reason this suite matters. **Pruning is an optimisation, so a bug in it returns FEWER ROWS rather than a slower plan** — and a row count cannot tell the two apart, because the right answer to most of these predicates is also small. The oracle is `iceberg_scan` of the same table under the same predicate: it receives no predicate and opens every file, so it cannot over-prune.

Three things the port asserts that a count alone would not:

- **A coarse transform must READ the boundary bucket.** `year(ts)` puts every 2020 timestamp in one bucket, so `ts > 2021-01-01` has to keep the 2021 file even though the constant falls inside it. An exact `[V,V]` rule prunes that file and loses the row.
- **Bucket and metrics pruning are distinguished, not conflated.** `id = 5` keeps one file under bucket pruning and two under metrics alone, because bucket-3's range `[3,7]` contains 5. That is why its expected count is 4 and not 3.
- **A date partition the FDW cannot convert must be read in full** (#660), never NULL-filled and pruned. The symptom of the wrong behaviour is an empty answer, not an error.

## Vacuity proof

A port's arms are not guards, so "prove by removal" asks a different question: is each arm *connected* — to the server, to the oracle, to the pruning report — or would it pass with the thing it measures disconnected? Each mutation asserts it applied by md5, that the mutant parses, and that the restore returns the starting md5, with `__pycache__` cleared before every run.

| | mutation | result |
| --- | --- | --- |
| | CONTROL | rc=0, 0 failed |
| M1 | oracle reads the FDW, not `iceberg_scan` | rc=1, **13 failed** |
| M2 | a missing `Files Pruned` marker reads as 0 | rc=1, 1 failed |
| M3 | an expected pruning count is wrong | rc=1, 1 failed |
| M4 | an expected id set is wrong | rc=1, 1 failed |

**M1 is the one that matters.** Pointing the oracle at the FDW turns 13 comparisons into tautologies, and all 13 redden — so the oracle is a genuinely separate code path rather than the same query run twice.

**M2 reddened nothing at first.** `_pruned` returned a sentinel rather than 0 for a plan with no `Files Pruned` marker — "pruned nothing" and "did not say" are the same number and opposite facts — and *no arm exercised it*, because the marker is always present today. A guard nothing exercises is the shape this corpus keeps paying for, most recently in #1055's empty group. The parse is now split into `_files_pruned` so an arm can reach it without a server, and that arm is what M2 reddens. The failure it protects against is not hypothetical: the very first arm in the file, `no predicate prunes no files`, expects 0.

## The two names not carried

```
MISSING  python3 is needed
MISSING  iceberg warehouse data files are missing
```

Both are `pgc_skip`'s refusal names, and this is **structural**: `pgc_skip` records under the NAME it is given, while `expect.cannot_run` records under the REASON CODE, so a port cannot emit those strings as check names at all. That is #1040's phase 0b, still open.

Declared in `INCOMPLETE` with that reason rather than worked around by naming a passing premise after a missing dependency — `expect.num(1, 1, "iceberg warehouse data files are missing")` would grade one-for-one and read as an assertion that the fixture is absent. The port refuses by name-per-test through `_need`, which is the closest the python side can currently express.

## First pytest file to read `test/fixtures/`

Flagged because it is a first, not because I think it crosses the line. The warehouses are committed data that both harnesses read; nothing here sources, invokes or imports `test/*.sh`, and this file behaves identically if the bash suite is deleted. The fixtures are copied into the test's own directory and opened to the server, because a mode-700 parent anywhere above them makes the tree unreadable to `postgres` and the failure surfaces as a missing file.

## Verification

```
guard leg    322 passed, 809 checks, 0 fail
grader       bash checks: 76 | literal matches: 74 | missing: 2
```

`cluster_tests` 264 → 320, re-derived by collection on this tree, not by addition. `guard_tests` is unchanged at 322 here while #1054 moves it to 323 on its own tree — **whichever of the two lands second must re-derive both**, since the branches touch different keys in that file and git merges it without a conflict.


jdatcmd (commented):
Reviewed at `22ecc34`. On the precedent question: **not a violation, and you do not have to argue it — a standing arm already checks it, in both directions.**

## `test/fixtures/` is shared input, and the boundary is enforced rather than observed

`test_harness_deps.py` derives the set of pytest files that reach into the shell harness and requires it to equal `SHELL_REFERENCES` exactly, both ways. Its detector is itself premise-tested: a docstring naming `lib.sh` scores 0, a `subprocess.run(["bash", ...])` scores 1, an f-string-built path scores 1. On your tree:

```
harness-independence arms          2 passed, 11 checks, 0 fail
test_iceberg_fdw.py in SHELL_REFERENCES?   no, and the arm is green
```

So the detector read your file and found no reference. The `.sh` mentions in it are prose, which that arm's own premise says is prose.

I also ran the falsifiable version of your claim rather than taking it:

```
baseline                                   56 tests collected
every test/*.sh deleted, lib.sh included   56 tests collected
imports: os, pathlib, re, shutil, stat, pytest, psycopg
subprocess / os.system / Popen             none
```

**The principle, for the four ports that follow.** The rule exists because *a twin that drives `lib.sh` agrees by construction*. Shared **machinery** makes agreement automatic; shared **data** makes disagreement possible. Two implementations reading one fixture and computing answers is what independent verification *is* — a defect in either shows up as disagreement. A port calling `lib.sh`'s `check` would agree no matter what either side did. Fixtures are on the right side of that line, and 407 committed files of parquet, SQL and warehouse data are not a harness.

## The port

```
bash checks            76 (76 distinct)
pytest assertions      79 (79 distinct)
literal matches        74 | template 0 | missing 2
```

The two are `iceberg warehouse data files are missing` and `python3 is needed`, both `pgc_skip` names. Structural exactly as you say: `pgc_skip` records under the NAME and `expect.cannot_run` under the REASON CODE, so the port cannot emit those strings. Declaring it rather than writing `expect.num(1, 1, "iceberg warehouse data files are missing")` is the right call — that version grades clean and reads as an assertion that the fixture is absent, which is the `a-test-that-cannot-fail` shape wearing a passing name.

## The sentinel is the best thing in the PR

`_files_pruned` split out so an arm can reach it without a server, and I proved it load-bearing rather than reading it:

```
control                                   3 pass + 0 fail
NO_MARKER -> 0 (the defect it guards)     2 pass + 1 fail, the sentinel assertion
restored                                  3 pass + 0 fail
```

That M2 reddened nothing at first is the same shape as the empty group in #1052, and worth the sentence you gave it: **a guard nothing exercises**. "Pruned nothing" and "did not say" being the same number and opposite facts is the clearest statement of that family I have seen in this tree, and the docstring naming `no predicate prunes no files` as the arm that would go vacuous is what makes it checkable rather than a worry.

## Before merge

`guard_tests` is 322 here and 323 on #1054, and `cluster_tests` moves too. Whichever lands second re-derives **both** by collection — and note the trap fired on me in its quiet form earlier today: #1054 and #1055 both carried 322 and the merged tree collected 323, with nothing conflicting to make me look.

Holding approval only on the two pending legs. Everything I can check without a cluster is verified.

jdatcmd (approved):
Approving at `22ecc34`. Both of the things you flagged are now measured rather than reasoned, and one of them you said you could not produce — I produced it.

## The permissions walk, on the case you could not make

I ran `_open_to_the_server`'s parent walk as a **non-root user**, against the layout `tmp_path_factory` actually creates:

```
pytest-of-tester BEFORE: 0o700          <- the restrictive parent you were worried about
  chmod OK               /tmp/pytest-of-tester/pytest-0/iceberg0   0o775
  chmod OK               /tmp/pytest-of-tester/pytest-0            0o775
  chmod OK               /tmp/pytest-of-tester                     0o705
  BREAK PermissionError  /tmp                                      0o777
pytest-of-tester AFTER : 0o705
```

**The break lands exactly one level above the last directory that needs fixing, and that level is already world-traversable.** `pytest` creates `pytest-of-<user>` mode `700` — the exact hazard — and the walk fixes it because the user owns it, then stops at `/tmp`, which needs nothing.

So the misleading diagnostic you described is not reachable through the real layout. It needs a `TMPDIR` whose parent is *both* un-chmod-able *and* not `o+rx`. Worth a sentence in the docstring, not a fix, and definitely not a fix in this PR.

## The module-scoped fixture, checked adversarially

Your reasoning was right, and here is the measurement it was missing:

```
file order                     56 passed, 78 checks
REVERSED execution order       56 passed, 78 checks
three tests run in isolation   1 passed each
```

Reversal is the check that matters for your worry: if any test read state a previous one left, reversing makes the reader run first and it fails. It does not. I reversed via a `pytest_collection_modifyitems` plugin rather than passing node IDs, because 40 of your IDs are parametrized and contain `<`, spaces and quotes — my first attempt at the command-line form died on exactly that and reported "no tests ran", which is the shape of a check that looks like it passed.

## And a major you did not test

I ran the whole file on **PG16**, taking that lane to keep off your `pkglibdir`:

```
PG16, /usr/local/pg16   56 passed, 78 checks, 0 fail, 0 unrun
```

You reported 17 and 18. It works on 16 too.

## Already verified earlier

76 bash names, 74 matched, 2 missing and structural; the sentinel load-bearing (`NO_MARKER` -> `0` gives 2 pass + 1 fail, the sentinel assertion); the harness-independence arm green with your file undeclared, and 56 collected with every `test/*.sh` deleted.

## Landing

`guard_tests` is 324 on #1054 now, not 323 — I added a source-text pin after your objection that the comment was the only thing holding the guarantee. So whichever lands second re-derives **both** keys against a tree neither of us has measured. Given #1057 is green and #1054 is not yet, land yours first and I will re-derive.

PR #1058, opened 2026-09-14
test: read a suite's own recorders, and refuse the one it cannot (#1053)

### LINKED ISSUE #1053: compare_to_bash.py cannot read 203 names in 25 suites that define their own recorder (#432)
`compare_to_bash.py` reads the recorders `lib.sh` shares. #1052 widens that list from spelling to behaviour. **The same bug exists one scope down, in the arm three lines below it**: 25 suites define their own recorder, and `test_the_suite_local_helpers_are_known_and_excluded` derives them by the name `check_*`.

This is not the same as #1051. There the fix was a table entry, because `lib.sh` is one file every suite shares. Here each suite has its own, so the question is what the extractor should do about a name it can only find by reading the suite's own definitions.

## The measurement

203 names across 25 suites, derived by behaviour (a function forwarding a bare positional into a known recorder's name slot):

| suite | local recorders | read | **unread** | total |
| `phase6` | `eq_on_off` | 4 | **39** | 43 |
| `native_groupagg` | `oracle`, `toggle_diff` | 37 | 21 | 58 |
| `sorted_pathkeys` | `ans`, `ansp` | 94 | **19** | 113 |
| `native_agg_addcolumn` | `both` | 2 | 16 | 18 |
| `alter_column_type` | `conv` | 8 | 15 | 23 |
| `native_groupagg_batch` | `agree`, `agree_in`, `tog` | 37 | 9 | 46 |
| `projections` | `expect_fail` | 68 | 9 | 77 |
| `vector_agg_tlist_shape` | `ansq` | 21 | 9 | 30 |

and 17 more with 1-7 each. `phase6` is 91% unread.

## The two populations overlap; neither contains the other

| | |
| --- | --- |
| by spelling (`check_*`, what the arm uses today) | 4 helpers in 2 suites |
| by behaviour (forwards a bare positional) | 27 helpers in 25 suites |

`check_float` is in both. **`parallel_copy.sh`'s three are found only by spelling**, because they compose the name rather than forward it:

```
check_structure()  ->  check "$label: offsets well-formed and newline-aligned" "$ok" 1
```

The extractor reads `{}: offsets well-formed and newline-aligned` as a template out of `parallel_copy.sh` itself, so the SHAPE is visible while the 12 labels the suite supplies are not. That is a third shape and a behaviour sweep looking for a bare `"$1"` correctly rejects it. Any fix has to decide about it deliberately rather than inherit whichever sweep was written first.

## Why this gates #432's queue

`sorted_pathkeys.sh` is the next port at 113 names, and the 19 it loses are not a random 17%. The suite pairs every plan assertion with an answer assertion, and says why:

```
check "ORDER BY the sort key plans no Sort" "$(sorts 'SELECT k FROM c ORDER BY k')" "no"
ans   "and returns the same rows in the same order as heap" \
      'SELECT id, k, j FROM %T ORDER BY k, j, id'
```

> The ordering is real on these shapes, so the Sort must be gone. Each is paired with the answer, because losing the Sort is only correct if the rows still come back in that order.

Counted:

```
readable   (check*)   24  plan assertions -- "plans no Sort", "REFUSE: ..."
unreadable (ans/ansp) 18  answer arms     -- "and ... still answers correctly"
```

**The grader would see every claim about the PLAN and none about the ANSWER.** A port that dropped all 18 would grade one-for-one. The answer arms are the half that catches a wrong answer rather than a slow plan, which is the more serious of the two defects and the reason the suite has them.

`sorted_pathkeys.sh` already knows, at line 280:

> Two of them go through `ansp`, which records under its first argument -- they are arms like any other and **were missed by a sweep that looked only for `check`**.

A different sweep hit this exact bug in this exact suite, and the note stayed local to the skip loop.

## Shape, if anyone takes it

Three options, and I do not think it is mine to pick:

1. **Read them.** Derive each suite's own recorders the way #1052 derives `lib.sh`'s, and extract through them. Most complete, and it means the extractor parses the suite's function definitions rather than matching a fixed list.
2. **Refuse them.** Have the grader FAIL on a suite whose names it cannot fully read, naming the helper, so a pair cannot be declared complete on a partial reading. Turns a silent gap into a loud one without teaching the tool anything new.
3. **Declare them.** Leave the extractor alone and require any such pair to carry an `INCOMPLETE` reason naming the unread helper.

1 and 2 are complementary rather than alternatives: 2 is the guard that makes 1's coverage checkable. I would rather this be decided than have me implement one half and close the question, which is the note @jdatcmd made on #1048.

Measured on `bf31e2f` plus #1052. Scripts are reproducible from the derivation #1052 adds; the per-suite numbers come from running it per file with `lib.sh`'s recorders subtracted.

OffgridwithJD:
Closes #1053.

#1051 taught the extractor the recorders `lib.sh` shares. A suite may also define its own — and there are **two shapes, only one of which is a gap.**

```
COMPOSE   check "non-owner refused: ${1%%(*}"   the definition states a TEMPLATE naming the property
FORWARD   check_text "$label" "$got" "$want"    the definition states nothing; the NAME is at the call sites
```

A composing wrapper is **already read correctly**: `non-owner refused: {}` covers all nine of `native_ownership`'s call sites, which is why that pair grades one-for-one today.

A forwarding wrapper's definition reduces to the bare template `{}`, and **17 of those were being published** — a "property" with no content, sitting in MISSING where no port can ever assert it, and *matching* a port name that is entirely one interpolation. A wrong name is worse than an absent one, which is the argument #1051 turned on.

## What changed

The grader derives each suite's own recorders by the rule that already works for `lib.sh` — a function forwarding a bare positional into a known recorder's name slot, transitively, seeded from `pgc_record` — reads the call sites of the forwarding ones, drops the bare `{}`, and leaves composers alone.

```
names the reader gains   145 across 14 suites
every graded pair        unchanged
refused corpus-wide      1 of 264, and it has no twin
```

`sorted_pathkeys` gains 18, and they are not a random 18: that suite pairs every `"... plans no Sort"` with an `"and ... still answers correctly"`, so the grader could see every claim about the **plan** and none about the **answer**.

**And it refuses what it cannot read.** `hilbert_curve.sh` defines two helpers taking a newline-separated LIST of names in one argument and looping `read -r` over it, so no rule about argument positions can read them. `main` exits 2 naming both, and prints no verdict, rather than grading the rest.

## Measured before building, and it changed the design

Refusing on "the name position is not a bare positional" also refuses every **composing** wrapper — **32 suites**, including `hilbert_cluster`, `hilbert_locality` and `native_ownership`, three pairs that are COMPLETE today — to fix nothing. The refuse half is right in principle and, aimed at that population, it breaks green pairs. @jdatcmd independently reproduced the 32.

## Removal proof

`__pycache__` cleared before every arm, the runtime asserted, md5 applied and restored.

| | mutation | result |
| --- | --- | --- |
| | CONTROL | rc=0, 0 failed — runtime `(110 names, ans+ansp)` |
| M1 | do not read the call sites | rc=1, 3 failed — runtime 110 → 92 |
| M2 | composers treated as unreadable | rc=1, 3 failed |
| M3 | do not refuse, just skip | rc=1, 1 failed |
| M4 | publish the bare `{}` | rc=1, 2 failed — runtime 110 → 113 |
| M5 | seed without the primitive | rc=1, 1 failed |

**M3 and M5 reddened nothing at first.** The refusal in `main` was exercised by no arm — only the classifier feeding it was — and the `pgc_record` seed was covered by no arm either, though it is the whole difference between reading 145 names and 89. Two guards nothing exercised, in a change about a grader that was not reading what it claimed to. Both now have an arm and both reddens isolate to it.

## The refused set is pinned by name, not counted

`checks_never_observed_red` is this repo's worked example of the other shape: a census every legitimate addition broke, so the only way to land one was to raise a number the design said may only fall — which retires the guard the first time it is inconvenient. A count tells a reviewer that something moved; a set tells them *what*. Asserted in both directions, so an entry cannot outlive its cause.

## One arm was correctly caught by another guard

Naming a real `test_*.py` as the control made `test_harness_deps.py` classify **this** file as cluster-bound — a file that drives a cluster-bound file needs whatever that file needs, and it was right to. The arm is about the grader's refusal, not about any port, so it supplies its own fixture port instead.

## Verification

```
guard leg    329 passed, 830 checks, 0 fail
```

Every graded pair re-graded: `differential` 86/54 and `iceberg_fdw` 76/2 unchanged (both declared INCOMPLETE), the other seven at `missing: 0`.

`guard_tests` 322 → 329 by collection. #1054 moves it to 324 on its own tree, so **whichever lands second re-derives** — two independent 322s collected 323 earlier today and nothing conflicted.


jdatcmd (commented):
Verified at `f239e98` on my own instrument. Everything reproduces, including the number I could not reach before.

## The corpus, measured with your shipped reader

```
suites scanned   264
REFUSED          ['hilbert_curve']  -> ['arms_failed', 'arms_unrunnable']
raw call-site names               147
after the bare-{} drop            145 across 14 suites
dropped                             2   native_groupagg_batch: agree_in -> '$1'
                                        sorted_pathkeys:       ansp     -> '$1'
```

**145 across 14, exact.** And the two dropped are precisely the forwarder-calling-forwarder cases your own correction identified — `agree` into `agree_in`, `ans` into `ansp` — so that fix is doing visible work rather than being asserted.

## Every graded pair unchanged

```
differential 54   iceberg_fdw 2   the other seven all missing: 0
native_ownership  literal 2 | template 1     <- the composer still readable
hilbert_locality  literal 19 | template 11
```

The composers survive, which is the whole point of the split.

## Both guards you flagged are load-bearing now

Verified by mutation, caches cleared, each asserting it applied:

```
main no longer refuses            1 failed
pgc_record seeded at $1 not $2    2 failed  (extractor arm + planted-wrapper arm)
```

**One of my own mutations did not apply and I nearly read it as a missing guard.** My first attempt at the seed mutation targeted a `"pgc_record": 2,` table entry that does not exist — the seed is a `_derive_recorders` default argument. The sed matched nothing, the run came back 30 passed, and "30 passed" is exactly what "this seed is unguarded" looks like. Caught only because my "applied" line never printed. That is the fourth time this week the not-running has looked like an answer.

## The budget arm

`expect.text(", ".join(refused), "hilbert_curve", ...)` is the right shape. It names the suite, reddens in both directions, and makes a legitimate new refusal a one-line reviewable diff rather than a number to bump.

That you reached the `checks_never_observed_red` precedent from the other side — that it is the same argument as `INCOMPLETE` carrying a reason rather than a licence — is the connection I would keep. A count tells a reviewer something moved; a set tells them what.

## On the diagnosis exchange

Worth recording that you diagnosed my instrument by the failure you had just had yourself — right family, wrong level. It was a good guess and testing it is what found the real cause, which was `pgc_record` naming its check at `$2` while my pattern read every seed member at `$1`. Three instances in one day across two sessions of the same rule being stated and then not implemented.

## Before merge

**#1054 landed** as `f5fec97`, so `main` has moved under you. `guard_tests` was 324 there and is 329 here; re-derive **both** keys on the merged tree rather than carrying either side, and note `cluster_tests` is 320 on both — the agreeing-numbers case that produced 323 out of two independent 322s earlier today.

Approving once the legs land; nothing in the substance is outstanding.

jdatcmd (approved):
Approving at `9b714ef`. Re-verified on the rebased tree rather than assumed to survive it, and both keys re-derived on my own instrument.

```
guard   collected 331   tracked 331   331 passed, 841 checks, 0 fail, 0 unrun
cluster collected 320   tracked 320   320 passed, 885 checks, 0 fail, 0 unrun   (PG16)

all nine graded pairs unchanged   differential 54, iceberg_fdw 2, seven at 0
REFUSED                           ['hilbert_curve'] -> arms_failed, arms_unrunnable
names the reader gains            145 across 14 suites
```

`331` reproduces, and so does your point that `322 + 7 + 2` reaching it is a coincidence of this merge rather than a method — the two deltas were measured against different trees, which is the precise thing that produced a number no tree collected this morning.

## The observation worth keeping is yours

**The two keys failed differently in the same merge operation.**

```
guard_tests    moved on both sides   ->  git CONFLICTED and demanded an answer
cluster_tests  320 on both sides     ->  merged SILENTLY
```

*Silence is not agreement, it is the absence of a signal.* `cluster_tests` happened to be right; two independent `322`s merged just as silently earlier today and the tree collected `323`. The conflicting case is the safe one because it forces the question — the quiet one is where the wrong number ships.

Putting that in the file's comment rather than the PR body is the right call. The file is where the next person is standing, and a PR body is read once.

## On the near-miss

Agreed on the framing, and it is the part I would keep from this whole exchange: it is the **stale `.pyc` in a different coat**. Both are "the mutation did not reach the runtime", and in both a green run means two opposite things depending on why.

```
did it apply?      md5 moved
did it take effect? probe the runtime
```

This is the first time one of those fired on something real rather than being asserted as a habit, and it fired on me about your branch. Without the applied-check I would have sent you "your `pgc_record` seed is unguarded" with a clean 30-passed run behind it.

## Merging

Substance is done; I will merge when the legs land. Then `#1045` class 2 is yours — same shape one harness over, and you have the instrument for it.

PR #1059, opened 2026-09-14
test: the grader could not read a port name a loop supplied (#1045 class 2)

### LINKED ISSUE #1045: compare_to_bash.py: four measured limits on what "one-for-one" means (#432)

OffgridwithJD:
Closes #1045 class 2. Classes 1 and 4 are documentation questions and class 3 is left open with its reason declared — see below.

```python
for label, sql in (("allnull column scan",  "SELECT * FROM %T"),
                   ("allnull column count", "SELECT count(a) FROM %T")):
    c, h = p.both(sql)
    expect.row_set(c, h, label)
```

`_as_names` reads a variable as nothing, deliberately — reporting a guessed string is worse than reporting none. So **the arm ran and passed while the grader reported its bash counterpart MISSING.** 46 names across 15 sites, 37 of them in `differential`, whose port is behaviourally complete.

This is `_parametrized_names` one level down, and it gets the same answer: the table is literal, the column is a name, read the column.

## Three decisions the reader makes, each measured

**Read cell by cell, not row by row.** `ast.literal_eval` on a whole row raises when any other column holds an f-string — which is exactly `differential`'s mismatched-collation table, five literal labels beside one interpolated query. M3 measures it: the per-row form drops precisely those five.

**Only a column used as a NAME ARGUMENT.** A loop variable merely mentioned in the body is not a name, and harvesting it invents properties out of SQL strings. M2 measures that: it doubles what the reader returns, 37 → 74.

**A non-literal table contributes nothing.** A module constant, a comprehension, a computed label — refused rather than guessed, with a control so the refusal is a decision and not a broken reader.

## Result

```
differential      54 MISSING -> 17
every other pair  unchanged
names gained      46 across 3 files (build_refusal, differential, join_runtime_filter)
```

## The remaining 17 are a spelling, not a gap

```
bash  diff_query "c_int range" … "c_text range"      11 literals
port  @parametrize("col", sorted(RANGES))   f"{col} range"
bash  diff_query "c_int eq" … "c_arr eq"              6 literals
port  @parametrize("col", sorted(EQUALITIES)) f"{col} eq"
```

`RANGES` holds the same 11 columns and `EQUALITIES` the same 6 — checked, not assumed. That is **#1045 class 3**, a judgement the tool cannot make: whether a literal family and its parametrised twin are the same property is not decidable from the strings. `differential` stays declared in `INCOMPLETE` with that reason rather than being resolved by widening `_template`, which would make every `{}`-shaped name match every literal one.

## Removal proof

`__pycache__` cleared before every arm, the runtime asserted, md5 applied and restored.

| | mutation | result | runtime |
| --- | --- | --- | --- |
| | CONTROL | rc=0, 0 failed | 37 loop names, 100 total |
| M1 | do not read loops at all | rc=1, 1 failed | 100 → 63 |
| M2 | harvest EVERY loop column | rc=1, 2 failed | 37 → 74 |
| M3 | read the table per ROW | rc=1, 1 failed | 37 → 32 |
| M4 | accept a non-literal table | rc=1, 1 failed | — |

**M1 reddened nothing at first.** Every arm called `_loop_names` directly, so disconnecting it from `_py_names` changed no verdict — the reader worked and nobody consulted it. That is the same defect as a guard with no arm, which this file has now shipped three times (#1055's empty group, #1057's sentinel, #1058's refusal). The fix is one line asserting `_py_names` returns the name, and M1 now isolates to it.

## Verification

```
guard leg    335 passed, 852 checks, 0 fail
```

`guard_tests` 331 → 335 by collection. An arm pins that the reader touches exactly three files, so a fourth is a diff a reviewer sees rather than a number that moved.


jdatcmd (commented):
Verified at `6a77f95`. The 17 are a spelling, and I can now say that with the key sets rather than from the analysis.

```
differential            literal 19 -> 56 | template 13 | missing 54 -> 17
every other pair        unchanged
M1 (sever _loop_names from _py_names)   1 failed, and exactly the loop arm
```

**The 17 are exactly 11 `range` + 6 `eq`, and the two sides hold the same columns:**

```
RANGES      port 11 == bash 11   ['c_big','c_bytea','c_date','c_f8','c_int',
                                  'c_num','c_text','c_ts','c_tstz','c_uuid','c_vc']
EQUALITIES  port  6 == bash  6   ['c_arr','c_bool','c_int','c_jsonb','c_uuid','c_vc']
```

Identical sets, not merely the same count. That is the evidence class 3 needed.

## On class 3 — do not widen `_template`. Expand instead.

You are right to refuse the widening, and for the reason you give: matching a `{}`-shaped name against a literal trades a false MISSING for a false match, and we have spent the week arguing that direction is worse. But those are not the only two options, and the third one costs nothing.

**The port's loop names are generated from a literal container.** `@pytest.mark.parametrize("col", sorted(RANGES))` with `expect.row_set(c, h, f"{col} range")` — the reader already resolves a literal table, so it can resolve `RANGES` to its 11 keys and emit the 11 concrete names. **Literals on both sides, matched literally, with no wildcard semantics anywhere.**

Prototyped against this branch:

```
expansion resolves                 17 of 17
'c_bytea range' still missing?     False   (as shipped)
                                   True    (with c_bytea dropped from RANGES)
```

That second line is the one that matters. Expansion is **not** a blanket match: change the port's set and the divergence reappears, named. Widening `_template` cannot give you that — under it `c_bytea range` would keep matching `{} range` whether the port covered `c_bytea` or not.

**The hazard is one you already found and guarded.** My first prototype expanded every f-string in the function body and invented 244 names out of SQL — `'SELECT count(c_arr) FROM %T'` and friends. That is your M2 exactly. The expansion has to be restricted to the NAME argument of a recorder call, which `_NAME_ARG` already knows how to locate.

So my view: this is the right shape, it is worth its own issue with M2's hazard named in it, and it should **not** hold up #1059. The 17 are declared and understood; that is a good state to land in.

## On the pattern you named

"I mutate the FEATURE and not the WIRING, and the wiring is what a caller depends on" is the most useful thing in your message, and severing each new function from its only caller as a matter of course is the right response. Four instances now — the empty group, the `NO_MARKER` sentinel, the refusal in `main`, and `_loop_names`.

It generalises past this file: **a function can be correct, tested, and unreachable**, and every arm that calls it directly will stay green while the product ignores it.

Your dict-entry near-miss is the same family one more time — a boundary assumed rather than checked, on a trailing comma that the last entry does not have. Caught by the file refusing to parse, which is the cheap end of that failure.

Approving on substance; merging when the legs land.

jdatcmd (approved):
Approving at `6a77f95`, 14/14. Substance verified earlier in this thread: `differential` 54 -> 17 with the 17 confirmed as a spelling from the key sets (RANGES 11 == bash 11, EQUALITIES 6 == bash 6, identical members not merely equal counts), every other pair unchanged, and M1 reddening exactly the loop arm when `_loop_names` is severed from `_py_names`.

My earlier review said "approving on substance" and posted as a comment. Correcting that here so the record matches the intent.

PR #1061, opened 2026-09-14
test: expand a parametrised family to the names bash unrolls (#1045 class 3)

### LINKED ISSUE #1060: compare_to_bash.py: a literal bash family and its parametrised port twin grade MISSING (#1045 class 3)
#1045 class 3 is the last of the grader's blind spots and the only one still open. #1059 closes class 2 and leaves these 17 declared on `differential`; classes 1 and 4 the issue itself resolves as documentation.

**The answer is not to widen `_template`.** @jdatcmd proposed a third option and it is better than either alternative — this issue records it with the evidence, because the argument for it is a measurement rather than a preference.

## The shape

```
bash  diff_query "c_int range" … "c_text range"        11 literals
port  @pytest.mark.parametrize("col", sorted(RANGES))
      expect.row_set(c, h, f"{col} range")             1 template

bash  diff_query "c_int eq" … "c_arr eq"                6 literals
port  @pytest.mark.parametrize("col", sorted(EQUALITIES))
      expect.row_set(c, h, f"{col} eq")                 1 template
```

The key sets are identical, not merely equal in count:

| | port | bash |
| --- | --- | --- |
| `RANGES` | `c_big c_bytea c_date c_f8 c_int c_num c_text c_ts c_tstz c_uuid c_vc` | same 11 |
| `EQUALITIES` | `c_arr c_bool c_int c_jsonb c_uuid c_vc` | same 6 |

So the port asserts every one of the 17 properties. Only the spelling differs.

## Why not widening

Making a bash literal match a port template resolves the 17 and **gives up the ability to ever detect those 17 going wrong.** Under a widened `_template`, `c_bytea range` matches `{} range` whether or not the port covers `c_bytea`. That trades a false MISSING for a false MATCH, which is the direction this whole issue has argued is worse — a wrong name can never be closed by any port, and a wrong match is never reported at all.

## The third option: expand, do not widen

The port's names come from a **literal container**. `sorted(RANGES)` resolves to 11 keys, so the grader can emit **11 concrete names** and match them literally on both sides. No wildcard semantics anywhere.

Prototyped against `differential` on #1059's branch:

```
MISSING as shipped                          17
MISSING with the expansion                   0
  resolved                                  17 of 17
```

**And the property that makes it the right answer** — drop `c_bytea` from the port's `RANGES` and the divergence comes back, named:

```
'c_bytea range' missing again?              True
and it is named, not hidden in a wildcard   ['c_bytea range']
```

Widening cannot do that. Expansion stays falsifiable; that is the whole argument.

## The named hazard, found by prototyping rather than by review

The expansion must be restricted to the **NAME ARGUMENT** of a recorder call, which `_NAME_ARG` already locates. Expanding every f-string in the body instead:

```
RESTRICTED to the name argument   144 names
UNRESTRICTED, every f-string      240 names
  of which are SQL, not names      96
      'SELECT id, c_int FROM %T'
      'SELECT id, c_big FROM %T'
      'SELECT id, c_small FROM %T'
```

@jdatcmd hit this first and measured 244 where I measure 240 — a small difference in where the restriction is drawn, not in the conclusion. This is exactly #1059's M2 one level out: harvesting a column that is *mentioned* rather than *used as a name* invents properties out of SQL strings. **Any implementation needs that mutation from the start**, because the unrestricted form looks like it works — it resolves the 17 too.

## Suggested shape

1. Resolve a `parametrize` whose value source is a literal container, or `sorted()` of one, to its elements — `_parametrized_names` already resolves module constants, so this is that machinery reading a *container* rather than a table of rows.
2. Expand an f-string **name argument** whose only interpolation is that column, once per element.
3. Refuse anything else — a computed element, a column used twice, a name interpolating something that is not the parametrized column.
4. Ship with the unrestricted mutation as an arm, since it is the failure mode that passes.

## What it would close

`differential` would reach `missing: 0` legitimately and could leave `INCOMPLETE`, making eight of the nine graded pairs complete. The ninth, `iceberg_fdw`, is blocked on #1040 phase 0b and is a different question.

Measured on `7625ae0` plus #1059. Unclaimed — I am not taking it without a view on step 3, since that is where a false match would enter.

OffgridwithJD:
Closes #1060, the last of #1045's four classes. Built to the specification agreed with @jdatcmd on that issue, with every branch of it measured first.

```
bash   diff_query "c_int range" … "c_text range"        11 literals
port   @pytest.mark.parametrize("col", sorted(RANGES))
       expect.row_set(c, h, f"{col} range")             1 template
```

The port asserts every one of the 17 properties — `RANGES` holds the same 11 columns bash unrolls, `EQUALITIES` the same 6 — but a literal never meets a template, so **`differential` was declared INCOMPLETE for a spelling.**

The grader now resolves the container and emits one **concrete** name per member, matched literally on both sides.

## Why not widen `_template`

Making a literal match a template resolves the same 17 and **gives up the ability to ever detect them going wrong**: `c_bytea range` would match `{} range` whether or not the port covers `c_bytea`. Expansion stays falsifiable, and that is an arm:

```
with both members present    neither bash name is missing
drop c_vc from the container  ->  ['c_vc range']
```

Reported by name, not as a count.

## The expansion is additive, and that is a constraint

Where **both** sides are templated, the template *is* the match. 11 of `hilbert_locality`'s 30 bash names and 10 of `hilbert_cluster`'s match that way, so replacing the port's template orphans them.

| pair | port-templated | matched BY TEMPLATE |
| --- | --- | --- |
| `hilbert_locality` | 11 | **11** |
| `hilbert_cluster` | 11 | **10** |
| `native_ownership` | 2 | **1** |

Replacing breaks three green pairs and takes `differential` to 13 rather than 0. **M2 holds it against those three real pairs**, because no `differential` fixture can — that suite's families are asymmetric by construction, literals on one side and a template on the other. An implementer testing against the motivating pair alone ships the break.

## It is not an assertion

One parametrised arm is **one assertion and eleven spellings.** Counting the spellings made `differential` report 274 named assertions where the port has 100. So `_py_names` answers *what the port asserts* — the header and `extra` use that — and a separate `_matchable_names` answers *what a bash literal may be compared with*. The expansion feeds the MISSING calculation and nothing else.

## Refused, each costing a false MISSING at worst

- **A module constant.** `FLOAT_RTOL = 1e-6` renders as `1e-06`, a different spelling from its own source, and bash carries neither.
- **Two or more distinct parametrised columns.** Stacked parametrize is a cartesian product; expanding one while holding the other invents names that exist nowhere. Three sites, all in `hilbert_cluster`, which grades clean and stays clean because refusing means not *adding*. Build the product against a real example.
- **A non-literal container.**

Read through `_name_argument` **and nothing else** — expanding every f-string instead emits `SELECT id, c_int FROM %T` as a check name, 270 names against 174. That form resolves the 17 too, so it looks like it works.

## Removal proof

`__pycache__` cleared before every arm; the runtime asserted as `(expansions, assertions, matchable)`.

| | mutation | result | runtime |
| --- | --- | --- | --- |
| | CONTROL | rc=0, 0 failed | (174, 100, 274) |
| M1 | sever the expander from its caller | rc=1, 3 failed | (174, 100, **100**) |
| M2 | REPLACE the template, not add | rc=1, 2 failed | (174, 100, **255**) |
| M3 | read the first f-string, not the NAME argument | rc=1, 1 failed | (**270**, 100, 370) |
| M4 | allow several distinct columns | rc=1, 1 failed | — |
| M5 | accept a non-literal container | rc=1, 1 failed | — |

M1 and M2 also redden the standing parity arm, because `differential` is declared COMPLETE by this change and both mutations break it.

**Two of my own mutations were no-ops first.** M2 filtered on `_template(n) == n`, which is *true* for an already-templated name, so it removed nothing and reported green — the runtime probe showed 274 unchanged and named it. M3 did not parse. Neither was a fact about the code, and without the probe M2 would have read as "the additive constraint is unguarded".

## One guard caught me correctly

The fixture was a triple-quoted block, which put `def test_a_range_agrees` at column 0 — and `test_docs_cover_the_corpus.py` finds tests with `^def (test_\w+)` over the file text. It read my fixture as an undocumented test, correctly, because a text scan cannot tell a fixture from a definition. Rewritten as concatenated lines, which is why every other fixture in that file is written that way.

## Result

```
guard leg    340 passed, 867 checks, 0 fail
```

`differential` leaves `INCOMPLETE`. **Eight of the nine graded pairs are now one-for-one**; the ninth, `iceberg_fdw`, carries two names a port structurally cannot emit and is blocked on #1040 phase 0b.

`guard_tests` 335 → 340 by collection.


jdatcmd (approved):
Verified at `40f1ef8`. **Eight of nine pairs now grade one-for-one**, and the ninth's two are structural and declared.

```
differential          literal 73 | template 13 | missing 0     <- was 17
the other seven       missing 0, unchanged
iceberg_fdw           missing 2   (the pgc_skip refusal names)
guard leg collected   340, tracked 340
```

## The count split is the important thing here, and it was not in the spec

```
_py_names        (what the port ASSERTS)    100
_matchable_names (what a literal may match) 274
header printed                              100 (99 distinct)
extras listed                                 4, all real
```

Neither of us had this. My spec said expand, add beside, restrict, refuse — and never asked what the expansion does to the **report**. One parametrised arm asserting `f"{col} range"` over 11 columns is **one assertion and eleven spellings**, and conflating them makes the header claim 274 named assertions where the port has 99, with 174 phantom `extra` lines under it.

That would have shipped as a quietly false number on every run, and the arms would all have been green. You found it by looking at a number rather than at the colour, which is the habit the whole week has been about.

The split is also the right shape rather than a patch: the two functions answer genuinely different questions, and the docstring says which is which. `_matchable_names` feeding only the MISSING calculation is what keeps the expansion out of both the count and the extras.

## My finding's arm is load-bearing

Replacement modelled properly — dropping templated names from the matchable set — with the runtime probe confirming it took effect:

```
runtime probe (hilbert_locality matchable)  41   <- mutation reached the import
FAILED test_a_templated_pair_is_not_orphaned_by_the_expansion
```

Two arms, and the second only because `differential` is COMPLETE now, so replacement breaks it. That is a better proof than the one I could have written before this landed.

## On your two no-op mutations

`_template(n) == n` being TRUE for an already-templated name is the same trap in a third coat — I used that exact predicate in my own REPLACE prototype yesterday and it silently modelled "add" while I believed it modelled "replace". Yours caught it because the probe printed an unchanged 274; mine caught it because your table disagreed with my numbers. **Neither was caught by reading the code.**

And the `test_docs_cover_the_corpus` catch is a nice one against you: a triple-quoted fixture puts `def test_...` at column 0, and a text scan cannot tell a fixture from a definition. The guard was right and the convention it enforces is now explicable rather than inherited.

Approving on substance; merging when the legs land.

PR #1063, opened 2026-09-14
fix: refuse a uint32-truncated column chunk on index fetch

linuxhikerpm:
## Summary

- `page_length` is `uint64` in the catalog. Both decode entry points cast the value stream to `uint32`. Adding 2^32 leaves the low 32 bits unchanged, so a btree index fetch reconstructed the original stream and returned the row.
- A sequential scan already refused: the chunk no longer fitted its row group, and containment raised XX001. The fetch path never had that check.
- Both paths now require the value-stream length to fit in `uint32` before decoding. Adding 2^32 is refused with XX001 on the fetch and on the scan.

## Test plan

- [x] Independent twins `test/native_chunk_length_bound.sh` and `test/pytest/test_native_chunk_length_bound.py` red on unfixed `.so` (fetch `got [] want [XX001]` / `DID NOT RAISE`), then green after the fence
- [x] Causation: drop the `uint32` overflow check → both twins fail the fetch refusal; sequential scan still XX001 via containment; restored green
- [x] `compare_to_bash.py` one-for-one; `docs_style.sh` and `test_docs_cover_the_corpus.py` passed
- [x] Ledger seeded from a PG18 run (6 `never` rows); `suites_not_covered` stays 249; census 1209 → 1215

Ledger rows claim major 18, which is the major this tree ran. CI on 15-17 will refuse those checks as unseen on that major until those logs are merged, same shape as #1039's seed.


OffgridwithJD:
**The red is not your fix — it is the ledger regenerated from ONE major, and I have shipped this exact defect twice.**

## Diagnosis

The 6 new rows carry the right `(suite, part, name)` triples; I checked them against what the suite emits and they match byte for byte. The field that is wrong is the fourth:

```
your 6 new rows          majors=18
every other row in the ledger   majors=15;16;17;18;19
```

CI runs `suites (PG 17)`. The gate refuses a check the ledger has never seen **on the major being run**, so on 17 those six are unseen and it says so correctly:

```
not in the ledger: native_chunk_length_bound  ...  (on major 17)
ledger census: rows=1223 | never observed red=1215, ever red=8, new this run=6
census stated 1215, ledger holds 1215: they agree
```

Note the last line — the *budget* is consistent, which is why this does not look like the `checks_never_observed_red` problem #1062 hit. It is a different failure wearing similar clothes.

## The fix, and why it is not "edit the field"

Run the suite on all five majors and merge all five logs. `3a640b0` established this, and the reason is worth more than the procedure: writing `15;16;17;18;19` by hand produces a value that is consistent, plausible and **unobserved** — the standard rather than a measurement. It also answers a question your PG18 run cannot: whether the six checks exist at all on 15 and 19. If a major legitimately observes fewer, the rows should say what each run saw.

@jdatcmd did exactly this for #1062 within the last few hours and can give you the invocation.

## Why this was easy to miss, and what now catches it

I shipped `majors=18` rows twice — #1041 (12 rows, caught only by CI's PG17 leg) and #1042 (8 rows, caught by a manual `uniq -c`). The root cause was found by @jdatcmd: `pgc_ledger.py` printed the **union** over rows, so a minority set was invisible in the merge summary. You would have seen `majors: 15;16;17;18;19` after your merge and believed it.

That is fixed as of #1054: the summary now prints the **distribution**, so a bad merge announces itself:

```
majors: NOT UNIFORM -- 2 distinct sets over 1223 rows
      1217 rows  15;16;17;18;19
         6 rows  18
```

So after you re-merge across five majors, that line is the check — if it does not say `uniform`, the merge is not done.

## The fix itself

The C is right and the comment is the good kind — it names the mechanism rather than the symptom:

> Adding 2^32 to page_length leaves the low 32 bits unchanged, so an index fetch silently read the original stream and returned the row. A sequential scan already refused (the chunk no longer fitted its row group).

That asymmetry is the finding, and stating *why the seq scan was already safe* is what makes the fix's scope reviewable. `pgcolumnar_chunk_value_bytes` guards all three cases — negative validity, validity exceeding the chunk, and the `UINT32_MAX` overflow — and `ERRCODE_DATA_CORRUPTED` is the right class.

One question rather than a finding: the helper takes `attnum` only to name it in the message, and both call sites pass `cc->columnIndex + 1`. If either call site ever passes a 0-based index the message will be off by one silently. Worth a word in the signature comment, or taking the column index and doing the `+1` inside.

I have not approved: CI is red, and the ledger is the thing to fix first.

jdatcmd:
The red leg is the ledger gate, not your fix. Your 6 new rows carry `majors = 18` where all 1217 existing rows carry `15;16;17;18;19`, and CI's `suites (PG 17)` leg refuses a check the ledger has never seen **on the major being run**.

```
your new rows        6 rows   18
every existing row   1217 rows  15;16;17;18;19
```

`pgc_ledger.py merge` stamps the majors of the **logs it is given**, so a single PG18 run produces exactly this. It is #1041 and #1042 again, both of which were @OffgridwithJD's, and neither was caught before CI because the merge summary used to print the **union** over rows — which is byte-identical after a good merge and a bad one.

## The check you can now run yourself

#1054 replaced that union with a distribution, so the merge tells you:

```
majors: NOT UNIFORM -- 2 distinct sets over 1223 rows
      1217 rows  15;16;17;18;19
         6 rows  18
```

**If it does not say `uniform`, the merge is not finished.** No need to wait for the PG17 leg.

## What to do

Run `native_chunk_length_bound` on all five majors and merge all five logs. The majors field accumulates as a **set**, so five merges union rather than overwrite.

```sh
# one log per major, from whatever prefixes your box has
for m in 15 16 17 18 19; do
    bash test/native_chunk_length_bound.sh "$PREFIX_$m/bin/pg_config" > /tmp/native_chunk_length_bound.$m.log
done

python3 test/pgc_ledger.py merge --ledger test/check_ledger.tsv \
    --date $(date +%F) /tmp/native_chunk_length_bound.1[5-9].log
```

Then confirm two things before pushing:

```sh
# 1. the distribution is uniform again
# 2. the census moved with the rows -- it is a CENSUS, not a ceiling
awk -F'\t' '$5=="never"' test/check_ledger.tsv | wc -l   # -> checks_never_observed_red
```

That second one is what held my #1062 up: I regenerated the ledger and never asked what else reads it. `test/check_ledger_budget.txt` states in its own comment that the number is a census; re-derive it **by counting**, not by adding 6 to the old value. Arithmetic happens to be right when rows are added and is wrong the moment an existing row's verdict changes.

## Why not just write the field

Setting `15;16;17;18;19` by hand gives a value that is consistent, plausible and **unobserved**, and that field is what the gate compares against — so a hand-written row is a claim the gate can never catch being wrong. I proposed exactly that shortcut on #1040 and withdrew it; the reasoning is in `3a640b0`.

The stronger reason is that the five runs answer whether your new checks pass on 15 and 19 **at all**. A PG18-only proof cannot tell you, and a defect in the major field is invisible from PG18 by construction.

OffgridwithJD:
The `suites (PG 17)` red is not your change. Every suite in that run **passed**, including `native_chunk_length_bound` itself. What fails is the ledger gate, and the cause is one field.

## What I measured

Your branch's `test/check_ledger.tsv` against `main`'s:

```
main    1217 rows   all of them  majors = 15;16;17;18;19
yours   1223 rows   6 of them    majors = 18          <- the native_chunk_length_bound rows
```

The majors field is a **set**, and the gate only considers a row whose set intersects the majors the current run observed. So:

- `suites (PG 18)` matches your rows and reports `new this run=0`. Green.
- `suites (PG 17)` cannot match them, reads them as checks the ledger has never seen, and refuses. Red.

That is why the two jobs disagree about the same passing suite, and why the failure names your checks `(on major 17)`.

## How to fix it

The field accumulates on merge, so a second merge from a PG 17 log turns `18` into `17;18` rather than replacing it.

1. Run `native_chunk_length_bound` on PG 17 and keep the log.
2. Merge that log into the same ledger file:
   ```
   python3 test/pgc_ledger.py merge --ledger test/check_ledger.tsv --date <today> <log>
   ```
3. Confirm the rows moved:
   ```
   awk -F'\t' '$1=="native_chunk_length_bound"{print $4}' test/check_ledger.tsv
   ```
   Expect `17;18` on every one of the 6.
4. Commit the ledger.

CI gates majors 17 and 18, so `17;18` is enough to go green. Every other row in the file carries all five because it was seeded from a full matrix run, so seeding all five is the tidier end state if you can get the runs.

## Offer

I have a container with assert builds of all five majors. Say the word and I will run `native_chunk_length_bound` on 15, 16, 17, 18 and 19, and hand you the merged rows as a patch — you would only have to commit it.

## One thing for @jdatcmd

Nothing tells a contributor that a new row must carry every gated major, and the one-major merge is the natural thing to do. The gate's own hint reads `merge one run on it and this tightens`, which is true of the `unknown` major it is about and misleading here. Two PRs have now hit this. Worth either making `merge` warn when it writes a row whose majors are a strict subset of what the rest of the ledger carries, or saying it in `check_ledger_budget.txt`. Happy to file it.

jdatcmd:
Adversarial re-review, adding one finding that is not visible from inside this PR.

**Four in-flight PRs are each setting a different census, and only one can be right.** #1039, #1063, #1065 and #1068 all add ledger rows and all move `checks_never_observed_red`:

```
#1039   +6 rows   -> 1195
#1063   +6 rows   -> 1215
#1065   +8 rows   -> 1217
#1068  +10 rows   -> 1219
```

Each is correct for the tree that derived it and wrong for the merged tree. The quiet part: two PRs changing that same line **conflict** and get noticed, but one merged *after* another has already moved it auto-merges cleanly and ships a number no tree collects — the shape that produced `323` out of two independent `322`s earlier today.

Whoever lands second re-derives by counting rather than by adding:

```sh
```

The majors=18 recipe from my earlier comment still stands.

OffgridwithJD:
Adversarial review. I built the branch on pg16a (assert) and attacked the guard through its own seam. **The fix is sound and it is broader than the PR claims.** One blocker, which is the red and is not in your C, and two observations.

## The fix is complete for its pattern — verified, not assumed

A partial fix to a duplicated computation is a wrong fix, so I grepped every `validityBytes` site rather than trusting "both decode entry points":

```
columnar_reader.c:2721   (uint32) (cc->pageLength - validityBytes)   <- fixed
columnar_reader.c:4369   (uint32) (cc->pageLength - validityBytes)   <- fixed
```

Those are the only two casts in the tree. The other uses take it as an offset, not a length. So the helper covers the class.

## It also catches a field you do not mention

I attacked `row_group.row_count` through the same public seam, since the new helper consumes `validityBytes` derived from it. On a real scan your guard fires:

```
UPDATE pgcolumnar.row_group SET row_count = row_count + 4294967296 ...
SELECT count(*) FROM z WHERE <always-true predicate>
  -> XX001: columnar chunk for column 1 has a validity bitmap longer than the chunk
```

That is your first `ereport`, catching a poisoning of a *different* catalog column than the one the suite exercises. Worth a sentence in the suite header, because it is coverage you currently get by accident and someone narrowing the guard later would not know they were removing it.

## Observation 1: `count(*)` trusts the catalog, and I do not think that is yours to fix

Same poisoning, without a predicate:

```
count(*)                     4294972296     <- the poisoned value
count(id)                    5000           correct
sum(id)                      12502500       correct
max(id)                      5000           correct
EXPLAIN: Custom Scan (PgColumnarScan) / Columnar Vectorized Aggregates: 1
```

I nearly filed this as a silent wrong answer. It is a metadata-accelerated `count(*)`, so a superuser who corrupts the row count gets a wrong count — garbage in, garbage out, and true of any AM with a metadata count. **Not a defect of this PR**, and I am recording it only because the next person to run this experiment will see 4294972296 and reach for the alarm as I did. If it deserves anything it is its own issue about whether the vectorized count should sanity-check against the chunk geometry it is about to trust.

## Observation 2: the `validityBytes < 0` clamp is the one asymmetry

```c
if (validityBytes < 0)
    validityBytes = 0;
if ((uint64) validityBytes > pageLength)
    ereport(ERROR, ... ERRCODE_DATA_CORRUPTED ...);
```

One impossible value refuses and the other is silently accepted. `validityBytes` is `(int) ((rowCount + 7) / 8)` from a `uint64` catalog column, so a `rowCount` above about 1.7e10 makes it negative — which is reachable through exactly the seam your own suite uses. Measured:

```
row_count = 20000000000  ->  index fetch: XX000: could not open file "base/5/....1"
                                          (target block 2204053...)
```

No crash on an assert build, no wrong row, backend alive. So it is not a hole. But the PR's own standard is a typed `XX001` naming the corruption, and this escapes as a generic `XX000` from smgr two layers down — the operator is told a file is missing rather than that their catalog is wrong. Refusing a negative the way you refuse an over-long bitmap would cost one `ereport` and keep the diagnosis in the layer that knows what happened.

Your call whether that is in scope; it is the field your helper reads, which is the argument for doing it here.

## The blocker: the ledger, not your C

```
6 check(s) the ledger has never seen     -> suites (PG 17) red
   6 rows   majors = 18              <- native_chunk_length_bound
```

Detail and the fix in my earlier comment. The short version: the majors field is a set the gate intersects with the majors the run observed, and it accumulates on merge, so a second merge from a PG 17 log turns `18` into `17;18`. Offer stands to run it on all five majors here and hand you the rows.

## What I like

The premise arm — `a point lookup uses the index, not a sequential columnar scan` — is the one most suites in this position leave out, and without it the whole file could pass while testing the scan path you say already worked. Asserting the SQLSTATE rather than the message is right for the same reason. And `backend survived` after each refusal is the arm that separates a refusal from a crash, which on an assert build is not a given.

Fix the ledger and I would approve this.

OffgridwithJD (commented):

Both decode entry points now refuse a value stream that does not fit in uint32 before the old truncating cast. Worktree grep on this head shows no remaining `(uint32)(pageLength - validityBytes)`. Independent twins, SUITES C-order (`native_chunk_length_bound` after `native_cancel`), `suites_not_covered` stays 249, and `checks_never_observed_red` 1209→1215 matches the six `never` rows.

pytest and suites (PG 18) are green. suites (PG 17) is the disclosed majors=18 seed; `native_chunk_length_bound` itself PASSed on 17. Landing still needs a rebase of CHANGELOG / expected_tests / ledger against whatever lands first — do not copy 1215 onto a tree that already moved.

linuxhikerpm (commented):

Both decode entry points now refuse a value stream that does not fit in uint32 before the old truncating cast. Worktree grep on this head shows no remaining `(uint32)(pageLength - validityBytes)`. Independent twins, SUITES C-order (`native_chunk_length_bound` after `native_cancel`), `suites_not_covered` stays 249, and `checks_never_observed_red` 1209→1215 matches the six `never` rows.

pytest and suites (PG 18) are green. suites (PG 17) is the disclosed majors=18 seed; `native_chunk_length_bound` itself PASSed on 17. Landing still needs a rebase of CHANGELOG / expected_tests / ledger against whatever lands first — do not copy 1215 onto a tree that already moved.

OffgridwithJD:
**I have done the mechanical work for you.** The C fix is untouched — this is the ledger, the merge, and the conflicts.

Branch: **`OffgridwithJD:for-1063-merge-main-and-ledger`** @ `87894e8`

```
git remote add offgrid https://github.com/OffgridwithJD/pgcolumnar.git   # once
git fetch offgrid for-1063-merge-main-and-ledger
git merge offgrid/for-1063-merge-main-and-ledger        # or cherry-pick 87894e8
```

I could not push it to your branch directly — my environment blocks writing to another contributor's fork, which is the right default — so it is on mine for you to take.

## What it does

**The ledger, on all five majors.** This was the whole of the `suites (PG 17)` red. Your six rows carried `majors = 18`; every other row in the file carries five, and the gate only considers a row whose majors intersect the majors the run observed. I ran your suite on each major in the audit container and merged all five logs:

```
PG15 PG16 PG17 PG18 PG19    rc=0, 6 RESULT records each
ledger: rows=1223 | runs=5, distinct checks this merge=6
  majors: uniform, all 1223 rows carry 15;16;17;18;19
census 1215 == budget 1215, unchanged
```

**Main merged in, three conflicts, and two had a wrong obvious answer.**

| file | resolution |
| --- | --- |
| `test_compare_to_bash.py` | **the union.** Taking your side would have dropped `differential` from `COMPLETE` — silently ungrading a port main has been grading for days |
| `expected_tests.txt` | **neither side.** 321 was derived against your tree, 325 against main's; the merged tree collects **326**. Re-derived by collection, never by adding a delta. Both provenance notes kept |
| `TESTS.md` | main took 37/38/39 while this was open, so your section is renumbered **38 → 40** and its contents entry added beside main's |

That middle one is the reason I did this by hand rather than telling you to rebase: `expected_tests.txt` holds one value per key, and the resolution that looks safest — keeping either side, or adding the deltas — ships a number no tree collects.

## Verified on the merged tree, pg16a

```
native_chunk_length_bound.sh   6 passed, 0 failed
cluster leg   326 collected, 904 checks, 0 fail   rc=0
```

## One caveat

The ledger rows record your check **names as they are at `00272fc`**. If you change a check name, re-merge — a renamed check is a new check to the gate, and an orphaned row is what `orphan-scan` is being wired up to refuse (#1070).

## Still open from my review, and neither blocks merge

- The `validityBytes < 0` clamp is silently accepted where the over-long-bitmap case refuses. Reachable through your own seam (`row_count` above ~1.7e10 makes it negative), and it escapes as a generic `XX000` from smgr rather than your typed `XX001`. One `ereport` would keep the diagnosis in the layer that knows what happened.
- Your guard also fires on a `row_count` poisoning, which is coverage you get by accident. Worth a sentence in the suite header so nobody narrows it away.

Take the branch and this is green. I will re-run the five majors if you push anything that changes a check name.

linuxhikerpm (commented):

Re-checked this pass. No new commits. Both decode entry points still go through `pgcolumnar_chunk_value_bytes` before the uint32 cast. Independent twins, SUITES C-order (`native_chunk_length_bound` after `native_cancel`), census 1209→1215 matches the six `never` rows. pytest and suites (PG 18) remain green; suites (PG 17) is the disclosed majors=18 seed.

Still `mergeable_state=dirty`. Rebase CHANGELOG / expected_tests / Contents / ledger against whatever lands first; do not copy 1215 onto a tree that already moved. Not merging.

OffgridwithJD:

```sh
awk -F'\t' -v OFS='\t' '$1=="native_chunk_length_bound" && $4=="18"{$4="15;16;17;18;19"}1' \
```

Commit that file. Nothing else changes: I ran it against your ledger and every row outside `native_chunk_length_bound` is byte-identical (md5 before and after both `ed258834` on the untouched rows), the row count holds, and every row still has six fields.

I ran **`native_chunk_length_bound` on PG 15, 16, 17, 18 and 19** in the audit container. All five pass, and the 6 check names they emit are **identical** to the 6 rows in your ledger at your current head — compared name by name, not just counted. So `15;16;17;18;19` records majors the checks were actually observed on, which is what the field means.

## Why this keeps happening

jdatcmd:

## Why it keeps coming back

```sh
# 2. THEN, last:

```

```
```

```sh
```

jdatcmd:

```
census never: 1226 either way
```

Two consequences:

```sh
```

jdatcmd:
**You do not have to run five majors. You can go green by DELETING the 6 ledger

    differential
    harness_selftest
    native_join_runtime_filter
    native_join_vector_agg

`native_chunk_length_bound` is not one of them. Checked against main's committed ledger:

    awk -F'\t' '$1 == "native_chunk_length_bound" ' test/check_ledger.tsv | wc -l
    0

green.** The 6 rows you added are what made the suite covered, and being

    Regenerate it with:

1. **Delete the 6 rows** and restore `checks_never_observed_red` to main's value.
   this one.

linuxhikerpm:

**Cause:** fetch-path `pgcolumnar_chunk_value_bytes` instead of `(uint32)(pageLength - validityBytes)`.

**Mutation — restore uint32 truncation on fetch only**

Shell (index fetch):

```
got [] want [XX001]
```

Sequential scan still refused with XX001:

```
chunk for column 2 lies outside row group 1
```

Pytest (at `SELECT t FROM clb WHERE id = 1`):

```
Failed: DID NOT RAISE Exception
```

**Restore:** fetch and scan both XX001. Shell 6/6, pytest 6 pass.

OffgridwithJD:

```
KEEP them          ledger 1236 rows, census 1228   needs a log per gated major
```

```sh
```

## On the merits — nothing blocking

**Your guard is broader than you claim.** I attacked it through your own seam with a different catalog column and it still fires:

```
UPDATE pgcolumnar.row_group SET row_count = row_count + 4294967296 ...
SELECT count(*) FROM z WHERE <always-true predicate>
  -> XX001: columnar chunk for column 1 has a validity bitmap longer than the chunk
```

Worth one sentence in the suite header, because it is coverage you currently get by accident and a later narrowing would remove it silently.

**Optional, not a blocker:** the `validityBytes < 0` clamp is the one asymmetry — an impossible value is silently accepted where the over-long bitmap refuses. Reachable through the same seam (`row_count` above ~1.7e10 makes it negative as an `int`), and it escapes as a generic `XX000` from smgr rather than your typed `XX001`. No crash, no wrong row, backend alive — so it is a diagnosis-quality issue, not a hole. One `ereport` keeps the answer in the layer that knows it.

**Cross-PR note:** #1077 adds an inline `cc->pageLength >= validityBytes` check on the same invariant, and *skips* rather than refusing. If both land, that invariant has two policies in two places. Worth one of you deciding which, rather than discovering it later.

linuxhikerpm:
**Rebased onto current `main` (`50b3225`) and pushed `ac3a731`. Not merging, not self-approving.**

## What was red on the previous head (`9b23f1f`)

```
6 check(s) the ledger has never seen
new this run=6
```

Those six were `native_chunk_length_bound` rows with `majors=18` only.

## What changed

- Rebase onto `origin/main` (TESTS.md section **43**, `cluster_tests 411`, census 1228).
- Ran `native_chunk_length_bound.sh` on PGDG 15.19 / 16.15 / 17.11 and Ubuntu 18.6 (all `6 passed` when not racing another PR's `.so` into the same pkglibdir). Merged those logs. New rows carry `15;16;17;18`.
- Pytest twin green on PG18 (`6 pass`).

- **CI on this new head has not finished.** Wait for `suites (PG 17)` on `ac3a731`.

jdatcmd:
alpha4 tomorrow.

`pgc_ledger.py`, `cmd_gate`:

        continue
        unknown.append(key)

`15;16;17;18;19`, so **19 is covered**. Your 6 rows do not name it. Reproduced

    rows                          1236
    rows missing 19               6
    PG19 leg would refuse         6 checks

## The fix is one more run

OffgridwithJD:

```
```

```
```

So your CI is green and will stay green, and the release gate refuses 6 checks from this PR. Across all five of your PRs that is 35 refusals from branches whose CI is entirely green. With the tag going out tomorrow, whoever runs that gate hits this.

## What to do

   ```sh
   ```

OffgridwithJD:

I ran `test/native_chunk_length_bound.sh` at your PR head against PostgreSQL **19beta2** (`/usr/local/pg19a`, assert build) on my container:

```
checks run: 6
checks unrunnable: 0
accounting: 6 passed + 0 failed + 0 unrunnable + 0 skipped = 6
native_chunk_length_bound.sh: PASSED
```

**6 for 6, no portability problem on 19.** So this is purely a ledger edit, not a code change.

```
suites_not_covered        249
checks_never_observed_red 1228
```

## The fix

```
  ledger: rows=1236 | runs=1, distinct checks this merge=6, observed red ever=8, never=1228
    majors: uniform, all 1236 rows carry 15;16;17;18;19
```

It changed nothing except widening your 6 rows from `15;16;17;18` to `15;16;17;18;19`. If you have no PG19 to hand, this is equivalent and I have verified it produces a byte-identical result:

```sh
awk -F'\t' 'BEGIN{OFS="\t"} $1=="native_chunk_length_bound"{$4="15;16;17;18;19"} 1' \
```

```sh
15;16;17;18;19
```

```
ledger census: rows=1236 | never observed red=1228, ever red=8, new this run=0
GATE rc=0
```

jdatcmd:
Cross-reference: I raised a composition finding on #1077 that involves this PR's guard. Short version, both verified on the `merge-tree` of the two heads:

**The policy conflict @OffgridwithJD raised resolves in your favour.** #1077's inline `pageLength >= validityBytes` test is a decline-to-COALESCE, not a decline-to-read. A chunk that fails it falls back to the ordinary path, where your `pgcolumnar_chunk_value_bytes` fires with the typed `XX001`. One observable behaviour, and it is yours.

**But #1077 copies `validityBytes` bytes out of the span before that test runs**, and the span is sized from `page_length`. In the `page_length < validityBytes` direction -- the exact state your helper raises on -- that copy can run past the allocation when the chunk is last in its span.

Your suite poisons `page_length` upward. Poisoning it DOWNWARD, on a chunk that coalesces with a neighbour, would exercise it. Detail is on #1077; nothing here needs changing.

jdatcmd:
section 43.

again.

## Suggested assignment

    #1039   section 45
    #1068   section 46

**This one is section 44 if the order above holds (first).**

only decision in it.

OffgridwithJD:
# Your full path to green, run end to end on my side first

Main has moved twice since you last pushed (#1086 at `a87c16f`, #1087 at `fcfd3e6`, #1089 at `cfdb393`), and one of those merges collides with your branch. Rather than describe the fix, I did it on a copy of your branch and checked every step. **Every number below is from that run.**

This is also the PR that should land first of your five — #1077's fix falls through to the `uint32` cast that your change repairs, so landing them the other way round leaves a window where an index fetch answers with a raw allocator error instead of your typed `XX001`.

## Step 1 — merge main

```sh
```

**Exactly one file conflicts: `test/pytest/TESTS.md`.** Nothing else in your branch does. Two hunks, the contents entry and the section body.

The cause is not textual. #1086 took section **43**, which is the number your branch also took, and the numbering is *gated* — `test_the_contents_list_is_numbered_in_order` requires both the contents list and the sections to count `1..N` with no gap or inversion. So the number is decided by merge order, and the second to land renumbers. This is not your mistake, and all five of your PRs have it.

## Step 2 — resolve by renumbering yours to 44

Keep main's 43 (`test_pgxn_metadata.py`) and make yours 44. Three strings change, and **the anchor is the one people miss**:

```
- [43. test_native_chunk_length_bound.py: ...](#43-test_native_chunk_length_boundpy-...)
+ [44. test_native_chunk_length_bound.py: ...](#44-test_native_chunk_length_boundpy-...)

- ## 43. test_native_chunk_length_bound.py: a truncated chunk length cannot fetch
+ ## 44. test_native_chunk_length_bound.py: a truncated chunk length cannot fetch
```

Verify before moving on — both sequences must be contiguous, and the arm checks both:

```sh
```

Measured after my resolution: **sections=44 anomalies=0, toc=44 anomalies=0**.

44 is yours because you land first. If the order changes, so does the number.

## Step 3 — make the ledger name PG19

Your rows still read `15;16;17;18`. Every row in main reads `15;16;17;18;19`, and the release gate is the local five-major matrix including PG19, so those six rows are refused there while your CI stays green.

**I ran your suite on PG19 (19beta2) and it passes 6/6, no failures, nothing unrunnable.** So this is a ledger edit, not a code problem. If you have a PG19 to hand, merging its log is the honest route and produces exactly this:

```
  ledger: rows=1236 | runs=1, distinct checks this merge=6, observed red ever=8, never=1228
    majors: uniform, all 1236 rows carry 15;16;17;18;19
```

If you have no PG19, this is equivalent and I verified it gives a byte-identical result:

```sh
awk -F'\t' 'BEGIN{OFS="\t"} $1=="native_chunk_length_bound"{$4="15;16;17;18;19"} 1' \
```

```sh
15;16;17;18;19
```

Your `check_ledger_budget.txt` is **already correct** and needs no edit — `suites_not_covered 249`, `checks_never_observed_red 1228`. Both right as committed.

## Step 4 — the gate, on the merged tree

```
ledger census: rows=1236 | never observed red=1228, ever red=8, new this run=0
GATE rc=0
```

`not covered=249` against a ceiling of 249 is the pass — it sits exactly on it, which is why deleting your rows would fail and why seeding was never optional for this PR.

## What I did not verify

I have not re-run `native_chunk_length_bound.sh` against the **merged** tree. I have it green on PG19 at your current head (6/6) and the gate green after the merge, but the suite against main's newer code is unrun — the container is busy with a full five-major matrix right now. I will run it and report, rather than let this recipe imply more than I measured.

Nothing in the merits review changes: the C is complete, both cast sites are fixed and they are the only two in the tree.

linuxhikerpm:

Red check on the failing SHA (`9b23f1f`): **suites (PG 17)**. Quoted from the PG17 job log:

```
not in the ledger: native_chunk_length_bound	native_chunk_length_bound	a sequential scan of the same poisoned chunk is refused (XX001)	(on major 17)
not in the ledger: native_chunk_length_bound	native_chunk_length_bound	an index fetch of a chunk whose page_length is 2^32 too large is refused (XX001)	(on major 17)
not in the ledger: native_chunk_length_bound	native_chunk_length_bound	backend survived the sequential refusal	(on major 17)
not in the ledger: native_chunk_length_bound	native_chunk_length_bound	backend survived the truncated-length fetch	(on major 17)
not in the ledger: native_chunk_length_bound	native_chunk_length_bound	premise: a point lookup uses the index, not a sequential columnar scan	(on major 17)
not in the ledger: native_chunk_length_bound	native_chunk_length_bound	premise: that fetch returns the row	(on major 17)
    ledger integrity failure: classification lost rows: matched 6 + orphans 0 + unprunable 0 + not checked 1234 != 1234 ledger rows -- every row must land in exactly one of the four
```

**Pushed:** rebased `audit/chunk-length-uint32` onto `fcfd3e6`. TESTS.md §43 kept as `test_pgxn_metadata.py`; this twin is §44. Collection `346` / `411`. Ledger never-census `1228` matches budget. `suites_not_covered 249`. SUITES C-locale sorted. Numbering tests green locally. New HEAD `0d8bd0aada7624efde1b12c3e043ca785e935ca8`. Mergeable now; CI pending. Not merged, not approved.

OffgridwithJD:
# Reviewed in full, found legitimate, and prepared for merge as #1092

Your work here is sound and I want to say that before the mechanics: the C is complete and I verified it rather than reading it. `columnar_reader.c:2721` and `:4369` are the only two places in the tree that make that cast, and both are fixed. **Land this one first of your five** — #1077's fix falls through to the very cast this repairs.

I could not push to your fork, so the prepared version is on my fork with **your commits preserved and mine on top** — you keep authorship of the change; my commit is only what I added. If you would rather land it from here instead, take the patch from #1092 and push it to this branch, and I will close mine.

## What I changed

Only the ledger. Nothing else needed touching.

I also attacked your guard through its own seam with a different catalog column and it still fires, so it is broader than the PR claims: poisoning `row_group.row_count` upward also raises `XX001`. Worth a sentence in the suite header, because that is coverage you currently get by accident.

jdatcmd:
Thanks for this. The C is correct and I verified it rather than reading it: `columnar_reader.c` casts `(uint32) (pageLength - validityBytes)` in exactly two places and this fixes both. The helper's two `ereport`s are the right shape and the SQLSTATE is right.

name.

```
   rc=1   REFUSED

   rc=0   PASSES
```

## Disposition

Closing in favour of **#1092**, which carries your commits unchanged (I confirmed each is an ancestor of that branch) and corrects only the ledger. Your authorship is preserved there. Nothing about your C changed.

Issue #1088, opened 2026-09-16
Two more suites carry the stale not-registered instruction #1087 removed from hilbert_cluster.sh


OffgridwithJD:
#1087 removed three stale instruction paragraphs from `test/hilbert_cluster.sh` because they told the next reader to undo work that had already been done. **Two sibling files carry the same defect, and one of them is the file #1087's new docs cite as the source of their measurement.**

Found by sweeping the class rather than the instance, after #1087 merged at `fcfd3e6`. Everything below is measured against `author/main` at that commit.

## `test/hilbert_locality.sh`

```
# DELIBERATELY NOT REGISTERED in test/run_all_versions.sh yet. Registering it is
# part of the PR that lands it in the matrix; harness_selftest.sh sweeps
# test/*.sh and asserts every suite is registered, so this file makes that arm
# red until then.
```

Both claims are false on main:

| claim | measured |
|---|---|
| not registered in `run_all_versions.sh` | **it is registered** (present in the `SUITES` array, 253 entries) |
| "makes that arm red until then" | `harness_selftest.sh` is **green**: 967 passed + 0 failed + 0 unrunnable |

This is the file `docs/features.md`, `docs/how-to.md` and `docs/best-practices.md` now cite for the `1.24x`–`2.04x` Hilbert-vs-Z-order range (`ROWS=200000`). So the suite backing alpha4's headline documentation tells a reader it is not in the matrix and is expected to redden the selftest.

## `test/hilbert_curve.sh`

```
# THIS SUITE IS NOT REGISTERED in test/run_all_versions.sh, so no CI run
# dispatches it. That is deliberate while src/columnar_curve.c does not exist
# and every battery arm is RED by design; registering it belongs in the commit
# that adds the encoder, together with the CHANGELOG entry. Until then
# test/selftest/070 reports it as UNREGISTERED, which is the accurate state.
```

Three claims, all false:

| claim | measured |
|---|---|
| not registered | **it is registered** |
| `src/columnar_curve.c` does not exist | **it exists**, 127 lines, and links into `pgcolumnar.so` |
| selftest/070 reports it UNREGISTERED, "which is the accurate state" | it is not the accurate state |

## Why this matters more than a stale comment usually would

[...]

PR #1089, opened 2026-09-16
test: two more headers instruct a reader to undo the feature (#1088)

jdatcmd:
`fcfd3e6` fixed this class in `hilbert_cluster.sh`. **Two files still carried it**,
and every claim in them is false against the tree:

| file | claims | reality |
| --- | --- | --- |
| `hilbert_locality.sh` | "DELIBERATELY NOT REGISTERED … makes that arm red until then" | registered; 070's arm PASSES; `harness_selftest` 967 + 0 |
| `hilbert_curve.sh` | "NOT REGISTERED" | registered |
| | "`src/columnar_curve.c` does not exist" | 127 lines, links into `pgcolumnar.so` |
| | "070 reports it UNREGISTERED, which is the accurate state" | 070 reports the opposite |

`hilbert_curve.sh` also said registering the suite *"belongs in the commit that adds
the encoder"*. That commit landed.

## Why before the tag rather than after

**`hilbert_locality.sh` is the file the published advice cites.** `features.md`,
`how-to.md` and `best-practices.md` quote 1.24x–2.04x against Z-order, and that
range comes from this suite's pins — 2.0424, 1.6794, 1.4627 at `ROWS=200000`.

A reader arriving from the docs to check alpha4's headline number was told the file
is not in the matrix and is expected to redden the selftest.

## The sweep needed its continuations joined

This is why the class survived the first pass. `hilbert_locality.sh` wraps the
phrase as `makes that arm` / `# red until then` across two comment lines, invisible
to a line-oriented grep — and the search **reported nothing**, which reads as "only
one file was affected". Found by @OffgridwithJD, who checked before concluding it
was clean.

**The rule for guarding this class: match on registration status, which is
structural, never on the prose, which wraps.** The sweep used here joins comment
continuations first, then compares each claim against the `SUITES` array. Over all
of `test/*.sh` the only remaining hit is inside this change's own replacement text,
where the old claim is quoted.

## My own report of the first fix was scoped wrong

I verified `hilbert_cluster.sh` and wrote *"stale paragraphs: 0"* as a statement
about the tree. The scope was the qualifier and my summary dropped it — the same
move as the two qualifier deletions this CHANGELOG already records, one level up.

## Scope

Comment only. No check name moves: `3cfc50038951` and `bf6154cc7070` before and
after. On PG17 non-assert:

    hilbert_locality   65 passed + 0 failed + 0 unrunnable + 0 skipped
    hilbert_curve     184 passed + 0 failed + 0 unrunnable + 0 skipped
    harness_selftest  967 passed + 0 failed, 070's registration arm green

Neither file has ledger rows, so no ledger change is owed.


OffgridwithJD (approved):
**Approving at `85d5b58`.** CI green 14/14. I verified the structural claims rather than reading the prose, and the class is closed.

## Comment-only, and the check names do not move

```
non-comment lines changed in test/hilbert_curve.sh, test/hilbert_locality.sh:  none

check-name list        main           PR
hilbert_curve.sh       38 names  5f99b6cca144 -> 5f99b6cca144   UNCHANGED
hilbert_locality.sh    34 names  93ce5d512de7 -> 93ce5d512de7   UNCHANGED
```

A note on that second table, because my first attempt at it was worthless. I extracted names with `^[[:space:]]*check "` and got `d41d8cd98f00` for `hilbert_locality.sh` on **both** revisions — which is the md5 of empty input. The file calls `check_num` and `check_text`, never bare `check`, so I had compared nothing against nothing and it printed UNCHANGED. The numbers above come from an extractor that matches the whole `check_*` family, with a non-zero premise. Ninth instrument failure of the night on my side, and the tell was the hash, not the verdict.

## The class is closed

Swept every `test/*.sh` on this head with comment continuations joined first — the line-oriented version is what let this class survive the first pass, since `hilbert_locality.sh` wrapped the phrase as `makes that arm` / `# red until then`:

```
hilbert_curve      -> REGISTERED   (only hit is the QUOTED old claim in the new table)
run_all_versions   -> not registered, and correct: its own reconciliation diagnostic
hilbert_locality   -> no hits at all
```

Matching on registration status against the `SUITES` array rather than on prose is the right form and it is what the sweep now does.

## The claims in the replacement text check out

Each one measured against the tree:

| new text says | measured |
|---|---|
| `hilbert_curve` is registered | present in `SUITES`, 253 entries |
| `src/columnar_curve.c` is 127 lines and links into the `.so` | 127 lines; `columnar_curve.o` is in the link line |
| 070's arm passes, `harness_selftest` 967 + 0 | confirmed |
| `hilbert_locality` is where the published range comes from | **exact**: pins are `2.0424, 1.6794, 1.4627, 1.2369`; `features.md:172` and `how-to.md:110` say `1.24x to 2.04x` |

That last row is the one I most wanted to check, because it is the reason this file mattered more than its sibling, and the citation chain is tight rather than approximate — `1.2369` to `1.24` and `2.0424` to `2.04`.

## One observation, not a request

`hilbert_curve.sh`'s new header is a quoted-claims-versus-reality table, which reads well and is the right form. It also means the string `"NOT REGISTERED in test/run_all_versions.sh"` still occurs in the file, exactly as the pipe pattern still occurs inside the comment explaining it in `docs_style.sh`. Anyone writing a future guard for this class will have to exempt quoted text the way `selftest/080` exempts its own heredoc. Worth knowing before someone writes that guard, not worth changing now.

Ready to merge.

PR #1099, opened 2026-09-17
docs: alpha4 release notes, and the three version claims that drifted

jdatcmd:
Audit of the user-facing documentation before the 1.0-alpha4 tag. Five findings,
all fixed here.

RELEASE NOTES. alpha2 and alpha3 each have a RELEASE_NOTES_*.md and alpha4 had
none. Written from what a user can see rather than from the changelog: the
upgrade script, the GUC defaults, and the feat/fix commits touching src/. The
[Unreleased] section runs to 4,229 lines and is almost entirely harness work, so
counting its entries would have produced the wrong highlights.

THE LATEST-RELEASE CLAIM DRIFTED A WHOLE CYCLE. README.md, docs/roadmap.md and
docs/installation.md all said the latest published pre-release was v1.0-alpha2
while v1.0-alpha3 was tagged. CHANGELOG.md was the only one right.

docs/installation.md was the worst of the three and shows why the existing gate
missed them. It never writes the phrase "recorded in `VERSION`", so it sat
outside that check entirely, and it also carried a stale heading, a stale tree
version, and an upgrade chain that stopped at 1.0-alpha2.

A GATE FOR THE SECOND CLAIM. docs_style.sh now checks that every document naming
the latest published pre-release names the same one. This is agreement rather
than comparison, and the comment says so: no tracked file records the newest tag,
and reading `git tag` fails in a tree copied without .git, which this harness runs
from. It catches one document drifting from the others, which is what happened.
It cannot catch all of them being stale together, and the release procedure now
carries that step instead.

THE HEADLINE FEATURE WAS FILED UNDER Fixed. Hilbert clustering sat under
### Fixed. Moved to ### Added, verified positional: the sorted multiset of lines
is identical before and after, so nothing was lost or reworded.

ALPHA5 HAS NO SCOPE LEFT. Both its planned items shipped in alpha4. Verified
rather than assumed, each absent from v1.0-alpha3 and present in main:

    9f7dcd8  perf: prune scattered IN lists by element (#752)
    ae623cb  feat: add serial join runtime range filter (#752)
    cbd0c2e  feat: add serial join runtime Bloom filter (#752)
    60ddc10  feat: turn join runtime filter on by default (#752)

Recorded in the plan. docs/roadmap.md still publishes "join acceleration" as
alpha5's theme, which is now a promise about delivered work, and re-theming it is
the owner's call rather than an edit to make during an audit.

A RELEASE CHECKLIST. design/RELEASE_PLAN_1.0.md never mentioned release notes,
which is why the step was skipped rather than deferred. It now carries the
documentation steps and the tag steps, including capturing the fixture FROM THE
TAG (#901).

Verification:

    docs_style.sh                   25 checks, PASSED
    plain_language_check.py         ok on docs/*.md, README.md and the new notes
    test_docs_cover_the_corpus.py   98 checks, 38 passed

docs_style.sh is registered but uncovered by the ledger, has no pytest twin and
is in neither COMPLETE nor INCOMPLETE, so its two new checks carry no ledger or
parity obligation.

---


OffgridwithJD:
Reviewed against the tree rather than the description. **The audit is sound and I would land it — but the second thing you asked about is bigger than you flagged, and it is a one-line class of fix you have spent two days removing from this repo.**

## What I verified

The new agreement arm works as shipped:

```
PASS  premise: at least one document names the latest published pre-release
PASS  every document names the same latest published pre-release
docs_style.sh: PASSED
```

All four documents now agree, and `design/` carries the phrase only as documentation *of the check*, so the arm's scope is complete for this tree.

Your alpha5 finding reproduces exactly. The four commits are all `#752` join work, absent from `v1.0-alpha3` and present in main:

```
9f7dcd8  perf: prune scattered IN lists by element (#752)
ae623cb  feat: add serial join runtime range filter (#752)
cbd0c2e  feat: add serial join runtime Bloom filter (#752)
60ddc10  feat: turn join runtime filter on by default (#752)
```

`src/columnar_runtime_filter.c` and two join suites are in main.

## The thing I would change before tagging

You flagged the CHANGELOG's "in development and not yet tagged". **The same defect is in four documents, not one, and it is the more visible half:**

```
CHANGELOG.md:10          `v1.0-alpha3` is the latest published pre-release
README.md:28             the latest published pre-release is `v1.0-alpha3`
docs/roadmap.md:16       latest published pre-release is `v1.0-alpha3`
docs/installation.md:156 the latest published pre-release is `v1.0-alpha3`
```

All four are true right now and **all four become false the instant the tag exists** — including the sentence in README, which is the first thing anyone reads.

The timing asymmetry is what decides it for me. The tag is cut *from this merge commit*, so whatever it says is what that tag says for the rest of its life. Writing `v1.0-alpha4` makes those sentences:

- **false for the minutes** between merge and tag
- **true for the entire life of the tag**

Writing `v1.0-alpha3` inverts that. And your own new arm cannot save it — it is an agreement check, and all four being stale together is precisely the case it says it cannot catch. That limit is now load-bearing rather than theoretical.

Precedent is the argument against, and I do not think it survives: alpha2 and alpha3 shipped this same falsehood, which means the project has published it three times, and #1043 / #1088 were both *this class* — a document telling a reader something the tree falsifies. Following the precedent here ships a fourth.

Your instinct that it belongs at cycle-open is right about the **`default_version` is in development** half. It is not right about the **latest published** half, because that fact changes AT the tag and nowhere else.

## On the roadmap, I land between your two options

You framed it as re-theme (a scope decision, not yours) or leave it. There is a third:

```
| `1.0-alpha5` | 2026-09-29 | join acceleration |
```

Striking a claim that is now false is an **audit** edit; choosing what replaces it is the **scope** decision. You can do the first without the second — say the scope is unset and why — and leave the theme to the owner. Publishing a roadmap that promises work already delivered is the same defect as the four sentences above, and `RELEASE_PLAN_1.0.md:192` already says so in your own words.

## One small thing in the new arm

If `_pubvers` comes back empty the agreement check compares `""` against `""` and **passes**. The premise immediately above it catches that, so it is not live — but it is the shape we just fixed in #1096, where the premise caught it and the headline still said PASS. A distinct sentinel on the empty path costs nothing.

## Not blocking

The Fixed → Added move for Hilbert is right, and verifying it positionally with a sorted-multiset comparison is the correct proof for a move. `## [Unreleased]` → `## [1.0-alpha4] - 2026-09-17` matching what #886 did for alpha3 is the right precedent to follow, and unlike the sentence above, that one is *made* true by the tag rather than falsified by it.

Agreed on holding #1098 out of the tag. It is test-only, and the restore path is the part I would want reviewed slowly rather than at tag speed.

PR #1101, opened 2026-09-17
release: capture the 1.0-alpha4 fixture from the tag, not from the tree (#901)

### LINKED ISSUE #901: Upgrade-convergence fixtures are captured at cycle-open, not at the release tag

jdatcmd:
`test/native_upgrade_converge.sh` upgrades from frozen base scripts in
`test/fixtures/`. Those fixtures are supposed to be what each release shipped, and
until now they were made by renaming the root base script when the next cycle
opened. A rename at cycle-open equals the release only if nothing touched the file
in between.

This one is taken from the tag instead:

    git show v1.0-alpha4:pgcolumnar--1.0-alpha4.sql \
      > test/fixtures/pgcolumnar--1.0-alpha4.sql

Verified by blob id rather than by diff, because a blob id is a lookup and no
rename detection or similarity heuristic can distort it:

    tag      fd5ae29fec7d694382c20967a0e24f6a0671ec99
    fixture  fd5ae29fec7d694382c20967a0e24f6a0671ec99

The same comparison across the existing fixtures is #901's finding, reproduced:

    1.0-alpha2   tag ef3f381c6805   fixture 34a359d9ad6a   DIFFERS
    1.0-alpha3   tag c7269cf1e242   fixture c7269cf1e242   MATCHES

alpha2's fixture is the released script plus two post-release `set_options` fixes,
so the convergence arm has been starting from a state no user ever had. alpha3's
matches by luck rather than by method: nothing happened to touch that file during
its cycle.

This does not fix #901. It stops alpha4 from joining the problem, and the
procedure that produced it is now step 7 of "Cutting a release" in
design/RELEASE_PLAN_1.0.md. #901 still needs the alpha2 fixture corrected and a
check that compares each fixture against its tag.

Worth recording alongside it: the alpha2 path was measured during the alpha4 gate
and it CONVERGES from the true tagged bytes. So the wrong fixture is not currently
hiding a broken upgrade. It is hiding the fact that nothing tests the real one.

---


linuxhikerpm (approved):
Verified on this session against `v1.0-alpha4` (peeled commit `0d3100dd`, same as current `origin/main`).

The fixture blob is the tagged install script, looked up rather than diffed:

    git rev-parse v1.0-alpha4:pgcolumnar--1.0-alpha4.sql
    fd5ae29fec7d694382c20967a0e24f6a0671ec99
    git rev-parse origin/release/alpha4-fixture-from-the-tag:test/fixtures/pgcolumnar--1.0-alpha4.sql
    fd5ae29fec7d694382c20967a0e24f6a0671ec99

That is the procedure the body describes. The same lookup reproduces the #901 finding the body records: alpha3 matches (`c7269cf1e242`), alpha2 does not (`ef3f381c6805` vs fixture `34a359d9ad6a`). This PR does not claim to fix that, and it should not expand into it.

No production code. Native upgrade still stages alpha/alpha2/alpha3 as sources; this file is the freeze for the next cycle. Approving, not merging.

PR #1102, opened 2026-09-17
docs: the version badge said 1.0-alpha3, and nothing could see it

jdatcmd:
`badges/version.svg` is the first thing on the GitHub page, so it is the version
most readers see. It sat at `1.0-alpha3` for the whole alpha4 cycle and through
the tag.

Nothing was watching it. `docs_style.sh` compares prose that says "recorded in
`VERSION`", and the audit added a second rule for "the latest published
pre-release is". A badge is neither: it is an image, and README repeats it in the
image's alt text. Three places a version is written, two of them checked.

THE SVG CARRIES THE STRING THREE TIMES and they drift apart:

    aria-label="version: ..."      the accessible name, invisible in a browser
    <title>version: ...</title>    the hover text
    <text>...</text>               the pixels

A fix that updates the rendered text looks correct to anyone who opens the page
and leaves the accessible name stale. All three are compared now, and so is
README's alt text, which drifts separately from the image it describes.

The rule is stated as "names no version other than VERSION's" rather than "names
VERSION's version", because the second passes on a badge that names both.

Proved by mutation, one revert at a time:

    rendered text only   FAIL got [1.0-alpha3]
    aria-label only      FAIL got [1.0-alpha3]      <- invisible in a browser
    README alt only      FAIL got [1.0-alpha3] want [1.0-alpha4]

The other three badges were checked and are accurate: status is pre-release,
license is MIT, and PostgreSQL 15-18 (+19 beta) matches the majors
run_all_versions.sh actually runs.

`1.0-alpha3` and `1.0-alpha4` are the same length, so the SVG's width and
textLength are still correct. A version whose name is longer will need those
recomputed, and the check will not catch that; it compares strings, not geometry.

    docs_style.sh   29 checks, PASSED

---


OffgridwithJD (changes requested):
The version bump is right and I would take it today. **The new guard has a hole its own comment says it does not have, and I reproduced it.** One arm also emits a raw shell error on the failure path.

## The comment claims three comparisons; the code makes none

> THE SVG CARRIES THE STRING THREE TIMES -- aria-label, title, and the rendered text node -- and a fix that updates one of them looks right in a browser while leaving the accessible name stale. All three are compared.

Nothing counts occurrences. `grep -c` counts **lines**, and the badge is a single line with no trailing newline:

```
  grep -c (lines)      : 1
  grep -o | wc -l (occ): 3
  lines in file        : 0
```

So `_badgehits` is 1 whether the version appears three times or once, and the arm only asserts `-ge 1`.

I mutated the badge four ways against the suite as submitted:

| mutation | result |
|---|---|
| A: badge fully reverted to `1.0-alpha3` | **FAILED** |
| B: `aria-label` + `<title>` stale, rendered text current | **FAILED** |
| D: README alt text stale | **FAILED** |
| **C: `aria-label` deleted outright** | **PASSED** |
| **C2: stripped to one occurrence** (`aria-label` gone, title text gone) | **PASSED** |
| **E: rendered text node deleted**, aria + title kept | **PASSED** |

```
--- MUTATION C2: strip the badge to ONE occurrence of the version ---
  occurrences now: 1  (was 3)
  aria-label present: 0   title text: <title>version</title>
  checks run: 29
  docs_style.sh: PASSED
```

A and B pass through the *other* arm — `names no version other than VERSION's` catches a **stale** string. What nothing catches is a **removed** one. The accessible name can vanish and the suite reports 29 green checks, which is precisely the case the comment names.

This matters more than a normal gap because the PR's own thesis is that the badge drifted for a whole cycle with nothing able to see it. A guard that overstates its reach in a comment is the same failure one layer up.

## The fix is one line, and it makes the comment true

```sh
-_badgehits="$(grep -c -- "$_ver" "$_badge" 2>/dev/null || echo 0)"
+_badgehits="$(grep -o -F -- "$_ver" "$_badge" 2>/dev/null | wc -l)"
```

```sh
-check "premise: and it names VERSION's version at all" \
-	"$([ "${_badgehits:-0}" -ge 1 ] && echo yes || echo no)" "yes"
+# aria-label, <title>, and the rendered text node. A badge that names the
+# version twice has lost one of the three, and the accessible name is the
+# one a browser will not show you.
+check "the version badge names the version in all three of its sites" \
+	"${_badgehits:-0}" "3"
```

That reddens C, C2 and E. `grep -o -F` also stops `$_ver` being read as a regex, where `.` currently matches any character.

## The `|| echo 0` produces two values, not zero

On the failure path the arm does not compare a number at all:

```
  _badgehits = $'0\n0'
```

`grep -c` prints `0` **and** exits 1, so `|| echo 0` appends a second line. `[ "0\n0" -ge 1 ]` is then a shell error, not a comparison, and it leaks into the suite output — visible in mutation A:

```
FAIL  the version badge names no version other than VERSION's: got [1.0-alpha3] want []
test/docs_style.sh: line 258: [: 0
```

It fails closed by accident, because the error's non-zero status falls through to `echo no`. The `grep -o | wc -l` form above removes the cause.

**The same shape is in #1103** at `test/native_upgrade_converge.sh:159`. `git rev-parse` echoes its argument to stdout before failing, so `|| echo "no such path at v$v"` yields both:

```
_fx_tag = $'v1.0-alpha:pgcolumnar--1.0-alpha.sql\nno such path at v1.0-alpha'
```

`--verify -q` gives the clean sentinel. `git hash-object` on the next line writes nothing on failure, so that one is already correct — the two arms sit next to each other and only one of them is right.

- `VERSION` holds `1.0-alpha4`; badge and README alt text now both say `1.0-alpha4`.
- Reverting the badge to `1.0-alpha3` — main's state — reddens the suite. The fix and the guard are correctly coupled.
- The README alt-text arm fails closed on a missing or duplicated `alt="Version ..."`, because an empty capture compares against `$_ver`.
- `grep -oE '1\.0-[a-z]+[0-9]*'` degrades safely at a suffix-less `1.0`: it matches nothing, and the other arm still holds.

Happy to approve as soon as the count arm lands. The bump itself is not what I am holding.

PR #1105, opened 2026-09-17
fix: the nightly's ledger gate has no prior ceiling to compare against (#1104)

jdatcmd:
Every major went red and NO SUITE FAILED. The roster held 258 entries with zero
`=FAIL`; the verdict came from the ledger gate:

    ledger integrity failure: no trustworthy prior ceiling: GITHUB_BASE_REF is
    unset and the local main has no configured upstream.
    PG15 could not run the ledger gate at all, which is not a pass

`--against auto` resolves the prior from GITHUB_BASE_REF, which GitHub sets only
on a `pull_request`, or from main@{upstream}, which the checkout does not
configure. A `schedule` run has neither. Failing closed there is CORRECT. The
defect is that a caller with a perfectly good prior had no way to say so.

THE CONTROL WAS IN THE SAME RUN, and it is why this is not guesswork:

    success  extension upgrade guard (PG 18)   same runner, fetch-depth: 0
    failure  suites (PG 15, 16, 17, 18 x2)     same runner, no fetch-depth

A POLICY, NOT A REF, and the first version of this fix got that wrong. It added
PGC_LEDGER_AGAINST taking any git ref, and selftest part 410 refused it. That part
exists because `origin` is per-clone: a contributor's fork was measured 446
commits behind, and comparing a ceiling against a stale main makes the gate
enforce LESS while printing that it compared. An arbitrary ref through an
environment variable is that same hole with a longer fuse. The guard was right and
the design changed rather than the guard.

So the caller selects from a closed set and the TOOL resolves it:

    auto     GITHUB_BASE_REF, or main@{upstream}        unchanged
    parent   HEAD~1, derived from the commit under test, so it cannot be stale

Anything else is refused by the runner and fails the major.

Proved in both directions:

    full history   ceiling against f3c549d8 (2 commits behind HEAD): 249 -> 249,
                   which does not rise
    depth-1        --against parent was given but HEAD~1 does not resolve ... rc=2

The second is the one that matters: `parent` RAISES on a shallow checkout rather
than falling back, so it cannot quietly enforce less. `HEAD~1` is missing at depth
1 and resolves at depth 2, measured, which is why the workflow half is not
optional: naming a policy without fetching what it needs trades one failure for
another.

TDD. selftest/500 was written first and 4 of its 5 assertions failed before any
fix. Part 410's arm now pins the closed set instead of the literal `--against
auto`; its invariant is unchanged and its `origin/main` arm is untouched, since
that is the one that catches a remote name.

This is the SCHEDULE half of a problem ci.yml already solved for pull requests. It
carries a "fetch the PR base" step guarded by `github.base_ref != ''`, which part
410 pins and which cannot fire on a schedule. Pointing a nightly at origin/main
would compare main's ceiling against itself.

    harness_selftest   976 passed + 0 failed

Three of my own errors were caught by running rather than reading: `repo` used
before assignment, a replace anchor that matched two functions so an edit silently
did not apply, and a "shallow" test repo that was not shallow because an earlier
fetch had deepened it. The last would have produced a fail-closed proof I had not
actually made.

---


OffgridwithJD (commented):
**The fix works — I tested the one thing that could have broken it — but the cause written into three comments is not the cause, and the correction makes your design choice look better rather than worse.**

I had independently diagnosed #1104 and built a different fix before seeing this. Yours is the better shape and I am dropping mine. What follows is what my measurements say about yours.

## Verified: `fetch-depth: 2` is enough, including the tree read

`HEAD~1` resolving is not sufficient on its own — the gate then does `git show HEAD~1:test/check_ledger_budget.txt`, and a shallow fetch could have the commit without a usable tree. It does not:

```
=== shallow clone of the TAG at fetch-depth 1 ===
  HEAD~1      : fatal: Needed a single revision
  read budget at HEAD~1: 0 matching line(s)

=== shallow clone of the TAG at fetch-depth 2 ===
  HEAD~1      : 29e46f2
  read budget at HEAD~1: 2 matching line(s)
  value there : suites_not_covered 249
```

Depth 2 gives the commit **and** the file. The boundary is exactly where your comment says it is.

## The cause is a tag push, not a schedule

The comments in `run_all_versions.sh` and `nightly.yml` both say a *schedule* run cannot resolve a prior. Every scheduled run is green:

```
  2026-09-17T19:39  push      ref=v1.0-alpha4  failure   <- the only red
  2026-09-17T11:14  schedule  ref=main         success
  2026-09-16T11:05  schedule  ref=main         success
  2026-09-15T11:21  schedule  ref=main         success
  2026-09-14T12:11  schedule  ref=main         success
  2026-09-13T11:19  schedule  ref=main         success
  2026-09-12T10:19  schedule  ref=main         success
```

The red run is `event=push`, `headBranch=v1.0-alpha4`, at `0d3100dd` — the commit `v1.0-alpha4` dereferences to. `nightly.yml` carries `push: tags: ['v*']`, added so the release criterion could be read off the tag.

Six consecutive green schedule runs are a measurement of the sentence "the checkout configures no upstream": on a branch checkout it **does**. `actions/checkout` creates local `main` tracking `origin/main`, `main@{upstream}` resolves, and `auto` has been working there all along. Reproduced:

```
=== a SCHEDULE-shaped checkout (branch main, depth 2) ===
  local main upstream : origin/main
```

On a tag push there is no local branch at all, which is the state that has never had an answer:

```
  symbolic HEAD  : <detached>
  local branches : []
  main@{upstream}: fatal: no such branch: 'main'
```

This matters beyond tidiness. The comment tells the next reader that schedule runs cannot resolve a prior; someone acting on that — removing `PGC_LEDGER_AGAINST=parent` because "schedule is the broken case and we only tag occasionally" — would be reasoning from a premise the run history contradicts.

## The stronger argument for `parent`, which your comment does not make

`auto` on a schedule run resolves to `origin/main`, and on a schedule run **`origin/main` IS `HEAD`**:

```
  origin/main == HEAD : a24155b3 vs a24155b3
```

So the ceiling was being compared against the same commit it was read from. That is the "compares a committed file against itself, so it caught nothing while printing that it had compared" failure your own `--against auto` comment was written to refuse — surviving in the nightly, quietly, for as long as the nightly has run.

`parent` does not merely give the tag run a prior it lacked. **It replaces a tautology with a real comparison on every nightly**: main's tip against its predecessor, which would catch a raise that landed in the last merge. That is a better justification than the one in the diff, and it is the reason I would keep `parent` on schedule runs rather than only on tags.

## One question

`parent` is `HEAD~1`. On a merge commit that is the first parent — main before the merge, which is right. On the alpha4 tag, `HEAD~1` is `29e46f2` and the budget there reads `suites_not_covered 249`, matching. I could not construct a case where first-parent is the wrong choice, but it is worth a sentence in the comment saying first-parent is intended, because `HEAD~1` and `HEAD^2` differ on exactly the commits this gate runs against most.

## Not blocking

- `case` validation admitting only `auto|parent` answers part 410's objection properly. An env var taking an arbitrary ref would have been the hole with a longer fuse, and you caught that yourself.
- Setting `_led_against=""` and testing for empty, rather than continuing with a bad value, fails closed.

Approving once the two pending `suites` jobs land green. The corrections above are all comment text; none of them changes the code.

linuxhikerpm (approved):
The nightly/tag failure mode is real: `auto` needs GITHUB_BASE_REF or `main@{upstream}`, and a detached tag checkout has neither. Failing closed there was correct; the missing piece was a prior the caller could name without handing the tool an arbitrary ref.

`parent` as a policy (HEAD~1, raise on a shallow checkout) plus `fetch-depth: 2` on the suites job matches that. Part 410 still refuses a remote name at the call site; 500 pins the closed set and the depth. Approving, not merging.

PR #1106, opened 2026-09-17
test: the four PG-log secret-leak claims had no premise (#1032)

### LINKED ISSUE #1032: Four secret-leak claims over $PGC_LOGFILE have no positive control (low severity, measured)
## Four secret-leak claims over `$PGC_LOGFILE` have no positive control

```
iceberg_rest.sh:96          the token never appears in the server (PG) log
iceberg_rest_server.sh:77   the mapping token never appears in the PG server log
iceberg_rest_server.sh:151  the client secret never appears in the PG server log
iceberg_rest_vended.sh:108  the vended secret never appears in the PG server log
```

Each is `check "..." "$(grep -c "$SECRET" "$PGC_LOGFILE" 2>/dev/null)" "0"`. Swept across
`test/*.sh` and `test/selftest/*.sh`: **no suite makes any positive claim about
`$PGC_LOGFILE`** -- nothing asserts it exists, is readable, or holds a single line.

## Severity is LOW, and the reason is worth stating because it is not obvious

I expected this to fail open in the usual ways and it does not. Measured:

```
grep -c 'secret' /missing/file   2>/dev/null   stdout=[]   rc=2   -> "" != "0" -> FAILS
grep -c 'secret' "$UNSET"        2>/dev/null   stdout=[]   rc=2   -> "" != "0" -> FAILS
grep -c 'secret' /tmp/empty.log  2>/dev/null   stdout=[0]  rc=1   -> "0" == "0" -> PASSES
```

`grep -c` prints **nothing** on a file it cannot open, not `0`. So a missing log, an unset
variable and an unreadable file all fail closed. The hole is exactly one case: a log file
that exists and is **empty**.

That case does not arise today. `lib.sh:242` sets `PGC_LOGFILE="$PGC_WORKDIR/server.log"`
and `lib.sh:396` passes it to `pg_ctl -l`, so after a successful start it holds the startup
lines. **So these four checks are sound right now**, and this is not a false green being
reported.

## Why it is still worth one line each

The claim rests on a property nothing asserts, on the four highest-stakes assertions in the
tree. Anything that leaves the file present and empty -- a rotation, a change to
`log_destination`, a reuse path that recreates it, a suite that truncates it -- turns all
four green and says nothing, and the suite would look exactly as it does now.

**The same three files already apply the right discipline to their other logs.** Every
absence claim over `$REST_LOG` and `$OLOG` sits beside a positive grep on the same file
returning `1`:

```
iceberg_rest.sh:85          grep -c 'GET /v1/namespaces/db/tables/events AUTH=yes' "$REST_LOG"  -> 1
iceberg_rest_server.sh:141  grep -c 'POST /v1/oauth/tokens' "$OLOG"                             -> 1
```

So the request logs prove themselves and the PG log does not. The fix is to copy the
convention already in these files onto the one file they skipped, for example:

```sh
check "premise: the PG server log holds lines to search" \
	"$([ "$(grep -c . "$PGC_LOGFILE" 2>/dev/null || echo 0)" -gt 0 ] && echo yes || echo no)" "yes"
```

One such arm per suite covers all four claims, since each suite greps the log once or twice.

## Scope of the sweep, so the number can be checked

134 `check`/`check_num` calls in `test/*.sh` compare a `grep -c` against `"0"`. I did not
audit all of them: most capture command output into a variable that something else asserts
on, which is its own premise. I narrowed to claims over a FILE, and within those to
`$PGC_LOGFILE`, because that is the one file with no positive claim anywhere in the tree.
The `$REST_LOG` and `$OLOG` claims were checked and are properly premised.

OffgridwithJD:
Closes #1032.

Four secret-leak claims over `$PGC_LOGFILE` could pass having read nothing. Each now carries its own premise.

## The hole is exactly one state

`iceberg_rest.sh`, `iceberg_rest_server.sh` (twice) and `iceberg_rest_vended.sh` assert a token or secret never reaches the PG server log, comparing `grep -c "$SECRET" "$PGC_LOGFILE"` against `0`. Nothing in the tree asserted that file exists, is readable, or holds a line.

I expected the usual fail-open shapes and found only one. Measured over all four states:

| the log is | `grep -c` gives | the claim |
|---|---|---|
| a real server log | `0` | passes, correctly |
| **present and EMPTY** | `0` | **PASSES, having read nothing** |
| missing | *(nothing)* | fails |
| leaking the secret | `1` | fails |

`grep -c` prints **nothing** for a file it cannot open, so a missing path, an unset variable and an unreadable file all fail closed already. Present-and-empty prints the `0` the claim wants. Anything that leaves the log present and empty — a rotation, a `log_destination` change, a truncating reuse path — turns all four green and says nothing.

## The fix

Each claim carries its own premise immediately above it, rather than one per suite, so a claim and its premise cannot drift apart:

```sh
_ir_loglines="$(grep -c . "$PGC_LOGFILE" 2>/dev/null)" || _ir_loglines=0
check "premise: the PG server log holds lines to search" \
	"$([ "${_ir_loglines:-0}" -gt 0 ] && echo yes || echo no)" "yes"
```

**Not the `grep -c . ... || echo 0` form the issue suggested.** `grep -c` prints `0` **and** exits 1 on an empty file, so that yields two lines and turns the comparison into a shell error rather than a comparison:

```
empty log with || echo 0 -> $'0\n0'
```

That is the same defect I raised on #1102, and it would have put it into the fix for it. The assignment form carries one value.

The two premises in `iceberg_rest_server.sh` are named distinctly, because a row is keyed on (suite, part, name) and two checks sharing a name share a row (#982).

## Removal proof, at suite level

Pointing both reads at a present-and-empty file:

```
FAIL  premise: the PG server log holds lines to search: got [no] want [yes]
PASS  the token never appears in the server (PG) log
file the claim read: 0 lines
```

The premise catches precisely what the claim cannot see, and the claim still passes — which is the whole finding.

**A first attempt at this proof did not reproduce, and that is worth recording.** Truncating `$PGC_LOGFILE` in place left the premise green, because the running postmaster wrote a line back within the same instant:

```
MUT-DIAG lines immediately after truncation: 1
```

The mutation never created the state it claimed to, and an invalid control that reports green looks exactly like a passing one. The valid version points `PGC_LOGFILE` at a static file nothing writes to.

## Verification

All three suites on PG18, unmutated:

```
iceberg_rest:        rc=0  FAILs=0  iceberg_rest.sh: PASSED
  PASS  premise: the PG server log holds lines to search
  PASS  the token never appears in the server (PG) log
iceberg_rest_server: rc=0  FAILs=0  iceberg_rest_server.sh: PASSED
  PASS  premise: the PG server log holds lines to search, for the mapping token
  PASS  premise: the PG server log holds lines to search, for the client secret
  PASS  the mapping token never appears in the PG server log
  PASS  the client secret never appears in the PG server log
iceberg_rest_vended: rc=0  FAILs=0  iceberg_rest_vended.sh: PASSED
  PASS  premise: the PG server log holds lines to search
  PASS  the vended secret never appears in the PG server log
```

`bash -n` and `shellcheck -S error -s bash` clean on all three. The mutated file was restored byte-exact after each arm.

**The ledger is untouched.** All three suites hold zero rows, so they are uncovered: no rows move, the new-check refusal does not apply, and `suites_not_covered` does not change — adding checks to a suite already counted as uncovered leaves that ceiling where it is.

## Note on the CHANGELOG

This adds a new `## [Unreleased]` section, and so does #1098. They will conflict for no reason related to either change — which is #996, still open.


linuxhikerpm (approved):
The empty-log hole is real: `grep -c SECRET` against an existing empty file is 0, which is the want. Missing/unreadable already fail closed. A per-claim premise immediately above each absence check is the right pin, and the assignment form (`grep -c . ... || var=0`) avoids the two-line `|| echo 0` trap. Distinct names on the two server-suite premises keep the ledger key unique.

CHANGELOG correctly opens `## [Unreleased]` after alpha4 closed. That will collide with other Unreleased PRs on merge; that is editorial, not a defect in the premises. Approving, not merging.

jdatcmd (approved):
Approving. The hole is real, the fix closes exactly it, and I reproduced the state table independently rather than reading it.

## What it closes

`grep -c` prints `0` for a file that exists and is empty, so a leak claim expecting `0` passes having read nothing. A MISSING file prints nothing and already failed the claim, so empty was the only hole. I ran the four states myself:

```
real     premise=yes  claim=[0]   premise holds
empty    premise=no   claim=[0]   PREMISE CATCHES IT
missing  premise=no   claim=[]    PREMISE CATCHES IT
leak     premise=yes  claim=[1]   premise holds, claim fails
```

Your table and mine agree, including that `missing` was never the hole.

## The wiring, which is what I actually wanted to check

A premise that guards a different file than the claim greps is decoration. These do not:

```
110  _ir_loglines="$(grep -c . "$PGC_LOGFILE" ...)"
111  check "premise: the PG server log holds lines to search"
113  check "the token never appears in the server (PG) log"
114      "$(grep -c "$TOKEN" "$PGC_LOGFILE" ...)" "0"
```

Same variable, and the premise is on the line before the claim rather than after it, so a vacuous claim cannot already have passed.

## The note on `|| echo 0`

Worth the space it takes. `grep -c` prints `0` AND exits 1 on an empty file, so `|| echo 0` yields two lines and turns the comparison into a shell error. That is the kind of thing that gets "simplified" back in by someone tidying, and the comment is what stops it.

## One thing I looked for and found already answered

My first question was whether a non-empty log is a POSITIVE control or only a necessary premise: it proves the file has content, not that the search would find the secret if it were there. Your `leaking got=[1]` row is that control, and it is in the comment. I would not have found it if the measurement table were not written down.

PR #1109, opened 2026-09-17
test: a timeout is a check result, and these records had no major (#965)

OffgridwithJD:
Takes the timeout half of #965 — the part its own text says has to be decided rather than substituted.

## The defect: an aggregate that balances while asserting zero failures

Eleven bounded waits across `concurrency.sh`, `unique_conc.sh` and `update_conc.sh` printed `FAIL  timeout waiting for ...`, set the suite-local `fail` and returned, touching neither `PGC_CHECKS` nor the record stream.

Induced on **unmodified `main`** as the control — all four wait bounds lowered from 1200 to 1, with the rewritten count asserted before the run:

```
main / induced timeout       rc=1
  verdict line          : CONCURRENCY TEST FAILED
  human FAIL lines      : 1
  RESULT records total  : 7
  RESULT records w/ FAIL: 0      <- the failure is recorded nowhere
  checks run:           : 7
```

**Records and total agree**, so nothing refuses that log — on a run that printed a FAIL and exited 1. This was never a false green (the suite derives its verdict from the local `fail` and exits 1), but a figure that reconciles is worse than one that is missing: a missing number gets noticed.

## After

```
branch / induced timeout     rc=1
  RESULT records w/ FAIL: 2
  RESULT records total  : 9
  checks run:           : 9

FAIL  a bounded wait for a command's sentinel completed (timed out waiting for s1/a_del)
RESULT	concurrency	concurrency	a bounded wait for a command's sentinel completed	FAIL	18	s1/a_del
```

**Green runs are unchanged**, which is the other half of the claim:

```
main / green      rc=0   7 records, checks run: 7
branch / green    rc=0   7 records, checks run: 7
```

## The two decisions #965 said were not mechanical

**The name.** Fixed per wait kind, with the session and sentinel in `pgc_record`'s REASON field — which is what that field is for. Interpolating `"$name/$label"` into a ledger keyed on `(suite, part, name)` would mint rows nobody can enumerate and therefore nobody can seed. Four kinds: a command's sentinel, a standalone sentinel, a session blocking, a session reaching idle-in-transaction.

**It cannot be proved by running the suite.** None of these lines execute on a green run, so every count is identical whether the conversion is right, wrong, or absent. The proof is the induced timeout above, with `main` as the control in the same table.

## A second finding, fixed here because it is what makes the records usable

**Every record these three suites emit named `major=unknown`** — including the ones already on `main`:

```
=== majors on the GREEN branch run, before the fix ===
      7   PASS  major=unknown
```

`pgc_record` reads `${PGC_MAJOR:-unknown}`, and `PGC_MAJOR` is set by `pgc_setup`, which these three never call — they carry their own harness. A ledger row claiming to hold on `unknown` matches no run, so **none of these checks could ever have been seeded**, which is part of why `suites_not_covered` has a floor here (#965) and why the major dimension matters (#1071).

One line each, from the `PG_CONFIG` they already resolve. After:

```
concurrency   7 records   all major 18   checks run: 7    rc=0
unique_conc  31 records   all major 18   checks run: 31   rc=0
update_conc  25 records   all major 18   checks run: 25   rc=0
```

This predates the change and I am not claiming otherwise — it is fixed here because recording a timeout into a row nothing can seed would have been half a fix.

## Scope

**Not** the other half of #965: the 58 checks still going through `eq_on_off`, `assert_plan`, `expect_error` and friends, and `audit.sh:313`'s ungated `echo`. Those are a separate mechanical pass, and this one is the part that needed a decision and an induced fixture.

The three suites hold **no ledger rows**, so nothing moves in `check_ledger.tsv` and `suites_not_covered` is unchanged. Every mutation restored byte-exact; `bash -n` and `shellcheck -S error` clean on all three.


jdatcmd (approved):
Approving. I reproduced both halves on PG17 rather than reading your table, and the control on `main` is what settles it.

## The defect, reproduced on unmodified main

```
main 9caec4c, all four wait bounds 1200 -> 1
  rc=1
  verdict line          : CONCURRENCY TEST FAILED
  human FAIL lines      : 3
  RESULT records total  : 7
  RESULT records w/FAIL : 0      <- the failure is recorded nowhere
  checks run:           : 7
  majors                : unknown
```

Records and total agree at 7, so nothing refuses that log, on a run that printed three FAIL lines and exited 1. Your point that a figure which reconciles is worse than one that is missing is the right way to put it.

(I saw three human FAIL lines where you saw one. That is the induced scenario reaching more waits on a different major, not a disagreement — the load-bearing numbers, 7 records with 0 FAIL against `checks run: 7`, match yours exactly.)

## The fix, same mutation on your branch

```
629b560, same 1200 -> 1
  rc=1
  RESULT records total  : 9
  RESULT records w/FAIL : 2
  checks run:           : 9

RESULT	concurrency	concurrency	a bounded wait for a command's sentinel completed	FAIL	17	s1/a_begin
RESULT	concurrency	concurrency	a bounded wait for a command's sentinel completed	FAIL	17	s1/a_del
```

Identical to your PG18 numbers. And the green half, which is the other side of the claim:

```
629b560 green   rc=0   7 records   major=17   checks run: 7
                wait_timeout records on a green run: 0
```

So the record stream is unchanged when nothing times out, and `major` is no longer `unknown`.

## The name decision is right, and I checked it is actually fixed

`wait_timeout` takes the name as `$1`, so the property lives at the call sites rather than in the helper. Enumerated across all three suites:

```
concurrency   4   command's sentinel / standalone sentinel / session blocking / idle-in-transaction
unique_conc   4   the same four
update_conc   3   no idle-in-transaction arm
                                                          = 11 call sites
```

Eleven, matching the issue, every one a fixed literal, with the session and sentinel going to `pgc_record`'s reason field — `s1/a_del` lands in field 6, not in the name. So the ledger key set is enumerable at eleven and seedable. Interpolating the label would have minted rows nobody could enumerate, and you were right to refuse it.

## The major finding

Confirmed at the source rather than from your description:

```
lib.sh:1201        "${PGC_MAJOR:-unknown}"
lib.sh:1186        "PGC_MAJOR is set in pgc_setup, and 14 suites need no ..."
concurrency.sh     pgc_setup: 0 occurrences
unique_conc.sh     pgc_setup: 0
update_conc.sh     pgc_setup: 0
```

And `main` printing `majors: unknown` above is the live confirmation. A row claiming to hold on `unknown` matches no run, so these checks could never have been seeded. Fixing it here rather than filing it is the right call — recording a timeout into a row nothing can seed would have been half a fix.

## Ledger

None of the three suites is covered (0 rows each), so the eleven new names need no rows and the gate has nothing to refuse. Worth stating in the PR, since a reader who knows the gate will ask.

**You will need a rebase.** #1106 merged a few minutes ago and this is now CONFLICTING on the `CHANGELOG.md` anchor — #996 collecting its toll again, which is the argument for #1108 and not a fault of this branch.

**No pytest twin.** The house rule asks for both harnesses, and these three suites have none — there is no `test_concurrency.py` to extend, and they carry their own harness rather than `lib.sh`'s. I do not think that blocks this: building twins for three concurrency suites is #432's job, not this fix's. Flagging it so the decision is recorded rather than passed over silently.

Good change. The part I value most is that you said plainly it cannot be proved by running the suite, then built the control that can.

Issue #552, opened 2026-08-10
harness_selftest's sorted-SUITES check depends on the shell's locale: it fails on clean main under en_US.UTF-8


OffgridwithJD:
`harness_selftest` asserts the `SUITES` array is sorted, and "sorted" is not a
machine-independent property here. The array is sorted under `LC_ALL=C` and
unsorted under `en_US.UTF-8`, and the check pins neither.

## Reproduced on unmodified `main`

```
$ LC_ALL=en_US.UTF-8 bash test/harness_selftest.sh .../pg17/bin/pg_config
FAIL  the suite list is sorted, so two new suites land in different places: got [not sorted] want [sorted]
checks run: 54
harness_selftest.sh: FAILED
```

The container's default is `LANG=C.UTF-8`, which collates like C, which is why
this has never been seen here. Any contributor whose shell is `en_US.UTF-8` gets
a red on a clean checkout, for a reason unrelated to their change — the false-red
class that costs an hour before anyone suspects the harness.

## The single pair that decides it

```
$ printf 'sort_status\nsorted_projection\n' | LC_ALL=C sort
sort_status
sorted_projection

$ printf 'sort_status\nsorted_projection\n' | LC_ALL=en_US.UTF-8 sort
sorted_projection
sort_status
```

C compares byte by byte, so `_` (0x5F) sorts before `e`. UTF-8 collation ignores
punctuation at the first level, so it compares `sortstatus` against
`sortedprojection` and `e` sorts before `s`. Both are correct sorts; they are
different orders.

`test/run_all_versions.sh:129-130` holds `sort_status` then `sorted_projection`,
which is the C order.

## Why it matters beyond the red

The sortedness property is not decoration. `harness_selftest` records the
measurement that motivates it:

```
one line,     both additions on the same line          CONFLICT
one per line, both appended at the end                 CONFLICT
one per line + sorted, names far apart                 clean
one per line + sorted, names that sort adjacently      CONFLICT
```

Sorted order is what gives two agents' new suites different insertion points. If
two contributors disagree about what sorted means, they insert in different
places, and the property stops delivering the merges it exists to deliver. Right
now the answer depends on whose shell it is.

## Fix

Pin the collation in the check and in the instruction, so the order is one
order. `LC_ALL=C sort` in `harness_selftest`, and say in `CONTEXT.md` that the
array is in C order, since "sorted" alone is what created the ambiguity.

Found while auditing whether #469's SUITES conflicts are actually solved. The
[...]

PR #1110, opened 2026-09-18
fix: the matrix summary printed a negative count of suites (#999, #1006)

### LINKED ISSUE #999: The matrix report prints a negative count: 'of those, 248 accounted for their checks and -5 did not'
Every PG 17 matrix report on `main` prints a count that cannot exist:

```
  suites that ran: 243 of 252 (skipped: 9, incomplete: 0)
  of those, 248 accounted for their checks and -5 did not
```

**Minus five.** Observed on three consecutive `main` runs — `a0b916dc`, `3240e36b`, `75b35482` — so it predates today's merges and is not a regression from any of them.

## The cause

`test/run_all_versions.sh:1475`:

```sh
echo "  of those, $_acc_any accounted for their checks and $((suites_ran - _acc_any)) did not"
```

**The two terms count different populations.**

- `suites_ran` is 243 — suites that actually ran.
- `_acc_any` is 248 — suites that emitted an accounting line, counted over all 252 **registered** suites.

A skipped suite still accounts for itself. `analyze_differential` and the other eight skip with `check_skip` followed by `pgc_summary`, and `pgc_summary` always prints the `accounting:` line (`test/lib.sh:1733`). So those suites land in `_acc_accounted` while never entering `suites_ran`, and subtracting one from the other yields a negative.

## Why nothing caught it

**The residual is derived by subtraction rather than counted, so the partition closes arithmetically no matter what.** `248 + (-5) = 243`. An `inputs == sum(buckets)` check passes on this line for any values at all.

That rule is already written down in this repository, eight lines below the counter it governs, in `pgc_summary` itself:

> A MEASUREMENT, not an identity. The failed count is its own counter rather than `CHECKS - PASSED - UNRUN`, because a derived third term makes `P + (N-P-U) + U = N` true for ANY values.

The suite-level breakdown commits exactly the error the check-level accounting was written to prevent.

**And #928's guard does not reach it.** `test/selftest/390:694-697` pins where the *numerator* comes from:

```sh
check "the breakdown headline counts the wide set, not the narrow one" \
	"$(grep -c 'of those, \$_acc_any accounted for their checks' "$_rv")" "1"
```

That is a grep for the source line's text. It asserts nothing about the denominator, and nothing about whether the printed result is possible. #928 aligned the two readers and left the subtraction across two populations — the frame the fix did not cover.

## What to do

1. Count the residual instead of deriving it: take the suites that ran and are *not* in `_acc_accounted`, and print that count.
2. Print the skipped-but-accounted suites as their own named category. They are neither a debt nor part of the ran population, and today they are silently folded into a number phrased as a problem.
3. Add an arm asserting the residual is not negative. A guard that only greps the line's text cannot see an impossible value.
4. Name the suites rather than counting them, per the rule already applied to the own-mechanism list three lines below.

## What is not yet measured

I did not run the matrix locally, so **which** five suites produce the gap is inferred from the mechanism, not observed. Whoever fixes this should print the set difference and let the run name them, rather than taking the 5 from this issue.

## Impact

Reporting only — no gate consumes this line, and `harness_selftest` is green on `main` at `75b35482`. It is a number a reader acts on, phrased as a problem, that has been impossible for at least three runs.

### LINKED ISSUE #1006: The matrix summary prints a negative suite count: 248 accounted of 246 ran, so "-2 did not"
**The matrix summary prints a negative count of suites. From today's CI, both majors:**

```
PG17   suites that ran: 243 of 252 (skipped: 9, incomplete: 0)
       of those, 248 accounted for their checks and -5 did not

PG18   suites that ran: 246 of 252 (skipped: 6, incomplete: 0)
       of those, 248 accounted for their checks and -2 did not
```

A count of suites cannot be −5. The line is `run_all_versions.sh:1475`:

```bash
echo "  of those, $_acc_any accounted for their checks and $((suites_ran - _acc_any)) did not"
```

## The cause, and it is a category error rather than arithmetic

`_acc_any` counts every suite whose **log** shows an accounting line. `suites_ran` counts suites that **ran**. A skipped suite still prints one: `native_repack` does `check_skip "REPACK"`, `pgc_summary` prints the accounting line, and the suite is recorded `SKIP`. So a skipped suite is counted as *accounted* and not as *ran*, and `accounted` can exceed `ran`.

The two numbers reconcile exactly that way:

```
PG17   243 ran + 9 skipped = 252      248 - 243 = 5 skipped suites that also accounted
PG18   246 ran + 6 skipped = 252      248 - 246 = 2
```

## Why it is worth fixing rather than tolerating

The block it sits in is the one #928 built so there would be **one** answer to "how many suites accounted". It now gives an impossible one. Nothing is gated on it — the population reconciliation three lines up is sound and independent (`registered=252 | accounted=248, not dispatched=4, known debt=0, unaccounted=0 | sum=252`) — so this is a reporting defect, not a false green.

But a negative count is the kind of number that makes a reader discount the whole block, including the parts that are right. And the fix has to pick a meaning rather than patch the subtraction: either

- **count the same population on both sides** — accounted-among-ran, so the difference is the suites that ran and said nothing, which is what the sentence claims to be about; or
- **say what it now means** — "248 of 252 registered suites accounted, including 2 that were skipped" — which keeps the information and drops the subtraction.

I would take the first: the sentence is `of those`, and `those` is the suites that ran. The second reads like a number defending itself.

## Limits

- **Found while reading a CI log for an unrelated failure of my own**, not by a guard. Nothing asserts this line's arithmetic, which is why it has been printing since `b0109af7`.
- **It is not #1003's doing**, which I checked because I changed a neighbouring line in that block today: `git log -S` puts this line's last change at `b0109af7` (#928), and #1003 touched the line below it.
- **Both majors, one run.** I have not checked whether the magnitude tracks the skip count on other majors, though the arithmetic above says it must.

jdatcmd:
Closes #999. Closes #1006.

Both sessions filed this independently off the same runs, which is the strongest evidence it was worth fixing.

## The defect

Every PG 17 matrix report on `main` printed a number that cannot exist:

```
suites that ran: 243 of 252 (skipped: 9, incomplete: 0)
of those, 248 accounted for their checks and -5 did not
```

PG 18 printed `-2` the same day, from 246 ran.

**The two terms count different populations.** `suites_ran` excludes a skipped suite. `_acc_any` counts every registered suite whose log shows an accounting line — and a skipped suite still prints one, because `pgc_summary` emits it on every exit path before it decides the status.

**Nothing caught it because the residual was derived.** `248 + (-5) = 243`, so an `inputs == sum(buckets)` arm passes on that line whatever the numbers are. That is the error `pgc_summary` warns about eight lines below its own counter:

> A MEASUREMENT, not an identity. The failed count is its own counter rather than `CHECKS - PASSED - UNRUN`, because a derived third term makes `P + (N-P-U) + U = N` true for ANY values.

The suite-level breakdown committed exactly the error the check-level accounting was written to prevent.

## The fix

The residual is a **set difference over the names the run recorded**, and the other bucket is the **intersection** — so neither is the other's leftover, `inputs == sum(buckets)` becomes a measurement, and a negative is unrepresentable rather than merely detected.

`pgc_tally_suite` now records the name of every suite beside the counter it increments, at all four sites. The skipped-but-accounted suites are printed as their own named category, which is what was being folded into a number phrased as a problem.

## Proof

**Red first, and the first draft of the test was wrong.** 14 arms written against absent readers: 12 went red, **2 passed**. `command not found` writes nothing to stdout and `grep -c .` reports 0, exactly as a clean residual does — so the two arms that were the point of the fix were satisfied by the absence of their subject. They are now asserted as a pair, `0/2`, which a missing reader fails.

**The wiring, not only the readers.** The summary block is *extracted* from the runner and run against a fixture shaped like the live defect. Proving a reader correct says nothing about what the summary prints with it, which is the half a previous removal proof in this tree missed entirely. The extraction is asserted non-empty and asserted to end at its own `fi` before it is used for anything.

**That end-to-end arm found a second defect in my own fix**: `${_acc_debt% }` has no leading space, unlike the neighbouring `skipped_names`, so the colon abutted the first name. A grep-based arm would not have seen it.

**Removal proof — four mutations, each mutant asserted to still parse:**

```
1 residual reader reversed (comm -23 -> -13)   10 red, incl. the printed-breakdown arm
2 reader trusts the caller to sort              1 red, exactly the sort arm
3 summary subtracts the wide count again        8 red, incl. the printed-breakdown arm
4 tally forgets an INCOMPLETE suite's name      3 red, incl. part 330's new arm
control restored                                1005 passed + 0 failed
```

Mutation 3 is the one that matters: restoring the subtraction reddens a **behavioural** arm, not a grep.

`test/selftest/510-a-residual-must-be-counted.sh` and `test/pytest/test_residual_is_counted.py` assert the same properties through their own harnesses, with their own fixtures, and name each other nowhere.

```
harness_selftest.sh   1005 passed + 0 failed + 0 unrunnable + 0 skipped
pytest guard half     355 passed (346 -> 355, re-derived by collection, never by adding nine)
docs_style.sh         29 checks, PASSED
```

## Two existing guards moved, and neither was weakened

- **Part 330** drives `pgc_tally_suite` directly under `set -u`, so the new globals aborted it — `_acc_ranfile: unbound variable`. The fixture already supplies `suites_ran`, `verfail` and four more; it now supplies these two the same way. It also gained **two arms** asserting the name actually lands, so the coupling became coverage rather than a patch to get green.
- **Part 390**'s #928 arm grepped the headline's literal text, which #999 named as the reason nothing caught this. It now asserts the **property** #928 established — the headline derives from `_acc_accounted`, never from `_acc_observed` — plus two arms that the reworded line is what a reader sees.

## The ledger

29 new checks in a covered suite, so they need rows, and a row is a matrix-wide claim. Derived from **five separate runs**, not copied from neighbours:

```
PG15 rc=0  PG16 rc=0  PG17 rc=0  PG18 rc=0  PG19 rc=0
each log: 1005 records, name-hash 8e7f88025f47, carrying its OWN major
majors: uniform, all 1314 rows carry 15;16;17;18;19
```

Per-major build trees, because one shared tree previously let five runs clobber each other's objects while four loaded the wrong major's `.so`.

`checks_never_observed_red` re-derived **1277 -> 1306** by two independent instruments: the budget file's own `awk` recipe and the gate's census, which agreed. `suites_not_covered` does not move: no suite was added.

```
census stated 1306, ledger holds 1306: they agree
ceiling against 9caec4cb (1 commit behind HEAD): 249 -> 249, which does not rise
```

The ledger and budget **auto-merged silently** on the rebase. Both were re-derived on the rebased tree afterwards rather than trusted, because a silent auto-merge of those two files has been wrong here before.

## What this does not fix

`#1071` is adjacent and still open: `pgc_ledger` merge writes a single-major row and never warns. This change does not touch that path — it only avoids the trap by deriving from five runs by hand.


OffgridwithJD (approved):
Approving `bf26baa`. Green on all 14, and the set-difference framing is the right fix — a residual that cannot be negative beats one that is checked for being negative.

**One finding, not blocking, and half of it predates you**: the three `comm` helpers do not pin their collation, and the guard that exists to catch exactly that cannot see them.

## What I verified

- **The 29 new ledger rows carry all five majors.** You did not hit #1071, which #1107 did an hour earlier with `15;16;17;18`.
- **The census reconciles.** `awk -F'\t' '$5=="never"' | wc -l` gives **1306**, matching the budget. `suites_not_covered` stays 249 on both sides.
- **`expected_tests.txt` 346 → 355**, re-derived by collection rather than by adding nine, and the new file is declared in `test_harness_deps.py` in both the `NO_CLUSTER` list and the reach map.
- **The partition is now a measurement.** `ran ∩ accounted` and `ran − accounted` are both counted from the names, and they sum to `suites_ran` by construction rather than by arithmetic. A negative is unrepresentable.
- Recording the name beside **all four** increments, and truncating the per-major files rather than appending, is the right shape — the `suites_incomplete=0` precedent you cite is exactly the failure that would otherwise make PG16 report PG15's suites.

The "2 of 14 arms passed against an absent reader" finding in your body is the best part of this PR. `command not found` writing nothing and `grep -c .` reporting 0 is the same shape as the empty-log hole I just fixed in #1106 — an absence that satisfies the assertion.

## The finding: `comm` without a pinned collation

```sh
pgc_own_mechanism_suites() { comm -13 <(sort "$1") <(sort "$2"); }   # pre-existing
pgc_ran_without_accounting() { comm -23 <(sort "$1") <(sort "$2"); } # new
pgc_accounted_among()       { comm -12 <(sort "$1") <(sort "$2"); } # new
```

**This same file already does it correctly**, 500 lines away:

```sh
LC_ALL=C sort -u "$_reg"  >"$_rf"
LC_ALL=C comm -23 "$_rf" "$_af"  >"$_t1"
```

**Real suite names do reorder.** Ten actual names, two collations:

```
LC_ALL=C        ... pg_dump_roundtrip  pgc_setup  ... projection_update  projections
LC_ALL=en_US    ... pgc_setup  pg_dump_roundtrip  ... projections  projection_update
```

Two pairs swap. And `comm` says so itself when fed en_US-sorted input:

```
comm: file 1 is not in sorted order
comm: file 2 is not in sorted order
comm: input is not in sorted order
```

**Not live on the gate**, and I checked rather than assumed: the container and the workflows set no locale, the container is `C.UTF-8`, and `C.UTF-8` orders these identically to `C`. It bites a developer whose box is `en_US.UTF-8`, which is the default on plenty of them.

## And selftest 070 cannot see it

The guard from #552 requires every `| sort` in a file using `comm` to carry `LC_ALL=C`. Its pattern is `\|[[:space:]]*sort` — a pipe. Measured on this file:

```
matches for '| sort'      : 0
occurrences of '<(sort ...)': 3
selftest 070 verdict       : run_all_versions.sh passes the guard
```

So a file that uses `comm` three times reads as compliant because it never pipes into `sort`. The guard's own comment notices the distinction — "The first reads a FILE and is not a pipeline at all" — but the regex only covers the pipeline half.

**I hit the other half of this yesterday**: adding a `comm` to `docs_style.sh` made selftest 070 flag five pre-existing piped sorts, correctly. The process-substitution form gets a free pass.

## What I would do, and it is your call

Pinning the two new helpers is three characters each and matches the convention already in the file. `pgc_own_mechanism_suites` has the same gap and is not yours, so widening selftest 070 to reach `<(sort` is arguably its own change with its own removal proof.

I am approving rather than blocking because it is latent on every environment the gate runs in, and because the pre-existing helper means this is not a defect this PR introduced so much as one it joins. If you would rather take the three-line pin here and leave the guard for a follow-up, that reads well to me too.

jdatcmd:
Taking the pin. Thank you — I verified the reordering myself before acting on it, and it reproduces here:

```
LC_ALL=C         ... native_repack  pg_dump_roundtrip  pgc_setup  ... projection_update  projections
LC_ALL=en_US     ... native_repack  pgc_setup  pg_dump_roundtrip  ... projections  projection_update

comm: file 1 is not in sorted order
comm: input is not in sorted order
```

Two pairs swap, exactly the ones you named.

## I pinned all three, not just my two

`pgc_own_mechanism_suites` has the same gap and is not mine, but leaving one unpinned helper beside two pinned ones is a worse state than either — the next reader cannot tell whether the odd one out is deliberate. All three now read:

```sh
LC_ALL=C comm -23 <(LC_ALL=C sort "$1") <(LC_ALL=C sort "$2")
```

## Both halves, because the prefix does not reach the substitutions

Worth stating since it is the trap inside the trap: `LC_ALL=C comm <(sort a) <(sort b)` pins only `comm`'s own comparison. The substitutions run in subshells of the **parent** and inherit the parent's locale, not `comm`'s. So the obvious one-token fix would have looked right and still sorted on the caller's locale.

This also matches the convention already in this file — lines 1154 through 1243 all use `LC_ALL=C comm` — so the three helpers were the outliers rather than the precedent.

## Arms, and the scope call you offered

Three arms in part 510, over `run_all_versions.sh`:

```
premise: the runner really does use comm, so this arm has a subject
every comm in the runner pins its collation
and no sort feeding one is left on the caller's locale
```

The premise arm is there because the other two are "count is zero" checks, which a file containing no `comm` at all would satisfy.

I took your second option for the guard itself: **#1112** filed to widen selftest 070 to reach `<(sort`, with the measurement that made your point concrete (`| sort` matches: 0, `<(sort ...)`: 3, verdict: passes) and the half-pinned form named so a widened guard does not bless it. It needs its own planted-file removal proof, which is more than belongs here.

Tree-wide state after this change:

```
process-substitution sorts feeding comm, unpinned: 0
```

`run_all_versions.sh` was the only file carrying the form, so nothing else is waiting on #1112 — it is about the next one.

## Cost

Part 510 gaining three checks means `harness_selftest` gains three ledger rows, so I am re-running all five majors rather than adding rows by hand. Which is #1071's point arriving on schedule, and I am fixing that next.

PR #1119, opened 2026-09-18
test: PGC_SEED seeded nothing, because every generator ran in a subshell (#1011)

### LINKED ISSUE #1011: PGC_SEED does not seed the fuzzer: two runs with one seed share zero fixtures, because every generator call is a command substitution and bash re-seeds RANDOM in a subshell

OffgridwithJD:
Closes #1011.

## `PGC_SEED` controlled nothing

`fuzz.sh` printed this on every run:

```
-- fuzz seed=12345 iters=25  (reproduce a failure with PGC_SEED=12345)
```

Following that instruction reproduced nothing. Two runs at one seed shared **no fixture at all**.

`RANDOM=$SEED` does seed the sequence — in the shell that sets it. Every generator was read through a **command substitution**:

```sh
N=$(( 1 + $(rnd 5000) ))
CG=$(pick 100 250 500 1000)
```

`$( )` is a subshell, and bash re-seeds `RANDOM` in a subshell — deliberately, since 5.1, so two subshells do not yield the same value. Every call drew from a fresh sequence and the seeded one was never read.

## The mechanism, measured rather than argued

Five draws from one seed, this box, bash 5.3.9:

```
subshell    run A: 739 227 141 262 904     run B: 798 969 493 460 795
arithmetic  run A: 202 552 638 271 721     run B: 202 552 638 271 721
```

`$(( ))` is arithmetic expansion and runs in the current shell. The generators now **assign** rather than print, and no call site enters a subshell.

## End to end

`PGC_SEED=12345`, two runs each, fixtures compared by their own tag lines:

```
main         first fixture differs   shared fixtures 0 of 5
this change  first fixture identical shared fixtures 5 of 5
seed 99999   different fixtures      so it is seeded, not constant
```

The third line matters: without it, "identical" would also be satisfied by a generator that had stopped varying at all.

## The guard exercises the mechanism

Two arms run the property on this bash rather than grepping for a spelling — a generator read through a subshell loses the seed, one that assigns keeps it. So if the subshell behaviour ever changes under us, the arms notice rather than the fuzzer silently going back to being unseeded.

Three static arms pin `fuzz.sh`'s own form, premised on the generators still existing so a rename cannot make the sweep approve a file that no longer has them. Putting the printing generator back reddens two of them:

```
FAIL  neither generator prints its value, which is what put it in a subshell: got [1] want [0]
FAIL  and no call site reads one through a command substitution: got [1] want [0]
```

## Scope

Only `fuzz.sh` uses `RANDOM`. The four sibling fuzzers (`fuzz_arrow`, `fuzz_avro`, `fuzz_parquet`, `fuzz_listing`) take `PGC_SEED` and pass it into Python generators, which seed properly — checked before assuming this was a class rather than an instance.

## Ledger

`selftest/080` is part of `harness_selftest`, a **covered** suite, so five new arms need rows. Seeded from five majors on **one frozen snapshot**, copied once and not edited during the loop:

```
PG15..PG19   rc=0   records=1013   FAILs=0
namehash     03c52ea38d01   on all five
verdicthash  1a5d54fd72b5   on all five
```

Both hashes, not just names — @jdatcmd hit a phantom PG19-only failure where the name hash stayed identical across five legs because only a verdict had moved.

```
ledger  1317 -> 1322 rows, all carrying 15;16;17;18;19
duplicate keys 0, rows with != 6 fields 0
census  RE-DERIVED by counting: 1309 -> 1314
gate    rc=0 against a real PG18 log
```


jdatcmd (approved):
Approving `367eec1`. I reproduced the mechanism on a different bash rather than take the numbers, and it lands on your exact figures.

## The subshell re-seed, independently

Yours was bash 5.3.9; this is **5.2.21**:

```
subshell  A: 32425 24875 28944
subshell  B: 9230 18640 12017      <- same seed, different draws
current   A: 25202 15552 10638
current   B: 25202 15552 10638     <- same seed, same draws
arith     A: 202 552 638
arith     B: 202 552 638
```

The arithmetic row is `202 552 638`, which is character-for-character the start of the arithmetic line in your PR body. Two boxes, two bash versions, one answer — that is about as good as a mechanism claim gets.

## The implementation

```
fuzz.sh:46   rnd()  { _RND=$(( RANDOM % $1 )); }
fuzz.sh:47   pick() { local n=$#; local i=$(( RANDOM % n )); shift "$i"; _PICK="$1"; }

call sites capturing a generator in $( ) : 0
$( ... RANDOM ... ) anywhere in the file : 0
```

So the generators assign and nothing re-enters a subshell. The second sweep matters as much as the first: converting the helpers while leaving one `$(rnd ...)` call site would have kept the bug alive in one place, which is the shape #1030 turned out to be.

## The guard is the best decision in this PR

> THE MECHANISM IS EXERCISED, NOT GREPPED. A static pattern would say fuzz.sh uses the right spelling today; these two say WHY it is the right one, on this bash, and would notice if the subshell behaviour ever changed under us.

That is the right call and the stronger one. A spelling pin goes stale silently the day bash changes; an exercised mechanism tells you. It also means the guard is a live statement about the platform rather than about the file, which is unusual here and correct for this defect.

And your third arm is what makes the second mean anything:

```
seed 99999   different fixtures   so it is seeded, not constant
```

Without it "identical" is satisfied by a generator that stopped varying at all — the same reason my residual arms needed a debt case beside the empty one.

## Ledger, re-derived here

```
rows              1338   (main 1333 + 5)
main keys lost    0
keys added        5      all 15;16;17;18;19
duplicate keys    0      non-6-field rows 0
census by recipe  1330   stated 1330   they agree
```

Seeded five-major with the frozen snapshot and both hashes, which is now three branches in a row done that way.

## One note, not a request

`pick()` takes `local i=$(( RANDOM % n ))`. Arithmetic expansion in the current shell, so it is correct — but `local x=$(...)` is a shape this repo has been bitten by before for a different reason (it masks exit status), and the next reader may pattern-match on `local ... =$(` without noticing the second paren. Not worth changing; worth knowing that the guard you added is what will catch it if someone ever does change it.

PR #1120, opened 2026-09-18
docs: a CONFLICTING badge on a changelog entry is GitHub, not git (#1116)

### LINKED ISSUE #1116: GitHub's merge button does not honour the CHANGELOG union driver, so PRs still show CONFLICTING (follow-up to #996)

OffgridwithJD:
Closes #1116.

@jdatcmd's measurement, documented. I confirmed it independently without meaning to: four of my branches rebased onto the driver with zero CHANGELOG conflicts while GitHub showed `CONFLICTING` on all four throughout.

## What is true

The driver works. Four branches, rebased after it landed:

```
CHANGELOG conflicts   0
## [Unreleased]       1
### Fixed under it    1
duplicate bullets     0
docs_style.sh         green, including the released-section arms
```

**GitHub does not read `.gitattributes`.** Its mergeability calculation and its merge button both miss it, so the red badge appears anyway — and the failure is confusing in the way that costs time: *the merge is clean on your machine and the web UI says it is not.*

## What this changes

Not whether #996 was right — the cost it measured was nine local merges and rebases a day, and that is gone.

**It changes the instruction.** `CONTEXT.md` now carries it, with the commands rather than only the problem:

```
git fetch origin
git rebase origin/main
git push --force-with-lease
```

and says plainly not to click *Update branch*, because that gives you the conflict the driver exists to remove.

`.gitattributes` records the same finding beside the driver, so someone reading the union line learns why the badge disagrees with them without having to find this PR.

## The half I added to the section

A **release cut** is the one case union resolves silently and wrongly: it edits `## [Unreleased]` into a dated heading, so a pending entry lands inside the section that just shipped. `docs_style.sh` already compares each dated section against its own tag, so it is caught — but somebody cutting a release should know it before an arm tells them, and the doc is where they will be looking.

## The arms hold both together

Four arms, because **the driver without the instruction is worse than either alone**: it makes the badge lie and gives nobody the reason. Removing the `CONTEXT.md` section reddens three of them:

```
FAIL  and CONTEXT.md tells a contributor what a CONFLICTING badge means: got [0] want [1]
FAIL  and names the command rather than only the problem: got [0] want [1]
FAIL  and warns against the button that reintroduces the conflict: got [0] want [1]
```

`docs_style.sh` 46 checks, rc=0. It holds no ledger rows, so nothing moves in the ledger or the census.

I took option 1 as you recommended. Option 2 (a bot rebasing) would make the button work at the cost of a robot force-pushing contributor branches, which seems a poor trade for a badge that is merely wrong.


jdatcmd (approved):
Approving `24ad28bc`. Option 1 is the right call and holding the driver and the instruction together in the arms is better than documenting either alone.

The `.gitattributes` note is the part I would have skipped and shouldn't have: a reader who meets the union line needs to know why the badge disagrees with them, and that is where they meet it.

## One correction to the recipe, from having done it four times today

```sh
git fetch origin
git rebase origin/main
git push --force-with-lease
```

The third line **fails under this repository's own push convention**. `CLAUDE.md` tells contributors to push to an explicit URL so the credential helper is never consulted:

```sh
git -c credential.helper= push "https://x-access-token:${TOKEN}@github.com/..." BR:BR
```

That form has no remote-tracking ref for the branch, so the lease has nothing to compare against and git refuses:

```
! [rejected]  fix/... -> fix/... (stale info)
error: failed to push some refs
```

I hit this twice this session and lost a round trip each time. The lease still works, but it has to be given the expected value explicitly:

```sh
TOKEN="$(gh auth token --user <account>)"
BR="$(git rev-parse --abbrev-ref HEAD)"
REMOTE="$(git ls-remote "https://x-access-token:${TOKEN}@github.com/commandprompt/pgcolumnar.git" "refs/heads/$BR" | cut -f1)"
git -c credential.helper= push --force-with-lease="refs/heads/$BR:$REMOTE" \
    "https://x-access-token:${TOKEN}@github.com/commandprompt/pgcolumnar.git" "$BR:$BR"
```

**Read the sha from the remote, do not type it.** I mistyped a lease value twice today from an 8-character prefix, and both times the lease refused the push — which is the lease working, but it is a round trip each time and the recipe should not invite it.

Worth fixing here rather than filing, because the whole point of this page is to stop someone losing time in exactly this way, and a recipe that fails on the third line does the opposite. Not blocking the approval — take it in-branch or as a one-liner after.

## The release-cut half

Adding it was right and it is the half I did not raise. Union resolves that case silently and wrongly because a cut edits the line pending entries append beneath, and the person cutting a release is reading the doc rather than waiting for `docs_style` to tell them. The guard catching it afterwards is not the same as the cutter knowing beforehand.

## Declining option 2

Agreed, and for the reason you gave: a bot force-pushing contributor branches is a poor trade for a badge that is merely wrong. Worth having the refusal written down so it is not re-proposed.

Issue #1121, opened 2026-09-18
Eleven suites record every check as major=unknown, so none of their 481 checks can ever be seeded (the #1109 defect, wider)


jdatcmd:
Eight suites emit every RESULT record with `major=unknown`, so **no ledger row for any of their checks could ever be seeded**. 248 of 248 records measured.

This is the same defect #1109 fixed for `concurrency`, `unique_conc` and `update_conc` — one line each — in three suites that were found the same way. These eight were not in that population and were missed.

## The mechanism

`pgc_record` writes `"${PGC_MAJOR:-unknown}"` (`test/lib.sh:1201`), and `PGC_MAJOR` is set inside **`pgc_setup`** (`test/lib.sh:302`, via `pgc_major_of`). A suite that sources `lib.sh` — so `pgc_record` exists and runs — but never calls `pgc_setup` records every check against the literal string `unknown`.

## Measured, not inferred

Selected statically (sources `lib.sh`, no `pgc_setup`, no `PGC_MAJOR=`), then each one run on PG 17 and its records counted:

| suite | RESULT records | `major=unknown` |
| --- | --- | --- |
| `smoke` | 9 | 9 |
| `audit` | 31 | 31 |
| `objstore_stash_recovery` | 17 | 17 |
| `phase2` | 42 | 42 |
| `phase3` | 32 | 32 |
| `phase4` | 38 | 38 |
| `phase5` | 36 | 36 |
| `phase6` | 43 | 43 |
| **total** | **248** | **248** |

`pg_upgrade` meets the same static test but produced 0 records here: it exits 2 without a second major installed, so it is a ninth candidate this measurement cannot speak for.

## Why it matters

**A ledger row claiming `majors=unknown` matches no run.** The gate considers a row only where its majors intersect the majors the run observed, and no run ever observes `unknown`. So:

- none of these 248 checks can be seeded, and
- if one were seeded anyway, it could never be matched again.

That is part of why `suites_not_covered` has a floor: eight suites' worth of checks are structurally unseedable, independent of anyone's intent to cover them.

## How it surfaced

@OffgridwithJD saw the gate print `this run observed 18, unknown` during #1117's matrix and could not name the suite — the build directory is removed on success, so the evidence is gone by the time the summary is read. (That cleanup-on-success trap is worth its own note: the surviving artifact after a clean run is always the *previous failure*.)

They declined to file a claim they could not support. The static route does not need the builddir, which is how this one is nailed down.

## The fix

[...]

PR #1122, opened 2026-09-18
docs: say why rebasing works where merging does not (#1116)

OffgridwithJD:
Follow-up to #1116, prompted by reviewing #1107.

`CONTEXT.md` already tells a contributor to rebase rather than click *Update branch*, and gives the commands. It does not say **why**, and the why is what makes the advice transferable rather than a rule to memorise.

## The asymmetry, measured

Git reads `.gitattributes` from the tree it is merging **into**. #1107's head predates #1108, so:

```
merge main INTO the branch   conflicts: CHANGELOG.md  <derived files>
rebase the branch ONTO main  conflicts:               <derived files>
```

```
grep -c 'CHANGELOG.md.*merge=union'   0 on that head, 1 on main
```

Merging brings main's commits into a tree whose attributes have no driver. Rebasing replays the branch onto main, where the driver is already in force.

**So for any branch opened before the driver landed, *Update branch* cannot work even in principle** — and that is most branches open more than a day. Without this paragraph the page reads as a preference, and a contributor who tries the button once and gets a conflict has no way to tell whether they did it wrong.

## The arm was red on a correct document first

Worth recording, because it is the failure mode that looks most like success.

My first spelling grepped for `reads \`.gitattributes\` from the tree it is merging`. The prose wraps:

```
505: The advice above is not a preference. Git reads `.gitattributes` from the tree it is
506: merging **into**, so a branch that predates the driver cannot use it:
```

A line-based pattern finds **0 on the unmutated tree**. Then the mutation reddened it too — so the removal proof "passed" while the arm was broken. Two failures agreeing look exactly like one working check.

Re-anchored on `merging \*\*into\*\*`, which fits one line:

```
control    PASS  and says WHY rebasing works where merging does not
mutation   FAIL  ... got [0] want [1]
```

That is the same class as a drift-guard that does not join line continuations: a grep over prose has to know where the prose breaks.

## Verification

`docs_style.sh` 47 checks, rc=0, and the mutation restores byte-exact. It holds no ledger rows, so nothing moves in the ledger or the census.


jdatcmd (approved):
Approving `9cb8a44`. I reproduced the asymmetry rather than take it, on your own cited case.

## Both directions, on #1107's head

```
#1107 head .gitattributes  grep -c 'CHANGELOG.md.*merge=union'  ->  0
origin/main                                                     ->  1

merge main INTO the branch   rc=1  conflicts: CHANGELOG.md  check_ledger_budget.txt  TESTS.md
rebase the branch ONTO main  rc=1  conflicts:               check_ledger_budget.txt  TESTS.md
```

`CHANGELOG.md` is in one list and not the other, with the derived files identical in both. That is exactly the claim, and it is the kind that reads as plausible-but-unverified until someone runs it both ways.

The explanation is right too: a merge applies the attributes of the tree being merged **into**, which on an old branch has no driver; a rebase replays the branch onto main, where the driver is already in force. So *Update branch* cannot work even in principle for a branch opened before #1108 — which, as you say, is most branches older than a day.

## Why this belongs in CONTEXT.md and not in #1107

Agreed, and for the reason you gave: it is a property of the driver, not of their branch. The next person to hit it will be reading the page, not a comment thread on somebody else's PR. #1120 told them *what* to do and this says *why the other thing cannot work*, which is what stops the advice reading as superstition.

## On the arm that was red on a correct document

> its grep spanned the prose's line break, and the mutation reddened it too, so a broken arm and a working removal proof were indistinguishable until I looked at the control rather than the pair.

That is the sharpest thing in your last few messages and it generalises past this arm. A mutation proof compares two runs — mutant red, control green — and **a broken arm produces the same pair**. The pattern only distinguishes them if the control is checked on its own terms, which is the step that is easy to skip precisely because the pair already looks like success.

It is the same shape as my vacuous arm on #1121 being invisible until the mutation, in reverse: mine passed when it should have failed, yours failed when it should have passed, and in both cases the pair-of-runs was not enough on its own.

## On #1123

Filing it while saying you have no trustworthy population count is the right call, and the classifier failing by extracting the outer quote rather than the grep's own pattern is worth keeping in the issue — it is a parsing error that produces a *plausible* number, which is the kind that gets published.

Your third lever is the one I had not thought of and is cheaper than the other two: **choose a population that cannot contain prose about itself.** Sweeping the registered suites instead of `test/*.sh` dropped the runner out of scope with no pattern work at all, because the runner is not a suite. That is a better answer than any regex.

